PR #11, opened 2015-03-30
Delta compression

alfhenrik:
- [x] Find package
- [x] Find nearest packages
- [x] Apply delta file to existing package
- [x] Add tests
 - [x] Add test for applying delta
- [x] Rebase `master` on top of this PR
- [x] Locking so that we don't apply same delta multiple times ~~(or upload the resulting package)~~

This goes :couple: with OctopusDeploy/OctopusDeploy#183


source/Calamari/Commands/ApplyDeltaCommand.cs
```diff
@@ -0,0 +1,104 @@
...
+            {
+                SkipHashCheck = true
+            };
+            using(var basisStream = new FileStream(basisFilePath, FileMode.Open, FileAccess.Read, FileShare.Read))
+            using(var deltaStream = new FileStream(deltaFilePath, FileMode.Open, FileAccess.Read, FileShare.Read))
+            using(var newFileStream = new FileStream(newFilePath, FileMode.CreateNew, FileAccess.Write, FileShare.Read))
+            {
+                Log.Info("Applying delta to {0} with hash {1} and storing as {2}", basisFilePath, fileHash, newFilePath);
+                Log.Info("##octopus[stdout-verbose]");
+                deltaApplier.Apply(basisStream, new BinaryDeltaReader(deltaStream, progressReporter), newFileStream);
```
alfhenrik: :goberserk:

source/Calamari/Integration/FileSystem/PackageStore.cs
```diff
@@ -0,0 +1,160 @@
...
+        string GetPackageRoot(string prefix)
+        {
+            return string.IsNullOrWhiteSpace(prefix) ? rootDirectory : Path.Combine(rootDirectory, prefix);
+        }
+
+        StoredPackage ReadPackageFile(string filePath)
+        {
+            try
+            {
+                var metadata = new ZipPackage(filePath);
```
alfhenrik: `ZipPackage` is not an `IDisposable`

mjrichardson:
Nice work on this.  Overall looks really good, and should be a win.

Inline comments to come, but one question first: do you think we should consider storing the signature and delta files alongside the packages, rather than in the Octopus working directory?
Which leads to a follow-up question, we might want to apply retention-policy to these?

alfhenrik:
Signature and delta files on the server are stored in the PackageCacheDirectory (which gets cleaned out every 4 hours or so) and on the Calamari side it's just stored in the working folder of the task so gets cleaned up after the task is finished.

mjrichardson:
Ahh right. I see now. Perfect ;)

source/Calamari/Commands/ApplyDeltaCommand.cs
```diff
@@ -0,0 +1,131 @@
...
+        string FindOctoDiffExecutable()
+        {
+            var basePath = Path.GetDirectoryName(GetType().Assembly.FullLocalPath());
+            var exePath = Path.Combine(basePath, "Octodiff.exe");
+            if (!File.Exists(exePath))
+                throw new CommandException("Unable to find Octodiff.exe in " + basePath);
+            return exePath;
+        }
+
+        string FormatCommandArguments(string basisFilePath, string deltaFilePath, string newFilePath)
```
mjrichardson: This method isn't used.
Given that, are the skip-verification and progress options may no longer be required either?

source/Calamari/Commands/ApplyDeltaCommand.cs
```diff
@@ -0,0 +1,131 @@
...
+            if (previousPackage.Metadata.Hash != fileHash)
+            {
+                throw new CommandException("Basis file hash " + previousPackage.Metadata.Hash +
+                                           " does not match the file hash specified " + fileHash);
+            }
+        }
+
+        string FindOctoDiffExecutable()
+        {
+            var basePath = Path.GetDirectoryName(GetType().Assembly.FullLocalPath());
```
mjrichardson: I had an issue with this method the other day, as it uses the `Assembly.CodeBase` property.  This gave me issues when the path was a UNC path. More likely than not, we actually want `Assembly.Location`
See the remarks section of https://msdn.microsoft.com/en-us/library/system.reflection.assembly.codebase%28v=vs.110%29.aspx

PR #14, opened 2015-04-12
Skip existing installations

PaulStovell:
This PR adds three features:

 - [x] Support for skipping installation of a package if the package is already installed
 - [x] Output variables for the staging directory
 - [x] Variables for the previous installation

**Skipping previous installations**: when users choose this:

2015-04-12 14_33_04-giantpackage settings - octopus deploy

This should happen:

2015-04-12 14_32_37-deploy giantpackage release 1 0 0 to test - octopus deploy

**Output variables for staging directory**: makes this variable available:

2015-04-12 14_36_39-new step - octopus deploy

And it works:

2015-04-12 14_48_40-deploy giantpackage release 1 0 0-a1 to test - octopus deploy

**Previous installation**: makes these useful variables available:

2015-04-12 14_52_48-deploy giantpackage release 1 0 0-a2 to test - octopus deploy


source/Calamari/Log.cs
```diff
@@ -32,6 +33,18 @@ public static void Verbose(string message)
             }
         }
 
+        public static void SetVariable(string name, string value)
```
mjrichardson: Perhaps this method should be named SetOutputVariable or something similar, given it doesn't actually set the variable.  Or perhaps an extension method the VariableDictionary that both sets the variable and writes the log message?

source/Calamari/Deployment/SpecialVariables.cs
```diff
@@ -106,6 +112,8 @@ public static class Features
         public static class Action
         {
             public const string Name = "Octopus.Action.Name";
+            public const string SkipRemainingConventions = "Octopus.Action.SkipRemainingConventions";
```
mjrichardson: Should the variables collection be used for passing variables between conventions?  If they are never used outside of the Calamari conventions.  It seems it should be more for passing between server and Tentacle/Calamari.  In the case of SkipRemainingConventions, and SkipJournal etc, we could simply set properties on the RunningDeployment?

source/Calamari/Deployment/Conventions/ContributePreviousInstallationConvention.cs
```diff
@@ -0,0 +1,40 @@
...
+            }
+            else
+            {
+                previousExtractedFrom = previous.ExtractedFrom;
+                previousExtractedTo = previous.ExtractedTo;
+                previousVersion = previous.PackageVersion;
+                previousCustom = previous.CustomInstallationDirectory;
+            }
+
+            deployment.Variables.Set(SpecialVariables.Tentacle.PreviousInstallation.OriginalInstalledPath, previousExtractedTo);
```
mjrichardson: Should these variables be made available to run-script actions?  
Is this possibly something the Octopus server should be setting?

mjrichardson:
Also, I think the ISemaphore is quite neat.

source/Calamari/Deployment/SpecialVariables.cs
PaulStovell: Good point, will change that

source/Calamari/Log.cs
PaulStovell: Fixed

source/Calamari/Deployment/Conventions/ContributePreviousInstallationConvention.cs
PaulStovell: Script actions don't run within the context of a step, so there would be no way to track it to a previous installation - e.g., if the project had 4 NuGet package steps followed by one standalone script step, which package would we use to calculate the previous install?

source/Calamari/Deployment/SpecialVariables.cs
PaulStovell: Actually on second thought, `SkipRemainingConventions` is sometimes set by PowerShell scripts by users - e.g., in their PreDeploy.ps1, they might decide conditionally not to proceed with the package deployment, in which case they would set this variable. `SkipJournal` is new, but a similar argument could be made for it (if they conditionally decide not to install, they might also decide not to have it counted towards the previous install).

mjrichardson: Fair enough.  My comment only applied if there was *zero* reason anything outside of Calamari needed to know about the variable.

PR #16, opened 2015-05-06
Feature mono

zentron:
Enables running calamari on a *Nix server on mono.


source/Calamari.Tests/Calamari.Tests.csproj
```diff
@@ -34,6 +34,10 @@
       <SpecificVersion>False</SpecificVersion>
       <HintPath>..\packages\MarkdownSharp.1.13.0.0\lib\35\MarkdownSharp.dll</HintPath>
     </Reference>
+    <Reference Include="Microsoft.Web.Administration, Version=7.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
+      <Private>True</Private>
+      <HintPath>..\packages\Microsoft.Web.Administration.7.0.0.0\lib\net20\Microsoft.Web.Administration.dll</HintPath>
```
PaulStovell: So we'll deploy this assembly on Mono (even though it won't work) to avoid errors around type discovery?

source/Calamari.Tests/Calamari.Tests.csproj
```diff
@@ -212,7 +256,8 @@
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <PropertyGroup>
-    <PostBuildEvent>xcopy "$(SolutionDir)..\tools\Octodiff\Octodiff.exe" "$(TargetDir)" /y</PostBuildEvent>
+    <PostBuildEvent>xcopy "$(SolutionDir)..\tools\Octodiff\Octodiff.exe" "$(TargetDir)" /y
+xcopy "$(SolutionDir)packages\NuGet.CommandLine.2.8.3\tools\Nuget.exe" "$(TargetDir)" /y</PostBuildEvent>
```
PaulStovell: What is Nuget.exe used for?

source/Calamari.Tests/Fixtures/Bash/BashFixture.cs
```diff
@@ -0,0 +1,51 @@
...
+using Calamari.Integration.Scripting.Bash;
+using Calamari.Integration.ServiceMessages;
+using Calamari.Tests.Helpers;
+using NUnit.Framework;
+using Octostache;
+
+namespace Calamari.Tests.Fixtures.Bash
+{
+    [TestFixture]
+    [Category(TestEnvironment.CompatableOS.Nix)]
```
PaulStovell: Nice!

source/Calamari.Tests/Calamari.Tests.csproj
zentron: No, including this reference in the test project ensures that the dll is copied over so that the tests run. As the assembly is referenced by classes loaded via reflection it doesn't get brought into the bin directory in Calamari.Tests

source/Calamari.Tests/Fixtures/Deployment/DeployPackageFixture.cs
```diff
@@ -49,10 +49,20 @@ public void ShouldDeployPackage()
...
-            result.AssertOutput("Extracted 7 files");
 
-            result.AssertOutput("Bonjour from PreDeploy.ps1");
+            result.AssertOutput("Extracting package to: " + Path.Combine(stagingDirectory, "Acme.Web", "1.0.0"));
+
+            if (CalamariEnvironment.IsRunningOnNix)
+            {
+                result.AssertOutput("Extracted 9 files");
+                result.AssertOutput("Bonjour from PreDeploy.sh");
+
```
PaulStovell: Whitespace!

source/Calamari.Tests/Calamari.Tests.csproj
zentron: One of the tests themselves directly reference this to pre-package a set of files for "deployment".  The existing tests do this but by referencing back into the packages directly rather than including it as a build dependency.
 suppose we could replace all this with the dlls directly.

PaulStovell:
Looks good to me! Merge when ready

PR #22, opened 2015-06-11
Log the working-directory contents after extracting the cspkg file

mjrichardson:
https://github.com/OctopusDeploy/Issues/issues/1632


mjrichardson:
Will now only log if special variable set, as per Vanessa's suggestion.

PR #25, opened 2015-06-16
Remove ExtractPackageToTemporaryDirectoryConvention for Azure tasks a…

zentron:
…s they will now be provided with the extraction location. This requires a related change in OctopusDeploy to provide said value. Relates to issue Issues#1643


mjrichardson:
We should also remove the final `DeleteStagingDirectoryConvention` from the Azure commands (the entire convention can be removed).

PR #31, opened 2015-07-13
Fix variable substitution under nix when target entries contain path

RyanMelenaNoesis:
Addresses https://github.com/OctopusDeploy/Issues/issues/1706 by:

1. Allowing Octopus Deploy to standardize variable substitution target entries on Windows path delimiter syntax while still allowing file matching to work on nix deployments.
2. Implementing a work around far a [mono bug](https://bugzilla.xamarin.com/show_bug.cgi?id=31885) which was causing FileNotFoundExceptions for variable substitution target entries which contain path.


RyanMelenaNoesis:
Please disregard, replaced with clean PR https://github.com/OctopusDeploy/Calamari/pull/32

PR #32, opened 2015-07-13
Fix variable substitution under nix when target entries contain path

RyanMelenaNoesis:
Addresses https://github.com/OctopusDeploy/Issues/issues/1706 by:

1. Allowing Octopus Deploy to standardize variable substitution target entries on Windows path delimiter syntax while still allowing file matching to work on nix deployments.
2. Implementing a work around far a [mono bug](https://bugzilla.xamarin.com/show_bug.cgi?id=31885) which was causing FileNotFoundExceptions for variable substitution target entries which contain path


zentron:
Thanks for this @RyanMelenaNoesis, So that we can give this code a test run in both Linux & Windows environments could I get you to put together a few tests exercising these code changes. There might have been none around this code but this sounds like a good place to start. :smile:

PaulStovell:
This is for OctopusDeploy/Issues#1706

RyanMelenaNoesis:
Added a unit test for relative path variable substitution.  I believe it should cover the described situation.

zentron:
Thanks for the tests @RyanMelenaNoesis . 
The Mono bugfix outlined in #2 will be merged into master branch but the changes made for #1 will not get included.
It's been decided that in order to keep things predictable the user can continue to enter the paths in *nix or windows format explicitly. A comment will be included in the UI to ensure the user understands this.

RyanMelenaNoesis:
@zentron, just to clarify the pull request as written does not force users to utilize the Windows path delimiter syntax in the UI.  It simply converts any Windows path delimiters to Nix delimiters while running the variable substitution.

zentron:
That's correct, we don't want to convert anything from what the user has explicitly provided. Our concern is that this could lead to confusion when we replace characters assuming that we know what the user wanted. 
Consider if the user path includes 2 windows path separators. This in Linux is a valid path but converting them both to Linux path separators would not map to the correct location. As the windows path separator can also be used for escaping characters this could lead to other issues.
see posts such as http://programmers.stackexchange.com/questions/245156/is-it-safe-to-convert-windows-file-paths-to-unix-file-paths-with-a-simple-replac
for more details.
We would prefer the code to be predictable and explicit where possible.

RyanMelenaNoesis:
@zentron, Any solution for those of us that may be deploying the same release to both Windows and Nix systems?

zentron:
@RyanMelenaNoesis at the moment we would suggest either scoping a variable, or including a PreDeploy script to modify the variables required.

RyanMelenaNoesis:
@zentron, the point you raised about Windows delimiters as legitimate characters in Nix paths is a good one and I agree that my naive replacement could cause problems.  As to your suggested solutions, is it possible to dynamically define/modify the substitution targets as part of the PreDeploy?  Thanks.

zentron:
I think the best thing you could do in this particular case @RyanMelenaNoes is feature detection rather than explicit target filtering. 
You may even be able to do the substitution in a C# script which would allow you to leverage the Environment variables.
e.g.
```
if(System.Environment.OSVersion.Platform == PlatformID.MacOSX ||
                       System.Environment.OSVersion.Platform == PlatformID.Unix){
... change or set path variables ...
}
```

zentron:
Contents of PR merged into Calamari bc9d7d91cdf35d100a1d7101cf74726bddee125b

RyanMelenaNoesis:
Thanks @zentron

PR #34, opened 2015-07-21
Include full stack trace in StdErr logging from Calamari

IanMercer:
By logging the exception stacktrace to StdOut but not StdErr we are not getting it in customer reports like http://community.octopusdeploy.com/t/error-object-reference-not-set-to-an-instance-of-an-object-on-nuget-deployment-to-iis/418

This change gets the full stacktrace from exceptions that happen in Calamari

[Fixes OctopusDeploy/Issues/issues/1731]

### DIFF
--- a/source/Calamari/Commands/Support/ConsoleFormatter.cs
+++ b/source/Calamari/Commands/Support/ConsoleFormatter.cs
@@ -34,8 +34,7 @@ namespace Calamari.Commands.Support
                 return 43;
             }
 
-            Log.Error(ex.Message);
-            Log.Info(ex.ToString());
+            Log.Error(ex.ToString());
             return 100;
         }
     }


michaelnoonan:
Hi Ian, I'd like to try and reproduce this locally because it looks to @mjrichardson and I that we should be getting the `Exception.Message` in `Log.Error`, and `Exception.ToString()` in `Log.Info` which should be visible in the ServerTask logs... I'm likely missing something.

michaelnoonan:
Fixes OctopusDeploy/Issues#1731

michaelnoonan:
Seems like there was no valuable reason to log the Message @ Error and the Exception.ToString() @ Info. Putting the whole thing @ Error seems fine.

This should only happen if there's an unhandled Exception or downloading Nuget Packages.

PR #39, opened 2015-08-06
Allow existing files to be preserved when deploying Azure Web App

mjrichardson:
Added `Octopus.Action.Azure.RemoveAdditionalFiles` and `Octopus.Action.Azure.PreserveAppData` variables.
Connects to OctopusDeploy/Issues#1852


michaelnoonan:
Just come comments inline, but it looks good to me. If that's what the StepTemplate is implementing I'd suggest we're best keeping it the same.

mjrichardson:
I renamed the rules as suggested.

PR #41, opened 2015-08-06
Filter Packages by semver + package + top 5 before calculating hash.

zentron:
Also removed a bit of useless code.

Fixes OctopusDeploy/Issues#1859

### DIFF
--- a/source/Calamari/Integration/FileSystem/PackageStore.cs
+++ b/source/Calamari/Integration/FileSystem/PackageStore.cs
@@ -77,7 +77,7 @@ namespace Calamari.Integration.FileSystem
             var zipPackages =
                 from filePath in fileSystem.EnumerateFilesRecursively(rootDirectory, packageId + "*.nupkg-*")
                 let zip = ReadZipPackage(filePath)
-                where zip != null && zip.Id == packageId && zip.Version < version
+                where zip != null && zip.Id == packageId && zip.Version <= version
                 orderby zip.Version descending
                 select new {zip, filePath};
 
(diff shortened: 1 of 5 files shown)


source/Calamari/Integration/FileSystem/PackageStore.cs
```diff
@@ -108,57 +72,56 @@ public StoredPackage GetPackage(string prefix, PackageMetadata metadata)
...
-            fileSystem.EnsureDirectoryExists(root);
+            fileSystem.EnsureDirectoryExists(rootDirectory);
 
-            return (
-                from file in fileSystem.EnumerateFilesRecursively(root, packageId + "*.nupkg-*")
-                let package = ReadPackageFile(file)
+            var zipPackages =
+                from filePath in fileSystem.EnumerateFilesRecursively(rootDirectory, packageId + "*.nupkg-*")
+                let zip = ReadZipPackage(filePath)
+                where zip != null && zip.Id == packageId && zip.Version < version
```
IanMercer: Should this be `< version` or `<= version`? Wouldn't an identical package version be *nearest* the one I want?

zentron: While in some sense its technically correct, it looks like the way this is used is it checks for exact match first (using `GetPackage` method), and then assuming no match it then does this call to find the _nearest_ match. This explains why no-one has brought this up as a bug before.

IanMercer: Maybe worth a <remark> on the method to say that as it's non-intuitive, or just make it `<=` so the method name isn't misleading.

PR #43, opened 2015-08-07
Fix Mono FileNotFoundException bug in Directory.EnumerateFiles

zentron:
Fix for [mono bug](https://bugzilla.xamarin.com/show_bug.cgi?id=31885) which was causing FileNotFoundExceptions for variable substitution target entries which contain path

Merged branch from @RyanMelenaNoesis
Part Fixes OctopusDeploy/Issues#1706


source/Calamari/Integration/FileSystem/CalamariPhysicalFileSystem.cs
```diff
@@ -125,18 +125,22 @@ public void DeleteDirectory(string path, DeletionOptions options)
...
-        public IEnumerable<string> EnumerateFiles(string parentDirectoryPath, params string[] searchPatterns)
+        public virtual IEnumerable<string> EnumerateFiles(string parentDirectoryPath, params string[] searchPatterns)
         {
+            var parentDirectoryInfo = new DirectoryInfo(parentDirectoryPath);
+
             return searchPatterns.Length == 0
-                ? Directory.EnumerateFiles(parentDirectoryPath, "*", SearchOption.TopDirectoryOnly)
-                : searchPatterns.SelectMany(pattern => Directory.EnumerateFiles(parentDirectoryPath, pattern, SearchOption.TopDirectoryOnly));
+                ? parentDirectoryInfo.GetFiles("*", SearchOption.TopDirectoryOnly).Select(fi => fi.FullName)
+                : searchPatterns.SelectMany(pattern => parentDirectoryInfo.GetFiles(pattern, SearchOption.TopDirectoryOnly).Select(fi => fi.FullName));
```
IanMercer: Should this have a `.Distinct()` on it? As it stands if I supplied multiple search patterns that overlapped wouldn't I get duplicate results back? [and same below] [not a new issue].

zentron: Good point. i suppose it _could_ lead to problems depending on how its used.

PR #44, opened 2015-08-11
Remove Azure cert and context script after executing PS script so tha…

zentron:
…t it doesn't get pushed up with the deployment

Fixes OctopusDeploy/Issues#1856


PaulStovell:
I must be blind, where does it get deleted/excluded?

zentron:
AzurePowershellContext.cs ln43/44 uses TemporaryFile to force it to be removed after the using block disposes

mjrichardson:
Yup. I like it.

PR #45, opened 2015-08-13
Split Azure functionality into Calamari.Azure executable

mjrichardson:
Because we package the Azure PowerShell modules, with the latest version of Azure PowerShell the size of the Calamari package exceeded 50MB.  The majority of this (>40MB) is Azure-related DLLs, which were being pushed to all Tentacles (and other targets), but were only ever used on the server (as Azure deployments run on the Octopus server).  So it was decided to split the Azure functionality into a separate executable.

This PR also upgrades the Azure PowerShell module to version `0.9.5`.  Which requires **.NET 4.5**.

Closes OctopusDeploy/Issues#1873


source/Calamari.Azure.Tests/Calamari.Azure.Tests.csproj
```diff
@@ -0,0 +1,176 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
mjrichardson: New `Calamari.Azure` project created

source/Calamari.Azure.Tests/CloudServicePackage/ManifestSchema/AzureCloudServicePackageManifestSchemaFixture.cs
```diff
@@ -1,14 +1,13 @@
 ﻿using System;
 using System.IO;
-using System.Text;
 using System.Xml;
 using System.Xml.Linq;
 using System.Xml.Schema;
-using Calamari.Integration.Azure.CloudServicePackage.ManifestSchema;
```
mjrichardson: Loads of namespace changes...

PR #48, opened 2015-08-14
Change Calamari file system retry logic to a maximum time not a number of retries

IanMercer:
When a service shuts down, even after it moves to the stopped status the process running it may hang on to files for ~30s. In some cases that process may host multiple services so we cannot wait for the process to exit. Moving the retry logic to time-based and ensuring that we give ample time to delete or copy files should solve this problem. Also, no point waiting 1-3 seconds between file system retries, can go faster. Also, log warnings when delete and copy operations have to do retries. Also improves some other aspects of logging.

* Change retry logic to allow a total time limit on retry operations instead of just a count
* Change retry back-off timing logic to allow linear, exponential back-off and retry options
* Separate retry count from failure options
* Add a warning log FeatureScriptConvention so you can see when it didn't overwrite a powershell script being unpacked from embedded resources (maybe you put one there deliberately, or maybe it was left behind by accident)
* Change file system to use new retry logic
* Add a count to file system operation for copying a directory (to improve logging)

Fixes OctopusDeploy/Issues#1892
Release note: Change Calamari file system retry logic to a maximum time not a number of retries


IanMercer:
During retries logging now happens on each retry 1-5 and then every 10s thereafter so as not to spam the log. Something like:

`00:00 Retry #1 on copy ...`
`00:00 Retry #2 on copy ...`
`00:00 Retry #3 on copy ...`
`00:00 Retry #4 on copy ...`
`00:10 Retry #54 on copy ...`
`00:20 Retry #104 on copy ...`
`00:30 Retry #154 on copy ...`

source/Calamari/Integration/FileSystem/CalamariPhysicalFileSystem.cs
```diff
@@ -335,41 +362,46 @@ public void EnsureDirectoryExists(string directoryPath)
...
+                count++;
             }
 
             foreach (var childSourceDirectory in Directory.GetDirectories(sourceDirectory))
             {
                 var name = Path.GetFileName(childSourceDirectory);
                 var childTargetDirectory = Path.Combine(targetDirectory, name);
-                CopyDirectory(childSourceDirectory, childTargetDirectory, cancel, overwriteFileRetryAttempts);
+                count += CopyDirectory(childSourceDirectory, childTargetDirectory, cancel);
             }
```
tothegills: Does the cancellation token need to be checked in this loop?

IanMercer: It could be, but a better place for an additional check might be above `Directory.CreateDirectory`. This loop does no  time consuming work of its own, it will quickly recurse back into the same method and as soon as it hits a file it will stop.

tothegills:
lgtm

PR #51, opened 2015-08-25
Feature crypto var

zentron:
Largely backwards compatible change (apart from expecting offline drops to include explicit encrypted variable file name) to allow providing sensitive variables into calamari.

Internally the sensitive and insensitive variables are combined however when written out to the scripts they are again encrypted in a format that works for .Net library or openssl for decryption.


source/Calamari/Integration/Processes/CalamariVariableDictionary.cs
```diff
@@ -0,0 +1,78 @@
...
+using Calamari.Integration.FileSystem;
+using Calamari.Util;
+using Newtonsoft.Json;
+using Octostache;
+
+namespace Calamari.Integration.Processes
+{
+    public class CalamariVariableDictionary : VariableDictionary
+    {
+        protected List<string> SensitiveVariableNames = new List<string>();
```
PaulStovell: Could this be a HashSet<string> (with a StringComparer.OrdinalIgnoreCase)?

PR #52, opened 2015-08-27
Allow setting physical path when deploying to Azure Web App 

mjrichardson:
Added `Octopus.Action.Azure.PhysicalPath` variable, which is the physical path, relative to the site root.

Connects to OctopusDeploy/Issues#1914


michaelnoonan:
LGTM

PR #55, opened 2015-09-01
Corrected file-system retry handling and attempt to ensure directory deletion.

mjrichardson:
Corrected `CanRetry()` logic.
Implemented `EnsureDirectoryDeleted()` in `CalamariPhysicalFileSystem`.
Connects to OctopusDeploy/Issues#1960

### DIFF
--- a/source/Calamari/Integration/FileSystem/CalamariPhysicalFileSystem.cs
+++ b/source/Calamari/Integration/FileSystem/CalamariPhysicalFileSystem.cs
@@ -87,13 +87,13 @@ namespace Calamari.Integration.FileSystem
                     }
                     break;
                 }
-                catch (Exception)
+                catch (Exception ex)
                 {
                     if (retry.CanRetry())
                     {
                         if (retry.ShouldLogWarning())
                         {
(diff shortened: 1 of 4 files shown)


source/Calamari/Integration/FileSystem/CalamariPhysicalFileSystem.cs
```diff
@@ -152,6 +153,27 @@ public void DeleteDirectory(string path, FailureOptions options)
...
+            while (retry.Try())
+            {
+                if (!Directory.Exists(path))
+                    return;
+
+                if (retry.CanRetry() && retry.ShouldLogWarning())
+                    Log.VerboseFormat("Waiting for directory '{0}' to be deleted", path);
+            }
+
+            var message = $"Unable to ensure directory '{path}' was deleted";
```
michaelnoonan: I'm not sure we were "unable" to, it's probably more a case of "The directory '{path}' was not deleted". What do you think?

michaelnoonan:
Comments inline.

I would really like to see a higher level tests like `AttemptingToCopyOverALockedFile_ShouldRetry_AndThenThrow` and deleting. That way we know not only the retry logic works, but how it is being invoked. It's the true end result.

PR #56, opened 2015-09-02
Preserve IIS bindings

mjrichardson:
Don't clear all IIS bindings. Remove those not required, and add those not present.  Preserve existing bindings.

Closes OctopusDeploy/Issues#1582

### DIFF
--- a/source/Calamari/Scripts/Octopus.Features.IISWebSite_BeforePostDeploy.ps1
+++ b/source/Calamari/Scripts/Octopus.Features.IISWebSite_BeforePostDeploy.ps1
@@ -264,13 +264,36 @@ Execute-WithRetry {
 	Set-ItemProperty $sitePath -name physicalPath -value "$webRoot"
 }
 
+function Bindings-AreEqual($bindingA, $bindingB) {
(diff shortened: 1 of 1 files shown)


tothegills:
Should we remove bindings that are not configured in Octopus? Following the conversation on the ticket is seems like part of the problem is that existing bindings are deleted.

mjrichardson:
The consensus was to remove non-configured bindings.  
I believe the two primary arguments for removing them are:
- consistency with former behaviour
- repeatable (i.e. stateless) deployments

tothegills:
:+1: LGTM

PaulStovell:
If you're confident that it works I'm totally fine with it, but just in case I regret not mentioning it.... :)

I was under the impression we'd do something like this:

```
Concat all expected bindings
Concat all actual bindings
Are there any differences?
If so, clear all bindings and add the expected ones
Otherwise do nothing
```

Are there any risks with the current approach?

mjrichardson:
I think your approach would be more tolerant in some error scenarios.  I'm happy to modify it to work like that.

mjrichardson:
I have modified this to work as per Paul's suggestion

PaulStovell:
:+1: looks good to me

PR #61, opened 2015-10-01
Bring back support for config transform files in a subfolder

alfhenrik:
Add back support for config transforms like `config\foo.bar.config => bar.config`

Wildcard config transforms like `config\*.bar.config => bar.config` is not supported.

Fixes OctopusDeploy/Issues#2032


mjrichardson:
Looks good Henrik.

PR #62, opened 2015-10-15
Using local AzureProfile.json file for Azure deployments

alfhenrik:
Create and use a local AzureProfile.json file to prevent locking issues on the global AzureProfile.json file when deploying multiple Azure steps in parallel, or using a cached certificate that is wrong.

Fixes OctopusDeploy/Issues#2054 **(hopefully, have been unable to replicate the issue reported, so can't confirm changes have fixed it)**
Fixes OctopusDeploy/Issues#2076

### DIFF
--- a/source/Calamari.Azure/Scripts/DeployAzureCloudService.ps1
+++ b/source/Calamari.Azure/Scripts/DeployAzureCloudService.ps1
@@ -26,7 +26,7 @@
 
 function CreateOrUpdate() 
 {
-	Set-AzureSubscription -SubscriptionId $OctopusAzureSubscriptionId -CurrentStorageAccount $OctopusAzureStorageAccountName
+	Set-AzureSubscription -SubscriptionName $OctopusAzureSubscriptionName -CurrentStorageAccount $OctopusAzureStorageAccountName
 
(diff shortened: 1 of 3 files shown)


mjrichardson:
We should probably make `AzureProfile.json` a TemporaryFile, similar to the certificate and script files in `AzurePowerShellContext.cs`.  Otherwise it will be uploaded as part of the deployments.

source/Calamari.Azure/Scripts/DeployAzureCloudService.ps1
```diff
@@ -26,7 +26,10 @@
 
 function CreateOrUpdate() 
 {
-	Set-AzureSubscription -SubscriptionName $OctopusAzureSubscriptionName -CurrentStorageAccount $OctopusAzureStorageAccountName 
+	$azureProfile = New-AzureProfile -Path ".\AzureProfile.json"
```
mjrichardson: I think this can be removed too.

mjrichardson:
I think the `New-AzureProfile` in `DeployAzureCloudService.ps1` can be removed too (see line note above).  
It will be added by the context.

PR #65, opened 2015-11-05
Convert IISWebsite_BeforePostDeploy script to allow for JSON encoded …

zentron:
…binding information.

Currently If IPv6 ip address is used (or variable with `:` in name), the binding information cannot be parsed. By changing to JSON encoded we can still test for the old method and extract the information in a backwards compatable way.

Connected to OctopusDeploy/Issues#1738

### DIFF
--- a/source/Calamari/Scripts/Octopus.Features.IISWebSite_BeforePostDeploy.ps1
+++ b/source/Calamari/Scripts/Octopus.Features.IISWebSite_BeforePostDeploy.ps1
@@ -81,32 +81,79 @@ $supportsSNI = $iisVersion -ge 8
 
 $wsbindings = new-object System.Collections.ArrayList
 
-# Each binding string consists of a protocol/binding information (IP, port, hostname)/SSL thumbprint/enabled/sslFlags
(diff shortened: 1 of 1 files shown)


source/Calamari/Scripts/Octopus.Features.IISWebSite_BeforePostDeploy.ps1
```diff
@@ -81,32 +81,79 @@ $supportsSNI = $iisVersion -ge 8
...
-		if (![String]::IsNullOrEmpty($bindingParts[3]) -and [Bool]::Parse($bindingParts[3]) -eq $false) {
-			$skip = $true
+Write-Output $bindingString
+
+if($bindingString.StartsWith("[{")) {
+	
+	if(Get-Command ConvertFrom-Json -errorAction SilentlyContinue){
+		$bindingArray = (ConvertFrom-Json $bindingString)
+	} else {
+		add-type -assembly system.web.extensions
```
zentron: `ConvertFrom-Json` is only in PS 3 but would be nice to not have to rely on `system.web.extensions` so fallback to this only if in PS 2 or less.

PR #66, opened 2015-11-09
Check if already using the right app pool before trying to set it.

bzbetty:
This should allow calamari to run using an account with less permissions when changes are not needed.

As per forum thread http://help.octopusdeploy.com/discussions/problems/41960-using-octopus-in-an-environment-with-cutdown-security-permissions


zentron:
Thanks for the PR. This will go out sometime in the next release.

PR #69, opened 2015-11-23
New convention for appSettings.json generation

PaulStovell:
To support ASP.NET 5 we will have a new convention to generate AppSettings.json files. Here's what the UI in Octopus will look like:

screen shot 2015-11-23 at 3 48 50 pm

### DIFF
--- a/source/Calamari/Commands/DeployPackageCommand.cs
+++ b/source/Calamari/Commands/DeployPackageCommand.cs
@@ -6,6 +6,7 @@ using Calamari.Commands.Support;
 using Calamari.Deployment;
 using Calamari.Deployment.Conventions;
 using Calamari.Deployment.Journal;
+using Calamari.Integration.AppSettingsJson;
 using Calamari.Integration.ConfigurationTransforms;
 using Calamari.Integration.ConfigurationVariables;
 using Calamari.Integration.EmbeddedResources;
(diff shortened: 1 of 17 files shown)


mjrichardson:
I like it. It all makes sense to me.

source/Calamari/Integration/AppSettingsJson/AppSettingsJsonGenerator.cs
```diff
@@ -0,0 +1,98 @@
...
+        public void Generate(string appSettingsFilePath, VariableDictionary variables)
+        {
+            var root = LoadJson(appSettingsFilePath);
+
+            var names = variables.GetNames();
+            names.Sort(StringComparer.OrdinalIgnoreCase);
+
+            foreach (var name in names)
+            {
+                if (name.StartsWith("Octopus", StringComparison.OrdinalIgnoreCase))
```
zentron: I'm still not entirely convinced that we want to dump _all_ variables to json. but for want of a better process then this will do.

michaelnoonan:
LGTM also. I like that we're dealing with the ambiguous cases sensibly. Is it worth a warning in this case though? I would want to fix the ambiguity if I was the user.

source/Calamari/Commands/DeployPackageCommand.cs
```diff
@@ -52,6 +53,7 @@ public override int Execute(string[] commandLineArguments)
             var variables = new CalamariVariableDictionary(variablesFile, sensitiveVariablesFile, sensitiveVariablesPassword);
             var scriptCapability = new CombinedScriptEngine();
             var replacer = new ConfigurationVariablesReplacer(variables.GetFlag(SpecialVariables.Package.IgnoreVariableReplacementErrors));
+            var generator = new AppSettingsJsonGenerator();
```
zentron: Are we sure we don't want to support `ini` files? :)

zentron:
Seems like a nice solution to vNext :).

PaulStovell:
@michaelnoonan done, here's the warning:

> Unable to set value for `EmailSettings:DefaultRecipients:To`. The property at `EmailSettings.DefaultRecipients` is a `String`.

PR #72, opened 2015-11-26
If a config transform rule is not performed then a verbose log entry …

zentron:
…is added. The choice for verbose over warning or error stems from the fact that the non-application of any specific rule may actually result from an overlapping rule and could generate false positives. A verbose entry is still more informative than no log at all.

Fixes OctopusDeploy/Issues#2091

image


michaelnoonan:
Looks good. I like the choice to go Verbose - I've seen lots of false-positives over the years.

PR #74, opened 2015-12-01
Feature nonnupkg

zentron:
`LightweightPackageExtractor` is now `OpenPackagingConventionExtractor` which is utilized through the `GenericPackageExtractor` whos job it is to defer the extraction to the appropriate class depending on format.

To support work that needs to be done on the server, the existing packages file extensions will be included in the `FindPackage` command.

Connects to OctopusDeploy/Issues#2183


.gitignore
```diff
@@ -133,6 +133,8 @@ publish/
 
 # NuGet Packages
 *.nupkg
+!**/*/Acme.Core.1.0.0.0-bugfix.nupkg
```
PaulStovell: Seems awfully specific :)

source/Calamari.Tests/Fixtures/Integration/Packages/GenericPackageExtractorFixture.cs
```diff
@@ -0,0 +1,49 @@
...
+
+
+        [Test]
+        [ExpectedException(typeof(FileFormatException), ExpectedMessage = "Package is missing file extension. This is needed to select the correct extraction algorithm.")]
+        public void FileWithNoExtensionThrowsError()
+        {
+            extractor.GetExtractor("blah");
+        }
+
+
```
PaulStovell: Too much whitespace! :)

source/Calamari/Integration/Packages/SimplePackageExtractor.cs
```diff
@@ -0,0 +1,86 @@
...
+            var metaDataSection = string.Empty;
+            if (matchingExtension != null)
+            {
+                metaDataSection = fileName.Substring(0, fileName.Length - matchingExtension.Length);
+            }
+            else
+            {
+                foreach (var ext in Extensions)
+                {
+                    var match = new Regex("(?<extension>" + ext.Replace(".", "\\.") + ")-[a-z0-9\\-]*$").Match(fileName);
```
PaulStovell: I think `Regex.Escape` might be better than `.Replace(".",...)` -- but then maybe string.IndexOf() >= 0 might work better too?

.gitignore
zentron: :) Just a resource used for a test, didn't see any reason to go more generic.

PR #75, opened 2015-12-14
Allow passing `AppOffline` special variable to set the AppOffline dep…

zentron:
…loyment rule when deploying Azure applications.

Solves: OctopusDeploy/Issues#2201


michaelnoonan:
LGTM - Nice pickup on the Azure support.

PR #79, opened 2016-01-04
Fun with config transform paths

tothegills:
This PR allows the specification of config transform rules when then config file is in a relative path.  For example:
```
Web.Base.config => config\octopus\Web.Config
```

Also enables wildcard transforms with relative paths:
```
transform\*.config => *.config
*.config => config\*.config
transform\*.config => config\*.config
```

And finally changes behaviour when there is a rule with an explicit and wildcard definition:
```
*.foo.config => bar.config
```
Will apply `abc.foo.config` and `xyz.foo.config` to `bar.config` instead of `abc.foo.config` to `abc.bar.config` and `xyz.foo.config` to `xyz.bar.config`.  If the latter case is desired it can be achieved with:
```
*.foo.config => *.bar.config
```


MJRichardson:
Very nice.

PR #80, opened 2016-01-05
Make sure we fail early if we can't load the WebAdministration PS module.

alfhenrik:
Connected to OctopusDeploy/Issues#2263
Connected to OctopusDeploy/Issues#2265


robpearson:
Nice error message! Change :+1:

PR #81, opened 2016-01-05
Fix display issue in PS 2.0 when logging IIS bindings.

alfhenrik:
Connected to OctopusDeploy/Issues#2268


robpearson:
LGTM! :+1:

PR #82, opened 2016-01-13
Allow a transform to be applied to multiple target files

tothegills:
Previously we only allowed a transform to be applied to a single target file.  This PR allows a transform to be applied to multiple target files, either through explicit definition or wildcards. eg;
```
transform.config => foo.config
transform.config => bar.config
```
or
```
transform.config => *.config
```

Relates to OctopusDeploy/Issues#2286


alfhenrik:
LGTM :+1:

PR #87, opened 2016-02-08
Display a helpful error when JSON config being applied to a folder.

tothegills:
image

Fixes OctopusDeploy/Issues#2354


michaelnoonan:
That looks good to me. Much better error than something that looks all the world like permissions.

PR #89, opened 2016-02-11
Generalise AppSettings JSON feature to a more JSON file agnostic feature

alfhenrik:
We're refactoring this feature to be more like the configuration variables feature where we match and replace (in a case-insensitive way) only variables that are already in the JSON file. Still keeping the functionality to allow for config nesting with variables in the format of `Foo:Bar:Baz` to replace the value for `Baz` in a JSON file looking like:
```
{
  "Foo": {
    "Bar" : {
      "Baz": "BazValue"
    }
  }
}
```

Goes hand in hand with UI changes being made in OctopusDeploy/Issues#2363

Connects to OctopusDeploy/Issues#2363
Connects to OctopusDeploy/Issues#2364
Connects to OctopusDeploy/Issues#2346


MJRichardson:
Looks good to me.

PR #92, opened 2016-02-26
Ensure the ScriptModules are written after Variables.

michaelnoonan:
Some customers write immediately-executing script inside their Script Modules that depend on Variables. However from Octopus 2.x -> 3.x we'd changed the ordering. This change ensures the Script Modules are written after the Variables. There should be no negative side-effects of this change, unless both Script Modules and Variables were declaring colliding local PowerShell variables.

Fixes OctopusDeploy/Issues#2395


alfhenrik:
LGTM :+1:

PR #94, opened 2016-03-04
Add variable to use the Release Number in the blob storage path instead of the nuget package version

jkschanke:
Related to http://help.octopusdeploy.com/discussions/questions/7226-is-there-a-way-to-change-override-the-package-name-when-uploading-to-azure-blob-storage

### DIFF
--- a/source/Calamari.Azure.Tests/Conventions/UploadAzureCloudServicePackageConventionFixture.cs
+++ b/source/Calamari.Azure.Tests/Conventions/UploadAzureCloudServicePackageConventionFixture.cs
@@ -73,5 +73,32 @@ namespace Calamari.Azure.Tests.Conventions
 
             Assert.AreEqual(uploadedUri.ToString(), variables.Get(SpecialVariables.Action.Azure.UploadedPackageUri));
         }
+
+        [Test]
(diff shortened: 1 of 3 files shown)


jkschanke:
I'm not entirely happen with the variable name, but it's the best I could think of.

michaelnoonan:
Hi @jkschanke, thanks for the PR! It looks pretty good, though with this kind of functionality, I would be inclined to let people completely customize the name, but I can see how you'd lose the Hash feature without doing some gymnastics.

@MJRichardson what are your thoughts on this?

zentron:
Thanks for the PR @jkschanke.
 Unfortunately after some discussion with other members of the team we have decided that the problem presented and solution proposed in this ticket is very user-specific and unlikely to be something that proves useful for other users.
In a similar vein to our standard [why was this closed](https://octopus.com/company/roadmap/why-was-this-closed) outline for tickets lodged against the main Octopus Deploy repository, it is our thought that this _suggestion is unique to your individual situation_ and as a result is not something that we should necessarily support going forward.
If you feel strongly about this feature please feel free to lodge a [UserVoice](https://octopusdeploy.uservoice.com/) ticket so that we can better gauge if this is something that has some demand.

PR #100, opened 2016-05-03
Allow semaphore access to everyone

tothegills:
Semaphore permissions are for current user (or admin).  When Calamari is run as different users and both users try to acquire the same semaphore an exception is thrown because one will not have permissions to access the semaphore.

This adds semaphore control to "Everyone".

Relates to OctopusDeploy/Issues#2485


michaelnoonan:
Good pickup on the "Everyone" (localisation wreaks havok with these)

LGTM

PR #104, opened 2016-06-13
Add support for deploying ASP.NET5 (Core RC1) apps OOTB

alfhenrik:
image


pawelpabich:
Won't be merging as RTM changed things a bit and is coming soon :pray:

PR #110, opened 2016-07-08
Filter Azure Web Apps by ResourceGroupName

slewis74:
Include ResourceGroupName when locating Azure Web Apps, to cater for ASEs using the same WebAppName.

Connects to OctopusDeploy/Issues#2554


MJRichardson:
Looks good to me.

PR #113, opened 2016-07-18
[WIP] Prevent downloading the same package version on a machine

alfhenrik:
Connects to OctopusDeploy/Issues#2597 
Goes hand-in-hand with OctopusDeploy/OctopusDeploy#552


alfhenrik:
This change is not required as it's all handled server side.

PR #119, opened 2016-09-01
Unescape zip entry before using it

tothegills:
When extracting files with unicode characters in their name we were attempting to extract the escaped file name (such as 21113.%D0%9A%D0%BB%D0%B0%D1%81%D1%81%D0%B8%D1%84%D0%B8%D0%BA%D0%B0%D1%82%D0%BE%D1%80%20%D0%B2%D0%B8%D0%B4%D0%BE%D0%B2%20%D0%B4%D0%BE%D0%BA%D1%83%D0%BC%D0%B5%D0%BD%D1%82%D0%BE%D0%B2%2C%20%D1%83%D0%B4%D0%BE%D1%81%D1%82%D0%BE%D0%B2%D0%B5%D1%80%D1%8F%D1%8E%D1%89%).  This PR escapes the file name before using it.

Relates to OctopusDeploy/Issues#2687

### DIFF
--- a/source/Calamari/Integration/Packages/NuGet/NupkgExtractor.cs
+++ b/source/Calamari/Integration/Packages/NuGet/NupkgExtractor.cs
@@ -37,10 +37,12 @@ namespace Calamari.Integration.Packages.NuGet
             {
                 while (reader.MoveToNextEntry())
                 {
-                    if (IsExcludedPath(reader.Entry.Key))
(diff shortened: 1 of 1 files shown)


michaelnoonan:
@MJRichardson if I understand this correctly, the code was ported from NuGet itself? Would they have the same problem? If so I wonder if we should offer a PR?

Pliner:
:+1:

PR #121, opened 2016-09-28
PurgeDirectory can now also check if a directory should be excluded.

dlemstra:
This patch changes the `FileInfoAdapter` to use `FileSystemInfo` instead of `FileInfo` so it can also be used with directories. I also think the name of `FileInfoAdapter` and `IFileInfo` should also be changed to `FileSystemInfoAdapter` and `IFileSystemInfo` but I did not include that in this pull request because you might not be happy with this breaking change. I can add another commit to rename those classes if you are okay with me renaming them. The `PurgeDirectory` method from `CalamariPhysicalFileSystem` has also been changed to check if the directory should be included or excluded.

The reason that I made these changes is because I would like to add an extra option to `CopyPackageToCustomInstallationDirectoryConvention` that would make it possible to exclude a set of files or directories when the `customInstallationDirectory` is purged. I already made a proof of concept for this but I would also like to know if you would be open to adding such a feature to the interface. I can send another pull request if you would be okay with me adding this option to Calamari.

### DIFF
--- a/source/Calamari/Integration/FileSystem/CalamariPhysicalFileSystem.cs
+++ b/source/Calamari/Integration/FileSystem/CalamariPhysicalFileSystem.cs
@@ -342,8 +342,8 @@ namespace Calamari.Integration.FileSystem
 
                 if (include != null)
                 {
-                    var info = new FileInfoAdapter(new FileInfo(file));
-                    if (!include(info))
+                    var includeInfo = new FileInfoAdapter(new FileInfo(file));
+                    if (!include(includeInfo))
                     {
                         continue;
                     }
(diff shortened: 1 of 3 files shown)


zentron:
Thanks for the PR @dlemstra, our apologies for the delay in replying back to you on this one. 
This does sound like it could be a useful feature and we already offer similar functionality for Azure deployments so looks like it could be worth merging in. 
I'll update you with its progress.

zentron:
@dlemstra I have added GitHub Ticket https://github.com/OctopusDeploy/Issues/issues/2943 to track the UI changes that need to be done to pass these values through. 
Thanks again.

dlemstra:
It has been a while since I last took a look at this but I will try to work on the other pull request this weekend.

zentron:
We will likely be hooking it up to the UI as part of the aforementioned github ticket so feel free to just subscribe to notifications for its progress.

PR #122, opened 2016-09-30
Throw an error if we can't evaluate the custom directory or it is a relative path

alfhenrik:
Using an incorrect variable:
image

Using a relative path:
image

Fixes OctopusDeploy/Issues#2763


alfhenrik:
Yep, just about to do that 👍

PR #124, opened 2016-10-05
Enable more mono tests

matt-richardson:
The bug in mono that stopped us using XML transforms has been fixed somewhere between 4.0 and 4.2.3, so I've re-enabled the tests to ensure we have coverage under mono.

Documentation (http://docs.octopusdeploy.com/display/OD/SSH+Targets) has been updated.


matt-richardson:
Feel free to sqaush the commits when this is merged.

PR #130, opened 2016-10-19
Enhance the IIS Step

michaelnoonan:
Relates to OctopusDeploy/Issues#2840
Relates to OctopusDeploy/Issues#2839
Relates to OctopusDeploy/Issues#2846

- [x] Make the use of WebSite clearer - create or assert

The Web Site configuration of this step will create the Site if missing, but the Virtual Directory and Web Application configurations expect the parent site to already exist.

image

- [x] Provide better collision detection, conversion, and messaging

This code detects misconfigured and duplicate items at the same path in IIS and will try to convert an existing directory to either a Virtual Directory or Web Application. If something goes wrong we now try to provide all the information required to fix the problem.

image

image

### DIFF
--- a/source/Calamari/Scripts/Octopus.Features.IISWebSite_BeforePostDeploy.ps1
+++ b/source/Calamari/Scripts/Octopus.Features.IISWebSite_BeforePostDeploy.ps1
@@ -198,6 +198,10 @@ function Set-Path($virtualPath, $physicalPath)
 	}
 }
 
+function Is-Directory($Path){
+	return Test-Path -Path $Path -PathType Container
+}
+
 if ($deployAsVirtualDirectory) 
 {
 	$webSiteName = $OctopusParameters["Octopus.Action.IISWebSite.VirtualDirectory.WebSiteName"]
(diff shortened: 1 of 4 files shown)


source/Calamari/Scripts/Octopus.Features.IISWebSite_BeforePostDeploy.ps1
```diff
@@ -220,16 +219,19 @@ if ($deployAsVirtualDirectory)
...
+			# It looks like the only reliable way to do the conversion is to delete the exsting application and then create a new virtual directory. http://stackoverflow.com/questions/16738995/powershell-convertto-webapplication-on-iis
+			# We don't want to delete anything as the customer might have handcrafted the settings and has no way of retrieving them.
+			throw "`"$virtualPath`" already exists in IIS and points to a Web Application. We cannot automatically change this to a Virtual Directory on your behalf. Please delete it and then re-deploy the project."
 		} else {
-			Write-Host "Virtual Directory `"$virtualPath`" already exists."
+			throw "`"$virtualPath`" already exists in IIS and points to an unknown item, perhaps a directory or file with the same path. We cannot automatically change this to a Virtual Directory on your behalf. Please delete it and then re-deploy the project. If you used the Custom Installation Directory feature to target this path we recommend removing the Custom Installation Directory feature, instead allowing Octopus to unpack the files into the default location and update the Physical Path of the Virtual Directory on your behalf."
```
pawelpabich: Not sure whether we can do it but if we can than it would be great if we could specify the physical path here so the user knows exactly where there unknown item is.

pawelpabich:
Love the messages. Very informative.

source/Calamari/Scripts/Octopus.Features.IISWebSite_BeforePostDeploy.ps1
michaelnoonan: I think at this point it's better they know where the path in IIS is, and then they can figure out what to do by following the mapping in IIS...

PR #132, opened 2016-10-21
Convert existing physical path into WebApp when deploying package as WebApp

bastien-hofmann:
Following discussion here:

http://help.octopusdeploy.com/discussions/problems/47717-cant-add-a-web-applicaiton

This change converts an existing physical path (existing because the package was deployed in it) that is seen as a virtual path (because it is contained in a web site or virtual directory) into a WebApp.


michaelnoonan:
HI @bastien-hofmann - thanks for the PR! We've implemented a similar approach over here https://github.com/OctopusDeploy/Calamari/pull/130. What do you think?

bastien-hofmann:
Hello @michaelnoonan,

 #130 seems great thanks, we'll test it as soon as we can !

michaelnoonan:
Great, let me know if you find any issues! I'll close out this PR now.

bastien-hofmann:
Our issue is solved with Octopus 3.4.14, thanks a lot !

PR #133, opened 2016-10-24
Detect if IIS Windows Feature is installed

michaelnoonan:
This avoids an ugly COM error when IIS is not installed on the server.

`Get-WindowsFeature` and `Install-WindowsFeature` were available since Windows Server 2008 so this should work on all supported platforms.


matt-richardson:
Should this fail the build, or just be a warning?

matt-richardson:
Task summary:
image

Task log:
image

PR #137, opened 2016-11-23
Adding environment information logging

thebigmoosey:
Adding environment information logs when Calamari executes. Each script bootstrapper (powershell, cs, bash and f#) also logs environment information to help with debugging support requests.

Relates to #2887 (will have a separate OctopusDeploy PR to include in the release notes)


source/Calamari/Integration/Scripting/Bash/BashScriptBootstrapper.cs
```diff
@@ -47,6 +49,18 @@ public static string PrepareConfigurationFile(string workingDirectory, CalamariV
             return configurationFile;
         }
 
+        static string LogEnvironmentInformation()
+        {
+            var environmentInformationStamp = $"Bash Environment Information:{Environment.NewLine}" +
+                $"  {string.Join($"{Environment.NewLine}  ", EnvironmentHelper.SafelyGetEnvironmentInformation())}";
```
michaelnoonan: I think this is not quite what we want... This will tell the bash script to log the context  `Calamari.exe` is currently running in, not necessarily the environment the actual bash script is running in. I could be wrong though!

thebigmoosey:
This now logs environment information from each script-bootstrapper (each calculating their own info) and you have an option to suppress this logging by setting "Octopus.Action.Script.SuppressEnvironmentLogging" to "True".

There's exception handling (where possible) and fall-through logic so we always get some environment logging happening.

Bash was the only edge case where I couldn't do exception handling that would silence the errors, hence the introduction of the "Octopus.Action.Script.SuppressEnvironmentLogging" variable, because this new logging has the 'potential' to break deployments (Eg. if some custom linux distribution doesn't have the expected variables we're expecting, it'll error ... so they can use this variable to suppress the logging and continue silently).

source/Calamari/Util/Environments/EnvironmentHelper.cs
```diff
@@ -0,0 +1,88 @@
...
+            {
+                envVars.Add($"HostProcessName: {Process.GetCurrentProcess().ProcessName}");
+            }
+            catch
+            {
+                // silently fail.
+            }
+        }
+
+        static void SafelyAddComputerInfoVarsToList(ref List<string> envVars)
```
michaelnoonan: We really shouldn't use `ref`, pretty much ever. I think remove these... but a better long-term change would be to make this method, and its brothers/sisters into:

```
static IEnumerable<string> SafelyGetComputerInfo()
{
  try
  {
    var computerInfo = new ComputerInfo();
    yield return $"TotalPhysicalMemory: {computerInfo.TotalPhysicalMemory.ToFileSizeString()}"
    yield return $"AvailablePhysicalMemory: {computerInfo.AvailablePhysicalMemory.ToFileSizeString()}"
  }
  catch
  {
    // silently fail
  }
}
```

source/Calamari/Util/Environments/EnvironmentHelper.cs
```diff
@@ -0,0 +1,88 @@
...
+using Microsoft.VisualBasic.Devices;
+#endif
+
+namespace Calamari.Util.Environments /* This is in the 'Environments' namespace to avoid collisions with CrossPlatformExtensions - long story */
+{
+    public static class EnvironmentHelper
+    {
+        public static string[] SafelyGetEnvironmentInformation()
+        {
+            var envVars = new List<string>();
```
michaelnoonan: If you make all these methods return `IEnumerable<string>` you can do this:
```
  var envVars = SafelyGetEnvironmentVars()
    .Concat(SafeflyGetPathVars())
   ...etc
```

source/Calamari/Integration/Scripting/WindowsPowerShell/Bootstrap.ps1
```diff
@@ -12,46 +12,137 @@ $ErrorActionPreference = 'Stop'
...
 
+function Log-EnvironmentInformation
+{
+	if ($OctopusParameters.ContainsKey("Octopus.Action.Script.SuppressEnvironmentLogging")) {
+		if ($OctopusParameters["Octopus.Action.Script.SuppressEnvironmentLogging"] -eq "True") {
+			return;
+		}
+	}
+
+	Write-Host "##octopus[stdout-verbose]"
```
michaelnoonan: To be consistent, should all of these be `Write-Verbose` instead of `Write-Host` wrapped in the service message? I'm sure there's a good reason I'm missing here...

Hmm, I guess if any of those calls write to the host without you realising, they'll all end up at Verbose - so maybe this is a good approach. :)

source/Calamari/Integration/Scripting/ScriptCS/ScriptCSBootstrapper.cs
```diff
@@ -73,7 +72,7 @@ public static string PrepareConfigurationFile(string workingDirectory, CalamariV
             var configurationFile = Path.Combine(workingDirectory, "Configure." + Guid.NewGuid().ToString().Substring(10) + ".csx");
 
             var builder = new StringBuilder(BootstrapScriptTemplate);
-            builder.Replace("{{VariableDeclarations}}", WriteVariableDictionary(variables));
+            builder.Replace("/*{{VariableDeclarations}}*/", WriteVariableDictionary(variables));
```
michaelnoonan: Is this an accidental change?

source/Calamari/Integration/Scripting/FSharp/FSharpBootstrapper.cs
```diff
@@ -65,7 +65,7 @@ public static string PrepareConfigurationFile(string workingDirectory, CalamariV
             var configurationFile = Path.Combine(workingDirectory, "Configure." + Guid.NewGuid().ToString().Substring(10) + ".fsx");
 
             var builder = new StringBuilder(BootstrapScriptTemplate);
-            builder.Replace("{{VariableDeclarations}}", WritePatternMatching(variables));
+            builder.Replace("(*{{VariableDeclarations}}*)", WritePatternMatching(variables));
```
michaelnoonan: Is this an accidental change?

source/Calamari/Integration/Scripting/FSharp/Bootstrap.fsx
```diff
@@ -40,7 +42,7 @@ let private decryptString encrypted iv =
 
 let tryFindVariable name =
     match name |> encode with
-{{VariableDeclarations}}
```
michaelnoonan: Is this an accidental change?

source/Calamari/Integration/Scripting/FSharp/Bootstrap.fsx
```diff
@@ -2,10 +2,12 @@
```
michaelnoonan: Would be good if @pawelpabich could review this too. :)

source/Calamari.Tests/project.json
```diff
@@ -60,7 +60,7 @@
          "Polly": "4.3.0" 
        },
       "buildOptions": {
-        "define": [ "NET40", "IIS_SUPPORT", "USE_NUGET_V2_LIBS", "USE_OCTODIFF_EXE" ]
+        "define": [ "NET40", "IIS_SUPPORT", "USE_NUGET_V2_LIBS", "USE_OCTODIFF_EXE", "VISUALBASIC_SUPPORT" ]
```
michaelnoonan: Consider naming this something like `CAN_GET_PHYSICALMEMORY`

source/Calamari/Integration/Scripting/WindowsPowerShell/Bootstrap.ps1
matt-richardson: `Write-Host` is frowned upon generally in PowerShell, as it may not always be redirected.

source/Calamari/Util/Environments/EnvironmentHelper.cs
thebigmoosey: Old habits die hard, thx good catch

source/Calamari/Util/Environments/EnvironmentHelper.cs
thebigmoosey: It won't let you yield return within a try/catch (we had a similar thing in auto-deploy from memory) ... I've nested it as follows to get around it:

```
    static IEnumerable<string> SafelyGetPathVars()
    {
        try
        {
            return GetPathVars();
        }
        catch
        {
            // Fail silently
            return Enumerable.Empty<string>();
        }
    }
    static IEnumerable<string> GetPathVars()
    {
        yield return $"CurrentDirectory: {Directory.GetCurrentDirectory()}";
        yield return $"TempDirectory: {Path.GetTempPath()}";
    }
```

source/Calamari/Integration/Scripting/WindowsPowerShell/Bootstrap.ps1
thebigmoosey: @michaelnoonan Yes, our `Write-Verbose` function simply wraps a call to `Write-Host` with `##octopus[stdout-verbose]` ... 

```
function Write-Verbose([string]$message)
{
	Write-Host "##octopus[stdout-verbose]"
	Write-Host $message
	Write-Host "##octopus[stdout-default]"
}
```
So this ensures they all end up being verbose logs without additional overhead (eg. If you make each of these a `Write-Verbose`, you end up with a lot of additional calls to `Write-Host` for no reason).

@matt-richardson - If you can recommend an alternative we should be using, let us know. This seems to be how we're doing it in all PowerShell scripts in Octopus too (see HealthCheck.ps1).

source/Calamari/Integration/Scripting/FSharp/FSharpBootstrapper.cs
thebigmoosey: @michaelnoonan - Same reason, see above.

source/Calamari/Integration/Scripting/FSharp/Bootstrap.fsx
thebigmoosey: Same reason as other bootstrappers

source/Calamari/Integration/Scripting/ScriptCS/ScriptCSBootstrapper.cs
michaelnoonan: Perfecto!

source/Calamari/Integration/Scripting/WindowsPowerShell/Bootstrap.ps1
michaelnoonan: @matt-richardson yeah, I've read http://www.jsnover.com/blog/2013/12/07/write-host-considered-harmful/ and had the same reaction. We've actually seen a couple of bugs around sensitive variable handling where PowerShell functions used `Write-Output` which affected the return value from the function in certain surprising circumstances, and took a long time to diagnose.

Since that time I've been quite deliberately using `Write-Host` where I only want a `Console.WriteLine()` equivalent - and so far, so good. :)

In this case I'd err on the side of caution and stick with `Write-Host` - also for consistency.

matt-richardson: Ahh, cool. All good then.

source/Calamari/Integration/Scripting/FSharp/FSharpBootstrapper.cs
pawelpabich: Looks good to me

PR #140, opened 2016-11-30
Fix issue with wildcards in the middle of transform filename

matt-richardson:
Was working fine when wildcard was at the start (`*.mytransform.config`) but not in the middle (`MyApp.*.mytransform.config`).

Fixes OctopusDeploy/Issues#2970


zentron:
Bit of a hairy method in there we might need to revisit at some point but otherwise the change LGTM

PR #141, opened 2016-12-03
Added option to select the files that should be purged.

dlemstra:
This patch adds an extra option to select the files that should be purged and is a follow-up of #121 and OctopusDeploy/Issues#2943. But instead of an exclusion list this is an inclusion list. Our situation is that we have a website with a known structure. Extra files are added by an external system and we don't know the files/folders that are added. This means that we need to make sure that those files are left behind. Maybe it should be an option if the list is an exclude or include list though. I could add an extra setting to the special variables to set that.

### DIFF
--- a/source/Calamari.Tests/Fixtures/Conventions/CopyPackageToCustomInstallationDirectoryConventionFixture.cs
+++ b/source/Calamari.Tests/Fixtures/Conventions/CopyPackageToCustomInstallationDirectoryConventionFixture.cs
@@ -1,4 +1,6 @@
-﻿using Calamari.Deployment;
+﻿using System;
+using System.Linq;
+using Calamari.Deployment;
 using Calamari.Deployment.Conventions;
 using Calamari.Integration.FileSystem;
 using Calamari.Integration.Processes;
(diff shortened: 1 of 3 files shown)


zentron:
Thanks for the PR @dlemstra, As noted in OctopusDeploy/Issues#2943, We would ideally aim to get a similar behaviour to how the [Azure preserve paths](http://docs.octopus.com/display/OD/Preserve+Specific+Paths+When+Deploying+Azure+Web+App) logic works, using multiple regex-style paths to support a glob-like behaviour.

dlemstra:
I can could change the code to support regex-style paths. But what should we do with preserving or purging the specified paths? What are your thoughts on that @zentron?

zentron:
for consistency i think it would make sense to supporting a "preserve path x" rather than "purge path x" semantics to align with the currently available Azure feature.

dlemstra:
Would it be possible to add an extra option (boolean) that would allow the user to do a preserve instead of a purge? Not sure if that can easily be added to the Azure feature but adding this to `CopyPackageToCustomInstallationDirectoryConvention` should be easy. But what should be the name of `CustomInstallationDirectoryPurgePaths` parts then? Any good ideas for a new name @zentron?

PR #145, opened 2017-01-03
Fixed Azure Web App deploy to work with non-global Azure regions

sebastianroesch:
I am working with Azure Germany, which requires specific API endpoints in order to deploy web apps. So far, Calamari did not fully respect the endpoints specified as variables. Therefore, I am passing the endpoint to the resource and website management clients.


droyad:
[Fixed Azure Web App deploy to work with non-global Azure regions #145](https://trello.com/c/FZaY40X9/72-fixed-azure-web-app-deploy-to-work-with-non-global-azure-regions-145)

PR #147, opened 2017-01-23
Check to make sure custom package deployment script types are compatible with the OS

droyad:
… and throw an error if it not supported. Refactored how Engines declare the scripts they support.

Resolves https://github.com/OctopusDeploy/Issues/issues/2905


droyad:
[Running a powershell script on an SSH target is silently ignored #2905](https://trello.com/c/xjFbmAZ7/34-running-a-powershell-script-on-an-ssh-target-is-silently-ignored-2905)

PR #149, opened 2017-02-02
Config changes to support long paths in Windows 10+

YogirajA:
- Calamari fails to deploy if the path of folders that longer than 240 characters.
- I updated the config file based on https://blogs.msdn.microsoft.com/jeremykuhne/2016/07/30/net-4-6-2-and-long-paths-on-windows-10/ and it works

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-Calamari is the command-line tool invoked by Tentacle during a deployment. It knows how to extract and install NuGet packages, run the build.cmd etc. conventions, modify configuration files, and all the other things that happen during an deployment.
+Calamari is the command-line tool invoked by Tentacle during a deployment. It knows how to extract and install NuGet packages, run the Deploy.ps1 etc. conventions, modify configuration files, and all the other things that happen during an deployment.
 
 When the solution is built, a new Calamari package is created in the `built-packages` directory.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,4 +1,4 @@
-Calamari is the command-line tool invoked by Tentacle during a deployment. It knows how to extract and install NuGet packages, run the Deploy.ps1 etc. conventions, modify configuration files, and all the other things that happen during an deployment.
+Calamari is the command-line tool invoked by Tentacle during a deployment. It knows how to extract and install NuGet packages, run the build.cmd etc. conventions, modify configuration files, and all the other things that happen during an deployment.
```
droyad: Hi, this line is incorrect. Deploy.ps1 refers to the scripts it runs at runtime, not the build script

droyad:
@YogirajA Could you please detail how to reproduce the problem that this solves?

I tried adding the config value to Calamari and then deploying a package to a custom directory that was 252 characters long. I still got a path too long error or could not find part of the path error. (Windows 10)

YogirajA:
@droyad 
- Thanks for your feedback. I have reverted the ReadMe changes
- You seem to have recreated the issue as you got the correct error.
- Have you enabled the LongPaths in Windows? It was shipped with the anniversary update last year. They are required on top of this config change per msdn blog post.  You can do that with steps below:
      - run regedit
      - Go to HKEY_LOCAL_MACHINE -> System -> CurrentControlSet -> Control ->FileSystem
      - Open LogPathsEnabled and set its value to 1
  
longpaths
 
I'd make sure to restart the services after the changes

After this, I pushed the updated Calamari package to the tentacles.

YogirajA:
@droyad Here's the exact error I was getting. Do you get the same? I have some node modules with very long paths.

screen shot 2017-02-02 at 7 47 43 pm

droyad:
@YogirajA Thanks for that. I rechecked my error and realised it was coming from the csscript engine as I had a C# deployment script in my package.

I was able to successfully reproduce and apply this change, it'll be included in the next Octopus release.

I've also added an overarching issue https://github.com/OctopusDeploy/Issues/issues/3129

droyad:
Fixes https://github.com/OctopusDeploy/Issues/issues/3130

PR #157, opened 2017-02-20
Globbing support in 'Substitute variables in files' and 'JSON configuration variables'

NickJosevski:
https://octopusdeploy.uservoice.com/forums/170787-general/suggestions/6261527-allow-for-wildcard-directory-searching-for-substit

http://help.octopusdeploy.com/discussions/problems/39746-substitute-variables-in-files-recursive-octopus-312

Implemented (via the 'Target Files' inputs) for, `JSON and Substitute Variables in Files`

image

### DIFF
--- a/source/Calamari.Tests/Fixtures/Integration/FileSystem/CalamariPhysicalFileSytemFixture.cs
+++ b/source/Calamari.Tests/Fixtures/Integration/FileSystem/CalamariPhysicalFileSytemFixture.cs
@@ -91,5 +91,57 @@ namespace Calamari.Tests.Fixtures.Integration.FileSystem
             return filename;
         }
 
+        [Test]
+        [TestCase(@"**/*.txt", "f1.txt", 2)]
+        [TestCase(@"**/*.txt", "r.txt", 2)]
+        [TestCase(@"*.txt", "r.txt")]
+        [TestCase(@"**/*.config", "root.config", 5)]
+        [TestCase(@"*.config", "root.config")]
+        [TestCase(@"Config/*.config", "c.config")]
+        [TestCase(@"Config/Feature1/*.config", "f1-a.config", 2)]
+        [TestCase(@"Config/Feature1/*.config", "f1-b.config", 2)]
+        [TestCase(@"Config/Feature2/*.config", "f2.config")]
+        public void GlobTestMutiple(string pattern, string expectedFileMatchName, int expectedQty = 1)
+        {
+            var realFileSystem = CalamariPhysicalFileSystem.GetPhysicalFileSystem();
+            var rootPath = realFileSystem.CreateTemporaryDirectory();
+            var content = "file-content" + Environment.NewLine;
+
+            if (CalamariEnvironment.IsRunningOnWindows)
+            {
+                pattern = pattern.Replace(@"/", @"\");
+            }
+
+            try
+            {
+                var configPath = Path.Combine(rootPath, "Config");
+
+                realFileSystem.CreateDirectory(configPath);
+                realFileSystem.CreateDirectory(Path.Combine(configPath, "Feature1"));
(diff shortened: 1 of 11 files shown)


NickJosevski:
The other 2 options, do not currently do GLOB:

 - `Configuration Variables` is already recursive.
 - `Configuration transforms` uses a `from => to` notation and has heaps of rules around siblings and soforth. It is also recursive if you use releative paths.

image

source/Calamari/Integration/FileSystem/CalamariPhysicalFileSystem.cs
```diff
@@ -342,7 +351,15 @@ static string GetTempBasePath()
         public string CreateTemporaryDirectory()
         {
             var path = Path.Combine(GetTempBasePath(), Guid.NewGuid().ToString());
-            Directory.CreateDirectory(path);
+            return CreateDirectory(path);
+        }
+
+        public string CreateDirectory(string path)
```
NickJosevski: I wanted to use CalamariPhysicalFileSystem for it's cross platform logic it was already being used in tests.

source/Calamari/Util/CrossPlatformExtensions.cs
```diff
@@ -16,11 +16,17 @@ public static string GetApplicationTempDir()
...
-            var path = Environment.GetEnvironmentVariable("LOCALAPPDATA") ?? Environment.GetEnvironmentVariable("TMPDIR") ?? "/tmp";
+            var path = Environment.GetEnvironmentVariable("LOCALAPPDATA")
+                   ?? Environment.GetEnvironmentVariable("TMPDIR")
+                   ?? Environment.GetEnvironmentVariable("TEMP")
+                   ?? "/tmp";
 #endif
-            path = Path.Combine(path, Assembly.GetEntryAssembly().GetName().Name);
-            path = Path.Combine(path, "Temp");
-            return path;
+            if (Assembly.GetEntryAssembly() != null)
```
NickJosevski: GetApplicationTempDir() was existing but the first time used in tests, and was failing on the build server because how the test runs.

Do you mean the `?:` instead of `if-else`? It doesn't read as well... If there is a assembly name is there based on how it's running then it will work as it was originally written, the else case covers the case where team city is running it, and just another temp directory name will be fine.

droyad: `Assembly.GetEntryAssembly()?.GetName()?.Name ?? Guid.NewGuid().ToString()`

source/Calamari.Tests/Fixtures/Integration/FileSystem/CalamariPhysicalFileSytemFixture.cs
```diff
@@ -91,5 +91,57 @@ string CreateFile(params string[] relativePath)
...
+            if (CalamariEnvironment.IsRunningOnWindows)
+            {
+                pattern = pattern.Replace(@"/", @"\");
+            }
+
+            try
+            {
+                var configPath = Path.Combine(rootPath, "Config");
+
+                realFileSystem.CreateDirectory(configPath);
```
droyad: Why not just use `Directory.CreateDirectory`, then you wouldn't need to change CalamariPhysicalFileSystem

NickJosevski: I see yes, will do

source/Calamari/Integration/FileSystem/CalamariPhysicalFileSystem.cs
NickJosevski: yes agree, not necessary.

source/Calamari/Util/CrossPlatformExtensions.cs
NickJosevski: I see, yes that goes to 1 line ok.

PR #159, opened 2017-02-24
Clarify config transform logging

tothegills:
As of Octopus 3.1 we elevate config transform warnings to errors. The workaround is to use the variable `Octopus.Action.Package.IgnoreConfigTransformationErrors` which downgrades warnings to info, errors to warnings and prevents config transform exceptions for failing the deployment.

In the cases I've dealt with someone has upgraded to 3.1+ and their config transforms stop working.  They have found the variable but are nervous about using it because they are not sure what it does and want any errors to fail the deployment but do not want warnings to fail the deployment.

This PR (hopefully) helps clear up the logging options and gives information about why certain logging happened and how to fix it. It introduces the special variable `Octopus.Action.Package.TreatConfigTransformationWarningsAsErrors` which can be set to false to restore pre-3.1 behaviour.

Config transforms should behave the same except the special case where there is no root element will abide by the logging options.  By default this will cause an error instead of a warning.

Fixes OctopusDeploy/Issues#3207


tothegills:
Need to add the special variable to Octopus.

PR #160, opened 2017-02-24
Added code to allow untrusted certificates

wryanpni:
Related to #3075


droyad:
Related to https://github.com/OctopusDeploy/Issues/issues/3075

slewis74:
The need to support un-trusted certificates is possibly being driven by a bug in Azure, where it uses the incorrect certificates on the Web App's SCM site.  The SCM site should always be using a trusted certificate issued from Microsoft.  https://social.msdn.microsoft.com/Forums/sqlserver/en-US/ebd2bcf4-09ee-4b80-9492-82a465dbb345/incorrect-custom-certificate-on-mysitescmazurewebsitesnet?forum=windowsazurewebsitespreview details a fix.

johnsimons:
We have fixed this as part of #189

PR #162, opened 2017-03-02
Added logging of those packages and directories that are to be kept.

droyad:
Resolves https://github.com/OctopusDeploy/Issues/issues/2937


tothegills:
Worth including the reason why something was removed?

Current:
```
Keeping deployments from the last 1 days
Removing directory 'C:\Octopus\Applications\Test\web\1.0.0'
```

Suggested:
```
Keeping deployments from the last 1 days
Removing directory 'C:\Octopus\Applications\Test\web\1.0.0' as it was installed 2 days and 4 hours ago
```

PR #163, opened 2017-03-07
Handled the Linux root directory when Sanitizing directory paths.

droyad:
Fixes https://github.com/OctopusDeploy/Issues/issues/3264


NickJosevski:
👍

PR #170, opened 2017-03-15
obtain semaphore before executing TargetScriptFile

NickJosevski:
Fixes https://github.com/OctopusDeploy/Issues/issues/2824

### DIFF
--- a/source/Calamari/Integration/Scripting/WindowsPowerShell/Bootstrap.ps1
+++ b/source/Calamari/Integration/Scripting/WindowsPowerShell/Bootstrap.ps1
@@ -12,6 +12,32 @@ $ErrorActionPreference = 'Stop'
 # Functions
 # -----------------------------------------------------------------
 
+function Wait-OnSemaphore
+{
+	param(
+	[parameter(Mandatory = $true)][string] $SemaphoreId
+	)
+
+	Try
+	{
+		$SemaphoreInstance = New-Object System.Threading.Semaphore -ArgumentList  1, 1, $SemaphoreId
+
+		Write-Verbose "Acquired SemaphoreInstance $SemaphoreId"
+
+		while (-not $SemaphoreInstance.WaitOne(50))
+		{
+			Start-Sleep -m 50;
+		}
+
+		return $SemaphoreInstance
+	}
+	Catch [System.Threading.AbandonedMutexException]
+	{
+		$SemaphoreInstance = New-Object System.Threading.Semaphore -ArgumentList 1, 1, $SemaphoreId
+		return Wait-OnSemaphore -SemaphoreId $SemaphoreId
+	}
+}
+
 function Log-VersionTable
 {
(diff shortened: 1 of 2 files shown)


tothegills:
By default scripts need to acquire a mutex on the Tentacle to run.  This can overridden with the variable `OctopusBypassDeploymentMutex` which allows scripts to run in parallel on the same Tentacle.  It looks like this change requires all scripts to acquire this semaphore which can not be overridden by `OctopusBypassDeploymentMutex` - seems like we are doubling up a bit.
Can the IIS semaphore be more targeted, only required before IIS operations?

NickJosevski:
Instead of checking for `DeployAsWebSite`, `DeployAsWebApplication` or `DeployAsVirtualDirectory` in `WindowsPowerShell/Bootstrap.ps1` I've moved the semaphore function, and obtaining it inside `iiswebsite_beforepostdeploy.ps1`.

That file doesn't have a single entry point function, it's lots of conditional logic based on state, so the bottom two-thirds of it are now in try catch in order to release the semaphore.

I've made the sleep time total 400ms for obtaining the semaphore, if there's a few instances of this running on the same machine and they take at least a few seconds to run.

droyad:
Should the semaphore perhaps only go around the bit where we have trouble, specifically modification code? (Which are those where we do the retries)?

Otherwise we serialize the starting of Websites, etc, which can be parallel.

NickJosevski:
@droyad thanks that change is there now, just watching run a few times, it seemed faster (anecdotal evidence).

NickJosevski:
Will add logging so people don't get worried about why it's possibly taking longer under certain scenarios.

NickJosevski:
Next item, improved logging and longer wait time:

```
12:29:28   Verbose  |       Unable to determine if IIS is installed on this server but will optimistically continue.
12:29:28   Info     |       Making sure a Website "web1" is configured in IIS...
12:29:28   Verbose  |       Cannot start this IIS website related task yet. There is already another task running that cannot be run in conjunction with any other task. Please wait...
12:29:33   Verbose  |       Cannot start this IIS website related task yet. There is already another task running that cannot be run in conjunction with any other task. Please wait...
12:29:40   Verbose  |       Cannot start this IIS website related task yet. There is already another task running that cannot be run in conjunction with any other task. Please wait...
12:29:47   Verbose  |       Cannot start this IIS website related task yet. There is already another task running that cannot be run in conjunction with any other task. Please wait...
12:29:51   Verbose  |       Cannot start this IIS website related task yet. There is already another task running that cannot be run in conjunction with any other task. Please wait...
12:29:56   Verbose  |       Acquired SemaphoreInstance Global\Octopus-IIS-Metabase

```

NickJosevski:
@droyad if the last commit looks good this ok to merge?

https://github.com/OctopusDeploy/Calamari/pull/170/commits/70af59be3eb48070bf3d3aad9bb5714efc6eb682

PR #173, opened 2017-03-22
JsonConfigurationVariableReplacer.MapNumber - value from tryparse not being used

NickJosevski:
Fixes https://github.com/OctopusDeploy/Issues/issues/3324


NickJosevski:
extra change was for a float test case `variables.Set("FloatValue", "456e-5");`

PR #175, opened 2017-03-29
Powershell script debugging support [WIP]

matt-richardson:


PaulStovell:
Looks low risk enough to me, I'd have no concerns merging this.

source/Calamari/Integration/Scripting/WindowsPowerShell/PowerShellBootstrapper.cs
```diff
@@ -58,20 +60,30 @@ public static string FormatCommandArguments(string bootstrapFile, CalamariVariab
             {
                 commandArguments.Append($"-Version {customPowerShellVersion} ");
             }
-            var executeWithoutProfile = variables[SpecialVariables.Action.PowerShell.ExecuteWithoutProfile];
-            bool noProfile;
-            if (bool.TryParse(executeWithoutProfile, out noProfile) && noProfile)
+            var debug = variables[SpecialVariables.Action.PowerShell.ExecuteWithoutProfile];
```
PaulStovell: This seems wrong - why is this "debug"1

source/Calamari/Integration/Scripting/WindowsPowerShell/PowerShellBootstrapper.cs
```diff
@@ -58,20 +60,30 @@ public static string FormatCommandArguments(string bootstrapFile, CalamariVariab
...
+                ? debuggingBootstrapFile.Replace("'", "''")
+                : bootstrapFile.Replace("'", "''");
+
+            commandArguments.AppendFormat("-Command \"Try {{. {{. '{0}' -OctopusKey '{1}'; if ((test-path variable:global:lastexitcode)) {{ exit $LastExitCode }}}};}} catch {{ throw }}\"", filetoExecute, encryptionKey);
             return commandArguments.ToString();
         }
 
+        private static bool IsDebuggingEnabled(CalamariVariableDictionary variables)
+        {
+            var powershellDebugMode = variables[SpecialVariables.Action.PowerShell.DebugMode];
```
PaulStovell: I think you should test if the value is "truthy" not just if it has a value - people will expect to be able to set it to "false"

source/Calamari/Integration/Scripting/WindowsPowerShell/PowerShellBootstrapper.cs
matt-richardson: Yeah, thats definitely broken. Bad refactor.

source/Calamari/Integration/Scripting/WindowsPowerShell/PowerShellBootstrapper.cs
matt-richardson: Worth comparing against an enum?

source/Calamari/Integration/Scripting/WindowsPowerShell/Bootstrap.ps1
```diff
@@ -246,13 +247,13 @@ Log-VersionTable
 # -----------------------------------------------------------------
 # Variables
 # -----------------------------------------------------------------
-
+{{BeforeVariablesDebugLocation}}
 {{VariableDeclarations}}
```
PaulStovell: Think these are necessary? Feels like it would be to debug our stuff - I think the use case for this feature is just to debug their stuff.

source/Calamari/Integration/Scripting/WindowsPowerShell/PowerShellBootstrapper.cs
PaulStovell: I don't mind the strings - ultimately the enum will only be used in PS anyway so it may as well be magic strings. Just need to support false

source/Calamari/Integration/Scripting/WindowsPowerShell/Bootstrap.ps1
matt-richardson: Was debating with myself whether it was overkill having these...
But... The ability for us to debug our stuff is super helpful too. Maybe we just dont call these options out in the documentation.

PR #181, opened 2017-04-11
Long path support

markryd:
This is .net 4 only (not .net core) and Windows only.

### DIFF
--- a/source/Calamari/Integration/FileSystem/CalamariPhysicalFileSystem.cs
+++ b/source/Calamari/Integration/FileSystem/CalamariPhysicalFileSystem.cs
@@ -9,10 +9,6 @@ using Calamari.Deployment;
 using Calamari.Integration.Retry;
 using Calamari.Util;
 using System.Runtime.InteropServices;
-#if LONG_FILE_PATHS
-using File = Alphaleonis.Win32.Filesystem.File;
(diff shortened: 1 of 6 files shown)


source/Calamari/Integration/FileSystem/CalamariPhysicalFileSystem.cs
```diff
@@ -114,7 +117,7 @@ public void DeleteFile(string path, FailureOptions options)
 
         public void DeleteDirectory(string path)
         {
-            Directory.Delete(path, true);
+            Directory.Delete(path);
```
NickJosevski: The second parameter is recurse=true, why is it no longer required? I guess nothing did depend on it being recursive

I see the usage of does do `fileSystem.PurgeDirectory` first:
https://github.com/OctopusDeploy/Calamari/blob/2b45299e50393cb6153ea9b3ae159f7ed7c1d557/source/Calamari/Deployment/Retention/RetentionPolicy.cs#L85

the rest of the calls to `DeleteDirectory` are in tests.

markryd: I've just hard-coded `true` in the implementation of Delete, since we pass `true` in all our uses. I might change it back though so we keep the same API as the underlying Dir.Delete. I did that with the other functions, but this was one of the first I wrapped.

PR #184, opened 2017-04-19
Cannot remove SSL cert brindings here if SNI is not enabled

jacobnlsn:
[In this commit](https://github.com/jacobnlsn/Calamari/commit/5fc29d6d31714eac9aea3099a1abe7c08adbcfd2) a chunk of code was added to attempt to remove SSL cert bindings from IP:Port combinations that are no longer being used.  This had a impact that if any other bindings on that site, or any other sites on the IIS server was using the same IP:Port combination it cleared the SSL cert for all of the sites, breaking SSL for those sites.

This pull request is related to this git issue: https://github.com/OctopusDeploy/Issues/issues/3414


alfhenrik:
Hi Jacob,

Thanks for taking the time to send us this PR, we are actually in the process of getting a fix in that will make this work in both non-SNI and SNI situations as the code was put in a fix for another customers issue when updating SSL certs (this [support ticket](http://help.octopusdeploy.com/discussions/problems/50976) and this [PR](https://github.com/OctopusDeploy/Calamari/pull/167)).

Again, thank you for taking the time to put together this PR but I'll close it and we will have a fix for your situation ready in the next few days.

Thank you and best regards,
Henrik

PR #190, opened 2017-05-08
Added web start mode parameters to Octopus.Features.IISWebSite_BeforePostDeploy.ps1

johnsimons:
Relates to https://github.com/OctopusDeploy/Issues/issues/3459

Modified the PS to stop the apppool and/or website based on the used parameter at the start of the script this is required because the `New-Item` actually starts them or they could already be running.

We then check again the parameter to figure out if we should start the site and/or pool at the end.

There are two new parameters used:

- Deploy as website uses both `Octopus.Action.IISWebSite.StartApplicationPool` and `Octopus.Action.IISWebSite.StartWebSite`, boolean flags
- Deploy as WebApplication uses `Octopus.Action.IISWebSite.StartApplicationPool` boolean flag


source/Calamari/Scripts/Octopus.Features.IISWebSite_BeforePostDeploy.ps1
```diff
@@ -760,16 +803,20 @@ if ($deployAsWebSite)
 		throw
 	}
 
-	Start-ApplicationPool $applicationPoolName
+	if($startAppPool -eq $true) {
```
tothegills: why not `if($startAppPool)`?

johnsimons: can be either right ?
Is the code convention to shorten it ?

PR #192, opened 2017-05-12
[WIP] Upgrade solution to VS2017

johnsimons:
Solution opens and builds in VS2017
Build script is not working yet
The package build is strange, it is not including all the files in the output directory, everything becomes a dependency, I can't work it out 😞 

Majority of shell commands have been migrated to msbuild.


NuGet.Config
```diff
@@ -4,7 +4,12 @@
     <add key="disableSourceControlIntegration" value="true" />
   </solution>
   <packageSources>
+    <add key="LocalPackages" value="../LocalPackages" />
+    <add key="NuGet.org v3" value="https://api.nuget.org/v3/index.json" />
     <add key="Octopus Dependencies" value="https://octopus.myget.org/F/octopus-dependencies/" />
     <add key="NuGet.org" value="https://www.nuget.org/api/v2/" />
   </packageSources>
-</configuration>
+  <disabledPackageSources>
```
droyad: We could just remove the feed from the package source path

NuGet.Config
```diff
@@ -4,7 +4,12 @@
     <add key="disableSourceControlIntegration" value="true" />
   </solution>
   <packageSources>
+    <add key="LocalPackages" value="../LocalPackages" />
+    <add key="NuGet.org v3" value="https://api.nuget.org/v3/index.json" />
```
droyad: Ordering the OD Dependencies feed before NuGet might be better. I did notice this is the same in the OD sln

build.cake
```diff
@@ -22,156 +20,78 @@ var signingCertificatePassword = Argument("signing_certificate_password", "");
...
-        Configuration = configuration,
-        OutputDirectory = outputDirectory,
-        Framework = framework,
-		NoBuild = true
-    });
-
-	SignBinaries(outputDirectory);
+		 DotNetCoreBuild("./source", new DotNetCoreBuildSettings
+		{
+			Configuration = configuration
```
droyad: We should add the version number here, eg: https://github.com/OctopusDeploy/Halibut/blob/master/build.cake#L74

build.cake
```diff
@@ -22,156 +20,78 @@ var signingCertificatePassword = Argument("signing_certificate_password", "");
...
-{
-    DoPackage("Calamari", "net40", nugetVersion);
-    DoPackage("Calamari.Azure", "net45", nugetVersion);   
-});
-
-private void DoPackage(string project, string framework, string version)
-{ 
-	var outputDirectory = Path.Combine(artifactsDir, project);
-	
-    DotNetCorePublish(Path.Combine("./source", project), new DotNetCorePublishSettings
```
droyad: This is the step that pulls together all the dependent DLLs into one folder

build.cake
```diff
@@ -22,156 +20,78 @@ var signingCertificatePassword = Argument("signing_certificate_password", "");
...
-        settings.ArgumentCustomization = f => {
-            f.Append("-where");
-            f.AppendQuoted(testFilter);
-            return f;
-        };
-
-     DotNetCoreTest("source/Calamari.Tests/project.json", settings);
-});
-
-Task("__Pack")
```
droyad: This is still required to build the zip packages

build.cake
```diff
@@ -22,156 +20,78 @@ var signingCertificatePassword = Argument("signing_certificate_password", "");
...
+					return args.Append("--logger:trx");
+				}
+			});
+	});
 	
-    TransformConfig(Path.Combine(artifactsDir, project, "project.json"), new TransformationCollection {
-        { "version", version }
-    });
-
-    DotNetCorePack(Path.Combine(artifactsDir, project), new DotNetCorePackSettings
```
droyad: Again we should add the version number here: https://github.com/OctopusDeploy/Halibut/blob/master/build.cake#L99

build.cake
```diff
@@ -22,156 +20,78 @@ var signingCertificatePassword = Argument("signing_certificate_password", "");
...
-			ToolPath = signtoolPath,
-            TimeStampUri = new Uri("http://timestamp.globalsign.com/scripts/timestamp.dll"),
-            CertPath = signingCertificatePath,
-            Password = signingCertificatePassword
-    });
-}
-
-Task("__Publish")
-    .WithCriteria(isContinuousIntegrationBuild)
+Task("Publish")
```
droyad: This task can be removed altogether, it was never called in the previous script. TC is currently pushing packages for us after all the platform tests pass

source/Calamari.Azure/Calamari.Azure.csproj
```diff
@@ -0,0 +1,106 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Version>1.0.0.0</Version>
+    <TargetFramework>net452</TargetFramework>
```
droyad: This should be `net45` as this is our current minimum requirements for tentacle.

source/Calamari.Azure/Integration/Security/ServicePrincipal.cs
```diff
@@ -9,7 +9,7 @@ public static string GetAuthorizationToken(string tenantId, string applicationId
             var authContext = GetContextUri(activeDirectoryEndPoint, tenantId);
             Log.Verbose($"Authentication Context: {authContext}");
             var context = new AuthenticationContext(authContext);
-            var result = context.AcquireToken(managementEndPoint, new ClientCredential(applicationId, password));
+            var result = context.AcquireTokenAsync(managementEndPoint, new ClientCredential(applicationId, password)).GetAwaiter().GetResult();
```
droyad: Any reason we switched to the Async version when a sync is available?

source/Calamari.Tests/Some sample text
```diff
@@ -0,0 +1 @@
+C:\Users\john\AppData\Local\Temp\tmpCE92.tmp\ZZZZabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz\ZZZZabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz\ZZZZabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz\ZZZZabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz\ZZZZabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz\ZZZZabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz\ZZZZabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz\ZZZZabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz\ZZZZabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz\ZZZZabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz\ZZZZabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz\ZZZZabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz\ZZZZabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz\ZZZZabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz\ZZZZabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz\ZZZZabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz\test.txt
```
droyad: What is this file for?

source/Calamari.Tests/Calamari.Tests.csproj
```diff
@@ -0,0 +1,128 @@
...
+      <FSharpFiles Include="$(FSharpCompilerTools)" />
+      <FSharpFilesExe Condition="'$(TargetFramework)' == 'netcoreapp1.1'" Include="$(FSharpCompilerToolsExe)" />
+      <ScriptCSFiles Include="$(ScriptCS)" />
+      <ScriptCSFilesExe Condition="'$(TargetFramework)' == 'netcoreapp1.1'" Include="$(ScriptCSExe)" />
+      <NuGetFiles Include="$(NuGetCommandLine)" Condition=" '$(TargetFramework)' == 'net452'" />
+    </ItemGroup>
+    <Copy SourceFiles="@(FSharpFiles)" DestinationFolder="$(OutDir)/FSharp/" />
+    <Copy SourceFiles="@(ScriptCSFiles)" DestinationFolder="$(OutDir)/ScriptCS/" />
+    <Copy SourceFiles="@(NuGetFiles)" DestinationFolder="$(OutDir)/NuGet/" Condition="'$(TargetFramework)' == 'net452'" />
+    <Exec Command="chmod +x %(FSharpFilesExe.Identity)" IgnoreExitCode="true" Condition="'$(TargetFramework)' == 'netcoreapp1.1'" />
```
droyad: This should only run on Linux. Not sure whether this would really help, we can build the Linux binaries on Windows. Is there a way to do this while/after tar.gzing?

source/Calamari/Calamari.csproj
```diff
@@ -0,0 +1,127 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Version>1.0.0.0</Version>
+    <TargetFrameworks>net452;netcoreapp1.0</TargetFrameworks>
```
droyad: This should be `net40` as we still support Windows 2003 via older Tentacles.

source/Calamari/Util/CrossPlatformExtensions.cs
```diff
@@ -78,13 +78,6 @@ public static string GetManifestExtension()
 #endif
         }
 
-#if NET40
```
droyad: This is probably still needed once the project targets net40

source/Calamari/Calamari.csproj
```diff
@@ -0,0 +1,127 @@
...
+    <PackageProjectUrl>https://github.com/OctopusDeploy/Calamari/</PackageProjectUrl>
+    <PackageLicenseUrl>https://github.com/OctopusDeploy/Calamari/blob/master/LICENSE</PackageLicenseUrl>
+    <Authors>Octopus Deploy</Authors>
+    <RepositoryType>git</RepositoryType>
+    <RepositoryUrl>https://github.com/OctopusDeploy/Calamari/</RepositoryUrl>
+    <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
+    <GeneratePackageOnBuild>True</GeneratePackageOnBuild>
+    <Description>Contains the command-line Calamari tool that is used by Tentacle to perform depoyment actions on machines.</Description>
+    <Product>Calamari</Product>
+  </PropertyGroup>
```
droyad: The RIDs are missing (see original project.json and https://github.com/OctopusDeploy/DataCenterManager/blob/master/source/Service/Service.csproj#L8)

build.cake
johnsimons: no need, we use Gitversion directly in the individual projects

build.cake
johnsimons: no need, we use Gitversion directly in the individual projects
And the package is build as part of it, see https://github.com/OctopusDeploy/Calamari/pull/192/files#diff-5c2422bfd230f086caac361ea9cfa9a2R17

source/Calamari.Azure/Calamari.Azure.csproj
johnsimons: it actually needs to be net451

source/Calamari.Azure/Integration/Security/ServicePrincipal.cs
johnsimons: maybe i stuffed up something but there is no sync version last time I checked

source/Calamari.Tests/Some sample text
johnsimons: oops, part of running the tests

source/Calamari.Tests/Calamari.Tests.csproj
johnsimons: yeah we can toggle it based on OS

source/Calamari/Calamari.csproj
johnsimons: Do we need them ?

build.cake
johnsimons: Do we need the zip ?

build.cake
droyad: Adding the version number here will include it in the nuspec and nupkg filename IIRC. Skip it if you puts it in automatically

build.cake
droyad: I should have said nupkgs

build.cake
droyad: Ah we are using Attribute generation, in that case we don't then :). If you set the version here though, dotnet will generate the attributes for you.

source/Calamari.Azure/Calamari.Azure.csproj
droyad: Before the conversion this project targeted net45, why does it need to be upgraded now?

source/Calamari/Calamari.csproj
droyad: Ah looks like that work may never have been completed. The intention was to create `coreclr` builds that would not be dependent on .NET Framework (and target the Unicies). Looks like we are only building the .NET Framework one at the moment.

source/Calamari.Azure/Calamari.Azure.csproj
johnsimons: .net451 is part of [Microsoft Server 2012 r2](https://msdn.microsoft.com/en-us/library/8z6watww(v=vs.110).aspx#Supported server operating systems), I thought that was the minimum requirement correct ?

BTW, net451 is the minimum for OD.

source/Calamari.Azure/Integration/Security/ServicePrincipal.cs
johnsimons: It looks like the sync version is no longer available in the latest version and since I have upgraded this dependency, https://github.com/OctopusDeploy/Calamari/pull/192/files#diff-5c2422bfd230f086caac361ea9cfa9a2R49, the code changed slightly, do u want me to go back ?

source/Calamari/Calamari.csproj
johnsimons: I thought the minimum is Windows Server 2012 R2 ?

source/Calamari.Azure/Integration/Security/ServicePrincipal.cs
droyad: Probably, otherwise we'd have to retest that usage.

source/Calamari/Calamari.csproj
droyad: Tentacle 3.0 can run on 2003 (https://octopus.com/blog/octopus-3.1-will-be-.net-4.5), and people are doing that with the latest version of OD server (https://octopusdeploy.slack.com/archives/C072C2D5L/p1470097608000604)

PR #193, opened 2017-05-22
Generate unique deployment names for Azure Resource Group deployments

robpearson:
## Issue Summary:
Calamari `DeployAzureResourceGroupConvention` calculates the deployment name inconsistently where a random name isn't always random.  After discussions w/ @MJRichardson, I learned that the real thing this needs is to be unique and not random so we decided to use a guid.  This is long and not quite as user friendly but it always unique thus addressing the core issue.

This change also includes a small tweak to the AesEncryption.RandomString() method to ensure calls made to it used a single static System.Random instance to avoid collisions. 

This was confirmed by my own local testing and the customer who raised this (awaiting response).

Updated deployment names:

Fixes OctopusDeploy/Issues#3449
Fixes OctopusDeploy/Issues#3473

## Tests:
- Added `AesEncryptionFixture` to test this for uniqueness.
- Manually tested the random deployment name calculation.  All tests pass!


robpearson:
["Move into..." is breaking the rules around supported child step types](https://trello.com/c/O7YSNIQA/528-move-into-is-breaking-the-rules-around-supported-child-step-types)

thebigmoosey:
@robpearson Nice one! I think this could also _potentially_ fix [this issue](https://github.com/OctopusDeploy/Issues/issues/3473).

It may be worth following the same lock logic that was implemented in Octopus.Server (and abstracting this out into a `RandomStringGenerator` utility class, in case anyone else tries to write their own random string logic in Calamari in the future), as we ran into issues there last year that took forever to figure out. Eg. https://github.com/OctopusDeploy/OctopusDeploy/blob/f35360d1f067db4fc8db49cbefb6202bbf54b459/source/Octopus.Shared/Util/RandomStringGenerator.cs#L14

robpearson:
@thebigmoosey Yup, I think that'll fix it too.  Same issue.

robpearson:
@thebigmoosey I'll split that out as well.  Good point as it's not really related to AesEncryption.

robpearson:
@thebigmoosey I've left the `RandomString() code as is b/c I didn't really need to change it.  The biggest change needed was to ensure deployment names were unique and after discussions @MJRichardson and others it was decided the simplest change was to use a GUID.  

Any other comments appreciated.

source/Calamari.Tests/Fixtures/Util/AesEncryptionFixture.cs
```diff
@@ -0,0 +1,23 @@
...
+    [TestFixture]
+    public class AesEncryptionFixture
+    {
+        [Test]
+        public void CanGetRandomStrings() 
+        {
+            var randomString1 = AesEncryption.RandomString(8);
+            var randomString2 = AesEncryption.RandomString(8);
+
+            Assert.That(randomString1, Is.Not.EqualTo(randomString2));
```
pawelpabich: This assertion can fail because random is not the same as unique.

robpearson: @pawelpabich The change to `AesEncryption.RandomString()` now use a single static `Random` instance and both of these use `.Next()` to get the random number.  This should be random.  This SO post is the best source on this.  

https://stackoverflow.com/questions/767999/random-number-generator-only-generating-one-random-number

robpearson:
@thebigmoosey Good questions.  In regards to the `Guid.NewGuid()` method.  I had to investigate but this SO post pretty much confirms that it's pretty much safe.  

> Possible (as in, could it ever happen, in the lifetime of the universe)? Yes.
> Likely (at all)? No.

https://stackoverflow.com/questions/2827362/can-two-threads-of-the-same-process-produce-the-same-guid

The readability thing is the downside but it tests fine so I think it's good.  If customers complain, we can always iterate pretty quickly.

PR #195, opened 2017-06-01
Change to use ZipArchive instead of ZipReader in NuGet package reader.

EssentialNRG:
Change to use ZipArchive instead of ZipReader in NuGet package reader.

Changed LocalNuGetPackage and NupkgExtractor to use ZipArchive instead of ZipReader.

This is to support (nested) zip files in a nuget package. When the NuGet package contains a zip file the and the ZipReader is used this will result in a "System.NotSupportedException: Unknown header: ..." error message during deployment.


tothegills:
Hi,

Thanks for the PR, please sign the CLA so we can accept.

Cheers,
Shane

EssentialNRG:
I've signed the CLA.

PR #200, opened 2017-06-15
Azure Cloud Region Deployment Name calc

robpearson:
## Issue Summary
Simple change to ensure azure resource group deployment names are unique and less than 64 characters long.

Fixes OctopusDeploy/Issues#3545

## Tests
Added Calamari.Azure.Tests and added a new test fixture for this.


robpearson:
[Azure Resource Management Deployment Task generates deployment name that's too long](https://trello.com/c/OwXW6d5K/619-azure-resource-management-deployment-task-generates-deployment-name-thats-too-long)

source/Calamari.Azure.Tests/Fixtures/Conventions/DeployAzureResourceGroupConventionFixture.cs
```diff
@@ -0,0 +1,58 @@
+﻿using Calamari.Azure.Deployment.Conventions;
+using Calamari.Azure.Deployment.Integration.ResourceGroups;
+using Calamari.Integration.FileSystem;
+using NUnit.Framework;
+
+namespace Calamari.Azure.Tests.Fixtures.Conventions
+{
+    public class GenerateDeploymentNameFromStepNameTestWrapper : DeployAzureResourceGroupConvention
```
pawelpabich: Can we extract the code that generates the deployment name to a separate static function and test is directly? In that case we would not need the additional wrapper and reduce the noise

source/Calamari.Azure.Tests/Fixtures/Conventions/DeployAzureResourceGroupConventionFixture.cs
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        [Test]
+        public void GivenLongStepName_Then_Can_Generate_Deployment_Name_Appropriately()
+        {
+            // Given / When
+            var deploymentName = GenerateDeploymentNameFromStepNameTestWrapper.TestGenerateDeploymentNameFromStepName("123456789012345678901234567890"); // 30 chars
+
+            // Then
+            Assert.That(deploymentName, Has.Length.EqualTo(64));
```
pawelpabich: With FluentAssertions this line would become: `deploymentName.Length.Should().Be(64)`.  Totally personal opinion but it feels cleaner

source/Calamari.Azure/Deployment/Conventions/DeployAzureResourceGroupConvention.cs
```diff
@@ -79,16 +78,17 @@ public void Install(RunningDeployment deployment)
...
             deploymentName = Regex.Replace(deploymentName, "\\s", "-");
-            deploymentName =
-                new string(deploymentName.Select(x => (char.IsLetterOrDigit(x) || x == '-') ? x : '-').ToArray());
+            deploymentName = new string(deploymentName.Select(x => (char.IsLetterOrDigit(x) || x == '-') ? x : '-').ToArray());
             deploymentName = Regex.Replace(deploymentName, "-+", "-");
             deploymentName = deploymentName.Trim('-', '/');
-            deploymentName = deploymentName + "-" + Guid.NewGuid().ToString();
+            // Azure Deployment Namese can only be 64 characters == 27 chars + "-" (1) + Guid (36 chars)
+            deploymentName = deploymentName.Length <= 27 ? deploymentName : deploymentName.Substring(0, 27);
+            deploymentName = deploymentName + "-" + Guid.NewGuid();
```
pawelpabich: We could use `guid.ToString("N")` which would allow us to use a few more characters that were provided by the end user. We could also base the number of available characters on the length of the unique value.  Then we don't have to have the comment above.

robpearson: Great suggestion!  Updated!

PR #202, opened 2017-06-20
Used a temporary filename when download nuget packages 

droyad:
In case the process gets killed (eg through cancellation)

Fixes https://github.com/OctopusDeploy/Issues/issues/3575

See issue for more details

### DIFF
--- a/source/Calamari/Deployment/Retention/RetentionPolicy.cs
+++ b/source/Calamari/Deployment/Retention/RetentionPolicy.cs
@@ -1,8 +1,10 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using Calamari.Deployment.Journal;
 using Calamari.Integration.FileSystem;
+using Calamari.Integration.Packages.Download;
 using Calamari.Integration.Time;
 
 namespace Calamari.Deployment.Retention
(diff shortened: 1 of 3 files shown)


source/Calamari/Deployment/Retention/RetentionPolicy.cs
```diff
@@ -149,5 +153,23 @@ static string FormatWithThPostfix(int value)
                     return value + "th ";
             }
         }
+
+        private void RemovedFailedPackageDownloads()
+        {
+            var pattern = "*" + PackageDownloader.DownloadingExtension;
```
thebigmoosey: This could be inlined or declared inside the `if` block

source/Calamari/Integration/Packages/NuGet/NuGetPackageDownloader.cs
```diff
@@ -34,13 +37,15 @@ public void DownloadPackage(string packageId, NuGetVersion version, Uri feedUri,
 
                 try
                 {
-                    action(packageId, version, feedUri, feedCredentials, targetFilePath);
+                    action(packageId, version, feedUri, feedCredentials, tempTargetFilePath);
+                    fileSystem.MoveFile(tempTargetFilePath, targetFilePath);
```
thebigmoosey: Good solution, nice and simple 👍

PR #205, opened 2017-06-21
Fixes in Readme (missing '=' and wrong folder name) + automated creation of artifacts folder

ShaKann:
Hi

I have noticed few flaws

- missing '=' in command adding custom folder
- wrong folder name (currently package is created in artifacts folder)
- I changed build.ps1 so artifacts folder will be created automatically like tools folder


droyad:
@ShaKann Thanks for that. The `=` isn't strictly necessary, but I'll keep it in.

PR #206, opened 2017-06-26
Port to netcoreapp2.0

droyad:
TODO:

- [x] See if we need long file name support under windows and `netcoreapp2.0`
- [x] Determine if WORKAROUND_FOR_EMPTY_STRING_BUG is required still
- [x] Get Linux build agents working
- [x] Test it out

### DIFF
--- a/source/Calamari/Integration/Processes/Semaphores/SystemSemaphoreManager.cs
+++ b/source/Calamari/Integration/Processes/Semaphores/SystemSemaphoreManager.cs
@@ -24,6 +24,13 @@ namespace Calamari.Integration.Processes.Semaphores
 
         public IDisposable Acquire(string name, string waitMessage)
         {
+            return CalamariEnvironment.IsRunningOnWindows
+                ? AcquireSemaphore(name, waitMessage)
+                : AcquireMutex(name, waitMessage);
+        }
+
+        IDisposable AcquireSemaphore(string name, string waitMessage)
+        {
             Semaphore semaphore;
             var globalName = $"Global\\{name}";
             try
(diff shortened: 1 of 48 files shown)


droyad:
The WORKAROUND_FOR_EMPTY_STRING_BUG workaround causes the `ApplyDeltaFixture`, `FindPackageFixture` and `PackageDownloadFixture` to fail (https://build.octopushq.com/viewLog.html?buildId=252933&tab=buildResultsDiv&buildTypeId=OctopusDeploy_CalamariNetCore_DotnetTesting_BuildAndTestRhel72DotNet)

droyad:
netcoreapp2.0 supports long filenames OOTB

droyad:
WORKAROUND_FOR_EMPTY_STRING_BUG is still needed

source/Calamari/Integration/Processes/Semaphores/SystemSemaphoreManager.cs
```diff
@@ -22,9 +22,15 @@ public SystemSemaphoreManager(ILog log, TimeSpan initialWaitBeforeShowingLogMess
             this.initialWaitBeforeShowingLogMessage = (int)initialWaitBeforeShowingLogMessage.TotalMilliseconds;
         }
 
-#if NET40
         public IDisposable Acquire(string name, string waitMessage)
         {
+            return CalamariEnvironment.IsRunningOnWindows
+                ? AcquireSemaphore(name, waitMessage)
+                : AcquireMutex(name, waitMessage);
```
MJRichardson: Why can we not just use the Mutex on both Windows and Linux?  It's global right?

droyad: Yeh that would be good (this sparked my #feature-discussion thread). I think the initial approach was not to change functionality in the existing code, but this sounds like a good change.

droyad: Thinking about it, we should do it as a separate change and add a release note for it

PR #211, opened 2017-07-12
Make WindowsX509CertificateStore safer

michaelnoonan:
This change makes the WindowsX509CertificateStore safe for concurrent access where it matters. We were seeing exceptions thrown when multiple threads/processes attempted to access/set private key ACLs.

Fixes OctopusDeploy/Issues#3644


michaelnoonan:
[Intermittent System.AccessViolationException using built-in Octopus Certificate feature](https://trello.com/c/RmUMNth4/564-intermittent-systemaccessviolationexception-using-built-in-octopus-certificate-feature)

PR #215, opened 2017-08-02
Add journal entry if script is in a package so the package can be cleaned up by retention policy

alfhenrik:
Relates to OctopusDeploy/Issues#3708

### DIFF
--- a/source/Calamari/Commands/RunScriptCommand.cs
+++ b/source/Calamari/Commands/RunScriptCommand.cs
@@ -2,9 +2,11 @@
 using System.IO;
 using Calamari.Commands.Support;
 using Calamari.Deployment;
+using Calamari.Deployment.Journal;
 using Calamari.Integration.FileSystem;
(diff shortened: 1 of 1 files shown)


source/Calamari/Commands/RunScriptCommand.cs
```diff
@@ -87,9 +96,15 @@ private int InvokeScript(CalamariVariableDictionary variables)
 
             if (result.ExitCode == 0 && result.HasErrors && variables.GetFlag(SpecialVariables.Action.FailScriptOnErrorOutput, false))
             {
+                if(!(deployment?.SkipJournal ?? true))
```
droyad: Would this line be better as `if(deployment?.SkipJournal != false)`

alfhenrik: Yes, yes it would...I went through a couple of different options but totally missed that one 😳

PR #217, opened 2017-08-23
Returned empty string instead of `Unrecognized variable` if a variable does not exist

droyad:
Fixes https://github.com/OctopusDeploy/Issues/issues/3747


droyad:
On hold until 4.0

PR #218, opened 2017-08-28
Adhere to PowerShell common output preferences

tothegills:
Will not log verbose and warning output if the common parameters `$VerbosePreference` and `$WarningPreference` (respectively) are set to `SilentlyContinue`.

The parameter `-WarningAction` for the method `Write-Warning` can be set to `SilentlyContinue` to avoid writing warning output.

A limitation of this implementation is that passing `-WarningAction` to any other method will still result in the warning being written.  For example this will result in a warning being written:
```
Test-NetConnection -ComputerName abc123 -WarningAction SilentlyContinue
```


tothegills:
Fixes OctopusDeploy/Issues#3736

PR #224, opened 2017-10-06
If reading package metadata with schema validation fails, try without schema validation

alfhenrik:
- [x] Add test

I'm not 100% sure this is the right solution, but it's the only way I could come up with...~(might be worth investigating if using newer NuGet dependencies resolves this issue without the need for calling `Manifest.ReadFrom(path, false)`)~

Tested with latest (`4.3.0`) `NuGet.Packaging` version and it works correctly with schema validation. So the question is, do we want to update our custom build of NuGet with latest NuGet dependencies...?

### DIFF
--- a/source/Calamari.Tests/Fixtures/PackageDownload/PackageDownloadFixture.cs
+++ b/source/Calamari.Tests/Fixtures/PackageDownload/PackageDownloadFixture.cs
@@ -20,6 +20,7 @@ namespace Calamari.Tests.Fixtures.PackageDownload
         static readonly string DownloadPath = TestEnvironment.GetTestPath(TentacleHome, "Files");
 
         static readonly string PublicFeedUri = "https://www.myget.org/F/octopusdeploy-tests";
+        static readonly string NuGetFeedUri = "https://api.nuget.org/v3/index.json";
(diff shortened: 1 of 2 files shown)


source/Calamari.Tests/Fixtures/PackageDownload/PackageDownloadFixture.cs
```diff
@@ -68,6 +70,21 @@ public void ShouldDownloadPackage()
...
+
+            result.AssertSuccess();
+
+            result.AssertOutput(
+                $"Downloading NuGet package {NuGetFeed.PackageId} {NuGetFeed.Version} from feed: '{NuGetFeedUri}'");
+            result.AssertOutput($"Downloaded package will be stored in: '{NuGetFeed.DownloadFolder}");
+
+            AssertStagePackageOutputVariableSet(result, NuGetFeed.File);
+            result.AssertOutput($"Package {NuGetFeed.PackageId} {NuGetFeed.Version} successfully downloaded from feed: '{NuGetFeedUri}'");
+        }
```
tothegills: Do we need a similar test on the server?

alfhenrik: That'd probably be a good idea, I'll add it before merging the server PR

PR #229, opened 2017-11-01
Bug fail script on error output variable causes successful script to fail 

tw17:
Fixes the issue where setting the Octopus.Action.FailScriptOnErrorOutput variable to true causes successful script to fail.
Fixes OctopusDeploy/Issues#3773


tothegills:
The changes look great. It would be nice to have some tests, maybe in `PowerShellFixture`?

tw17:
There are existing tests that check writing to std error doesn't fail when Octopus.Action.FailScriptOnErrorOutput isnt set, and does fail when `Octopus.Action.FailScriptOnErrorOutput=true`, we were just missing the scenario to check successful steps still pass when `Octopus.Action.FailScriptOnErrorOutput=true`

PR #232, opened 2017-11-07
Fix for DeployFailed.ps1 not automatically cleaned up after successful deployment

tw17:
- Added an implementation for the method Cleanup on RollbackScriptConvention to delete the script. Cleanup was already being called, it was just missing an implementation. 
- Removed RollbackScriptConvention from TransferPackageCommand as it doesn't extract the package, so rollback scripts aren't available.

Fixes OctopusDeploy/Issues#3608


tothegills:
Don't forget to update Calamari refs in Octopus

PR #239, opened 2017-12-04
update to Azure csproj file to copy tools correctly

slewis74:
F# and ScriptCS tooling wasn't being included because the file copy was only being done after the Build target but also needs to be done after the Publish target.

Fixes OctopusDeploy/issues#4023

### DIFF
--- a/source/Calamari.Azure/Calamari.Azure.csproj
+++ b/source/Calamari.Azure/Calamari.Azure.csproj
@@ -33,7 +33,6 @@
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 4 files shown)


matt-richardson:
Any way we can add a test for this? Feels like it could break again and we’d not know.

NickJosevski:
Yes could an easy test be just detecting the presence of these extra elements to the Calamari Azure package? 

A better test would be a more comprehensive E2E integration test using scriptcs, F# and a few other combinations, wonder how flaky that would be...

droyad:
I don't think having those extra files is a reason to not ship if it's working.

PR #256, opened 2018-02-28
Extracted the body of the webexception and printed it to the logs

mcasperson:
Fixes https://github.com/OctopusDeploy/Issues/issues/4332

### DIFF
--- a/source/Calamari.Aws/Deployment/Conventions/DeployAwsCloudFormationConvention.cs
+++ b/source/Calamari.Aws/Deployment/Conventions/DeployAwsCloudFormationConvention.cs
@@ -1,6 +1,8 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
+using System.Net;
 using System.Text.RegularExpressions;
 using System.Threading;
 using Amazon.CloudFormation;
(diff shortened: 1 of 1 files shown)


source/Calamari.Aws/Deployment/Conventions/DeployAwsCloudFormationConvention.cs
```diff
@@ -865,5 +892,21 @@ private bool DisplayWarning(string errorCode, string message)
...
+        /// <summary>
+        /// The AmazonServiceException can hold additional information that is useful to include in
+        /// the log.
+        /// </summary>
+        /// <param name="exception">The exception</param>
+        private void HandleAmazonServiceException(AmazonServiceException exception)
+        {
+            if (exception.InnerException is WebException)
+            {
+                ((WebException) exception.InnerException).Response.GetResponseStream()
```
shaunmarx: Btw you can avoid the additional cast here if you wanted to by using pattern matching i.e.

```
private void HandleAmazonServiceException(AmazonServiceException exception)
        {
            if (exception.InnerException is WebException innerException)
            {
                (innerException.Response.GetResponseStream()
                    .Map(stream => new StreamReader(stream).ReadToEnd())
                    .Tee(message => DisplayWarning("AWS-CLOUDFORMATION-ERROR-0014",
                        "An exception was thrown while contacting the AWS API.\n" + message));
            }
        }
```

PR #262, opened 2018-03-20
Add AWS S3 uploads

shaunmarx:


source/Calamari.Aws/Util/ConstantClassExtensions.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using Amazon.Runtime;
+
+namespace Calamari.Aws.Util
+{
+    public static class ConstantHelpers
+    {
+        public static IEnumerable<T> GetConstantValues<T>() where T : ConstantClass
```
shaunmarx: The AWS SDK provides these constant value objects as static members of a class, but provide no way of getting a definitive list of these.

source/Calamari/Deployment/Conventions/SubstituteInFilesConvention.cs
```diff
@@ -9,21 +10,33 @@ namespace Calamari.Deployment.Conventions
...
+        private readonly Func<RunningDeployment, IEnumerable<string>> fileTargets;
         private readonly ICalamariFileSystem fileSystem;
         readonly IFileSubstituter substituter;
 
         public SubstituteInFilesConvention(ICalamariFileSystem fileSystem, IFileSubstituter substituter)
         {
             this.fileSystem = fileSystem;
             this.substituter = substituter;
+            predicate = (deployment) => deployment.Variables.GetFlag(SpecialVariables.Package.SubstituteInFilesEnabled);
+            fileTargets = (deployment) => deployment.Variables.GetPaths(SpecialVariables.Package.SubstituteInFilesTargets);
```
shaunmarx: Refactored from what we had before to keep the default behaviour

source/Calamari/Serialization/InheritedClassConverter.cs
```diff
@@ -0,0 +1,102 @@
...
+using System.Linq;
+using System.Reflection;
+using System.Text;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using Newtonsoft.Json.Serialization;
+
+namespace Calamari.Serialization
+{
+    public abstract class InheritedClassConverter<TBaseResource, TEnumType> : JsonConverter
```
shaunmarx: Taken from the server, however, existing contract resolver will be used. We are dealing with a nested object serialized into the property for file selections for example.

source/Calamari/Deployment/Conventions/ConditonalConvention.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+
+namespace Calamari.Deployment.Conventions
+{
+    public class ConditionalInstallationConvention<T>: IInstallConvention where T: IInstallConvention
```
shaunmarx: Just a convenience to conditionally run an installation convention based on a predicate

source/Calamari.Aws/Deployment/Conventions/CreateS3BucketConvention.cs
```diff
@@ -0,0 +1,55 @@
...
+        public void EnsureBucketExists(Func<AmazonS3Client> clientFactory, string bucketName)
+        {
+            Guard.NotNull(clientFactory, "Client factory should not be null");
+            Guard.NotNullOrWhiteSpace(bucketName, "Bucket name should not be null or empty");
+
+            using (var client = clientFactory())
+            {
+                if (Amazon.S3.Util.AmazonS3Util.DoesS3BucketExist(client, bucketName))
+                {
+                    Log.Info($"Bucket {bucketName} exists in region {awsEnvironmentGeneration.AwsRegion}. Skipping creation.");
```
mcasperson: I'd make this message verbose. It seems quite likely that the bucket would exist (probably for every time the step is run after the first step).

PR #270, opened 2018-04-09
S3 bugbash suggestions

shaunmarx:
This PR enables the bucket key for single file selections and packages to use the filename with a bucket key prefix or a "custom" bucket key value. It also modifies the upload behaviour to skip files where the content hash is equal to the ETag.

AWS specifies the following for the ETag:

- Objects created by the PUT Object, POST Object, or Copy operation, or through the AWS Management Console, and are encrypted by SSE-S3 or plaintext, have ETags that are an MD5 digest of their object data.
- Objects created by the PUT Object, POST Object, or Copy operation, or through the AWS Management Console, and are encrypted by SSE-C or SSE-KMS, have ETags that are not an MD5 digest of their object data.
-If an object is created by either the Multipart Upload or Part Copy operation, the ETag is not an MD5 digest, regardless of the method of encryption.

In our case, we aren't using multipart uploads and in the cases where this does not work, the fallback is to overwrite which is no different from the previous behaviour. We may make this behaviour more flexible in the future but should cover 80% for now.


source/Calamari.Aws/Deployment/Conventions/UploadAwsS3Convention.cs
```diff
@@ -90,14 +89,13 @@ public void Install(RunningDeployment deployment)
...
-                            Log.Info(
-                                $"Saving variable \"Octopus.Action[{deployment.Variables["Octopus.Action.Name"]}].Output.Files[{response.BucketKey}]\"");
+                            Log.Info($"Saving variable \"Octopus.Action[{deployment.Variables["Octopus.Action.Name"]}].Output.Files[{response.BucketKey}]\"");
                             Log.SetOutputVariable($"Files[{response.BucketKey}]", response.Version);
                         }
                     }
                 });
             }
-            catch (AmazonS3Exception exception)
+             catch (AmazonS3Exception exception)
```
thebigmoosey: This additional space _may_ cause the internet to explode

source/Calamari.Aws/Deployment/Conventions/UploadAwsS3Convention.cs
```diff
@@ -236,22 +234,36 @@ public S3UploadResult UploadUsingPackage(Func<AmazonS3Client> clientFactory, Run
...
+                Key = bucketKey()?.Trim(),
+                StorageClass = S3StorageClass.FindValue(properties.StorageClass?.Trim()),
+                CannedACL = S3CannedACL.FindValue(properties.CannedAcl?.Trim())
+            }
+            .WithMetadata(properties)
+            .WithTags(properties)
+            .WithMd5Digest(fileSystem);
+        }
+
+        private static Func<string> GetBucketKey(ICalamariFileSystem fileSystem, string filePath, IHaveBucketKeyBehaviour behaviour)
```
thebigmoosey: Yea, good having this method, I was confused by why there were differences when calling CreateRequest previously, with the enum this helps

source/Calamari.Aws/Deployment/Conventions/UploadAwsS3Convention.cs
shaunmarx: I think it may already have if my internet's behaviour from yesterday is anything to go by! 😆

shaunmarx:
Relates to OctopusDeploy/Issues#4482

PR #272, opened 2018-04-11
Skip content hash checking for s3 step FIPS

shaunmarx:
Skip content hash checking for s3 step when md5 hash algorithm is not available such as when FIPS is enabled.

### DIFF
--- a/source/Calamari.Aws/Deployment/Conventions/UploadAwsS3Convention.cs
+++ b/source/Calamari.Aws/Deployment/Conventions/UploadAwsS3Convention.cs
@@ -84,7 +84,7 @@ namespace Calamari.Aws.Deployment.Conventions
 
             if (!md5HashSupported)
             {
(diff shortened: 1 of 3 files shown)


source/Calamari.Aws/Deployment/Conventions/UploadAwsS3Convention.cs
```diff
@@ -78,6 +82,11 @@ public void Install(RunningDeployment deployment)
             //The bucket should exist at this point
             Guard.NotNull(deployment, "deployment can not be null");
 
+            if (!md5HashSupported)
+            {
+                Log.Info("Md5 hashes are not supported in executing environment. Files will always be uploaded.");
```
michaelnoonan: `Md5` should be `MD5` :)

shaunmarx:
Relates to OctopusDeploy/Issues#4482

PR #274, opened 2018-04-17
Changed AWS and Azure to be plugins

droyad:
MVP for them to be plugins. They both still depend on the main Calamari project, but this allows Octopus to load them plugin style

### DIFF
--- a/source/Calamari/Plugins/PluginLoader.cs
+++ b/source/Calamari/Plugins/PluginLoader.cs
@@ -1,0 +1,50 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using Calamari.Commands.Support;
+
+namespace Calamari.Plugins
+{
+    public static class PluginLoader
+    {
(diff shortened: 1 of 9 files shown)


source/Calamari/Plugins/PluginLoader.cs
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        private static List<Assembly> LoadAssemblies(string[] paths)
+        {
+            var load = new List<Assembly>();
+            foreach (var path in paths)
+            {
+                try
+                {
+                    load.Add(Assembly.LoadFrom(path));
```
zentron: So long as we develop and update all our plugins in lockstep this will work, otherwise we need to be careful about needing to load different versions of dependencies which leads to needing to use `AssemblyLoader` etc

zentron: given that these plugins will be _ours_ unlike our thoughts a year or two ago, i think its fair to assume we can control the dependencies properly which will allow this to work 👍

PR #286, opened 2018-05-24
Change run-script command to allow using scriptBody from variables

zentron:
Currently exists a few inconsistencies with inline scripts where the variable replacement takes place on the server, while scripts from packages are replaced on calamari.

Since we already pass the script body through in the provided variable collection, this change allows calamari to pull the body out of  the variables and write it to disk, replacing the variables in the process.

The old `--script` param will still be supported for the time being for backwards compatibility  however a warning will be logged since we will be no longer doing the replace on the server (to avoid double substitution).

### DIFF
--- a/source/Calamari.Shared/Integration/Scripting/ScriptSyntax.cs
+++ b/source/Calamari.Shared/Integration/Scripting/ScriptSyntax.cs
@@ -1,16 +1,15 @@
 ﻿using System.Linq;
-using System.Reflection;
 using Calamari.Commands.Support;
 
 namespace Calamari.Integration.Scripting
 {
-    public enum ScriptType
+    public enum ScriptSyntax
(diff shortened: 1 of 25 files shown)


source/Calamari.Shared/Integration/Scripting/ScriptSyntax.cs
```diff
@@ -1,16 +1,15 @@
 ﻿using System.Linq;
-using System.Reflection;
 using Calamari.Commands.Support;
 
 namespace Calamari.Integration.Scripting
 {
-    public enum ScriptType
+    public enum ScriptSyntax
```
zentron: To allow calamari to write the correct script (with the right extension) we are synchronizing the script syntax enum with the server.

PR #289, opened 2018-05-29
Correct script engine in Azure ARM step

benPearce1:


benPearce1:
Fixes: https://github.com/OctopusDeploy/Issues/issues/4604

PR #298, opened 2018-07-02
Fix problem with Service Fabric application deployment when there are multiple apps with the same Type

valdisz:
This PR solves issue when  you want to deploy several applications with different names, but the same type.

Current script checks only for Type existence to decide to do Upgrade or not. This results in situation when second application is always deployed to Service Fabric as Update by Octopus, even when there are nothing to update.

Proposed changes will check also for App Name to start Update.


valdisz:
FYI we have tested this patch on our Octopus installation, works as expected...

PR #300, opened 2018-07-16
Fix compatibility with servers that only have .net4.0 installed as TLS11/TLS12 security protocol types didn't exist

alfhenrik:
image

### DIFF
--- a/source/Calamari/Program.cs
+++ b/source/Calamari/Program.cs
@@ -68,7 +68,7 @@ namespace Calamari
         {
             this.command = command;
             this.helpCommand = helpCommand;
-        }        
+        }
 
         public int Execute(string[] args)
         {
@@ -90,7 +90,7 @@ namespace Calamari
             catch (Exception ex)
             {
                 return ConsoleFormatter.PrintError(ex);
-            }            
+            }
         }
 
         private int PrintHelp(string action)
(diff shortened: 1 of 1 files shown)


source/Calamari/Program.cs
```diff
@@ -107,13 +107,25 @@ public static void EnableAllSecurityProtocols()
...
+            if (Enum.IsDefined(typeof(SecurityProtocolType), 768))
+                securityProcotolTypes = securityProcotolTypes | (SecurityProtocolType)768;
+
+            if (Enum.IsDefined(typeof(SecurityProtocolType), 3072))
+            {
+                securityProcotolTypes = securityProcotolTypes | (SecurityProtocolType)3072;
+            }
+            else
+            {
+                Log.Verbose($"TLS1.2 is not supported, this means that some outgoing connections to third party endpoints will not work as they now only support TLS1.2.{Environment.NewLine}This includes GitHub feeds and Maven feeds.");
```
matt-richardson: Is this worth logging as warning? or at least info?

matt-richardson:
The more I think about it, I'm thinking we probably need an E2E test or something that explicitly checks that it runs against PS2 and fails if it doesn't. (You're probably all over this already though 😄)

source/Calamari/Program.cs
zentron: I dont think warning is great if for 99% of use cases it wont be a problem. When their script _does_ fail, at least we\they can get some info about this in the logs

PR #301, opened 2018-07-16
Enh create namespace

mcasperson:


zentron:
Assuming it works for both then looks good. What does the log output look like? WE should probably log out non-verbose if we create a namespace

mcasperson:
Fixed up the script to display the namespace creation as a default message

PR #303, opened 2018-07-17
Added support for packaged scripts in ARM template deployment.

dpiessens:
Added the builtin package convention so scripts can be controlled with the template.


dpiessens:
@MJRichardson Mind getting this merged in? It would be helpful since we can't manage calamari versions in Octopus Cloud

PR #305, opened 2018-07-26
Azure web app publish profile with no resource group supplied

benPearce1:
Fixes https://github.com/OctopusDeploy/Issues/issues/4752

### DIFF
--- a/source/Calamari.Azure/Integration/Websites/Publishing/ResourceManagerPublishProfileProvider.cs
+++ b/source/Calamari.Azure/Integration/Websites/Publishing/ResourceManagerPublishProfileProvider.cs
@@ -56,7 +56,6 @@ namespace Calamari.Azure.Integration.Websites.Publishing
                     throw new CommandException($"Could not find Azure WebSite '{azureTargetSite.Site}'{resourceGroupMessage} subscription '{subscriptionId}'");
                 }
                     
-                
                 // if more than one site, fail
                 if (matchingSites.Count > 1)
                     throw new CommandException(
--- a/source/Calamari.Azure/Integration/Websites/Publishing/ResourceManagerPublishProfileProvider.cs
+++ b/source/Calamari.Azure/Integration/Websites/Publishing/ResourceManagerPublishProfileProvider.cs
@@ -49,19 +49,21 @@ namespace Calamari.Azure.Integration.Websites.Publishing
                                          .ToList();
 
                 if (!matchingSites.Any())
-                    throw new CommandException($"Could not find Azure WebSite '{azureTargetSite.Site}' in subscription '{subscriptionId}'");
+                {
+                    var resourceGroupMessage = !string.IsNullOrWhiteSpace(resourceGroupName)
+                        ? $" in resource group '{resourceGroupName}' and"
+                        : " in";
+                    throw new CommandException($"Could not find Azure WebSite '{azureTargetSite.Site}'{resourceGroupMessage} subscription '{subscriptionId}'");
+                }
+                    
                 
                 // if more than one site, fail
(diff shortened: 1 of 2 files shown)


source/Calamari.Azure/Integration/Websites/Publishing/ResourceManagerPublishProfileProvider.cs
```diff
@@ -43,12 +43,25 @@ public static SitePublishProfile GetPublishProperties(string subscriptionId, str
...
-                    .FirstOrDefault(webApp => string.Equals(webApp.Name, azureTargetSite.Site, StringComparison.CurrentCultureIgnoreCase) &&
-                                         (string.IsNullOrWhiteSpace(resourceGroupName) || string.Equals(webApp.ResourceGroup, resourceGroupName, StringComparison.InvariantCultureIgnoreCase)));
+                var matchingSites = sites
+                    .Where(webApp => string.Equals(webApp.Name, azureTargetSite.Site, StringComparison.CurrentCultureIgnoreCase) &&
+                                         (string.IsNullOrWhiteSpace(resourceGroupName) || string.Equals(webApp.ResourceGroup, resourceGroupName, StringComparison.InvariantCultureIgnoreCase)))
+                                         .ToList();
 
-                if (matchingSite == null)
+                if (!matchingSites.Any())
                     throw new CommandException($"Could not find Azure WebSite '{azureTargetSite.Site}' in subscription '{subscriptionId}'");
```
slewis74: should we output ResourceGroup in this message too, if it is supplied?

source/Calamari.Azure/Integration/Websites/Publishing/ResourceManagerPublishProfileProvider.cs
```diff
@@ -43,12 +43,25 @@ public static SitePublishProfile GetPublishProperties(string subscriptionId, str
...
+                
+                // if more than one site, fail
+                if (matchingSites.Count > 1)
+                    throw new CommandException(
+                        $"Found {matchingSites.Count} matching the site name '{azureTargetSite.Site}' in subscription '{subscriptionId}'.{(string.IsNullOrWhiteSpace(resourceGroupName) ? " Please supply a Resource Group name." : string.Empty)}");
+
+                var matchingSite = matchingSites.First();
```
slewis74: I'd do `matchingSites.Single()` here, if the previous checks ever got incorrectly modified that would still catch the case where multiples are returned and prevent the deployment. Less friendly error, but at least doesn't incorrectly deploy to the wrong place.

benPearce1: the less friendly message was what stopped me putting single here in the first place

source/Calamari.Azure/Integration/Websites/Publishing/ResourceManagerPublishProfileProvider.cs
```diff
@@ -43,12 +43,25 @@ public static SitePublishProfile GetPublishProperties(string subscriptionId, str
...
+
+                if (string.IsNullOrWhiteSpace(resourceGroupName))
+                {
+                    resourceGroupName = matchingSite.ResourceGroup;
```
slewis74: Should we just always do this assignment? It'll always give you the right answer, and 1 less `if` statement is always a good thing, right? :)

benPearce1: similar to your comment about incorrect modifications in earlier code, if the matching site query gets borked, then this will ensure that the resource group supplied will take precedence

PR #306, opened 2018-07-27
Moving health checks for Azure targets out of server

slewis74:
Implemented new commands in Calamari.Azure that do health checks for Azure targets.

Created some abstractions around Accounts to share/reuse the places where we're pulling account info out of the variables.

Also fixed an issue with the container setup for commands, so we differentiate "you gave me a command that I can't find" from "I found the command but it's registration is wrong and I went 💥 trying to resolve it"

There's a corresponding PR coming in server, for offloading the health checks to Calamari.

Relates to OctopusDeploy/Issues/issues/4765


source/Calamari.Azure/Accounts/AccountFactory.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using Calamari.Deployment;
+using Octostache;
+
+namespace Calamari.Azure.Accounts
+{
+    public static class AccountFactory
+    {
+        public static Account Create(VariableDictionary variables)
```
MichaelJCompton: Wasn't clear on this.  Does this mean a variable set can contain only one variable of Account type?  Would that meant that the same deployment process can't access two azure accounts?  Did I miss something here?

source/Calamari.Azure/Commands/HealthCheckAzureCloudServiceCommand.cs
```diff
@@ -0,0 +1,62 @@
...
+using Calamari.Commands.Support;
+using Calamari.Deployment;
+using Calamari.Integration.Certificates;
+using Calamari.Integration.Processes;
+using Microsoft.WindowsAzure.Management.Compute;
+
+namespace Calamari.Azure.Commands
+{
+    [Command("hc-azure-cs", Description = "Run a health check on an Azure Cloud Service")]
+    public class HealthCheckAzureCloudServiceCommand : Command
```
MichaelJCompton: Yeah, I like it!  Health check commands like this was what I was expecting for cloud workers.

source/Calamari.Azure/Accounts/AccountFactory.cs
slewis74: Yeah, that is how things currently work, you can only have 1 Account per variable set. Which equates to 1 account type per step in a process, because it comes down to the step running on a machine. There is some thinking going in to supporting multiple accounts, for things like connect to AWS and Azure at the same time, but this will be a big refactor across many things to make that work.

PR #307, opened 2018-08-03
Enable changesets feature for CloudFormation

shaunmarx:
This PR adds some capabilities to use CloudFormation changesets and also allows the RoleARN to be specified for the given stack at creation time.

Fixes OctopusDeploy/issues#4695
Fixes OctopusDeploy/Issues#4560


source/Calamari.Aws/Integration/CloudFormation/StackEventLogger.cs
```diff
@@ -0,0 +1,105 @@
...
+            bool missingIsFailure)
+        {
+            var isSuccess = status.Select(x => x.MaybeIndicatesSuccess()).SelectValueOr(x => x.Value, !missingIsFailure);
+            var isStackType = status.SelectValueOr(x => x.ResourceType.Equals("AWS::CloudFormation::Stack"), true);
+            
+            if (expectSuccess && !isSuccess && isStackType)
+            {
+                
+                log.Warn(
+                    "Stack was either missing, in a rollback state, or in a failed state. This means that the stack was not processed correctly. " +
```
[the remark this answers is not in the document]
matt-richardson: Not sure :)
I've just got a pet peeved about "X was either in a blah state or a blah state" - we know what the state is, so we can include it. I see you've fixed it 👍

source/Calamari.Shared/Deployment/Conventions/AggregateInstallationConvention.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+
+namespace Calamari.Deployment.Conventions
+{
+    public class AggregateInstallationConvention: IInstallConvention
```
thebigmoosey: This is a nice addition :)

source/Calamari.Aws/Integration/CloudFormation/CloudFormationObjectExtensions.cs
```diff
@@ -0,0 +1,327 @@
...
+            try
+            {
+                return clientFactory()
+                    .Map(client => client.CreateStack(request))
+                    // Narrow the response to the stack ID
+                    .Map(response => response.StackId);
+            }
+            catch (AmazonCloudFormationException ex)
+            {
+                if (ex.ErrorCode == "AccessDenied")
```
thebigmoosey: Does AWS have an enum for these kinds of error codes rather than string matching? Probably not I'm guessing :)

source/Calamari.Aws/Integration/CloudFormation/CloudFormationObjectExtensions.cs
```diff
@@ -0,0 +1,327 @@
...
+        /// <param name="status">The status to check</param>
+        /// <returns>true if the status indcates a failed create or update, and false otherwise</returns>
+        public static Maybe<bool> MaybeIndicatesSuccess(this StackEvent status)
+        {
+            return status.ToMaybe().Select(x => !UnsuccessfulStackEvents.Contains(x.ResourceStatus.Value));
+        }
+
+        public static bool StackIsUnrecoverable(this StackEvent status)
+        {
+            Guard.NotNull(status, "Status should not be null");
```
thebigmoosey: I like these guards from a dev point-of-view :) Reminds me of `assert` in ObjectiveC

source/Calamari.Aws/Deployment/Conventions/LogAwsUserInfoConvention.cs
```diff
@@ -39,5 +63,34 @@ public void Install(RunningDeployment deployment)
...
+                    // Client becomes the response of the API call
+                    .Map(client => client.GetCallerIdentity(new GetCallerIdentityRequest()))
+                    // The response is narrowed to the Aen
+                    .Map(response => response.Arn)
+                    // Try and match the response to get just the role
+                    .Map(arn => ArnNameRe.Match(arn))
+                    // Extract the role name, or a default
+                    .Map(match => match.Success ? match.Groups[1].Value : "Unknown")
+                    // Log the output
+                    .Tee(role => Log.Info($"Running the step as the AWS role {role}"));
```
thebigmoosey: There's this "Tee" again, this time in C#! TIL

source/Calamari.Aws/Deployment/Conventions/DeployAwsCloudFormationConvention.cs
```diff
@@ -33,734 +27,178 @@ public enum StackStatus
         InProgress
     }
 
-    public class DeployAwsCloudFormationConvention : IInstallConvention
```
thebigmoosey: Holy dumptruck batman ... that was one giant convention :) Good to see this getting more modular

source/Calamari.Aws/Deployment/Conventions/DeployAwsCloudFormationConvention.cs
```diff
@@ -33,734 +27,178 @@ public enum StackStatus
...
-            var retValue = TemplateReplacement.ResolveAndSubstituteFile(
-                    fileSystem,
-                    templateParametersFile,
-                    filesInPackage,
-                    deployment.Variables)
-                .Map(JsonConvert.DeserializeObject<List<Parameter>>);
-
-            return retValue;
+           
+            clientFactory.WaitForStackToComplete(CloudFormationDefaults.StatusWaitPeriod, stack, LogAndThrowRollbacks(clientFactory, stack, false));
```
thebigmoosey: Much easier to follow 👍

source/Calamari.Aws/Deployment/Conventions/CreateCloudFormationChangeSetConvention.cs
```diff
@@ -0,0 +1,137 @@
...
+            {
+                return clientFactory.CreateChangeSet(request)
+                    .Map(x => new RunningChangeSet(new StackArn(x.StackId), new ChangeSetArn(x.Id)));
+            }
+            catch (AmazonCloudFormationException ex)
+            {
+                if (ex.ErrorCode == "AccessDenied")
+                {
+                    throw new PermissionException(
+                        @"AWS-CLOUDFORMATION-ERROR-0017: The AWS account used to perform the operation does not have " +
```
thebigmoosey: Super minor, but it feels like these could be abstracted into some sort of template function that accepts: `errorNumber, errorSummary, exceptionMessage, includeDocTextAndLink`. Some of this just feels very duplicated, even though I know they're all slightly different based on error number

source/Calamari.Aws/Deployment/Conventions/CloudFormationInstallationConventionBase.cs
```diff
@@ -0,0 +1,136 @@
...
+using Calamari.Deployment;
+using Calamari.Deployment.Conventions;
+using Calamari.Integration.Processes;
+using Microsoft.Data.OData.Query.SemanticAst;
+using Octopus.CoreUtilities;
+using Octopus.CoreUtilities.Extensions;
+
+namespace Calamari.Aws.Deployment.Conventions
+{
+    public abstract class CloudFormationInstallationConventionBase: IInstallConvention
```
thebigmoosey: This is cool 👍 Abstract convention class ftw

source/Calamari.Aws/Commands/DeployAwsCloudFormationCommand.cs
```diff
@@ -84,5 +125,26 @@ public override int Execute(string[] commandLineArguments)
             return 0;
         }
 
+        private bool ChangesetsDeferred(RunningDeployment deployment)
+        {
+            return string.Compare(deployment.Variables[AwsSpecialVariables.CloudFormation.Changesets.Defer], bool.TrueString,
```
thebigmoosey: TIL about bool.TrueString and bool.FalseString 😂

source/Calamari.Aws/Deployment/Conventions/CreateCloudFormationChangeSetConvention.cs
shaunmarx: Yeah, definitely agree there. This is partly what I was referring to reviewing error codes in https://github.com/OctopusDeploy/Issues/issues/4778. These error codes in general need to be revisited and possibly changed so that we aren't being so repetitious. There are a number of locations where the error codes probably should be the same as well such as the Permission Exception above. The error message should be enough to provide some context, right? :)

source/Calamari.Aws/Integration/CloudFormation/CloudFormationObjectExtensions.cs
shaunmarx: Unfortunately not... The only way we can basically deal with this without taking any more dependencies on things / going completely crazy is by catching and rethrowing as an Exception which doesn't have things hidden in some obscure property as a string.

PR #312, opened 2018-08-22
Added the DisableNameChecking param when importing modules

tw17:
Fixes OctopusDeploy/Issues#4830


matt-richardson:
I don't have the full context on this, but this sounds like a very dangerous thing to do... You may randomly get the wrong cmdlets...

tothegills:
@matt-richardson How will you end up with the wrong cmdlets?

matt-richardson:
Sorry, wrong param. I got it mixed up with `-AllowClobber` on `Install-Module`. Dont mind me.

PR #313, opened 2018-08-22
Keep paths of uploaded files

johnsimons:


source/Calamari.Shared/Calamari.Shared.csproj
```diff
@@ -71,7 +71,7 @@
     <PackageReference Include="Microsoft.Web.Xdt" Version="2.1.1" />
     <PackageReference Include="NuGet.CommandLine" Version="2.8.6" />
     <PackageReference Include="NuGet.Core" Version="2.14.0" />
-    <PackageReference Include="AlphaFS" Version="2.1.3-octopus0006" />
+    <PackageReference Include="AlphaFS" Version="2.2.6" />
```
shaunmarx: This is probably a safe upgrade since the changes we had in our own branch has been merged.  We just need to double check to be 100% sure.

shaunmarx: Double checked and this is all good 👍 - we should probably look at updating other dependencies to be consistent and then we can also get rid of our own fork.

PR #320, opened 2018-09-07
Add HTTP_PROXY support to script steps

benPearce1:
This will set the `HTTP_PROXY`, `HTTPS_PROXY` and `NO_PROXY` environment variables based on the web request proxy configuration for the worker context that is executing the script. If using the built-in worker, this will be the server, for external workers, it will be the Tentacle web request proxy.
This will support AWS CLI, Azure CLI, and Kubernetes `kubectl`.

If authentication details are provided in the relevant proxy configuration the proxy environment variables will use the URI format `http://username:password@host:port/`
Port will default to 80 if not supplied.

If the proxy environment variables are already set on in the system variables, they will not be overridden.

The `NO_PROXY` variable also explicitly includes the AWS EC2 Metadata endpoint, as per https://docs.aws.amazon.com/cli/latest/userguide/cli-http-proxy.html

**Exception**: `kubectl` does not seem to support authentication in the proxy configuration.

The changes made to `bootstrap.sh` are future proofing for Linux Tentacle and will only work if the `TentacleProxyHost` environment variable is set. The current Windows Tentacle also sets these as an environment variable.

Scenarios tested:
AWS CLI and Azure CLI through a proxy from both built-in worker and external worker
`kubectl` through a proxy (without auth) from the built-in worker and external worker.

### DIFF
--- a/source/Calamari.Shared/Integration/Scripting/Bash/Bootstrap.sh
+++ b/source/Calamari.Shared/Integration/Scripting/Bash/Bootstrap.sh
@@ -197,22 +197,24 @@ function setup_proxy_configuration
     if [ $HTTP_PROXY ] || [ $HTTPS_PROXY ]
     then
         echo "##octopus[stdout-verbose]"    
-        echo "HTTP_PROXY already set"
+        echo "Proxy already set"
         echo "##octopus[stdout-default]"
         return 0
     fi 
 
     if [ $TentacleProxyHost ] 
     then
+        proxyHost=$TentacleProxyHost:${TentacleProxyPort:-80}
+        
         echo "##octopus[stdout-verbose]"
-        echo "Setting HTTP_PROXY to $TentacleProxyHost:${TentacleProxyPort:-80}"
+        echo "Setting HTTP_PROXY to $proxyHost"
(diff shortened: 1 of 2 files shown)


source/Calamari.Shared/Integration/Scripting/Bash/Bootstrap.sh
```diff
@@ -188,4 +188,38 @@ function log_environment_information
...
 
-log_environment_information
\ No newline at end of file
+function setup_proxy_configuration
+{
+    echo "##octopus[stdout-verbose]"    
+    echo "Configuring proxy"
+    echo "##octopus[stdout-default]"
+        
+    if [ $HTTP_PROXY ] || [ $HTTPS_PROXY ]
```
johnsimons: Should this be an if per proxy ?

johnsimons: Same as the PS version

benPearce1:
Closes https://github.com/OctopusDeploy/Issues/issues/4860
Closes https://github.com/OctopusDeploy/Issues/issues/4855

PR #329, opened 2018-09-27
Application exists based on Application name

prebenh:
Use application type & application name to check if the application exists

If the application doesn't exists, and the type and version match, do a Create instead of Register(AndCreate)

Fix for #4688

### DIFF
--- a/source/Calamari.Azure/Scripts/DeployAzureServiceFabricApplication.ps1
+++ b/source/Calamari.Azure/Scripts/DeployAzureServiceFabricApplication.ps1
@@ -1,4 +1,4 @@
-﻿## Octopus Azure Service Fabric Application script, version 1.0
+## Octopus Azure Service Fabric Application script, version 1.0
 ## --------------------------------------------------------------------------------------
 ##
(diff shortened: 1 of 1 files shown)


benPearce1:
@prebenh Thanks for the PR
Would you be able to give some further explanation to the version checking behavior so we can understand the use-case?

prebenh:
@benPearce1 Here are the use case we are currently using with this script:

1.  **One application type - one application instance:** Use case that is currently supported
1.  **one application type - multiple application instances of the the same version:** New use case that is not support. Have one application type with multiple instances of the application (with possible different configurations). This way we can run multi tenant solutions in one cluster.
1.  **one application type - multiple application instances of different types:** The same as use case 2, but with multiple versions of the type in the cluster.

We want to be able to create these scenarios if nothing exists in the cluster and we also want to be able to update the applications if the are already running.

For use case 2 & 3 we want to create the application instead of registering when the type already exists and the application is not yet in the cluster.


We tested scenario 1,2,3 with Stateless services, Statefull services, Actors and ASP.NET Core applications. All seem.

PR #331, opened 2018-10-08
Kubectl check

mcasperson:


mcasperson:
[Better error message if kubectl is not installed during health check](https://trello.com/c/dYPChEmi/55-better-error-message-if-kubectl-is-not-installed-during-health-check)

mcasperson:
Fixed https://github.com/OctopusDeploy/Issues/issues/4955

PR #332, opened 2018-10-09
Fix for UTF8Encoding.GetString throwing 'The method or operation is not implemented.` on mono >= 5.10

hnrkndrssn:
Fixes OctopusDeploy/Issues#4404

### DIFF
--- a/source/Calamari.Shared/Integration/Scripting/ScriptCS/ScriptCSBootstrapper.cs
+++ b/source/Calamari.Shared/Integration/Scripting/ScriptCS/ScriptCSBootstrapper.cs
@@ -104,9 +104,10 @@ namespace Calamari.Integration.Scripting.ScriptCS
                 return "null;";
 
             var bytes = Encoding.UTF8.GetBytes(value);
-            return CalamariEnvironment.IsRunningOnMono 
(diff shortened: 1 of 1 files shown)


source/Calamari.Shared/Integration/Scripting/ScriptCS/ScriptCSBootstrapper.cs
```diff
@@ -104,7 +104,9 @@ static string EncodeValue(string value)
                 return "null;";
 
             var bytes = Encoding.UTF8.GetBytes(value);
-            return string.Format("System.Text.Encoding.UTF8.GetString(Convert.FromBase64String(\"{0}\"))", Convert.ToBase64String(bytes));
+            return CalamariEnvironment.IsRunningOnMono 
+                ? $"System.Text.Encoding.UTF8.GetString(Convert.FromBase64String(\"{Convert.ToBase64String(bytes)}\"), 0, {bytes.Length})" 
+                : $"System.Text.Encoding.UTF8.GetString(Convert.FromBase64String(\"{Convert.ToBase64String(bytes)}\"))";
```
MJRichardson: Can we not make the same call on both platforms?

PR #333, opened 2018-10-09
Supported k8s logins with aws and azure

mcasperson:
Added support for prioritising the context scripts so the kubernetes context is generated after the cloud login context.


mcasperson:
[Support script wrapper priorities](https://trello.com/c/N4PMIcg8/76-support-script-wrapper-priorities)

PR #337, opened 2018-11-06
Add ability to enumerate variables in bash

espenwa:
This PR adds the ability to enumerate variable names (and subsequently get their value) in bash:

```
for variablename in "${OCTOPUS_VARIABLENAMES[@]}"
do
  echo ${variablename} = $( get_octopusvariable "${variablename}")
done
```

Due to limitations in bash, returning an array from a function is not practical, and hash-objects/dictionaries does not exists. The implementation is therefore done by adding a global array with variable names. Not very elegant, but is apparently working.
Two tests added.

Current implementation does _not_ use encoded (base64) variable names, as `get_octopusvariable` expects un-encoded variable name. This might result in obscure characters in the variable name (mainly `"`) breaking the script.
Making `OCTOPUS_VARIABLENAMES` contain encoded variables is trivial, but will require an additional function call to get the un-encoded variable name before calling `get_octopusvariable` (or some other means to counter the fact that `get_octopusvariable` encodes the name before doing the lookup).


distantcam:
[Add ability to enumerate variables in bash](https://trello.com/c/VoHPJxG8/13-add-ability-to-enumerate-variables-in-bash)

PR #345, opened 2018-11-28
Terraform run on linux

johnsimons:
Added support for Azure accounts
Made step work in linux
Added detailed exit code for apply


source/Calamari.Terraform/ApplyCommand.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Calamari.Commands.Support;
+
+namespace Calamari.Terraform
```
zentron: Rather than having a seperate command for every terraform step, might we be able to simply have a single step `terraform` that looks for the operation type from the variables (supplied by the ActionHandlers with something like `.WithVariable("terraform-type", "apply")`. All these commands seem to differ _only_ due to the convention being used. 

We can reduce the number of classes here if the base `TerraformCommand` simply checks what `terraform-type` value is used and supplies the appropriate convention to its chain. Ideally id even consider the type sent down by server to be a FQN that the command can just rehydrate so it doesn't need to even deal with switch statements. Remember we can currently always assume there is a direct relationship between an action handler and a specific version of Calamari.

source/Calamari.Terraform/TerraformConvention.cs
```diff
@@ -0,0 +1,109 @@
...
+            variables.Add(SpecialVariables.Package.EnableNoMatchWarning,
+                !String.IsNullOrEmpty(enableNoMatchWarning) ? enableNoMatchWarning : (!String.IsNullOrEmpty(additionalFileSubstitution)).ToString());
+
+            var substitutionPatterns = (DefaultTerraformFileSubstitution +
+                                        (string.IsNullOrWhiteSpace(additionalFileSubstitution)
+                                            ? String.Empty
+                                            : "\n" + additionalFileSubstitution))
+                .Split(new[] {"\r", "\n"}, StringSplitOptions.RemoveEmptyEntries);
+
+            new SubstituteInFilesConvention(fileSystem, fileSubstituter,
```
zentron: This convention within convention doesn't feel right. Should the SubstituteInFilesConvention itself be on the terraform command like its use for all other commands?

source/Calamari.Shared/Integration/Proxies/CommandLineToolsProxyEnvironmentVariables.cs
```diff
@@ -0,0 +1,80 @@
+using System;
+using System.Collections.Generic;
+using System.Net;
+using System.Net.Sockets;
+
+namespace Calamari.Integration.Proxies
+{
+    public class CommandLineToolsProxyEnvironmentVariables
```
michaelnoonan: This makes sense since you need the proxy configuration in the .NET Runtime/process.

source/Calamari.Shared/Log.cs
```diff
@@ -101,21 +100,28 @@ public static void Verbose(string message)
             }
         }
 
-        public static void SetOutputVariable(string name, string value)
+        public static void SetOutputVariable(string name, string value, bool isSensitive = false)
```
michaelnoonan: Nice to see this get parity with the PS/C#/F#/bash equivalents.

source/Calamari.Terraform/TerraformCliExecutor.cs
```diff
@@ -0,0 +1,212 @@
...
+                {
+                    throw new Exception($"Directory {TemplateDirectory} does not exist.");
+                }
+
+                TemplateDirectory = templateDirectoryTemp;
+            }
+
+            InitialiseTerraformEnvironmentVariables();
+
+            InitialisePlugins();
```
zentron: Initiali`z`ePlugins (same above\below)

source/Calamari.Terraform/Calamari.Terraform.csproj
```diff
@@ -3,7 +3,17 @@
     <TargetFrameworks>net452;netstandard2.0</TargetFrameworks>
   </PropertyGroup>
   <ItemGroup>
-    <ProjectReference Include="..\Calamari.Aws\Calamari.Aws.csproj" />
+    <Compile Include="..\Calamari.Aws\Integration\AwsEnvironmentGeneration.cs" Link="AwsEnvironmentGeneration.cs" />
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="AWSSDK.Core" Version="3.3.29.12" />
+    <PackageReference Include="AWSSDK.SecurityToken" Version="3.3.4.34" />
+    <PackageReference Include="Microsoft.DotNet.Analyzers.Compatibility" Version="0.2.12-alpha">
```
michaelnoonan: Just confirming you expected a pre-release here.

source/Calamari.Terraform/PlanTerraformConvention.cs
```diff
@@ -0,0 +1,40 @@
...
+            using (var cli = new TerraformCLIExecutor(fileSystem, deployment, environmentVariables))
+            {
+                var resultCode = cli.ExecuteCommand(out results, out var commandResult, "plan", "-no-color", "-detailed-exitcode", extraParameter, cli.TerraformVariableFiles, cli.ActionParams);
+
+                if (resultCode == 1)
+                {
+                    commandResult.VerifySuccess();
+                }
+
+                Log.Info(
```
michaelnoonan: Would these be better logged at `Verbose`?

source/Calamari.Terraform/TerraformCliExecutor.cs
```diff
@@ -0,0 +1,212 @@
...
+            return commandResult.ExitCode;
+        }
+
+        public void Dispose()
+        {
+            if (AttachLogFile)
+            {
+                if (fileSystem.FileExists(logPath))
+                {
+                    Log.NewOctopusArtifact(fileSystem.GetFullPath(logPath), fileSystem.GetFileName(logPath), fileSystem.GetFileSize(logPath));
```
michaelnoonan: Nice!

source/Calamari.Terraform/TerraformCommand.cs
```diff
@@ -0,0 +1,66 @@
...
+using Calamari.Deployment;
+using Calamari.Deployment.Conventions;
+using Calamari.Integration.FileSystem;
+using Calamari.Integration.Packages;
+using Calamari.Integration.Processes;
+using Calamari.Integration.Substitutions;
+
+namespace Calamari.Terraform
+{
+    public abstract class TerraformCommand : Command
```
michaelnoonan: Talking to @zentron I think we want to move away from adding a new `*Command` for each capability of Calamari. I'm not certain it's worth holding up this particular PR since the Calamari.exe API isn't something we keep compatible over time.

source/Calamari.Tests/Terraform/TerraformFixture.cs
```diff
@@ -0,0 +1,358 @@
...
+                customTerraformExecutable = Directory.EnumerateFiles(destinationDirectoryName).FirstOrDefault();
+                Console.WriteLine($"Downloaded terraform to {customTerraformExecutable}");
+            }
+        }
+
+        [Test]
+        public void ExtraInitParametersAreSet()
+        {
+            ExecuteAndReturnLogOutput<PlanCommand>(_ =>
+                    _.Set(TerraformSpecialVariables.Action.Terraform.AdditionalInitParams, "-upgrade"), "Simple")
```
zentron: While this `Action<Dictionary>` approach seems to work, it feels like its not immediately obvious to me, whereas just passing a 
```
new VariableDictionary()
                {
                    {TerraformSpecialVariables.Action.Terraform.AdditionalInitParam, "-upgrade"}
                };
```
makes it more clear what im passing in (you can then take that dictionary and add more defaults to it if need be before its passed along to be invoked)

source/Calamari.Tests/Terraform/TerraformFixture.cs
```diff
@@ -0,0 +1,358 @@
...
+                {
+                    customTerraformExecutable = path;
+                    Console.WriteLine($"Using existing terraform located in {customTerraformExecutable}");
+                    return;
+                }
+            }
+
+            Console.WriteLine("Downloading terraform cli...");
+
+            using (var client = new HttpClient())
```
michaelnoonan: It would be good to put some retries around this to make it more resilient to network outages. It's been a source of flaky tests in other areas. A better option might be to commit the CLI version you want to test.

source/Calamari.Tests/Terraform/TerraformFixture.cs
```diff
@@ -0,0 +1,358 @@
...
+
+        [Test]
+        public async Task AzureIntegration()
+        {
+            var random = Guid.NewGuid().ToString("N").Substring(0, 6);
+            var appName = $"cfe2e-{random}";
+
+            void PopulateVariables(VariableDictionary _)
+            {
+                _.Set(SpecialVariables.Action.Azure.SubscriptionId, Environment.GetEnvironmentVariable("Azure_OctopusAPITester_SubscriptionId"));
```
michaelnoonan: It would be good if this test failed with a helpful message when certain environment variables aren't present. It should tell the developer where to get the secrets in LastPass.

source/Calamari.Terraform/ApplyTerraformConvention.cs
```diff
@@ -0,0 +1,64 @@
...
+{
+    public class ApplyTerraformConvention : TerraformConvention
+    {
+        public ApplyTerraformConvention(ICalamariFileSystem fileSystem, IFileSubstituter fileSubstituter) : base(fileSystem, fileSubstituter)
+        {
+        }
+
+        protected override void Execute(RunningDeployment deployment, Dictionary<string, string> environmentVariables)
+        {
+            IEnumerable<(string, JToken)> OutputVariables(string result)
```
zentron: Does this need to be an inner function? feels like its contributing to the `Execute` method feeling bloated

source/Calamari.Terraform/DestroyTerraformConvention.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        public DestroyTerraformConvention(ICalamariFileSystem fileSystem, IFileSubstituter fileSubstituter) : base(fileSystem, fileSubstituter)
+        {
+        }
+
+        protected override void Execute(RunningDeployment deployment, Dictionary<string, string> environmentVariables)
+        {
+            using (var cli = new TerraformCLIExecutor(fileSystem, deployment, environmentVariables))
+            {
+                cli.ExecuteCommand("destroy", "-force", "-no-color", cli.TerraformVariableFiles,
```
zentron: The other two commands we take the return type and verify success. Do we not need to do that for `Destroy`?

source/Calamari.Terraform/TerraformCliExecutor.cs
```diff
@@ -0,0 +1,212 @@
...
+            var commandResult = ExecuteCommandInternal(ToSpaceSeparated(arguments), out result);
+
+            return commandResult.ExitCode;
+        }
+
+        public int ExecuteCommand(out string result, out CommandResult commandResult, params string[] arguments)
+        {
+            commandResult = ExecuteCommandInternal(ToSpaceSeparated(arguments), out result);
+
+            return commandResult.ExitCode;
```
zentron: rather than using an `out` parameter, then returning a value that comes off that same object, can we just set the return type to be `CommandResult` then consuming code can do with it what it wants (as a plus, its more obvious that the returned value is actually the exit code, since if you need the value you would use something like `result.ExitCode`

source/Calamari.Terraform/TerraformCliExecutor.cs
```diff
@@ -0,0 +1,212 @@
...
+        }
+
+        static string ToSpaceSeparated(IEnumerable<string> items)
+        {
+            return string.Join(" ", items.Where(_ => !String.IsNullOrEmpty(_)));
+        }
+
+        CommandResult ExecuteCommandInternal(string arguments, out string result)
+        {
+            var environmentVar = defaultEnvironmentVariables;
```
zentron: Since we initialize the `defaultEnvironmentVariables` variable in the contructor, do we ever need to check if its null?

source/Calamari.Terraform/TerraformCliExecutor.cs
```diff
@@ -0,0 +1,212 @@
...
+                        ExecuteCommandInternal($"workspace select \"{Workspace}\"", out _).VerifySuccess();
+                        return;
+                    }
+                }
+
+                ExecuteCommandInternal($"workspace new \"{Workspace}\"", out _).VerifySuccess();
+            }
+        }
+
+        class CaptureOutput : ICommandOutput
```
zentron: Would it be worth just using the existing `CaptureCommandOutput`  class and extracting the information off `Infos`, or are you concerned about the perf\service message parsing?

source/Calamari.Terraform/PlanTerraformConvention.cs
johnsimons: This is how the original script on the server was logging it, so I didn't want to introduce braking changes

source/Calamari.Tests/Terraform/TerraformFixture.cs
johnsimons: The CLI is ~80Mbs, I didn't want to add such a big binary. But I can add the retries.

source/Calamari.Terraform/DestroyTerraformConvention.cs
johnsimons: We are, the `ExecuteCommand` has a bunch of overloads, and this one verifies success

source/Calamari.Terraform/TerraformCliExecutor.cs
johnsimons: I did look at `CaptureCommandOutput` but that is in the Test project, so it didn't feel like I should bring it in given my requirements were a lot simpler

source/Calamari.Terraform/TerraformCliExecutor.cs
johnsimons: For public apis I would definitely, but given this is an internal api I don't tend to do it

source/Calamari.Terraform/TerraformCliExecutor.cs
johnsimons: fixed

source/Calamari.Terraform/ApplyTerraformConvention.cs
johnsimons: fixed

source/Calamari.Terraform/TerraformCliExecutor.cs
johnsimons: Do we prefer American spelling for internal apis?

source/Calamari.Terraform/TerraformConvention.cs
johnsimons: fixed

source/Calamari.Terraform/ApplyCommand.cs
johnsimons: I see where we going but given we have plans to refactor the commands, I am somehow inclined to leave this the way it is and do the big refactor later on once we decide which way to go, does that sound reasonable for now?

source/Calamari.Tests/Terraform/TerraformFixture.cs
johnsimons: fixed

source/Calamari.Terraform/TerraformCliExecutor.cs
zentron: we tend to use it throughout most of the code. just a consistency thing

source/Calamari.Terraform/ApplyCommand.cs
zentron: suppose i wouldn't think of it as a blocker to change it to use 1 Terraform command, although i think it simplifies things by removing a bunch of unnecessary classes that only exist to define the convention to use. Hopefully for future work we can take this into consideration. Whilst we dont need to solve the tech debt now its worthwhile to not get further from the goal.

source/Calamari.Terraform/TerraformCliExecutor.cs
zentron: right sorry, mis-read `defaultEnvironmentVariables` as `enviromentVariables`

source/Calamari.Terraform/TerraformCliExecutor.cs
michaelnoonan: Yeah, US English spelling for code generally. It's a standing code convention.

johnsimons: fixed

PR #347, opened 2018-12-05
Python3 script support

tothegills:
Adds support for running pythons scripts. There is a dependency on the `pycryptodome` python module to provide functionality for decrypting sensitive variables. This requires python 3.4+.

Relates to https://github.com/OctopusDeploy/Issues/issues/5234


source/Calamari.Shared/Integration/FileSystem/TemporaryFile.cs
```diff
@@ -48,7 +48,7 @@ public long Size
         }
         public void Dispose()
         {
-            fileSystem.DeleteFile(filePath, FailureOptions.IgnoreFailure);
+            //fileSystem.DeleteFile(filePath, FailureOptions.IgnoreFailure);
```
matt-richardson: Not sure this was meant to be committed :)

source/Calamari.Shared/Integration/Scripting/Python/InstallDependencies.py
```diff
@@ -0,0 +1,19 @@
...
+import sys
+
+def install_missing_dependency(module, package):
+    try:
+        printverbose("Checking for dependency {}".format(package))
+        importlib.import_module(module)
+        printverbose("{} was found".format(package))
+    except:
+        printverbose("Did not find dependency, attempting to install {} for the current user using pip".format(package))
+        subprocess.call([sys.executable, "-m", "pip", "install", package, "--user"])
```
matt-richardson: pip isn't always installed by default with python... Not sure if you're handling this case.

source/Calamari.Tests/Fixtures/RequiresMinimumPython3VersionAttribute.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            this.minor = minor;
+        }
+
+        public void BeforeTest(ITest test)
+        {
+            var python3Version = GetPython3Version();
+            if (python3Version.None())
+            {
+                Assert.Ignore("Requires Python3 to be installed");
```
matt-richardson: I prefer `Assert.Inconclusive` for these kind of ones - they show up better for this scenario

hnrkndrssn:
A few notes from our screen sharing demo this morning:
1. Provide a helpful message if we can't for some reason install the crypto dependency (i.e. `pip` not being installed on the target), similar to what we do for IIS [here](https://github.com/OctopusDeploy/Calamari/blob/master/source/Calamari/Scripts/Octopus.Features.IISWebSite_BeforePostDeploy.ps1#L38)
2. Variable insertion in the script editor doesn't insert the python specific variable format (i.e. `octopusvariables.get('Octopus.Environment.Name')`)
3. Script console does not yet support python scripts

source/Calamari.Shared/Integration/Scripting/Python/InstallDependencies.py
hnrkndrssn: Suggested we provide a helpful message with some steps the user can take (similar to what we do for IIS) to ensure that they can run python scripts on their targets (i.e. install `pip` or install the `pycryptodome` package using their package manager of choice on the target)

PR #349, opened 2018-12-18
Fixed Windows Service deployment delay issue

brinkie2004:
This script used WMI to check the current service status. On one of our machines this caused the deployment step to take 15 minutes or more. Therefor we replaced the WMI with a standard PS equivalent command. Field testing shows this works as expected. The 15 minute delay has gone.

### DIFF
--- a/source/Calamari/Scripts/Octopus.Features.WindowsService_BeforePostDeploy.ps1
+++ b/source/Calamari/Scripts/Octopus.Features.WindowsService_BeforePostDeploy.ps1
@@ -177,8 +177,9 @@ if ($description)
 	}
 }
 
-$wmiServiceName = $serviceName -replace "'", "\'"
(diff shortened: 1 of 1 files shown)


tothegills:
Hi @brinkie2004,

Thank you for the contribution.  The `StartType` property was introduced in PowerShell 5 so for the sake of backwards compatibility we are using `Get-WMIObject`. Until we raise the minimum PowerShell requirement I'm afraid I can't accept your PR.

Cheers

brinkie2004:
Has the minimal Powershell limit been raised yet? We still face this issue today.

EjaYF:
@brinkie2004 Maybe you can reopen this pull request by adding a check to use your new code for PowerShell>5 only. See https://stackoverflow.com/a/1825807/3883467 on how to get that done ;)

PR #350, opened 2018-12-19
Script syntax conditions on script wrappers

tothegills:
Currently we are attempting to run script types that are not PowerShell in the Azure PowerShell Context. The Azure PowerShell Context is only compatible with PowerShell scripts, so add a condition to ensure the Azure PowerShell wrapper and other wrappers are only included in the chain for the script syntax they are compatible with.

Relates to OctopusDeploy/Issues#5178

### DIFF
--- a/source/Calamari/Kubernetes/KubernetesContextScriptWrapper.cs
+++ b/source/Calamari/Kubernetes/KubernetesContextScriptWrapper.cs
@@ -31,9 +31,9 @@ namespace Calamari.Kubernetes
         /// <summary>
         /// One of these fields must be present for a k8s step
         /// </summary>
-        public bool IsEnabled(ScriptSyntax syntax) => !string.IsNullOrEmpty(variables.Get(SpecialVariables.ClusterUrl, "")) ||
+        public bool IsEnabled(ScriptSyntax syntax) => (!string.IsNullOrEmpty(variables.Get(SpecialVariables.ClusterUrl, "")) ||
                                !string.IsNullOrEmpty(variables.Get(SpecialVariables.AksClusterName, "")) ||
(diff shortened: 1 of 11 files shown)


source/Calamari/Kubernetes/KubernetesContextScriptWrapper.cs
```diff
@@ -28,9 +31,11 @@ public KubernetesContextScriptWrapper(CalamariVariableDictionary variables)
         /// <summary>
         /// One of these fields must be present for a k8s step
         /// </summary>
-        public bool Enabled => !string.IsNullOrEmpty(variables.Get(SpecialVariables.ClusterUrl, "")) ||
+        public bool IsEnabled(ScriptSyntax syntax) => !string.IsNullOrEmpty(variables.Get(SpecialVariables.ClusterUrl, "")) ||
                                !string.IsNullOrEmpty(variables.Get(SpecialVariables.AksClusterName, "")) ||
-                               !string.IsNullOrEmpty(variables.Get(SpecialVariables.EksClusterName, ""));
+                               !string.IsNullOrEmpty(variables.Get(SpecialVariables.EksClusterName, "")) &&
+                                supportedScriptSyntax.Contains(syntax);
```
hnrkndrssn: This needs to have `()` around the `!string.IsNullOrEmpty` calls otherwise if either `ClusterUrl` or `AksClusterName` is set and the `syntax` is not supported it will return `true`. E.g.

```
(!string.IsNullOrEmpty(variables.Get(SpecialVariables.ClusterUrl, "")) ||
    !string.IsNullOrEmpty(variables.Get(SpecialVariables.AksClusterName, "")) ||
    !string.IsNullOrEmpty(variables.Get(SpecialVariables.EksClusterName, ""))) &&
    supportedScriptSyntax.Contains(syntax)
```

PR #351, opened 2019-01-02
Helm chart values

zentron:
Added change in Octostache & Server such that the "default" package will now show up as an "empty indexed" item in the packages variable

### DIFF
--- a/source/Calamari/Deployment/Journal/DeployedPackage.cs
+++ b/source/Calamari/Deployment/Journal/DeployedPackage.cs
@@ -43,13 +43,22 @@ namespace Calamari.Deployment.Journal
         {
             var variables = deployment.Variables;
 
+            if (!string.IsNullOrWhiteSpace(variables.Get(SpecialVariables.Package.NuGetPackageId)))
+            {
+                yield return new DeployedPackage(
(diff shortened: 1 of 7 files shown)


zentron:
Related to https://github.com/OctopusDeploy/OctopusDeploy/pull/3272

source/Calamari/Deployment/Journal/DeployedPackage.cs
```diff
@@ -54,6 +54,11 @@ public static IEnumerable<DeployedPackage> GetDeployedPackages(RunningDeployment
 
             foreach (var packageReferenceName in variables.GetIndexes(SpecialVariables.Packages.PackageCollection))
             {
+                if (string.IsNullOrEmpty(packageReferenceName) && variables.IsSet(SpecialVariables.Package.NuGetPackageId))
```
MJRichardson: I don't understand this line?  Why are we checking `Octopus.Action.Package.NuGetPackageId` is set?

zentron: This is to check if the code above has already handled the default package. I'm thinking it makes sense to allow calamari to be backwards compatible with the old variable location.

PR #354, opened 2019-01-15
Only run PowerShell version test on OSes that support PowerShell 2

tothegills:
Missing Server 2016 but I think there is such a low risk it's not worth the time it would take to implement.

### DIFF
--- a/source/Calamari.Tests/Fixtures/PowerShell/PowerShellFixture.cs
+++ b/source/Calamari.Tests/Fixtures/PowerShell/PowerShellFixture.cs
@@ -26,6 +26,7 @@ namespace Calamari.Tests.Fixtures.PowerShell
         [Test]
         [Platform]
         [Category(TestCategory.CompatibleOS.Windows)]
+        // Windows 2016 (has PowerShell 2) will also match Windows 2019 (no PowerShell 2) so have omitted it.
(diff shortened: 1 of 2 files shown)


source/Calamari.Tests/Fixtures/PowerShell/PowerShellFixture.cs
```diff
@@ -28,9 +24,10 @@ public void Teardown()
         }
 
         [Test]
+        [Platform]
         [Category(TestCategory.CompatibleOS.Windows)]
-        [TestCase("2", "PSVersion                      2.0")]
-        [TestCase("2.0", "PSVersion                      2.0")]
+        [TestCase("2", "PSVersion                      2.0", IncludePlatform = "Win2008Server,Win2008ServerR2,Win2012Server,Win2012ServerR2")]
```
matt-richardson: Bit scared about that massive space in the `expectedLogMessage`, but I think I'll go 🙈.

The expected platforms should probably include windows 10 (for local dev) and Windows 2016.

tothegills: Added Windows 10. The nunit implementation to detect for Server 2016 will also match Server 2019 so we need to roll our own. My thinking was that for an obscure feature with such a low risk of failing on just one server version it was not worth rolling our own.

matt-richardson: Sounds reasonable. Wanna add a comment to that effect?

PR #356, opened 2019-01-17
Call Disable-AzureRMContextAutosave

johnsimons:
As mentioned in https://github.co…m/Azure/azure-powershell/issues/7110#issuecomment-426750910

Fixes https://github.com/OctopusDeploy/Issues/issues/5216


johnsimons:
@benPearce1 done, so much better :)

PR #361, opened 2019-01-29
Allow overriding namespace on Helm Upgrade step

MJRichardson:
See https://github.com/OctopusDeploy/Issues/issues/5141


MJRichardson:
[Allow overriding namespace in Helm Upgrade step](https://trello.com/c/SK9GpW5r/167-allow-overriding-namespace-in-helm-upgrade-step)

PR #371, opened 2019-02-21
Add handling for special headers when uploading to S3 buckets

shaunmarx:
This PR addresses the following issues:
* https://github.com/OctopusDeploy/Issues/issues/5302
* https://github.com/OctopusDeploy/Issues/issues/5323

It adds special handling where the AWS SDK has failed to do so by moving common Metadata into headers where it's necessary. This PR adds support for the following headers from metadata:

* Cache-Control
* Content-Disposition
* Content-Encoding
* Expires
* x-amz-website​-redirect-location
* x-amz-object-lock-mode

As part of this PR, metadata comparisons have also been added in order to ensure that changes will be applied to S3 objects when their content has not changed.

### DIFF
--- a/source/Calamari.Aws/Deployment/Conventions/UploadAwsS3Convention.cs
+++ b/source/Calamari.Aws/Deployment/Conventions/UploadAwsS3Convention.cs
@@ -379,9 +379,6 @@ namespace Calamari.Aws.Deployment.Conventions
             }
         }
 
- 
-
-
         /// <summary>
         /// The AmazonServiceException can hold additional information that is useful to include in
         /// the log.
--- a/source/Calamari.Aws/Deployment/Conventions/UploadAwsS3Convention.cs
+++ b/source/Calamari.Aws/Deployment/Conventions/UploadAwsS3Convention.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Net;
@@ -316,7 +317,7 @@ namespace Calamari.Aws.Deployment.Conventions
                 if (!await ShouldUpload(client, request))
                 {
                     Log.Verbose(
-                        $"Object key {request.Key} exists for bucket {request.BucketName} with same content hash. Skipping upload.");
+                        $"Object key {request.Key} exists for bucket {request.BucketName} with same content hash and metadata. Skipping upload.");
                     return new S3UploadResult(request, Maybe<PutObjectResponse>.None);
                 }
 
@@ -365,7 +366,7 @@ namespace Calamari.Aws.Deployment.Conventions
                     return true;
 
                 var metadata = await client.GetObjectMetadataAsync(request.BucketName, request.Key);
-                return !metadata.GetEtag().IsSameAsRequestMd5Digest(request);
+                return !metadata.GetEtag().IsSameAsRequestMd5Digest(request) || !S3ObjectExtensions.MetadataEq(request.ToCombinedMetadata(), metadata.ToCombinedMetadata());
             }
             catch (AmazonServiceException exception)
             {
@@ -377,6 +378,8 @@ namespace Calamari.Aws.Deployment.Conventions
                 throw;
             }
         }
+
+ 
 
 
         /// <summary>
--- a/source/Calamari.Aws/Deployment/Conventions/UploadAwsS3Convention.cs
+++ b/source/Calamari.Aws/Deployment/Conventions/UploadAwsS3Convention.cs
@@ -365,8 +365,8 @@ namespace Calamari.Aws.Deployment.Conventions
                 if (!md5HashSupported)
                     return true;
 
-                var metadata = await client.GetObjectMetadataAsync(request.BucketName, request.Key);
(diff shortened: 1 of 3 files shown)


source/Calamari.Aws/Deployment/Conventions/UploadAwsS3Convention.cs
```diff
@@ -365,7 +366,7 @@ private async Task<bool> ShouldUpload(AmazonS3Client client, PutObjectRequest re
                     return true;
 
                 var metadata = await client.GetObjectMetadataAsync(request.BucketName, request.Key);
-                return !metadata.GetEtag().IsSameAsRequestMd5Digest(request);
+                return !metadata.GetEtag().IsSameAsRequestMd5Digest(request) || !S3ObjectExtensions.MetadataEq(request.ToCombinedMetadata(), metadata.ToCombinedMetadata());
```
michaelnoonan: Would this read more clearly as `metadata.HasSameMetadata(request)` or similar (using the extension method)? I had to read the line several times to realise it's trying to express "has the file content changed, or has the metadata changed".

source/Calamari.Aws/Deployment/Conventions/UploadAwsS3Convention.cs
```diff
@@ -378,6 +379,8 @@ private async Task<bool> ShouldUpload(AmazonS3Client client, PutObjectRequest re
             }
         }
 
+
```
michaelnoonan: Whitespace? ;)

source/Calamari.Aws/Integration/S3/S3ObjectExtensions.cs
```diff
@@ -4,16 +4,98 @@
...
+        {
+            return GetCombinedMetadata(request.Headers, request.Metadata);
+        }
+
+        public static IDictionary<string, string> ToCombinedMetadata(this GetObjectMetadataResponse response)
+        {
+            return GetCombinedMetadata(response.Headers, response.Metadata);
+        }
+
+        private static (IList<T> filtered, IList<T> excluded) SelectWithExclusions<T>(this IEnumerable<T> values, Func<T, bool> predicate)
```
michaelnoonan: I'm admittedly feeling confused by this method. It's only used in a single place, and I wonder if you could inline and simplify it into a few simple linq expressions inside `MetadataEq`?

shaunmarx:
Yes, it will attempt to re-upload the file as unfortunately there isn't a metadata API we can use to set it out of band. We could maybe change it to use the CopyObject api instead and provide the new metadata, but not sure if that will do what we want it to, could be worth a quick test.

source/Calamari.Aws/Integration/S3/S3ObjectExtensions.cs
shaunmarx: I have moved the method inline for clarity and got rid of the aggregate call

source/Calamari.Aws/Deployment/Conventions/UploadAwsS3Convention.cs
shaunmarx: Moved this into an extension method, so should read a bit better in the location where it is actually being used

source/Calamari.Aws/Integration/S3/S3ObjectExtensions.cs
```diff
@@ -4,16 +4,101 @@
...
+
+        public static IEnumerable<string> GetKnownSpecialHeaderKeys()
+        {
+            return SupportedSpecialHeaders.Keys;
+        }
+
+        public static bool MetadataEq(IDictionary<string, string> next, IDictionary<string, string> current)
+        {
+            var allKeys = next.Keys.Union(current.Keys).Distinct();
+            var keysInBoth = new List<string>();
```
michaelnoonan: Nice! Now this is where I get all functional! :)

`keysInBoth = allKeys.Where(key => next.ContainsKey(key) && current.ContainsKey(key)).ToArray();`
`missingKeys = allKeys.Except(keysInBoth).ToArray()`
`differentValues = keysInBoth.Where(key => ...).ToArray()`
`return missingKeys.Count == 0 && differentValues == 0;`

:)

shaunmarx: Ah I think I see where this is going, getting the differing values gives us an opportunity to also add that information to the logs to give some indication at least of what has actually changed. Is that what you were thinking?

michaelnoonan: Well, it's a certain way of structuring these kinds of functions into decomposed queries. The knock on effects are:
1. The code becomes more readable since each query means something sensible, and the final result is usually an aggregate of those things
2. If it's easier to reason about each query, it's easier to maintain the code later
3. You can use this information (if you really want to) for logging/tracing/debugging.

It's just a habit I've built after falling in love with LINQ and idiomatic functional programming in C#. I find future self thanking my past self when I take the time to think this way.

shaunmarx:
@michaelnoonan you happy for this one to get merged? I think we can look at finding an alternative to update object metadata without uploading a file as a separate action. Does that sound reasonable?

SeanLMcCullough:
@shaunmarx Any idea when this will hit production?

shaunmarx:
@SeanLMcCullough this should have been released as part of [2019.1.10](https://github.com/OctopusDeploy/Issues/milestone/464?closed=1) and was released last night so you can get it from the [downloads page](https://octopus.com/downloads/2019.1.10) now :)

PR #373, opened 2019-02-28
Don't display certificate issues as warnings.

johnsimons:
As majority of time code works without accessing private key

Fixes https://github.com/OctopusDeploy/Issues/issues/5340


michaelnoonan:
Thanks @johnsimons. To help with the review, can you attach a before and after screenshot of the task logs? I want to see the real user experience change.

johnsimons:
@michaelnoonan we have never been able to exactly repro this issue 😞 
But I can explain it 😄 

So the before would be something like:
image

And the After:
image
So the red crossed square is gone, as you can see from the before, it was being logged twice.
And the blue square is now logged as `Info`

I hope this makes sense?

PR #378, opened 2019-03-21
Support additional args in Helm Upgrade Command

MJRichardson:
Relates to https://github.com/OctopusDeploy/Issues/issues/5407

### DIFF
--- a/source/Calamari/Kubernetes/Conventions/HelmUpgradeConvention.cs
+++ b/source/Calamari/Kubernetes/Conventions/HelmUpgradeConvention.cs
@@ -11,13 +11,13 @@ using Calamari.Integration.Processes;
 using Calamari.Integration.Scripting;
  using Newtonsoft.Json;
 
- namespace Calamari.Kubernetes.Conventions
+namespace Calamari.Kubernetes.Conventions
 {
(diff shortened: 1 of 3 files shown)


source/Calamari/Kubernetes/Conventions/HelmUpgradeConvention.cs
```diff
@@ -141,15 +142,25 @@ private void SetValuesParameters(RunningDeployment deployment, StringBuilder sb)
...
-        private static void SetTillerNamespaceParameter(RunningDeployment deployment, StringBuilder sb)
+        void SetAdditionalArguments(RunningDeployment deployment, StringBuilder sb)
+        {
+            var additionalArguments = deployment.Variables.Get(SpecialVariables.Helm.AdditionalArguments);
+
+            if (!string.IsNullOrWhiteSpace(additionalArguments))
+            {
+                sb.Append($" {additionalArguments}");
+            }
+        }
```
MJRichardson: This is the only actual change (plus the test case)

zentron:
I _think_ that looks ok. We don't really escape much of the rest of the parameters but that tents to be a little more controlled. Will sort that out if it becomes a problem. 🚢

PR #383, opened 2019-04-17
Allow Calamari's Write-* functions to take pipeline input 

stopthatastronaut:
Previously, we could only use Write-Verbose thus

`Write-Verbose "this is a message"`

Calling it in a pipeline would produce no output

Now, we can do the following:

```
Write-Verbose "this is a message"
"this is a message" | Write-Verbose
@("this is a message", "this is another message") | Write-Verbose
```


stopthatastronaut:
🙇 

I'll wait until the TC build comes through and then merge

stopthatastronaut:
tests blew up, so I made one tweak

PR #385, opened 2019-04-23
Add script module support for Python/C#/F#/Bash

matt-richardson:
Relates to https://github.com/OctopusDeploy/Issues/issues/5499

Also fixes support with the Python bootstrapper so that it works with `Octopus.Calamari.CopyWorkingDirectoryIncludingKeyTo`


source/Calamari.Shared/Integration/Scripting/Python/PythonScriptEngine.cs
```diff
@@ -5,44 +5,34 @@
...
+        public override ScriptSyntax[] GetSupportedTypes()
         {
             return new[] {ScriptSyntax.Python};
         }
 
-        public CommandResult Execute(
-            Script script,
-            CalamariVariableDictionary variables,
-            ICommandLineRunner commandLineRunner,
+        protected override IEnumerable<ScriptExecution> PrepareExecution(Script script, CalamariVariableDictionary variables,
```
tothegills: `IEnumerable<ScriptExecution>` bothers me a bit because I am expecting just a single execution. I understand it's required to run the python dependency script. I don't have a good suggestion for an alternative.

source/Calamari.Shared/Deployment/SpecialVariables.cs
```diff
@@ -15,6 +18,15 @@ public static string GetLibraryScriptModuleName(string variableName)
             return variableName.Replace("Octopus.Script.Module[", "").TrimEnd(']');
         }
 
+        public static ScriptSyntax GetLibraryScriptModuleLangauge(VariableDictionary variables, string variableName)
+        {
+            var expectedName = variableName.Replace("Octopus.Script.Module[", "Octopus.Script.Module.Language[");
+            var syntaxVariable = variables.GetNames().FirstOrDefault(x => x == expectedName);
```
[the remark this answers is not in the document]
matt-richardson: You know, I'm not convinced about that as an approach. (I know that's whats being talked about).
I'm really a fan of the capabilities approach, especially when we think about things like pre-preemptively showing a warning "your target needs to have powershell 5.1 on it", before executing a script, rather than during executing.
I also think the user has the most context on what script they want to run, and what language they want to use.

source/Calamari.Shared/Integration/Scripting/Bash/BashScriptBootstrapper.cs
```diff
@@ -110,5 +117,23 @@ public static string PrepareBootstrapFile(Script script, string configurationFil
             EnsureValidUnixFile(script.File);
             return bootstrapFile;
         }
+        
+        static IEnumerable<string> PrepareScriptModules(VariableDictionary variables, string workingDirectory)
+        {
+            foreach (var variableName in variables.GetNames().Where(SpecialVariables.IsLibraryScriptModule))
```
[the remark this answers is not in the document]
matt-richardson: I was going to, but... imagine a world where we have 2 script modules (bash and f#) and 3 steps (bash, f# and powershell).
Do you really want to log a message in this scenario? TBH, I cant think of a nice way of handling the mismatch between languages....

source/Calamari.Shared/Integration/Scripting/Python/PythonScriptEngine.cs
matt-richardson: I wonder if we (eventually) need to refactor the python dependency install and script execution into the one call somehow. Food for thought.

matt-richardson: Either that, or we introduce another method for "preparing an execution for installing dependencies"?
We could just not override it on the other methods? :thinking:
Stuff for the future.

source/Calamari.Shared/Integration/Scripting/WindowsPowerShell/Bootstrap.ps1
```diff
@@ -124,13 +124,6 @@ function Import-ScriptModule([string]$moduleName, [string]$moduleFilePath)
 		Write-Warning "Failed to import Script Module '$moduleName'"
 		Throw
 	}
-	Finally
```
matt-richardson: Related to https://github.com/OctopusDeploy/Issues/issues/5508

PR #386, opened 2019-04-24
Stop locking sdk version to 2.0.3

matt-richardson:
We had locked the sdk version to 2.0.3 as they made a breaking change in 2.1.3 - https://github.com/dotnet/sdk/issues/2342.

This PR removes the `global.json` files so that the build will just use the latest sdk available, and allow us to upgrade the SDK (to get security patches etc).

Driver for this was that I couldn't get calamari to build locally without deleting the `global.json` files, but, deleting them caused `ScriptCS` and `FSharp` not to be bundled.


zentron:
nice. this'll be much easier to use

PR #390, opened 2019-05-02
Fixed issue with using single quotes for terraform arguments

johnsimons:
So in Windows I am now executing Terraform.exe via PS


johnsimons:
How good is to have all these integration tests in Calamari 😄

PR #392, opened 2019-05-14
Bug 5560 terraform output

shaunmarx:
This PR drops the Powershell execution from Calamari and includes all output generated from executing the associated terraform commands. 

*Please Note* that dropping Powershell means that we now have the behavior as per the usual command line which means single quotes for additional parameters will no longer work, for example `-backend-config='backend.tfvars'` would not work whereas `-backend-config="backend-config"` would work on both windows and linux.

### DIFF
--- a/source/Calamari.Terraform/TerraformCliExecutor.cs
+++ b/source/Calamari.Terraform/TerraformCliExecutor.cs
@@ -38,7 +38,7 @@ namespace Calamari.Terraform
             AllowPluginDownloads = variables.GetFlag(TerraformSpecialVariables.Action.Terraform.AllowPluginDownloads, true);
             AttachLogFile = variables.GetFlag(TerraformSpecialVariables.Action.Terraform.AttachLogFile);
             TerraformVariableFiles = GenerateVarFiles();
-
+            
             logPath = Path.Combine(deployment.CurrentDirectory, "terraform.log");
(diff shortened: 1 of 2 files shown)


shaunmarx:
[Terraform step no longer generating outputs](https://trello.com/c/6UC695p4/842-terraform-step-no-longer-generating-outputs)

source/Calamari.Terraform/TerraformCliExecutor.cs
```diff
@@ -39,8 +38,9 @@ class TerraformCLIExecutor : IDisposable
             AllowPluginDownloads = variables.GetFlag(TerraformSpecialVariables.Action.Terraform.AllowPluginDownloads, true);
             AttachLogFile = variables.GetFlag(TerraformSpecialVariables.Action.Terraform.AttachLogFile);
             TerraformVariableFiles = GenerateVarFiles();
-
+            
             logPath = Path.Combine(deployment.CurrentDirectory, "terraform.log");
+            crashLogPath = Path.Combine(deployment.CurrentDirectory, "crash.log");
```
johnsimons: It may make sense to add a link to https://www.terraform.io/docs/internals/debugging.html#interpreting-a-crash-log so we know where this `crash.log` comes from

shaunmarx: Done :)

PR #397, opened 2019-05-22
Add optional flag to output kubeconfig as part of logs

distantcam:
Fixes a security issue where kubeconfig is dumped as part of logs, including private keys and tokens. This output is useful for debugging, but should not be done by default.

Now to turn on logging `Octopus.Action.Kubernetes.OutputKubeConfig' must be set and then the config is output in the verbose logs.


matt-richardson:
Randomly thought about this last night. Do we want to include a warning like we do for `OctopusPrintVariables` to say it's not safe?

PR #399, opened 2019-05-24
Lock python initialisation

johnsimons:


source/Calamari.Shared/Integration/Scripting/ScriptEngine.cs
```diff
@@ -26,7 +27,25 @@ public abstract class ScriptEngine : IScriptEngine
 
                 try
                 {
-                    result = commandLineRunner.Execute(execution.CommandLineInvocation);
+                    if (execution.CommandLineInvocation.Isolate)
+                    {
+                        using (var syncMutex = new Mutex(true, "PythonSynchronizeInitialization",
```
tothegills: The name is very specific to python but in a general implementation

johnsimons: what should we call it?

johnsimons: fixed

PR #402, opened 2019-05-31
Update Calamari to .Net Core 2.2

Waldo000000:
- Fixes https://github.com/OctopusDeploy/Issues/issues/4619
- [x] Flag end of Ubuntu 14.04 support ([docs](https://github.com/OctopusDeploy/docs/pull/416))
- [x] Check Ubuntu 14.04 behavior, to determine whether this is a patch or minor release --> an ad hoc test of script and package deployment steps completed successfully
- [x] Fix Mac OS X tests
- [ ] Upgrade Calamari in Server
- [ ] Retire Ubuntu 14.04 Calamari builds

### DIFF
--- a/source/Calamari.Shared/Integration/Packages/Download/MavenPackageDownloader.cs
+++ b/source/Calamari.Shared/Integration/Packages/Download/MavenPackageDownloader.cs
@@ -1,6 +1,5 @@
 ﻿using System;
-using System.Collections;
-using System.Collections.Generic;
+using System.Collections.Concurrent;
 using System.IO;
 using System.Linq;
 using System.Net;
@@ -205,7 +204,7 @@ namespace Calamari.Integration.Packages.Download
             Guard.NotNull(mavenPackageId, "mavenPackageId can not be null");
             Guard.NotNull(feedUri, "feedUri can not be null");
 
-            var errors = new HashSet<string>();
+            var errors = new ConcurrentDictionary<string, string>();
             var fileChecks = JarExtractor.EXTENSIONS
                 .Union(AdditionalExtensions)
                 .AsParallel()
@@ -214,7 +213,7 @@ namespace Calamari.Integration.Packages.Download
                     var packageId = new MavenPackageID(mavenPackageId.Group, mavenPackageId.Artifact,
                         mavenPackageId.Version, Regex.Replace(extension, "^\\.", ""));
                     var result = MavenPackageExists(packageId, feedUri, feedCredentials, snapshotMetadata);
-                    errors.Add(result.ErrorMsg);
+                    errors[result.ErrorMsg] = result.ErrorMsg;
                     return new
                     {
                         result.Found,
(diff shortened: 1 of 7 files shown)


Waldo000000:
Worth noting that when this ships, we should remove the `Test: Ubuntu 14.04 LTS - DotNet` build, as netcore 2.2 is [not supported](https://github.com/dotnet/core/blob/master/release-notes/2.2/2.2-supported-os.md) on Ubuntu 14.04.

johnsimons:
@Waldo000000 can we have a test for https://github.com/OctopusDeploy/Calamari/pull/402/commits/41269923d7a88df203243a726d5b4018de501e01 ?

Waldo000000:
@johnsimons It's already covered by `MavenPackageDownloaderFixture`, and `PackageDownloadFixture` tests which were previously flaky and intermittently failing, but should now pass reliably 🤞 

I'm happy to add more if you can think of any remaining gaps to cover?

johnsimons:
As long it is covered 👍

source/Calamari.Shared/Integration/Packages/Download/MavenPackageDownloader.cs
```diff
@@ -205,7 +204,7 @@ PackagePhysicalFileMetadata SourceFromCache(string packageId, IVersion version,
             Guard.NotNull(mavenPackageId, "mavenPackageId can not be null");
             Guard.NotNull(feedUri, "feedUri can not be null");
 
-            var errors = new HashSet<string>();
+            var errors = new ConcurrentDictionary<string, string>();
```
droyad: Can you use `ConcurrentBag` here instead?

source/Calamari/Program.cs
```diff
@@ -122,7 +122,7 @@ public static void EnableAllSecurityProtocols()
             //https://central.sonatype.org/articles/2018/May/04/discontinue-support-for-tlsv11-and-below/
 
             var securityProcotolTypes =
-#if !NETCOREAPP2_0
+#if !NETCOREAPP2_2
```
droyad: I think we should change this to a `HAS_SSL3` flag and add it here https://github.com/OctopusDeploy/Calamari/pull/402/files#diff-1ab97261a87deeb5e3c7c9f4ebf7bfcbR24

Then we don't need to keep updating it

droyad:
I'm unsure whether this should go in a minor or patch release since we are dropping support for Ubuntu 14

source/Calamari.Shared/Integration/Packages/Download/MavenPackageDownloader.cs
Waldo000000: Yeah I went with a dictionary to avoid `Distinct()`, but I don't really mind; have switched it over :+1:

Waldo000000:
>I'm unsure whether this should go in a minor or patch release since we are dropping support for Ubuntu 14

Let's go for a minor bump for Calamari, and just a patch bump for Server.

PR #404, opened 2019-06-10
Used "kubectl config view" instead of printing raw config

mcasperson:
Using `kubectl config view` will take advantage of the built in masking provided by kubectl. See the screenshot below - the top verbose output is our masking, and the bottom output is the result of  calling `kubectl config view` - note the `DATA+OMITTED` fields.

FYI - `kubectl` still leaks data, so we need to leave the variable guard in place. Details are at https://github.com/kubernetes/kubectl/issues/667.

image


mcasperson:
Fixes https://github.com/OctopusDeploy/Issues/issues/5616

PR #406, opened 2019-06-24
Improve recursive variable exception message

matt-richardson:
## Why

The current error message shown when a recursive variable is encountered is big, ugly and scary. It leads you to think that something really bad has gone wrong in Octopus:

image

## Result

Lets make the exception message nice and clear, and drop that stack trace.

## Screenshot

image

Relies on https://github.com/OctopusDeploy/Octostache/pull/34

### DIFF
--- a/source/Calamari.Shared/Commands/Support/ConsoleFormatter.cs
+++ b/source/Calamari.Shared/Commands/Support/ConsoleFormatter.cs
@@ -2,6 +2,7 @@
 using System.IO;
 using System.Reflection;
 using Calamari.Util;
+using Octostache.Templates;
 
 namespace Calamari.Commands.Support
 {
@@ -9,7 +10,7 @@ namespace Calamari.Commands.Support
     {
         public static int PrintError(Exception ex)
         {
-            if (ex is CommandException)
+            if (ex is CommandException || ex is RecursiveDefinitionException)
             {
                 Log.Error(ex.Message);
                 return 1;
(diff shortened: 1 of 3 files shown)


source/Calamari.Shared/Commands/Support/ConsoleFormatter.cs
```diff
@@ -14,6 +15,11 @@ public static int PrintError(Exception ex)
                 Log.Error(ex.Message);
                 return 1;
             }
+            if (ex is RecursiveDefinitionException)
+            {
+                //dont log these - they have already been logged earlier
+                return 101;
```
jburger: New return code... will this mean some assent changes in OD Server E2E tests req'd?

source/Calamari.Shared/Deployment/ConventionProcessor.cs
```diff
@@ -27,29 +29,39 @@ public void RunConventions()
...
+                    RunRollbackConventions();
 
+                    // Run cleanup for rollback conventions, for example: delete DeployFailed.ps1 script
+                    RunRollbackCleanup();
+                }
+                catch (Exception rollbackException)
+                {
+                    if (rollbackException is CommandException)
+                        Console.Error.WriteLine(rollbackException.Message);
+                    else if (rollbackException is RecursiveDefinitionException && rollbackException.Message != installException.Message)
```
jburger: Nice one avoiding any dupes :+1:

source/Calamari.Shared/Commands/Support/ConsoleFormatter.cs
matt-richardson: I doubt it...

PR #411, opened 2019-07-12
Added Update Progress helper methods.

droyad:
Resolves https://github.com/OctopusDeploy/Issues/issues/5696

### DIFF
--- a/source/Calamari.Shared/Integration/Scripting/Bash/Bootstrap.sh
+++ b/source/Calamari.Shared/Integration/Scripting/Bash/Bootstrap.sh
@@ -132,6 +132,17 @@ function new_octopusartifact
 }
 
 # -----------------------------------------------------------------------------
+# Function to update progress
+#	Accepts 2 arguments:
+#	  int: percentage progress
+#	  string: message to show
+# -----------------------------------------------------------------------------
+function update_progress
+{
+	echo "##octopus[progress percentage='$(encode_servicemessagevalue "$1")' message='$(encode_servicemessagevalue "$2")']"
+}
+
+# -----------------------------------------------------------------------------
 # Functions write a messages as different levels
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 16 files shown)


source/Calamari.Shared/Integration/Scripting/Bash/Bootstrap.sh
```diff
@@ -90,8 +90,8 @@ function set_octopusvariable
 		MESSAGE="$MESSAGE value='$(encode_servicemessagevalue "$2")'"
 	fi
 	
-	if [ ! -z "${3:-}" ] && [ "$3" = "-sensitive" ]
-	then
+	if [ ! -z "${3:-}" ] && [ "$3" = "-sensitive" ]
```
droyad: I think this is due to Rider normalising line endings

source/Calamari.Shared/Integration/Scripting/WindowsPowerShell/Bootstrap.ps1
```diff
@@ -356,6 +356,21 @@ function New-OctopusArtifact
     }
 }
 
+function Update-Progress
+{
+	[CmdletBinding()]
+	param(
+        [int] $percentage, 
+        [Parameter(ValueFromPipeline=$true)][string]$message
```
michaelnoonan: I wouldn't have picked the message as the `ValueFromPipeline`, but I also can't think of a reason not to pick the message.

droyad: Do yo think we should swap the parameters around? `Update-Process "Hello World" 50`

michaelnoonan: No, I like having the percentage first. :+1: I think it's good for go.

PR #414, opened 2019-07-19
Use WMI to set service username & password instead of sc.exe

matt-richardson:
Fixes https://github.com/OctopusDeploy/OctopusDeploy/issues/4099

Uses a bunch of tried-and-true code from Octopus & OctopusShared for the user creation

### DIFF
--- a/source/Calamari/Scripts/Octopus.Features.WindowsService_BeforePostDeploy.ps1
+++ b/source/Calamari/Scripts/Octopus.Features.WindowsService_BeforePostDeploy.ps1
@@ -97,19 +97,6 @@ if ($serviceAccount -ne "_CUSTOM")
 	}
 	$fullArgumentsSafeForConsole = $fullArguments
 }
-else
-{
-	if ($customAccountName)
-	{
-		$fullArguments += @("obj=", (ConvertTo-QuotedString $customAccountName))
-	}
-	$fullArgumentsSafeForConsole = $fullArguments
-	if ($customAccountPassword)
-	{
-		$fullArguments += @("password=", (ConvertTo-ConsoleEscapedArgument $customAccountPassword))
-		$fullArgumentsSafeForConsole += "password= `"************`""
-	}
-}
 
 $psServiceName = (ConvertTo-PowershellEscapedArgument $serviceName)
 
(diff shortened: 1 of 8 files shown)


source/Calamari/Scripts/Octopus.Features.WindowsService_BeforePostDeploy.ps1
```diff
@@ -158,6 +145,16 @@ else
 	}
 }
 
+if ($serviceAccount -eq "_CUSTOM") {
+	# dont use sc.exe to set the username / password, as it may be logged to the windows audit log if process creation event logs are enabled 
+	$wmiService = Get-WmiObject win32_service -filter "name='$serviceName'" -computer "."
+	$result = $wmiService.change($null, $null, $null, $null, $null, $null, $customAccountName, $customAccountPassword, $null, $null, $null)
+	if ($result.ReturnValue -ne "0") {
+		#return codes: https://docs.microsoft.com/en-us/windows/win32/cimwin32prov/change-method-in-class-win32-service#return-value
```
michaelnoonan: Maybe this URL could go into Octopurls and we reference that in the error message below? Not convinced it's too helpful though. 🤔

source/Calamari/Scripts/Octopus.Features.WindowsService_BeforePostDeploy.ps1
```diff
@@ -158,6 +145,16 @@ else
 	}
 }
 
+if ($serviceAccount -eq "_CUSTOM") {
+	# dont use sc.exe to set the username / password, as it may be logged to the windows audit log if process creation event logs are enabled 
+	$wmiService = Get-WmiObject win32_service -filter "name='$serviceName'" -computer "."
```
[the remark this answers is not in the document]
matt-richardson: @jburger - I've pushed a new commit to fix this

source/Calamari/Scripts/Octopus.Features.WindowsService_BeforePostDeploy.ps1
matt-richardson: I considered that, or even having a lookup here, but deferred that for the future to see how often we hit it.

PR #415, opened 2019-07-21
Upgraded YamlDotNet to fix license URL issue

michaelnoonan:
# Background

The versions of `YamlDotNet` we were using had an invalid license URL which broke our tests and our license attribution for external libraries. The author has since updated their packaging and fixed the problem. This PR updates the references in Octopus Server and Calamari.

See this conversation for more context: https://octopusdeploy.slack.com/archives/C9TUPEJ8K/p1563409987155800

Relates to https://github.com/OctopusDeploy/OctopusDeploy/pull/4105

# How to review this PR

@matt-richardson FYI
@mcasperson I couldn't see anything which would break, but this is a major version update of the library. Perhaps something will jump out at you? Otherwise, I'm relying on 💚 tests.

### DIFF
--- a/source/Calamari/Calamari.csproj
+++ b/source/Calamari/Calamari.csproj
@@ -45,7 +45,7 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
     </PackageReference>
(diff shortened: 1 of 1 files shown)


mcasperson:
We don't do anything clever with the YAML parsing (just extracting parameters for CloudFormation templates and generating YAML from classes for K8S), and I don't see anything in the release notes that would be an issue.

PR #420, opened 2019-08-19
Fix bug where Calamari doesn't respect the setting to bypass the default proxy in windows

flin-8:
There's a `TentacleUseDefaultProxy` environment variable that is not being respected when setting `WebRequest.DefaultProxy` as well as the standard proxy environment variable `NO_PROXY`.

The effect is that if you can't bypass the default proxy in windows.

In addition to fixing this behaviour, I've also:
- Moved the code that initializes `HTTP_PROXY`, `HTTPS_PROXY` and `NO_PROXY` into C# so all scripts get them now (previously only Powershell and Bash initalized them)
- Added tests for all scripting language to ensure this behaviour is tested
- Demoted the log message for `Setting Proxy Environment Variables` from INFO to VERBOSE level
It's probably easier to go through this commit by commit, as there are many file changes

Fixes:
https://github.com/OctopusDeploy/Issues/issues/5213
https://github.com/OctopusDeploy/Issues/issues/5784


source/Calamari.Tests/Fixtures/Integration/Proxies/ScriptProxyFixtureBase.cs
```diff
@@ -0,0 +1,140 @@
...
+#else
+        const string UrlEncodedProxyUserName = "some%40%3A%2Fuser";
+        const string UrlEncodedProxyPassword = "some%40%3A%2Fpassword";
+#endif
+        
+        protected const string proxyHost = "proxy-initializer-fixture-good-proxy";
+        protected const int proxyPort = 8888;
+
+        protected string proxyUrl = $"http://{proxyHost}:{proxyPort}";
+        string authentiatedProxyUrl = $"http://{UrlEncodedProxyUserName}:{UrlEncodedProxyPassword}@{proxyHost}:{proxyPort}";
```
tothegills: typo `authentiated` -> `authenticated`

source/Calamari.Shared/Integration/Proxies/ProxySettingsInitializer.cs
```diff
@@ -0,0 +1,41 @@
+using System;
+
+namespace Calamari.Integration.Proxies
+{
+    public static class ProxySettingsInitializer
+    {
+        public static ProxySettings GetProxySettingsFromEnvironment()
+        {
+            bool useDefaultProxy;
```
tothegills: I'm finding this part a little bit confusing.  We check if we should `useDefaultProxy`. We then check if we should `useCustomProxy`. The custom proxy setting then takes precedence over the default proxy setting. The order of the checks seems out of sync with the order of precedence which had me scratching me head a bit: why are the checks done in one order but the preference is the reverse order? The case where `TentacleUseDefaultProxy = true` and `TentacleProxyHost != empty` is not in the fixture, so I'm assuming the intended behaviour.

Also: `Boolean` can be `bool` for consistency, and the output variable can be inlined.

flin-8: yup it's a bit confusing, `TentacleProxyHost` takes precedence.

We only parse `TentacleUseDefaultProxy` at the top but don't do anything with it. I've moved it down below the `if (useCustomProxy)` check so it's clearer

PR #425, opened 2019-08-29
Allow for multiple sensitive variable files

tw17:
Due to the deployment pipeline change in Octopus Server, we can now send down multiple sensitive variable files.
This change to Calamari allows us to read multiple files

### DIFF
--- a/source/Calamari.Shared/Integration/Processes/CalamariVariableDictionary.cs
+++ b/source/Calamari.Shared/Integration/Processes/CalamariVariableDictionary.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Security.Cryptography;
 using System.Text;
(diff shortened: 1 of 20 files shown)


tothegills:
👍

source/Calamari.Shared/Integration/Processes/CalamariVariableDictionary.cs
```diff
@@ -31,23 +32,28 @@ public CalamariVariableDictionary(string storageFilePath, string sensitiveFilePa
                 nonSensitiveVariables.GetNames().ForEach(name => Set(name, nonSensitiveVariables.GetRaw(name)));
             }
 
-            if (!string.IsNullOrEmpty(sensitiveFilePath))
+            if (sensitiveFilePaths.Any())
```
droyad: You could have left out this block as the foreach will just be skipped on an empty collection

PR #426, opened 2019-09-02
Allow certificates with no public key to be installed into another users cert store

matt-richardson:
Relates to https://github.com/OctopusDeploy/Issues/issues/5825


source/Calamari.Tests/Fixtures/Certificates/WindowsX509CertificateStoreFixture.cs
```diff
@@ -176,6 +176,24 @@ public void CanImportCertificateForUser()
...
+            var sampleCertificate = SampleCertificate.CertWithNoPrivateKey;
+
+            sampleCertificate.EnsureCertificateNotInStore(storeName, StoreLocation.CurrentUser);
+
+            WindowsX509CertificateStore.ImportCertificateToStore(Convert.FromBase64String(sampleCertificate.Base64Bytes()), sampleCertificate.Password,
+                user, storeName, sampleCertificate.HasPrivateKey);
+
+            sampleCertificate.AssertCertificateIsInStore(storeName, StoreLocation.CurrentUser);
+
+            sampleCertificate.EnsureCertificateNotInStore(storeName, StoreLocation.CurrentUser);
```
michaelnoonan: How much do we care about wrapping this in a try/finally? This way we know the certificate was removed, or at least we tried, even if the test failed?

PR #429, opened 2019-09-04
Refactor powershell tests

TomPeters:
Some refactoring around powershell tests so that they can be parameterized (coming soon!) based on Powershell edition and environment (windows vs linux vs mac).

I have introduced a base class that each of the variants can inherit from in future
- windows powershell
- powershell core on windows
- powershell core on linux
- powershell core on mac

For now, only 3 of these variants exist, and the "powershell on [linux/mac]" variants are degenerate because we don't support powershell on these environments yet.

### DIFF
--- a/source/Calamari.Tests/Fixtures/PowerShell/PowerShellFixture.cs
+++ b/source/Calamari.Tests/Fixtures/PowerShell/PowerShellFixture.cs
@@ -7,21 +7,111 @@ using Calamari.Deployment;
 using Calamari.Integration.FileSystem;
 using Calamari.Integration.Scripting;
 using Calamari.Tests.Helpers;
-using Calamari.Util.Environments;
-using FluentAssertions;
 using Newtonsoft.Json;
 using NUnit.Framework;
 using Octostache;
-using SetProxy;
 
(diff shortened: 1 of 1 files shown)


gupta-kartik:
I noticed that we are running a C# script but no where do we have an assertion that checks that C# script runs. Should we have a test for that?

gupta-kartik:
Makes sense and LGTM 👍

TomPeters:
>I noticed that we are running a C# script but no where do we have an assertion that checks that C# script runs. Should we have a test for that?

I'm 90% sure that we aren't running a c# script. I think what's happening is that we are invoking calamari and saying "hey run this script" but then are giving it 3 different scripts to run. Since we are only telling calamari to run one script, it has to pick which of those scripts to run. The tests asserts that calamari takes into account the environment in which it is running to decide which of the scripts to run (i.e. on linux environments it ignores the powershell script).

I don't think this is a "real" scenario because you when you configure a Run script step in octopus, you can only ever give it a single inline script.

gupta-kartik:
What I mean is, we say "Hey run C# Script" but we are only ever testing against Bash and Powershell. Shouldn't we have a scenario where we test only a C# script (just like with PS and Bash)?

PR #435, opened 2019-09-10
Make proxies work with PowerShell Core

TomPeters:
Our existing proxy setup for windows powershell didn't work for powershell core

There are a few ways I'm aware of that you can do web requests in powershell
1. Using the powershell cmdlets `Invoke-WebRequest` and `Invoke-RestMethod`
2. Using `WebRequest`
3. Using `WebClient`
4. Using `HttpClient`

There are 5 interesting ways to configure your proxy
1. No proxy
2. Custom proxy (no auth)
3. Custom proxy (with auth)
4. System proxy (no auth)
5. System proxy (with auth)

The previous way we were configuring the proxy (for windows powershell) was `[System.Net.WebRequest]::DefaultWebProxy = $proxy`

This worked in all 4 of the ways you can do web requests in windows powershell, but does not work in all cases for powershell core. For 2 and 3 `WebRequest` and `WebClient`, it works well for powershell core.

For case 4 (`HttpClient`), PowerShell Core does not have a way of configuring a static default proxy in the current version of powershell (pre version 7.0.0), so there is no way for us to set this up globally for customers scripts. This has changed significantly from powershell vs powershell core.

For case 1, `Invoke-WebRequest` and `Invoke-RestMethod` use `HttpClient` under the hood, so they also suffer from the same problem. However, because these are cmdlets there is another option available, which is using [$PSDefaultParameterValues](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_parameters_default_values?view=powershell-6) to set a default proxy whenever these cmdlets are invoked.

So in summary, for PowerShell Core v6.x scripts
1. Using the powershell cmdlets `Invoke-WebRequest` and `Invoke-RestMethod` will work (through the magic of `$PSDefaultParameterValues`)
2. Using `WebRequest` works
3. Using `WebClient` works
4. Using `HttpClient` does not work. Customers will have to configure their `HttpClient` instances manually. This PR adds a new preconfigured `$OctopusProxy` parameter which customers can use in their scripts to configure `HttpClient`.

However, for PowerShell Core v7.0.0 and beyond, there exists a new mechanism for setting the Default Proxy for httpclient, which solves all of the above issues. See [this Issue](https://github.com/dotnet/corefx/issues/36553) and [this PR](https://github.com/dotnet/corefx/pull/37333) for more information. So this PR also uses this new feature if the detected version of powershell core is v7 or later, which means this PR becomes "future proof" and we won't need to come back and add more proxy support in a months time when PSCore v7 ships :)

I have manually tested this PR on windows powershell, powershell core v6, and powershell core v7, for the 5 proxy configuration scenarios described above, across the 4 different ways we can perform web request.

It would be nice to add automated tests but there's a bunch of work involved there. For now, I think we should just get by with manual testing until we have time to set up proxies on our test agents.

### DIFF
--- a/source/Calamari.Shared/Integration/Scripting/WindowsPowerShell/Bootstrap.ps1
+++ b/source/Calamari.Shared/Integration/Scripting/WindowsPowerShell/Bootstrap.ps1
@@ -543,7 +543,8 @@ function Initialize-ProxySettings()
 		#system proxy		
 		if ($useDefaultProxy)
 		{
-			$proxy = [System.Net.WebRequest]::GetSystemWebProxy()
+		    $proxyUri = New-Object Uri($env:HTTP_PROXY)
+            $proxy = New-Object System.Net.WebProxy($proxyUri)
 
 			if ($hasCredentials)
 			{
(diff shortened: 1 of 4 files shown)


gupta-kartik:
Looks good on first pass, need to check it out and try and break it. Will do that on Monday.

TomPeters:
@gupta-kartik It's quite a lot of effort to go to in order to test this properly. I'm fairly confident that I've tested it thoroughly already, so I'm happy for you to not spend the time testing this.

TomPeters:
@gupta-kartik Can you have another look? I've just added support for future versions of powershell core (v7 and beyond), which are _much simpler_ :tada:

TomPeters:
I've written some docs, see https://github.com/OctopusDeploy/docs/pull/527

gupta-kartik:
This looks so much better. The explanation above is also very clear. Good work Tom!

PR #436, opened 2019-09-13
Add PowerShell Core Tests for Windows OS

gupta-kartik:
Adding tests for PowerShell Core on Windows.


gupta-kartik:
This is a duplicate PR.

source/Calamari.Tests/Fixtures/PowerShell/PowerShellFixture.cs
```diff
@@ -18,15 +18,34 @@ namespace Calamari.Tests.Fixtures.PowerShell
 {
     [TestFixture]
     [Category(TestCategory.CompatibleOS.Windows)]
-    public class PowerShellCoreOnWindows : PowerShellFixture
+    public class WindowsPowerShellCoreFixture : PowerShellFixture
```
TomPeters: I think the previous name was better. The term `Windows PowerShell` has a very specific meaning (i.e. the traditional full .net framework version of powershell), which makes `WindowsPowerShellCore` very ambiguous.

source/Calamari.Tests/Fixtures/PowerShell/PowerShellFixture.cs
```diff
@@ -18,15 +18,34 @@ namespace Calamari.Tests.Fixtures.PowerShell
...
     {
         protected override PowerShellEdition PowerShellEdition => PowerShellEdition.PowerShellCore;
+
+        bool IsPowerShellInstalled()
+        {
+            var (output, _) = RunScript("PSCoreVersion.ps1");
```
TomPeters: `RunScript` runs a powershell script through `Calamari`.

- If PS Core is installed and everything is working, then the script will return `PowerShell x` in the output and the test will continue ✔️ 
- If PS Core is not installed and everything is working normally, then the script will not return `PowerShell x` in the output an the test will return an inconclusive result ✔️ 
- If PS Core is installed and something isn't working correctly (eg. a bug in calamari which causes calamari to crash) then the script will not include `Powershell x` in its output, and the test will return an inconclusive result 🔴.

This last case is bad because it could mask real issues.

It would be better to not use calamari at all when it comes to detecting if powershell is installed a pre-requisite of the test.

Are you familiar with the Arrange, Act, Assert pattern of writing unit tests? We only really want to invoke production code (in this case calamari) during the `Act` phase of the test. This part of the test (checking if powershell is installed) is part of the `Arrange` phase and should not use Calamari in any way, so as to be a more stable test.

source/Calamari.Tests/Fixtures/PowerShell/PowerShellFixture.cs
```diff
@@ -658,23 +682,8 @@ public void Dispose()
...
-            else
-            {
-                // Checking for not containing 'Core' as Build Servers run on
-                // PowerShell 3 which does not have PSEdition in the output
-                trimmedOutput.Should().NotContain(powerShellCoreEdition);
-            }
-        }
 
-        (CalamariResult result, VariableDictionary variables) RunPowerShellScript(string scriptName,
+        protected (CalamariResult result, VariableDictionary variables) RunPowerShellScript(string scriptName,
```
TomPeters: I don't think this changed was needed. Can you revert it?

source/Calamari.Tests/Fixtures/PowerShell/PowerShellFixture.cs
```diff
@@ -21,12 +21,23 @@ namespace Calamari.Tests.Fixtures.PowerShell
...
     {
         protected override PowerShellEdition PowerShellEdition => PowerShellEdition.PowerShellCore;
+
+        [SetUp]
+        public void SetUp()
+        {
+            CommandLineRunner clr = new CommandLineRunner(new IgnoreCommandOutput());
+            var result = clr.Execute(new CommandLineInvocation("pwsh.exe", "--version")); 
+            if (result.HasErrors)
+                Assert.Inconclusive();
```
TomPeters: Can you provide a message here? So that someone that only views the test results can see at a glance why this was an inconclusive result.

Something like:
```
Assert.Inconclusive("PowerShell Core is not installed in this machine");
```

PR #438, opened 2019-09-13
Allow usage of CompressPackage parameter during deployment to Azure ServiceFabric

vindmi:
This enables usage of `CopyPackageParameters/CompressPackage` parameter from application publish profile during deployment.
We needed in our organization to set this parameter `true` to solve problem when Octopus deployment occasionally got stuck during -
> Copying application package to image store...

As there is no possibility to specify this value in Octopus step parameters, I added it to the deployment script (similarly as it is in the original ServiceFabric deployment script).


slewis74:
We're closing this as we aren't willing to modify this file at the current time. We would highly recommend using your own version of the file when creating your package, see https://octopus.com/docs/deployment-examples/azure-deployments/deploying-to-service-fabric/deploying-a-package-to-a-service-fabric-cluster#customizing-the-deployment-process for details on how to override this file.

PR #439, opened 2019-09-16
Pick the latest installed powershell core version

TomPeters:
Users can install multiple powershell core versions side by side.

In this case, we should pick the latest powershell core version to run. We should also provide a means for those users to customize which version gets used, but this will come in a separate PR.

Right now, there are only two versions that can be installed, and these end up as different directories in `Program Files/PowerShell`:
- `6`
- `7-preview`

From this we can extrapolate a pattern for the directory structure:
`[Major Version]-[prerelease tag]`

Where the prerelease tag is optional.

PowerShell Core can be installed as a 64 or 32 bit program, which means it could end up in `Program Files` or `Program Files (x86)`.

Putting all these pieces together, this PR detects all installed versions of powershell across the two program files directories. It will order the installed versions based on the major version, and secondarily by the prerelease tag, in order to find the "latest" version installed. For example, if a user has both `6` and `7-preview` installed, it should pick `7-preview`. If they have both `7-preview` and `7` installed, it should pick `7`.

### DIFF
--- a/source/Calamari.Shared/Integration/Scripting/WindowsPowerShell/PowerShellBootstrapper.cs
+++ b/source/Calamari.Shared/Integration/Scripting/WindowsPowerShell/PowerShellBootstrapper.cs
@@ -56,13 +56,36 @@ namespace Calamari.Integration.Scripting.WindowsPowerShell
 
             try
             {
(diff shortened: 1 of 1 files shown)


gupta-kartik:
The code looks good. However, I think some unit tests there would be useful.

PR #440, opened 2019-09-16
Ability to specify which version of PowerShell Core is invoked

TomPeters:
By providing the `Octopus.Action.PowerShell.CustomPowerShellVersion` variable, users can specify which version of PowerShell Core is invoked if they have multiple installed. If this variable is provided, but the version can not be found, then an explicit error message is thrown which reveals the available version options on that target.

If they have a version installed with a prerelease tag, like `7-preview`, then they must include the prerelease tag in the version.

We can add tests as part of the https://trello.com/c/kPULmE6e/24-add-tests-for-detecting-the-latest-ps-core-version task.

### DIFF
--- a/source/Calamari.Shared/Integration/Scripting/WindowsPowerShell/PowerShellBootstrapper.cs
+++ b/source/Calamari.Shared/Integration/Scripting/WindowsPowerShell/PowerShellBootstrapper.cs
@@ -16,7 +16,7 @@ namespace Calamari.Integration.Scripting.WindowsPowerShell
         static string powerShellPath;
         const string EnvPowerShellPath = "PowerShell.exe";
         
-        public override string PathToPowerShellExecutable()
+        public override string PathToPowerShellExecutable(CalamariVariableDictionary variables)
         {
             if (powerShellPath != null)
             {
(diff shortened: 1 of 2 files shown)


source/Calamari.Shared/Integration/Scripting/WindowsPowerShell/PowerShellBootstrapper.cs
```diff
@@ -65,37 +71,63 @@ public override string PathToPowerShellExecutable()
...
             {
-                powerShellPath = EnvPowerShellPath;
+                return EnvPowerShellPath;
             }
+        }
 
-            return powerShellPath;
+        protected override IEnumerable<string> ContributeCommandArguments(CalamariVariableDictionary variables)
+        {
+            yield break;
```
gupta-kartik: What does this do?

source/Calamari.Shared/Integration/Scripting/WindowsPowerShell/PowerShellBootstrapper.cs
```diff
@@ -110,22 +142,21 @@ static PowerShellBootstrapper()
...
             var executeWithoutProfile = variables[SpecialVariables.Action.PowerShell.ExecuteWithoutProfile];
+            
+            foreach (var argument in ContributeCommandArguments(variables))
+                commandArguments.Append(argument);
+            
             bool noProfile;
             if (bool.TryParse(executeWithoutProfile, out noProfile) && noProfile)
             {
-                commandArguments.Append("-NoProfile ");
+                commandArguments.Append("-NoProfile ");;
```
gupta-kartik: minor - extra semicolon?

source/Calamari.Shared/Integration/Scripting/WindowsPowerShell/PowerShellBootstrapper.cs
TomPeters: Remind me at our next catch up and I'll run you through it

PR #442, opened 2019-09-18
Use sudo path when checking nginx is installed

thedewi:
This check was requiring nginx to be on the non-sudo `PATH`, which is a bit atypical and unnecessary, since our invocations of nginx use sudo. This adjustment checks the sudo `PATH` instead.

### DIFF
--- a/source/Calamari/Scripts/Octopus.Features.Nginx_BeforePreDeploy.sh
+++ b/source/Calamari/Scripts/Octopus.Features.Nginx_BeforePreDeploy.sh
@@ -32,14 +32,15 @@ function check_user_has_sudo_access_without_password_to_required_commands {
     fi
 }
 
-function check_app_exists {
-	command -v $1 > /dev/null 2>&1
+function check_nginx_exists {
+	sudo -n bash -c 'command -v nginx' > /dev/null 2>&1
 	if [[ $? -ne 0 ]]; then
-		echo >&2 "The executable $1 does not exist, or is not on the path"
-        echo >&2 "See https://g.octopushq.com/NginxInstall from more information"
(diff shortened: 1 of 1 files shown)


source/Calamari/Scripts/Octopus.Features.Nginx_BeforePreDeploy.sh
```diff
@@ -32,14 +32,15 @@ function check_user_has_sudo_access_without_password_to_required_commands {
...
-function check_app_exists {
-	command -v $1 > /dev/null 2>&1
+function check_nginx_exists {
+	sudo -n bash -c 'command -v nginx' > /dev/null 2>&1
 	if [[ $? -ne 0 ]]; then
-		echo >&2 "The executable $1 does not exist, or is not on the path"
-        echo >&2 "See https://g.octopushq.com/NginxInstall from more information"
+		echo >&2 "The executable 'nginx' does not exist, or is not on the sudo path,"
+		echo >&2 "or the user does not have 'sudo' access to it without a password."
+		echo >&2 "See https://g.octopushq.com/NginxInstall from more information."
```
matt-richardson: This feels like it needs to point to a page on our docs about "how to diagnose why nginx isn't found" which explains what the "sudo path" is. 
Directing them straight to "how to install nginx" doesn't quite feel right. (the page on our side could be pretty short - include a few sentences and then point them off to the nginx docs if they're not sure how to install)

thedewi: I agree "sudo path" is a little vague. Sudo environment configuration is a bit of a can of worms though, varying by distro/vintage/security preferences. I'll give it some more thought.

PR #443, opened 2019-09-22
Nginx temp path fixes

thedewi:
This adds 3 related fixes to the handling of Nginx temp paths:

- Avoid a misleading success state when nginx scripts can't find their temp files; now raising an error instead.
- Automatically create `Environment.SpecialFolder.LocalApplicationData` as needed (instead of returning an empty string, which caused temporary files to land in the current working directory on my test system).
- Use an absolute path for the base temp directory, to avoid referencing different locations as it is passed between contexts.

### DIFF
--- a/source/Calamari.Shared/Integration/FileSystem/CalamariPhysicalFileSystem.cs
+++ b/source/Calamari.Shared/Integration/FileSystem/CalamariPhysicalFileSystem.cs
@@ -349,7 +349,7 @@ namespace Calamari.Integration.FileSystem
 
         string GetTempBasePath()
         {
-            var path1 = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData);
+            var path1 = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData, Environment.SpecialFolderOption.Create);
             path1 = Path.Combine(path1, Assembly.GetEntryAssembly()?.GetName().Name ?? "Octopus.Calamari");
             path1 = Path.Combine(path1, "Temp");
             var path = path1;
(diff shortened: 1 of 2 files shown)


thedewi:
Although this solves one case where `Environment.GetFolderPath()` returns an empty string, perhaps we should guard against it happening any other way, because accidentally storing configuration under the working directory could have a significant impact.

Should we fall back to another location? Abort and fail the script?

droyad:
Any idea why we are using `LocalApplicationData` for a temp path instead of `Temp`? Can we fall back to `Temp`?

In what cases would `LocalApplicationData` not exist?

Is this related to an Issue?

thedewi:
> Any idea why we are using `LocalApplicationData` for a temp path instead of `Temp`? Can we fall back to `Temp`?

I'm not sure why, but it might be because of concern about `/tmp` being shared between users in linux? `LocalApplicationData` is a bit more isolated by being under `$HOME`, but that's not much security since it's often world-readable by default too. A better mitigation would be to revoke read permissions, although I'm not sure if there's a tidy cross-platform way to do that.

> In what cases would `LocalApplicationData` not exist?

In linux it is `$HOME/.local/share`, which doesn't exist by default on my systems.

> Is this related to an Issue?

Sorry, yes. I think this customer probably hit this bug: https://secure.helpscout.net/conversation/952149767/49961?folderId=2845840

thedewi:
Now I've remembered `GetTempBasePath()` was written far earlier with Windows in mind, so `LocalApplicationData` would have been chosen for other reasons in that context.

thedewi:
Fixes OctopusDeploy/Issues#5874.

PR #448, opened 2019-09-26
Combined os test categories

tothegills:
When we run Calamari unit tests, we exclude the categories that do not make sense for the agent the tests are running on.

For example: `/TestCaseFilter:"TestCategory != Windows & TestCategory != fsharp & TestCategory != scriptcs & TestCategory != PlatformAgnostic"`

We were assigning multiple OS categories to the same test, which would work if our test runs were inclusive rather than exclusive. Because they are exclusive, the test agents would skip tests where there were two OS categories but one was not applicable to the current build agent.

Give a category rule: `TestCategory != Nix`
The following test would not run on a Mac agent:
```
[Category("Nix")]
[Category("Mac")]
public void ATestThatCanRunOnLinuxOrMac
```

This PR adds a new category for tests that can run on Mac and Nix, and sets only a single category for all OS specific tests.

### DIFF
--- a/source/Calamari.Tests/AWS/S3Fixture.cs
+++ b/source/Calamari.Tests/AWS/S3Fixture.cs
@@ -40,7 +40,7 @@ namespace Calamari.Tests.AWS
         }
 
         [Test]
-        [Category(TestCategory.CompatibleOS.Windows)]
+        [Category(TestCategory.CompatibleOS.OnlyWindows)]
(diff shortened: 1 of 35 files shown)


tothegills:
[Fix Calamari build - test categories are not being observed correctly](https://trello.com/c/TL4nQ3xV/1533-fix-calamari-build-test-categories-are-not-being-observed-correctly)

PR #449, opened 2019-09-30
Power Shell Core Support

TomPeters:
This PR adds support for running scripts through PowerShell Core (`pwsh`) on Windows and non-Windows platforms.

# Feature Description

If Calamari is running on non-Windows environments, PowerShell Scripts (`.ps1` files) can now be executed. They will be run through `pwsh`, which is assumed to be available in the `$PATH` variable for the user running Calamari.

On Windows environments, PowerShell scripts will still be run through Windows Powershell (the desktop, full .net framework edition of powershell), unless Calamari has been explicitly configured to run the script on PowerShell Core. This is to preserve backwards compatibility.

To configure the edition that Calamari will use on Windows environments, provide the `Octopus.Action.PowerShell.Edition` variable with a value of either `Core` or `Desktop`.

All existing Windows PowerShell features are fully supported on PowerShell Core on both Windows and non-Windows targets, with a few exceptions and caveats listed below.

# Proxy

For version 6 of PowerShell Core, it does not appear possible to configure all http traffic to go through the proxy server configured by the `Tentacle*` on `HTTP_PROXY` environment variables. In particular, using the [HttpClient class](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=netframework-4.8) directly requires some manual set up in the powershell scripts in order for its web requests to use the proxy.

To make this set up more convenient, a new `$OctopusProxy` variable has been made available for scripts to use. For consistency, this variable has been made available in all versions and editions of PowerShell, across all platforms.

# Multiple versions

## Windows
Multiple versions of PowerShell Core can be installed side-by-side on Windows. If multiple versions are detected, the latest installed version will be used, even if that version is a pre-release version. By providing the `Octopus.Action.PowerShell.CustomPowerShellVersion` variable, you can customize the version that is used if multiple versions are installed.

## Non-windows
Multiple versions of PowerShell core can be installed side-by-side on many non-Windows targets. This may be difficult to do depending on what OS you are using, and what method you have used to install PowerShell Core. For example, if you have installed PowerShell Core through your OS's package manager, it may only be possible to installed one version at a time. However, if you installed PowerShell Core manually or through a snap package, you may be able to install multiple versions side-by-side. 

Due to the large variance in install locations, methods, and versioning schemes, the `Octopus.Action.PowerShell.CustomPowerShellVersion` variable is not supported for non-Windows targets and instead the `pwsh` executable must be available in the `$PATH` variable for the user running Calamari.

# Impersonation

For windows deployments using Windows PowerShell or PowerShell core, the scripts can be configured to run as another user by using the `Octopus.Action.PowerShell.UserName` and `Octopus.Action.PowerShell.Password` variables. These variables are not supported for non-Windows deployments


source/Calamari.Shared/Integration/Packages/Download/DockerImagePackageDownloader.cs
```diff
@@ -126,9 +124,9 @@ void PerformPull(string username, string password, string fullImageName, string
         
         string GetFetchScript(IScriptEngine scriptEngine)
         {
-            var syntax = new[] {ScriptSyntax.PowerShell, ScriptSyntax.Bash}
-                .First(syntx => scriptEngine.GetSupportedTypes().Contains(syntx));
-            
+            var syntax = scriptEngine.GetSupportedTypes()
```
TomPeters: This should instead simply be:
```
var syntax = ScriptSyntaxHelper.GetPreferredScriptSyntaxForEnvironment()
```

source/Calamari.Shared/Integration/Scripting/CombinedScriptEngine.cs
```diff
@@ -29,13 +29,15 @@ public CombinedScriptEngine(IEnumerable<IScriptWrapper> scriptWrapperHooks)
...
 
+
         public ScriptSyntax[] GetSupportedTypes()
         {
-            return (CalamariEnvironment.IsRunningOnNix || CalamariEnvironment.IsRunningOnMac)
-                ? new[] { ScriptSyntax.Bash, ScriptSyntax.PowerShell, ScriptSyntax.CSharp, ScriptSyntax.FSharp, ScriptSyntax.Python }
-                : new[] { ScriptSyntax.PowerShell, ScriptSyntax.CSharp, ScriptSyntax.FSharp, ScriptSyntax.Python };
-        }
+            var preferredScriptSyntax = new [] { ScriptSyntaxHelper.GetPreferredScriptSyntaxForEnvironment() };
+            var commonScriptSyntax =  new[] { ScriptSyntax.PowerShell, ScriptSyntax.CSharp, ScriptSyntax.FSharp, ScriptSyntax.Python };
```
TomPeters: You might rename this to be even more descriptive: `scriptSyntaxesSupportedOnAllPlatforms`

source/Calamari.Shared/Integration/Scripting/CombinedScriptEngine.cs
```diff
@@ -29,13 +29,15 @@ public CombinedScriptEngine(IEnumerable<IScriptWrapper> scriptWrapperHooks)
...
 
+            return preferredScriptSyntax.Concat(commonScriptSyntax).Distinct() as ScriptSyntax[];
```
TomPeters: Don't cast. The `.Distinct()` returns an `IEnumerable<ScriptSyntax>`, yet the return type needs to be a `ScriptSyntax[]`. Remember that an `IEnumerable<ScriptSyntax>` is a very thin interface that just says "This thing can be enumerated" and nothing else. This is opposed to an array like `ScriptSyntax[]` which is a much more specific type that has a very specific way of storing its data in memory. An `IEnumerable<T>` might not even be in memory all at once, like an array is.

While an array implements `IEnumerable<T>`, it is not always true that an `IEnumerable<T>` is an array. In fact, in this case it as actually not true, so this "soft cast" to fail, resulting in a `null`. (Side note: Look up the difference between a soft and hard cast).

The correct way to do this is to just got `.Distinct().ToArray()`

source/Calamari/Kubernetes/KubernetesContextScriptWrapper.cs
```diff
@@ -31,10 +31,13 @@ public KubernetesContextScriptWrapper(CalamariVariableDictionary variables)
...
-        public bool IsEnabled(ScriptSyntax syntax) => (!string.IsNullOrEmpty(variables.Get(SpecialVariables.ClusterUrl, "")) ||
-                               !string.IsNullOrEmpty(variables.Get(SpecialVariables.AksClusterName, "")) ||
-                               !string.IsNullOrEmpty(variables.Get(SpecialVariables.EksClusterName, ""))) &&
-                                supportedScriptSyntax.Contains(syntax);
+        public bool IsEnabled(ScriptSyntax syntax)
+        {
+            return (!string.IsNullOrEmpty(variables.Get(SpecialVariables.ClusterUrl, "")) ||
+                    !string.IsNullOrEmpty(variables.Get(SpecialVariables.AksClusterName, "")) ||
+                    !string.IsNullOrEmpty(variables.Get(SpecialVariables.EksClusterName, ""))) &&
+                   supportedScriptSyntax.Contains(ScriptSyntaxHelper.GetPreferredScriptSyntaxForEnvironment());
```
TomPeters: This should just be `syntax == ScriptSyntaxHelper.GetPreferredScriptSyntaxForEnvironment()`, and the `supportedScriptSyntax` field can be entirely removed.

PR #450, opened 2019-10-10
Printed out the process ID of Calamari and the shell

droyad:
Instead of printing
```
HostProcessName: Calamari
HostProcessName: powershell
```

it will print
```
HostProcess: Calamari (1234)
HostProcess: powershell (5678)
```


droyad:
[Print process Id](https://trello.com/c/p5IilAWI/1578-print-process-id)

PR #451, opened 2019-10-18
Update SharpCompress from 0.15.2 to 0.24.0

tw17:
The PR is to Update SharpCompress to latest version.

## How to test
Generally if all the tests are passing there is not much else to test.  The only other change in this PR is to provide a handler to WriteSymbolicLink, this handler prevents exceptions from being thrown when extracting an archive that contains Symbolic links.
There is a unit test making sure no exception is thrown, however if might be work doing a manual test to make sure a deployment of a package containing a symbolic link is successful.


tw17:
[Security vuln in SharpCompress](https://trello.com/c/YPFYLCZN/1594-security-vuln-in-sharpcompress)

PR #453, opened 2019-11-01
Warning about Management Certificate use for Web Apps

slewis74:
Updated the AzureWebAppConvention to emit a warning when it detects a Management Certificate is being used.

Relates to OctopusDeploy/Issues#5950


matt-richardson:
is it worth having a `try`/`catch` around the bit that fails?

slewis74:
The bit that fails just returns that the web app couldn't be found. Which is a legitimate case if the web app name or something like that was wrong. I.e. we can't distinguish it from other expected cases.

PR #455, opened 2019-11-06
Fixing relative file path converter that was url encoding the path

distantcam:
OctopusDeploy/Issues#5978

### DIFF
--- a/source/Calamari.Tests/Fixtures/Util/StringExtensionsFixture.cs
+++ b/source/Calamari.Tests/Fixtures/Util/StringExtensionsFixture.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using Calamari.Tests.Helpers;
 using Calamari.Util;
 using FluentAssertions;
 using NUnit.Framework;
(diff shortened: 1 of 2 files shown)


distantcam:
[S3 step doesn't handle spaces in filenames correctly](https://trello.com/c/kjIb8d1Q/2445-s3-step-doesnt-handle-spaces-in-filenames-correctly)

gupta-kartik:
LGTM as it's quite straightforward but (being new to the world of .NET) I would get a second pair of  👀 on it from @matt-richardson

source/Calamari.Tests/Fixtures/Util/StringExtensionsFixture.cs
```diff
@@ -27,5 +28,19 @@ public void NullValueShouldThrowException()
             Action action = () => "foo".ContainsIgnoreCase(null);
             action.Should().Throw<ArgumentNullException>();
         }
+
+        // This method has some weird legacy issues which we now rely on
+        [TestCase(@"C:\Path\To\File1.txt", @"C:\Path\", @"To/File1.txt")]
+        [TestCase(@"C:\Path\To\File2.txt", @"C:\Path", @"To/File2.txt")]
+        [TestCase(@"C:\Path\To\File3 With Spaces.txt", @"C:\Path", @"To/File3 With Spaces.txt")]
+        //[TestCase(@"C:/Path/To/File4.txt", @"C:/Path/", @"To/File4.txt")]
```
matt-richardson: Whats the go with this test case? Can we uncomment or delete?

gupta-kartik: I was a bit confused here with the `/` `\` changing in assertions in the tests 1-3. Are we doing this to check that both types of slashes have the same result?

distantcam: It's a weird case which produces `../To/File4.txt` which is an unexpected output. I'd prefer to delete.

This method returns some unexpected results when it comes to non-windows style paths. I'm uncomfortable changing the behaviour given we rely on it elsewhere.

distantcam: I've removed that comment

PR #458, opened 2019-11-25
Update all variables so long as they're in the mapped document

distantcam:
OctopusDeploy/Issues#6029

https://octopusdeploy.slack.com/archives/CNHBHV2BX/p1574314747210800

As per slack discussion, removing the prefix filter.

### DIFF
--- a/source/Calamari.Shared/Integration/JsonVariables/JsonConfigurationVariableReplacer.cs
+++ b/source/Calamari.Shared/Integration/JsonVariables/JsonConfigurationVariableReplacer.cs
@@ -188,8 +188,14 @@ namespace Calamari.Integration.JsonVariables
                 {
                     // Only include variables starting with 'Octopus'
                     // if it also has a colon (:)
-                    return v.StartsWith("Octopus:", StringComparison.OrdinalIgnoreCase)
-                        && map.ContainsKey(v);
+                    if (v.StartsWith("Octopus:", StringComparison.OrdinalIgnoreCase))
+                    {
+                        return map.ContainsKey(v);
+                    }
+                    else
+                    {
+                        return false;
+                    }
                 }
                 return map.ContainsKey(v);
             }
--- a/source/Calamari.Shared/Integration/JsonVariables/JsonConfigurationVariableReplacer.cs
+++ b/source/Calamari.Shared/Integration/JsonVariables/JsonConfigurationVariableReplacer.cs
@@ -182,7 +182,18 @@ namespace Calamari.Integration.JsonVariables
 
         public void Update(VariableDictionary variables)
         {
-            foreach (var name in variables.GetNames().Where(map.ContainsKey))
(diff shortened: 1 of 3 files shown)


distantcam:
[Allow variable replacement when variables are prefixed by Octopus](https://trello.com/c/KnDtfwmG/2555-allow-variable-replacement-when-variables-are-prefixed-by-octopus)

distantcam:
@droyad I've added to the checks to also look for a colon `:` in the variable after octopus, so that existing scenarios won't be broken. I've also added some test cases around this. Does this address your concern?

droyad:
/cc @MJRichardson 

I think we should document this behaviour and maybe add a log message if there is a root property in the json called "Octopus".

source/Calamari.Shared/Integration/JsonVariables/JsonConfigurationVariableReplacer.cs
```diff
@@ -182,18 +182,26 @@ private void PopPath()
...
-                    {
-                        map[name](variables.Get(name));
-                    }
-                    catch (Exception e)
-                    {
-                        Log.WarnFormat("Unable to set value for {0}. The following error occurred: {1}", name, e.Message);
-                    }
+                    // Only include variables starting with 'Octopus'
+                    // if it also has a colon (:)
+                    return v.StartsWith("Octopus:");
```
slewis74: Should there be a `StringComparison.OrdinalIgnoreCase` here?

source/Calamari.Shared/Integration/JsonVariables/JsonConfigurationVariableReplacer.cs
```diff
@@ -182,18 +182,26 @@ private void PopPath()
...
+                }
+                return map.ContainsKey(v);
+            }
```
slewis74: Can we rename `filter` to `VariableNameIsNotASystemVariable`?

source/Calamari.Shared/Integration/JsonVariables/JsonConfigurationVariableReplacer.cs
```diff
@@ -182,18 +182,26 @@ private void PopPath()
...
-                    }
-                    catch (Exception e)
-                    {
-                        Log.WarnFormat("Unable to set value for {0}. The following error occurred: {1}", name, e.Message);
-                    }
+                    // Only include variables starting with 'Octopus'
+                    // if it also has a colon (:)
+                    return v.StartsWith("Octopus:");
+                }
+                return map.ContainsKey(v);
```
slewis74: I think there is a logic gap here in the filter method. In the case where it starts with `Octopus:` should we not then be doing a `map.ContainsKey(v)`? I was expecting to just see something like ` || x.StartsWith("Octopus:", StringComparisn.OrdinalIgnoredCase` added to the original Where clause

PR #459, opened 2019-11-27
Support `Set-PSDebug -trace` to enable super detailed diagnostic traces

matt-richardson:
# background
I've been dealing with a few tickets lately where we've really had trouble understand whats going wrong.

# results 
This PR adds support for two new variables, `Octopus.Action.PowerShell.PSDebug.Trace` and `Octopus.Action.PowerShell.PSDebug.Strict`, allowing better diagnostics.

`Octopus.Action.PowerShell.PSDebug.Trace` allows both `int` values and `bool` values:
1. If the value is `0`, `1`, or `2` pass that directly through to the cmdlet, giving "power under the hood"
1. If the value is `false`(or `False`) make it `0`
1. If the value is `true` (or `True`) make it a default of `2`

This gives an uber detailed output of what powershell is doing:
image

`Octopus.Action.PowerShell.PSDebug.Strict` can be set to `true` (or `True`), and puts powershell into a stricter parsing mode where 
> variables must be assigned a value before being referenced in a script. If a variable is referenced before a value is assigned, an exception is thrown.

with the following code:
image

We get this:
image

# powershell compatibility

I've checked, and this is supported back to PS2.0, according to [this blog post[(https://www.jonathanmedd.net/2010/06/powershell-2-0-one-cmdlet-at-a-time-105-set-strictmode.html).

# first class support?

If we decide we like this - we could promote this, and powershell debugging up to be full on features like [powershell core is](https://octopus.com/docs/deployment-examples/custom-scripts/powershell-core).

# links

[internal discussion thread](https://octopusdeploy.slack.com/archives/CLQNM80U9/p1574748811032300)

### DIFF
--- a/source/Calamari.Shared/Integration/Scripting/WindowsPowerShell/PowerShellBootstrapper.cs
+++ b/source/Calamari.Shared/Integration/Scripting/WindowsPowerShell/PowerShellBootstrapper.cs
@@ -206,15 +206,21 @@ namespace Calamari.Integration.Scripting.WindowsPowerShell
             //https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/set-psdebug?view=powershell-6
             var traceArg = variables[SpecialVariables.Action.PowerShell.PSDebug.Trace];
             var traceCommand = "-Trace 0";
-            if (int.TryParse(traceArg, out var traceArgAsInt))
+            var powerShellVersion = ScriptingEnvironment.SafelyGetPowerShellVersion();
+            int.TryParse(traceArg, out var traceArgAsInt);
+            bool.TryParse(traceArg, out var traceArgAsBool);
+            if (traceArgAsInt > 0 || traceArgAsBool)
             {
-                Log.Warn($"{SpecialVariables.Action.PowerShell.PSDebug.Trace} is enabled,  This should only be used for debugging, and then disabled again for normal deployments.");
-                traceCommand = $"-Trace {traceArgAsInt}";
-            }
-            else if (bool.TryParse(traceArg, out var traceArgAsBool) && traceArgAsBool)
-            {
-                Log.Warn($"{SpecialVariables.Action.PowerShell.PSDebug.Trace} is enabled,  This should only be used for debugging, and then disabled again for normal deployments.");
-                traceCommand = "-Trace 2";
+                if (powerShellVersion.Major < 5 && powerShellVersion.Major > 0)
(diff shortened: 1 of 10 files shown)


source/Calamari.Shared/Integration/Scripting/WindowsPowerShell/PowerShellBootstrapper.cs
```diff
@@ -192,14 +193,41 @@ public string FormatCommandArguments(string bootstrapFile, string debuggingBoots
...
+            {
+                Log.Warn($"{SpecialVariables.Action.PowerShell.PSDebug.Trace} is enabled,  This should only be used for debugging, and then disabled again for normal deployments.");
+                traceCommand = "-Trace 2";
+            }
+            
+            var strictArg = variables[SpecialVariables.Action.PowerShell.PSDebug.Strict];
+            var strictCommand = "";
+            if (bool.TryParse(strictArg, out var strictArgAsBool) && strictArgAsBool)
+            {
+                Log.Info($"{SpecialVariables.Action.PowerShell.PSDebug.Strict} is enabled, putting PowerShell into strict mode where variables must be assigned a value before being referenced in a script. If a variable is referenced before a value is assigned, an exception will be thrown. This feature is experimental.");
```
michaelnoonan: Is the experimental bit necessary/helpful, or is it likely stay experimental forever?
```suggestion
                Log.Info($"{SpecialVariables.Action.PowerShell.PSDebug.Strict} is enabled, putting PowerShell into strict mode where variables must be assigned a value before being referenced in a script. If a variable is referenced before a value is assigned, an exception will be thrown.");
```

matt-richardson:
> However, untested. Is there a nice cheap/fast way we can test this behaviour inside Calamari?

Yeah, my bad. Added some tests.

source/Calamari.Shared/Integration/Scripting/WindowsPowerShell/PowerShellBootstrapper.cs
matt-richardson: I added that to give us the equivalence of the "EAP" way out if things dont work so well.
Also a way to see if it works in the real world of issue diagnosis - we'll be able to remove it / change it a lot.

source/Calamari.Tests/Fixtures/Integration/Scripting/ScriptEngineFixture.cs
```diff
@@ -25,6 +26,95 @@ public void PowershellDecryptsSensitiveVariables()
                 result.AssertOutput("KingKong");
             }
         }
+        
+        [Test]
+        [Category(TestCategory.CompatibleOS.OnlyWindows)]
```
michaelnoonan: Will this also work for PowerShell.Core on non-Windows?

PR #463, opened 2019-12-11
Feedz.io now needs mono 4.8 or higher to work

distantcam:
Feedz.io changed it's ssl certs, which is causing older versions of Mono to fail. The usual guidance here is to upgrade mono, so that's what we'll do.

### DIFF
--- a/source/Calamari.Tests/Fixtures/PackageDownload/PackageDownloadFixture.cs
+++ b/source/Calamari.Tests/Fixtures/PackageDownload/PackageDownloadFixture.cs
@@ -65,7 +65,7 @@ namespace Calamari.Tests.Fixtures.PackageDownload
         }
 
         [Test]
-        [Ignore("Failing on < Mono 5 Need to fix ASAP", Until = "2019-12-11")]
+        [RequiresMonoVersion480OrAbove]
         public void ShouldDownloadPackage()
         {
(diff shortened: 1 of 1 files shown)


distantcam:
> What if we catch the ugly ugly ugly exception and wrap it in something useful? "You appear to be using an old version of mono - this aint gonna work. please upgrade your version of mono"???

Originally I thought this would be a good idea, but then I started to think about how to implement this. The implementation would need code to tell if it's running on mono, and then what version of mono. I'm not sure the justification is worth it.

Instead, we currently have a section in our docs about how and why this is failing. http://octopus.com/docs/infrastructure/deployment-targets/linux/requirements#calamari-on-mono-limitations
Also I don't know how many customers are likely to face this situation, but surely it's ~0.

matt-richardson:
We've already got the code though? It's inside the `RequiresMonoVersion480OrAbove` attribute?

I'm thinking along the lines of reducing potential support calls. A little effort now saves a lot of pain in the future type thing.

All that said, I think i'd happily trade this effort for some real numbers about mono usage and versions...

distantcam:
I didn't realize that was calamari code checking for mono, not test code. Ok adding a cleaner exception should be possible, but I'll do it in another PR.

PR #468, opened 2020-01-13
Add timeout support to SilentProcessRunner via special variable to terminate stuck processes

miguelelvir:
**Could use a documentation update if accepted?**

# Issue
Tentacles wait indefinitely on actions which run scripts. For the most part, this is desirable by design and handled [here](https://github.com/OctopusDeploy/Calamari/blob/8.1.4/source/Calamari.Shared/Integration/Processes/SilentProcessRunner.cs#L152). There are many valid cases were a timeout on scripts should be supported to prevent tentacles from being indefinitely stuck processing action scripts (and thus preventing other operations such as clean up via Script Console, Runbooks, or other deployments).

One example of this is described in [this](https://stackoverflow.com/a/7588790) SO post and [this](https://forums.iis.net/t/1181715.aspx) IIS post when calling `CommitChanges` on IIS's ServerManager objects. This is important for Octopus scripts such as [IISWebSite_BeforePostDeploy](https://github.com/OctopusDeploy/Calamari/blob/master/source/Calamari/Scripts/Octopus.Features.IISWebSite_BeforePostDeploy.ps1#L273) that utilize the WebAdministration PowerShell module and are run on systems with high load. Cmdlets from this module internally call `appcmd`, which then calls the `CommitChanges` method of the `ServerManager` object. `CommitChanges` and appcmd may exit before the `Set-ItemProperty` method has called a Wait on it. As a result, random deployments can get stuck waiting for appcmd.
We have seen this behavior in many site deployments when using virtual paths.

## Samples
We captured some sample dumps and images showing this behavior. Images are included below:

Debugging Calamari dump showing that it gets stuck in `SilentProcessRunner`

Calamari Shared_WaitsForever

Powershell dump showing that it gets stuck in `Set-ItemProperty` setting a virtual path

powershell_IISWebSite_BeforePostDeploy_Set-ItemProperty_VirtualPath_Stuck

ProcessExplorer investigation to confirm IISWebSite_BeforePostDeploy is stuck on appcmd.
ProcessExplorerTree

# PR Changes
This PR adds the special variable `Octopus.Action.Script.Timeout` to allow a millisecond timeout to be defined via a process's variables that will be used to limit the amount of time a process can wait. If no timeout is defined, the default behavior of waiting indefinitely will be used. When it is defined, a message will be displayed on the Verbose log indicating a timeout value was used.
If the timeout is hit

# Tests
A test was added using Windows' [Timeout](https://ss64.com/nt/timeout.html) command to test that the timeout values are honored. Also deployed a Calamari package with the changes internally to our deployment server and confirmed it honors the timeout values when present.

Output after timeout (redacted):
image


source/Calamari.Shared/Integration/Processes/SilentProcessRunner.cs
```diff
@@ -149,11 +152,27 @@ static SilentProcessRunner()
...
 
-                        process.WaitForExit();
-                        outputWaitHandle.WaitOne();
-                        errorWaitHandle.WaitOne();
+                        // Some processes can have race conditions. Between the call to Start and WaitForExit part of the process may have exited already.
+                        // This can happen when calling CommitChanges in dotnet's ServerManager as shown here: https://stackoverflow.com/questions/7446632/servermanager-commitchanges-makes-changes-with-a-slight-delay
+                        // Commit changes is called by appcmd from IIS, which is called by Set-ItemProperty from the WebAdministration module in Powershell
+                        // https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.waitforexit?view=netframework-4.8#System_Diagnostics_Process_WaitForExit_System_Int32_
+                        // Add a timeout passed down from some default configuration or context up the call stack, -1 will never timeout.
+                        var timedOut = !process.WaitForExit(timeoutMilliseconds);
```
miguelelvir: Primary change here

source/Calamari.Shared/Integration/Processes/SilentProcessRunner.cs
```diff
@@ -149,11 +152,27 @@ static SilentProcessRunner()
...
 
-                        return new SilentProcessRunnerResult(process.ExitCode, errorData.ToString());
+                        if (!timedOut)
+                        {
+                            // Only wait when the process has not timed out, otherwise we will be stuck here as well.
+                            outputWaitHandle.WaitOne();
```
droyad: I think we still want to do this, as it'll finish the reading from the console. Would just adding the following work?
```
if(!processExited)
{
    Log.Error($"Process with ID {process.Id} exceeded the max allowed runtime of {timeoutMilliseconds} milliseconds and will be killed.");
    process.Kill();
    process.WaitForExit();
}
```

source/Calamari.Tests/Fixtures/Integration/Process/CommandLineRunnerFixture.cs
```diff
@@ -18,5 +20,24 @@ public void ScriptShouldFailIfExecutableDoesNotExist()
             result.HasErrors.Should().BeTrue();
             output.Errors.Should().Contain(CommandLineRunner.ConstructWin32ExceptionMessage(executable));
         }
+
+        [Test]
+        public void ScriptShouldFailWhenTimeoutIsSpecifiedAfterSomeTime()
```
droyad: This test would not work on Linux

droyad:
These kinds of timeouts have a habit of not working on Linux, a test should uncover that though.

source/Calamari.Shared/Integration/Processes/SilentProcessRunner.cs
miguelelvir: Works much better and cleaner to read. Much obliged :)

source/Calamari.Shared/Integration/Scripting/ScriptEngine.cs
```diff
@@ -31,6 +31,21 @@ public abstract class ScriptEngine : IScriptEngine
                         execution.CommandLineInvocation.Arguments);
                 }
 
+                if (variables.IsSet(SpecialVariables.Action.Script.Timeout))
+                {
+                    var timeout = variables.GetInt32(SpecialVariables.Action.Script.Timeout);
+                    execution.CommandLineInvocation.TimeoutMilliseconds = timeout ?? Timeout.Infinite;
```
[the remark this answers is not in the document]
miguelelvir: That's a much better idea. Consider it converted.

source/Calamari.Tests/Fixtures/Integration/Process/CommandLineRunnerFixture.cs
miguelelvir: I was thinking along similar lines. I'm not too sure if there are commands both platforms would change, so this test was converted into two instead: one for Windows platforms and one for Linux/Mac. It was done using the TestCategory.CompatibleOS attribute, though I'm only guessing at this point that it is what it is meant for.

miguelelvir:
> These kinds of timeouts have a habit of not working on Linux, a test should uncover that though.

Do you mean the timeout test command or the use of timeout in the Process.WaitForExit call? In *Nix the timeout one does indeed behave different, so the test would have failed. If needed, I'll have to spin up a linux tentacle container to test the Process.WaitForExit though.

miguelelvir:
> Thanks for the contribution. A few changes please.

Thanks for taking the time @droyad :) Really appreciate it.

miguelelvir:
@droyad Had reviewed the previous one too.

miguelelvir:
Closing in favor of #688

droyad (commented):
Oops, I commented but didn't submit

PR #470, opened 2020-01-15
Update CalamariPhysicalFileSystem.cs

i-know-beer:
Handle writing to NAS devices that the process has generic write permissions, but not write_dac permissions.

See https://stackoverflow.com/questions/59062659/file-replace-doesnt-work-on-shared-driver for reference.

### DIFF
--- a/source/Calamari.Shared/Integration/FileSystem/CalamariPhysicalFileSystem.cs
+++ b/source/Calamari.Shared/Integration/FileSystem/CalamariPhysicalFileSystem.cs
@@ -457,7 +457,7 @@ namespace Calamari.Integration.FileSystem
                 File.Copy(temporaryReplacement, originalFile, true);
             else
             {
-                System.IO.File.Replace(temporaryReplacement, originalFile, backup);
+                System.IO.File.Replace(temporaryReplacement, originalFile, backup, true);
             }
 
             File.Delete(temporaryReplacement);


source/Calamari.Shared/Integration/FileSystem/CalamariPhysicalFileSystem.cs
```diff
@@ -457,7 +457,7 @@ public void OverwriteAndDelete(string originalFile, string temporaryReplacement)
                 File.Copy(temporaryReplacement, originalFile, true);
             else
             {
-                System.IO.File.Replace(temporaryReplacement, originalFile, backup);
+                System.IO.File.Replace(temporaryReplacement, originalFile, backup, true);
```
matt-richardson: Hard to tell from the diff in github, but this is the `ignoreMetadataErrors` param, correct?

> ignoreMetadataErrors
> Boolean
> true to ignore merge errors (such as attributes and access control lists (ACLs)) from the replaced file to the replacement file; otherwise, false.

If so, this seems... dangerous. We could end up with files with incorrect ACLs, either breaking Octopus or ending up with files readable by people who arent meant to.

i-know-beer: Both files have the same acl so wouldn’t be an issue on the merge. Other option would be to set true for unc paths only.

PR #471, opened 2020-01-15
Exclude od-temp.example.com when merging bindings

tw17:
Fixes https://github.com/OctopusDeploy/Issues/issues/6110

When we create a IIS site with the `Octopus.Action.IISWebSite.ExistingBindings` variable set to true, we also merge our temp binding `:81:od-temp.example.com` with existing bindings.

This change explicitly excludes that binding from all potential bindings that we merge.


tw17:
[Unwanted od-temp.example.com binding in new IIS Website with merge existing bindings](https://trello.com/c/egAKFAcH/2733-unwanted-od-tempexamplecom-binding-in-new-iis-website-with-merge-existing-bindings)

PR #472, opened 2020-01-17
Remove use of File.Replace

i-know-beer:
Allow file operations to work on network drives that don't have full admin access (write_dec), and just the 'normal' write/modify file permission set


i-know-beer:
See https://stackoverflow.com/questions/59062659/file-replace-doesnt-work-on-shared-driver

PR #473, opened 2020-01-20
Use GetSystemWebProxy() when using default proxy

tw17:
Fixes: https://github.com/OctopusDeploy/Issues/issues/5953

This change brings back functionality to use `[System.Net.WebRequest]::GetSystemWebProxy()` for classic PowerShell on Windows targets.

For windows targets the only way to construct a proxy that also contains the bypass list as configured in IE / Internet Options is to make use of `GetSystemWebProxy()`

Tests have been update to include a test case for proxy bypasses.

To test manually:
1. Configure an environment to use a proxy for internet access
2. Add '127.0.0.1 proxytest.octopus.com' to your HOSTS file
3. Add proxytest.octopus.com to the proxy bypass list in Internet Options
4. Create a script step to run invoke-webrequest -uri http://proxytest.octopus.com/api
5. Create a release and deploy

### DIFF
--- a/source/Calamari.Shared/Integration/Scripting/WindowsPowerShell/Bootstrap.ps1
+++ b/source/Calamari.Shared/Integration/Scripting/WindowsPowerShell/Bootstrap.ps1
@@ -569,6 +569,7 @@ function Initialize-ProxySettings() {
 		if ($useDefaultProxy) {
 			# The system proxy should be provided through an environment variable, which has been used to initialize $proxyHost
 			if ($proxyUri -ne $null) {
+				# isWindows was introduced in PSCore and is not available for standard PowerShell
 				if (-not (Test-Path variable:IsWindows) -or $isWindows){
 					$proxy = [System.Net.WebRequest]::GetSystemWebProxy()
 				}
--- a/source/Calamari.Shared/Integration/Scripting/WindowsPowerShell/Bootstrap.ps1
+++ b/source/Calamari.Shared/Integration/Scripting/WindowsPowerShell/Bootstrap.ps1
@@ -569,7 +569,7 @@ function Initialize-ProxySettings() {
 		if ($useDefaultProxy) {
 			# The system proxy should be provided through an environment variable, which has been used to initialize $proxyHost
 			if ($proxyUri -ne $null) {
-				if($Env:OS -like "Windows*"){
+				if (-not (Test-Path variable:IsWindows) -or $isWindows){
 					$proxy = [System.Net.WebRequest]::GetSystemWebProxy()
 				}
 				else {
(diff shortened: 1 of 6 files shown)


tw17:
[no_proxy issue](https://trello.com/c/5WDYgVpU/2661-noproxy-issue)

source/Calamari.Shared/Integration/Scripting/WindowsPowerShell/Bootstrap.ps1
```diff
@@ -569,7 +569,12 @@ function Initialize-ProxySettings() {
 		if ($useDefaultProxy) {
 			# The system proxy should be provided through an environment variable, which has been used to initialize $proxyHost
 			if ($proxyUri -ne $null) {
-				$proxy = New-Object System.Net.WebProxy($proxyUri)
+				if($Env:OS -like "Windows*"){
```
flin-8: Would `if([System.Environment]::OSVersion.Platform -eq [System.PlatformID]::Win32NT){` be better?

source/Calamari.Tests/Fixtures/Integration/Proxies/WindowsScriptProxyFixtureBase.cs
```diff
@@ -97,6 +117,17 @@ protected override void AssertProxyBypassed(CalamariResult output)
                 output.AssertPropertyValue("WebRequest.DefaultProxy", "None");
         }
 
+        void AssertUnauthenticatedSystemProxyUsedWithException(CalamariResult output, string bypassedUrl)
+        {
+#if !NETCORE
+            AssertUnauthenticatedSystemProxyUsed(output);
+            if (IsRunningOnWindows && TestWebRequestDefaultProxy)
```
flin-8: Can remove `IsRunningOnWindows ` here, the test is only for Windows

source/Calamari.Shared/Integration/Scripting/WindowsPowerShell/Bootstrap.ps1
tw17: 👍 yea good call, that's a much better check

matt-richardson: What about `if ($IsWindows)`? Only works on netcore, so it'd have to be 
```
if (-not (Test-Path variable:IsWindows) || $isWindows)
```

source/Calamari.Shared/Integration/Scripting/WindowsPowerShell/Bootstrap.ps1
```diff
@@ -569,7 +569,12 @@ function Initialize-ProxySettings() {
 		if ($useDefaultProxy) {
 			# The system proxy should be provided through an environment variable, which has been used to initialize $proxyHost
 			if ($proxyUri -ne $null) {
-				$proxy = New-Object System.Net.WebProxy($proxyUri)
+				if (-not (Test-Path variable:IsWindows) -or $isWindows){
```
flin-8: It's not obvious that `$isWindows` is only defined for NetCore, please add a comment to explain.

PR #476, opened 2020-01-22
Fixed case where locations clobbered each other when sanitized

mcasperson:


source/Calamari.Tests/Fixtures/Nginx/NginxFixture.cs
```diff
@@ -214,5 +214,38 @@ public void ExecuteWorks()
             this.Assent(File.ReadAllText(rootConf), TestEnvironment.AssentConfiguration, $"{nameof(ExecuteWorks)}.rootLocation");
             this.Assent(File.ReadAllText(apiConf), TestEnvironment.AssentConfiguration, $"{nameof(ExecuteWorks)}.apiLocation");
         }
+        
+        [Test]
+        public void TestLocationsUnsitableForFilenames()
```
hnrkndrssn:
```suggestion
        public void TestLocationsUnsuitableForFilenames()
```

source/Calamari/Integration/Nginx/NginxServer.cs
```diff
@@ -141,9 +146,16 @@ public NginxServer WithAdditionalLocations(IEnumerable<Location> locations)
...
-            return match.Success 
-                ? match.Value.Replace("/", "_").Trim('_') 
-                : defaultValue;
+            if (match.Success)
+            {
+                // Watch out for cases where multiple locations both match a string like "/". Both
+                // of these will be sanitized down to an empty string.
+                var sanitizedResult = match.Value.Replace("/", "_").Trim('_');
+                // If we did not get an empty string, use the result. Otherwise fall back to the default.
+                if (!String.IsNullOrWhiteSpace(sanitizedResult)) return sanitizedResult;
```
hnrkndrssn: minor nitpick, but this should use `string` rather than `String`...See [this blog post](https://blog.paranoidcoding.com/2019/04/08/string-vs-String-is-not-about-style.html) for additional context as to why

```suggestion
                if (!string.IsNullOrWhiteSpace(sanitizedResult)) return sanitizedResult;
```

source/Calamari.Tests/Fixtures/Nginx/NginxFixture.cs
mcasperson: Good catch. I'll fix this up.

source/Calamari/Integration/Nginx/NginxServer.cs
mcasperson: 👍 I'll make this change.

mcasperson:
Fixes https://github.com/OctopusDeploy/Issues/issues/6121

PR #479, opened 2020-01-29
Fix Powershell $PSVersionTable log output on MaxOSX

flin-8:
After the Mac test agent was upgraded (OSX 10.12 to 10.15, Netcore to 3.1), the tests started failing.

`$PSVersionTable | Out-String` started resulting in blank lines for some reason, but only when running in the test. Invoking this in the `pwsh` terminal gives the right results.

Specifying the width manually fixed the problem, and shouldn't hurt other platforms, so I'm going with that.

### DIFF
--- a/source/Calamari.Shared/Integration/Scripting/WindowsPowerShell/Bootstrap.ps1
+++ b/source/Calamari.Shared/Integration/Scripting/WindowsPowerShell/Bootstrap.ps1
@@ -14,7 +14,7 @@ $ErrorActionPreference = 'Stop'
 
 function Log-VersionTable
 {
-	Write-Verbose ($PSVersionTable | Out-String)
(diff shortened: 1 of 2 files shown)


source/Calamari.Shared/Integration/Scripting/WindowsPowerShell/Bootstrap.ps1
```diff
@@ -14,7 +14,7 @@ $ErrorActionPreference = 'Stop'
 
 function Log-VersionTable
 {
-	Write-Verbose ($PSVersionTable | Out-String)
+	Write-Verbose ($PSVersionTable | Out-String -Width 200) # Calling Out-String without specify the width resulted in blank lines when running tests under MacOS for some reason
```
matt-richardson:
```suggestion
	Write-Verbose ($PSVersionTable | Out-String -Width 200) # Calling Out-String without specify the width resulted in blank lines when running tests under macOS 10.15 for some reason
```

PR #480, opened 2020-01-29
Package IDs with slashes break Nginx

mcasperson:
When a package ID includes a slash (for example, a GitHub release package), the NGINX step places the conf files in a subdirectory not scanned by NGINX. This PR removes any invalid characters from the package ID when generating a filename.

### DIFF
--- a/source/Calamari/Integration/Nginx/NginxServer.cs
+++ b/source/Calamari/Integration/Nginx/NginxServer.cs
@@ -18,14 +18,14 @@ namespace Calamari.Integration.Nginx
         private readonly string TempConfigRootDirectory = "conf";
         private readonly string TempSslRootDirectory = "ssl";
 
-        private bool useHostName;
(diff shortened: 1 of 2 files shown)


source/Calamari/Integration/Nginx/NginxServer.cs
```diff
@@ -37,7 +37,13 @@ public static NginxServer AutoDetect()
 
         public NginxServer WithVirtualServerName(string name)
         {
-            virtualServerName = name;
+            /*
+             * Ensure package ids with chars that are invalid for file names (for example, a GitHub package is in the format
+             * "owner/repository") do not generate unexpected file names.
+             */
```
mcasperson: This is the core change to filter out unexpected characters

mcasperson:
Fixes https://github.com/OctopusDeploy/Issues/issues/6137

PR #481, opened 2020-01-30
Download helm chart direct instead of using helm executable

shaunmarx:
This PR modifies our helm chart feed to download charts directly from an associated repository. This avoids issues with some external feeds such as bitbucket pages which has been returning the wrong headers for years now causing errors such as bad GZIP header errors.

Please note, this only deals with package references as the upgrade helm chart step pushes the chart from server to the tentacle.

### DIFF
--- a/source/Calamari.Shared/Integration/Packages/Download/PackageDownloaderStrategy.cs
+++ b/source/Calamari.Shared/Integration/Packages/Download/PackageDownloaderStrategy.cs
@@ -1,6 +1,8 @@
 ﻿using System;
 using System.Net;
+using System.Net.Http;
 using Calamari.Integration.FileSystem;
+using Calamari.Integration.Packages.Download.Helm;
 using Calamari.Integration.Processes;
 using Calamari.Integration.Scripting;
 using Octopus.Versioning;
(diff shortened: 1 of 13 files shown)


source/Calamari.Shared/Integration/Packages/Download/PackageDownloaderStrategy.cs
```diff
@@ -47,7 +49,8 @@ public PackageDownloaderStrategy(IScriptEngine engine, ICalamariFileSystem fileS
                     downloader = new GitHubPackageDownloader();
                     break;
                 case FeedType.Helm :
-                    downloader = new HelmChartPackageDownloader(fileSystem);
+                    var cred = feedCredentials.GetCredential(feedUri, "basic");
+                    downloader = new HelmChartPackageDownloader(fileSystem, new HelmEndpointProxy(new HttpClient(), feedUri, cred.UserName, cred.Password), new HttpClient());
```
flin-8: Since `feedCredentials` is a method argument to the downloader, can we get the downloader to pass it into the proxy via the method as well?

flin-8: Same with the Uri

PR #491, opened 2020-02-04
Use NormalizedFilename for s3 package upload

shaunmarx:
Use NormalizedFilename for s3 package upload since `deployment.PackageFilePath` or `OriginalFilename` may differ between built-in worker and external workers.


shaunmarx:
Alternate approach merged as part of https://github.com/OctopusDeploy/Calamari/pull/493

PR #500, opened 2020-02-18
Pulled free space check out of CalamariPhysicalFileSystem and into a service

droyad:
The skip and override were being resolved in multiple places and being set long before they were used. This extracts the free space check as a service so the entry points can be seen.

### DIFF
--- a/source/Calamari.Shared/Integration/FileSystem/FreeSpaceChecker.cs
+++ b/source/Calamari.Shared/Integration/FileSystem/FreeSpaceChecker.cs
@@ -1,0 +1,61 @@
+using System;
+using System.IO;
+using Calamari.Commands.Support;
+using Calamari.Deployment;
+using Calamari.Integration.Processes;
+using Octostache;
+
+namespace Calamari.Integration.FileSystem
+{
+    public interface IFreeSpaceChecker
+    {
+        void EnsureDiskHasEnoughFreeSpace(string directoryPath);
+    }
+
+    public class FreeSpaceChecker : IFreeSpaceChecker
+    {
+        readonly ICalamariFileSystem fileSystem;
+        readonly CalamariVariableDictionary variables;
+
+        public FreeSpaceChecker(ICalamariFileSystem fileSystem, CalamariVariableDictionary variables)
+        {
+            this.fileSystem = fileSystem;
+            this.variables = variables;
+        }
+
+        public void EnsureDiskHasEnoughFreeSpace(string directoryPath)
+        {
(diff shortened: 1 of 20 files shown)


source/Calamari.Shared/Integration/Packages/Download/PackageDownloaderStrategy.cs
```diff
@@ -40,18 +43,18 @@ public PackageDownloaderStrategy(IScriptEngine engine, ICalamariFileSystem fileS
             switch (feedType)
             {
                 case FeedType.Maven:
-                    downloader = new MavenPackageDownloader();
+                    downloader = new MavenPackageDownloader(fileSystem, freeSpaceChecker);
```
flin-8: Should the free space check & loading from cache be done by the strategy? that logic seems to be repeated (and missing) in some downloaders

source/Calamari.Shared/Integration/FileSystem/FreeSpaceChecker.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        void EnsureDiskHasEnoughFreeSpace(string directoryPath);
+    }
+
+    public class FreeSpaceChecker : IFreeSpaceChecker
+    {
+        readonly ICalamariFileSystem fileSystem;
+        readonly CalamariVariableDictionary variables;
+
+        public FreeSpaceChecker(ICalamariFileSystem fileSystem, CalamariVariableDictionary variables)
```
flin-8: Something feels off about using this bag of variables. Could we initialise some kind of settings class that gets initialised from the variables at startup? The intention being that less places know about and read from variables?

source/Calamari.Shared/Integration/Packages/Download/PackageDownloaderStrategy.cs
droyad: I did consider that, but Helm packages are not significant in size. Not sure about the Docker ones. Do they work on a remote target @MJRichardson  so local disk space isn't relevant?

source/Calamari.Shared/Integration/FileSystem/FreeSpaceChecker.cs
droyad: Yes, that is the intention as the refactor progresses. The Variables are actually already in the container as well.

source/Calamari.Shared/Integration/Packages/Download/PackageDownloaderStrategy.cs
flin-8: If it's not applicable for some, then we could declare it on `IPackageDownloader` and handle it in the strategy

PR #501, opened 2020-02-19
Removed IsSensitive from the variable dictionary

droyad:
image

This code no longer sparked joy. Since the server no longer sends down non-sensitive variable files, most variables will come into Calamari as "sensitive". The only time variables are not sensitive is if a offline drop does not have a password set, or it's an output variable passed from a previous offline step. 

This change removes the distinction between non-sensitive and sensitive variables and always treat the variable as sensitive. This means all variables will now be encrypted at in the bootstrapper file (at rest). 

Due to https://github.com/OctopusDeploy/Issues/issues/4840 sensitive output variables in offline drops were never included.

### DIFF
--- a/source/Calamari.Shared/Integration/Processes/CalamariVariableDictionary.cs
+++ b/source/Calamari.Shared/Integration/Processes/CalamariVariableDictionary.cs
@@ -13,8 +13,6 @@ namespace Calamari.Integration.Processes
 {
     public class CalamariVariableDictionary : VariableDictionary
     {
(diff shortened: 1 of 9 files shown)


droyad:
> There's a variables parameter to specify a variable file, which is passed to the VariableDictionary as storageFilePath. Should that get axed?

No because offline drops still use it when no password is supplied. However the sensitive parameter treats the file as plain text if no password is supplied. I'm unsure where that is actually used.

The input of the variables via files is separate to their resolution through the bootstrapper scripts. They all get dumped into the Variable Dictionary and then the variable dictionary is written into the bootstapper. At that point which file the variables came from is no longer relevant.

PR #503, opened 2020-02-20
Handle two-part extensions like "tar.gz" when parsing file names

emilol:
Fixes OctopusDeploy/Issues/issues/6199

.tar archives with a two-part extension aren't being parsed correctly when a package name includes release tags and/or build metadata.

This came up internally with a package named `helm-v2.9.1-windows-amd64.tar.gz`

## Before 

package name | .zip  .tar  .nupkg | .tar.gz  .tar.bz2 .tar.xyz | .abc.xyx 
--- | --- | --- | ---
foo.1.0.0  | ✔️ | ✔️  | ✔️ 
foo.1.0.0-tag   | ✔️ | ❌  | ❌ 
foo.1.0.0-tag-release.tag   | ✔️ | ❌  | ❌ 
foo.1.0.0+buildmeta   | ✔️ | ❌  | ❌
foo.1.0.0-tag-release.tag+buildmeta   | ✔️ | ❌  | ❌ 

## After

package name | .zip  .tar  .nupkg | .tar.gz  .tar.bz2 .tar.xyz | .abc.xyx 
--- | --- | --- | ---
foo.1.0.0  | ✔️ | ✔️  | ✔️ 
foo.1.0.0-tag   | ✔️ | ✔️ | ❌ 
foo.1.0.0-tag-release.tag   | ✔️ | ✔️ | ❌ 
foo.1.0.0+buildmeta   | ✔️ | ✔️ | ❌
foo.1.0.0-tag-release.tag+buildmeta   | ✔️ | ✔️ | ❌

### DIFF
--- a/source/Calamari.Shared/Integration/Packages/PackageName.cs
+++ b/source/Calamari.Shared/Integration/Packages/PackageName.cs
@@ -63,8 +63,6 @@ namespace Calamari.Integration.Packages
         /// <summary>
         ///  Old school parser for those file not yet sourced directly from the cache.
         /// Expects pattern {PackageId}.{Version}.{Extension}
-        /// Issue with extension parser if pre-release tag is present where two part extendsions are incorrectly split.
-        /// e.g. MyApp.1.0.0-beta.tar.gz  => .tar forms part of pre-release tag and not extension
         /// </summary>
         /// <param name="fileName"></param>
         /// <param name="packageId"></param>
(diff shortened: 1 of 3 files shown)


emilol:
[Calamari's GenericPackageExtractor doesn't handle extensions property (mistakes .tar.gz as .gz)](https://trello.com/c/KBNat530/2779-calamaris-genericpackageextractor-doesnt-handle-extensions-property-mistakes-targz-as-gz)

source/Calamari.Shared/Integration/Packages/PackageName.cs
```diff
@@ -80,13 +78,34 @@ static bool TryParseUnsafeFileName(string fileName, out string packageId, out IV
...
+            // so the below method strips out the extensions we know about before performing
+            // the match on the filename. If we don't find any extensions we know about
+            // we fall back to trying to pattern match one in the filename.
+            if (TryMatchKnownExtensions(fileName, out var strippedFileName, out var extensionMatch))
+            {
+                fileName = strippedFileName;
+            } else { 
+                const string extensionPattern = @"(?<extension>(\.([a-zA-Z0-9])+)+)"; //Extension (wont catch two part extensions like .tar.gz if there is a pre-release tag)
+                fileNamePattern += extensionPattern;
+            }
```
flin-8: What if we always stripped the extension off (not just `.tar.*`), then the regex to get the package ID can just be `$@"{packageIdPattern}\.{semanticVersionPattern}"`. It should make the code easier to follow

source/Calamari.Shared/Integration/Packages/PackageName.cs
```diff
@@ -100,6 +119,19 @@ static bool TryParseUnsafeFileName(string fileName, out string packageId, out IV
             return true;
         }
 
+        static bool TryMatchKnownExtensions(string fileName, out string strippedFileName, out Group extensionMatch)
```
flin-8: If we always strip the extension off, then we can just return a Tuple (filename, extension)

source/Calamari.Shared/Integration/Packages/PackageName.cs
emilol: it does, but we would miss out on any unknown two-part extensions for commonly formatted packages, i.e:

As Is

package name | .zip  .tar  .nupkg | .tar.gz  .tar.bz2 .tar.xyz | .abc.xyx 
--- | --- | --- | ---
foo.1.0.0  | ✔️ | ✔️  | ✔️ 
foo.1.0.0-tag   | ✔️ | ✔️ | ❌ 
foo.1.0.0-tag-release.tag   | ✔️ | ✔️ | ❌ 
foo.1.0.0+buildmeta   | ✔️ | ✔️ | ❌
foo.1.0.0-tag-release.tag+buildmeta   | ✔️ | ✔️ | ❌ 

If we change to make the code easier to follow

package name | .zip  .tar  .nupkg | .tar.gz  .tar.bz2 .tar.xyz | .abc.xyx 
--- | --- | --- | ---
foo.1.0.0  | ✔️ | ✔️  | ❌ 
foo.1.0.0-tag   | ✔️ | ✔️ | ❌ 
foo.1.0.0-tag-release.tag   | ✔️ | ✔️ | ❌ 
foo.1.0.0+buildmeta   | ✔️ | ✔️ | ❌
foo.1.0.0-tag-release.tag+buildmeta   | ✔️ | ✔️ | ❌

flin-8: I'm not sure how likely we are to come across new 2-part extensions, but fair point 👍

PR #504, opened 2020-02-20
Add warning message if select Helm version doesn't match the actual tool version

flin-8:
Otherwise the user might get confused when they don't match

image

### DIFF
--- a/source/Calamari.Tests/KubernetesFixtures/HelmVersionParserFixture.cs
+++ b/source/Calamari.Tests/KubernetesFixtures/HelmVersionParserFixture.cs
@@ -1,0 +1,34 @@
+using Calamari.Util;
+using FluentAssertions;
+using NUnit.Framework;
+
+namespace Calamari.Tests.KubernetesFixtures
+{
+    [TestFixture]
+    public class HelmVersionParserFixture
+    {
+        [Test]
+        public void ParseVersion_V2Version_V2()
+        {
+            HelmVersionParser.ParseVersion("Client: v2.16.1+gbbdfe5e").Should().Be(HelmVersion.V2);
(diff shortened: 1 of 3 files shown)


source/Calamari.Tests/KubernetesFixtures/HelmVersionParserFixture.cs
```diff
@@ -0,0 +1,34 @@
...
+        }
+        
+        [Test]
+        public void ParseVersion_OtherVersion_Null()
+        {
+            HelmVersionParser.ParseVersion("v4.0.2+g19e47ee").Should().BeNull();
+        }
+        
+        [Test]
+        public void ParseVersion_Rubbish_Null()
```
NickJosevski: do you see any value in testing the input string "V" you have a case in your code to check for it.

flin-8: sure, it's good document the limitations

flin-8: actually scrap that, it might be worth testing uppercase "V3.0.2" and asserting it returns null. I don't think a single "V" makes sense

NickJosevski: doesn't a single V ensure it enters the return statement of `if (indexOfVersionNumber >= versionString.Length)` not that it matters *that much*?

flin-8: Fair enough, I guess it's worth testing that it doesn't crash 😁

PR #510, opened 2020-02-24
Move Deployment Journal variable contributions into the factory

droyad:
Builds on #505 

Previously the `Previous` and `PreviousSuccessful` deployment variables were added as conventions on the deploy package step and java package step.

This change moves the addition of the variables into the `VariableFactory` and runs based on whether the `RetentionPolicySet` variable is supplied. This is only supplied for package deployments and is required to get the variables from the journal. 

It could be that some steps don't use the journal but supply the variable. In that case it's essentially a no-op.


source/Calamari/Variables/DeploymentJournalVariableContributor.cs
```diff
@@ -0,0 +1,69 @@
...
+            if (previous == null)
+            {
+                variables.Set(SpecialVariables.Tentacle.PreviousInstallation.OriginalInstalledPath, "");
+                variables.Set(SpecialVariables.Tentacle.PreviousInstallation.CustomInstallationDirectory, "");
+                variables.Set(SpecialVariables.Tentacle.PreviousInstallation.PackageFilePath, "");
+                variables.Set(SpecialVariables.Tentacle.PreviousInstallation.PackageVersion, "");
+            }
+            else
+            {
+                // This is assuming this convention is used only in steps with only one package
```
slewis74: Is this comment still valid? It's not a convention anymore, but is the single package still an assumption? I'd assume yes, but just checking

PR #511, opened 2020-02-24
Refactor how environment and additional file variables are contributed

droyad:
Builds on #507 

The majority of the changes lines are just shuffling from one file to another. Turn of whitespace changes in the GH diff, particularly for `Program.cs`.

Previously each command added a convention to contribute environment variable and print out the variables (`OctopusPrintVariables`). The only command that didn't do this were:
- The Extract Package Commands
- Download/Delta/Find package commands
- Check Health

The `Clean` command also didn't log variables.

This change makes the contribution of environment variables and reacting to the `OctopusPrintVariables` option global regardless of the command invoked.

The `AdditionalVariablesPath` variable is now also handled int the `VariableFactory` previously it was a convention that was only loaded by the `RunScript` command. This makes it always work, so less confusing.


source/Calamari/Commands/DeployPackageCommand.cs
```diff
@@ -73,10 +73,8 @@ public override int Execute(string[] commandLineArguments)
 
             var conventions = new List<IConvention>
             {
-                new ContributeEnvironmentVariablesConvention(),
                 new ContributePreviousInstallationConvention(journal),
                 new ContributePreviousSuccessfulInstallationConvention(journal),
```
droyad: This change does mean the variables contributed by this convention are not logged, however #510 re-instates this behaviour

source/Calamari/Variables/VariablesFactory.cs
```diff
@@ -13,66 +15,116 @@ namespace Calamari.Variables
 {
     public class VariablesFactory
```
droyad: Refactored this `Create` method into individual methods and cleaned up the code a bit

PR #512, opened 2020-02-25
Fix bug where multiple directives in NGINX step where getting overridden

tw17:
Calamari side fix for: https://github.com/OctopusDeploy/Issues/issues/6142
There is a bug in the NGINX step where we don't allow for multiple directives with the same name
We serialize the directives as a JSON object and duplicates with the same name/key get overridden.
example:
```
server {
    listen 80;
    location / {
        include fastcgi_params;
        include /etc/nginx/api_proxy.conf;
    }
}
``` 
would end up as 
```
server {
    listen 80;
    location / {
        include /etc/nginx/api_proxy.conf;
    }
}
```

We have made a change on the Server to serialize the directives as an array, and this change is for Calamari to handle that change.


tw17:
[NGINX location directives with the same name are overwritten](https://trello.com/c/aaNLOtWZ/2843-nginx-location-directives-with-the-same-name-are-overwritten)

PR #516, opened 2020-02-27
Respect the path component of the docker feed's Registry Path when calling `docker pull`

flin-8:
Fixes: https://github.com/OctopusDeploy/Issues/issues/6208

### DIFF
--- a/source/Calamari.Shared/Integration/Packages/Download/DockerImagePackageDownloader.cs
+++ b/source/Calamari.Shared/Integration/Packages/Download/DockerImagePackageDownloader.cs
@@ -37,18 +37,20 @@ namespace Calamari.Integration.Packages.Download
             ICredentials feedCredentials, bool forcePackageDownload, int maxDownloadAttempts, TimeSpan downloadAttemptBackoff)
         {
             //Always try re-pull image, docker engine can take care of the rest
-            var feedHost = GetFeedHost(feedUri);
-            var fullImageName =  GetFullImageName(packageId, version, feedUri, feedHost);
+            var fullImageName =  GetFullImageName(packageId, version, feedUri);
             var (username, password) = ExtractCredentials(feedCredentials, feedUri);
             
+            var feedHost = GetFeedHost(feedUri);
             PerformPull(username, password, fullImageName, feedHost);
             var (hash, size) = GetImageDetails(fullImageName);
             return new PackagePhysicalFileMetadata(new PackageFileNameMetadata(packageId, version, ""), fullImageName, hash, size );
         }
 
-        static string GetFullImageName(string packageId, IVersion version, Uri feedUri, string feedHost)
+        static string GetFullImageName(string packageId, IVersion version, Uri feedUri)
         {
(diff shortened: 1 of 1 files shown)


source/Calamari.Shared/Integration/Packages/Download/DockerImagePackageDownloader.cs
```diff
@@ -37,18 +37,20 @@ public DockerImagePackageDownloader(IScriptEngine scriptEngine, ICalamariFileSys
...
             return new PackagePhysicalFileMetadata(new PackageFileNameMetadata(packageId, version, ""), fullImageName, hash, size );
         }
 
-        static string GetFullImageName(string packageId, IVersion version, Uri feedUri, string feedHost)
+        static string GetFullImageName(string packageId, IVersion version, Uri feedUri)
         {
-            return feedUri.Host.Equals(DockerHubRegistry) ?  $"{packageId}:{version}" : $"{feedHost}/{packageId}:{version}";
+            return feedUri.Host.Equals(DockerHubRegistry) 
+                ? $"{packageId}:{version}" 
+                : $"{feedUri.Authority}{feedUri.AbsolutePath.TrimEnd('/')}/{packageId}:{version}";
```
flin-8: Keep the path component, are we going to break other people who have a path in their feedUrl but need the path stripped when calling `docker pull`?

source/Calamari.Shared/Integration/Packages/Download/DockerImagePackageDownloader.cs
```diff
@@ -37,18 +37,20 @@ public DockerImagePackageDownloader(IScriptEngine scriptEngine, ICalamariFileSys
...
         }
 
         static string GetFeedHost(Uri feedUri)
```
emilol: This can be removed if it's no longer being used. I'm assuming you've tested with an https/port 443 url?

flin-8: Yup, the tests cover that:
https://github.com/OctopusDeploy/Calamari/blob/e7819a76cd51eb9ea7a9e5dee2b8cecbc64ee1ac/source/Calamari.Tests/Fixtures/Integration/Packages/DockerImagePackageDownloaderFixture.cs#L16

flin-8: > I think this makes sense.
> 
> So this implies no one had ever used a docker registry hosted on a path before (at least where the image was pulled on the target)...
> 
> There is a little bit of confusion in the mapping between a feed and package ID in our world, and a repository and image in the docker world. The line between the two is arbitrary.
> 
> But ultimately it all becomes a URL in the end, and I don't see why it isn't valid to use a path in the Octopus feed: `https://artifactory.acme.com/foo/bar`
> 
> Worth testing with both DockerHub and an AWS ECR registry, but LGTM.
> 
> There is some confusion between

I couldn't get AWS ECR to work manually (both with and without my changes), but the E2E tests pass:
https://build.octopushq.com/viewLog.html?buildId=1247087&buildTypeId=OctopusDeploy_OctopusServer_TestOrchestration_TestE2eWindowsVersionAgnostic&tab=testsInfo

PR #517, opened 2020-02-28
Split Calamari Azure to allow running more things on Linux

hnrkndrssn:
### X-Plat all-the-things _(well, some more things)_
- Azure scripts
- K8S authentication with Azure service principals
- K8S authentication with AWS account

image


source/Calamari.Azure/HealthChecks/CloudServiceHealthChecker.cs
```diff
@@ -1,56 +0,0 @@
-﻿using System;
```
michaelnoonan: Bit of a bummer the renames weren't followed.

PR #520, opened 2020-03-02
Allowed NGINX config file paths to be overwritten

mcasperson:
Fixes https://github.com/OctopusDeploy/Issues/issues/6216

See https://github.com/OctopusDeploy/OctopusDeploy/pull/5363 for the UI changes.

The end result of these changes are a directory structure like this:

```
mcasperson@DESKTOP-3E4K4R8:/etc/nginx/conf.d$ ls -la
total 20
drwxr-xr-x 2 root root 4096 Mar  3 07:50 .
drwxr-xr-x 8 root root 4096 Mar  3 07:41 ..
-rw-r--r-- 1 root root   48 Mar  3 07:45 Dev.Octopus.TeamCity.conf
-rw-r--r-- 1 root root   48 Mar  3 07:42 Octopus.TeamCity.conf
-rw-r--r-- 1 root root   48 Mar  3 07:47 Test.Dev.Octopus.TeamCity.conf
```


mcasperson:
Ran through this implementation with Michael Richardson to verify the approach.

PR #523, opened 2020-03-06
Dont cleanup temp bootstrapper files if DeleteScriptsOnCleanup set to…

matt-richardson:
… false

We were cleaning up the bootstrapper files all the time, even if the `OctopusDeleteScriptsOnCleanup` variable was set to `false`.

@droyad - pinging you on this, as I believe you have context around this. Doesn't have to be you though.


matt-richardson:
Closing as stale - unfortunately I don't have time to invest in this right now (or in the foreseeable future), and as I've forgotten most of the context, I think closing is the wisest move.

PR #524, opened 2020-03-06
Refactored proxy initializing to remove the indirection

droyad:
Refactored proxy initializing to remove the indirection when dealing with `ProxySettings` to make it easer to figure out how `IWebProxy` and the environment variables are created.

The implementation of the visitor pattern made it really difficult to grok what was happening.

Added tests for proxy initialization and setting of the CalamariWorkingDirectory


source/Calamari.Shared/Integration/Proxies/IProxySettings.cs
```diff
@@ -0,0 +1,101 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
+using Octopus.CoreUtilities;
+
+namespace Calamari.Integration.Proxies
+{
+    public interface IProxySettings
```
flin-8: We should take more care in preserving file history when renaming the file (minimise add/deletes), it makes annotating harder in the future.

PR #525, opened 2020-03-09
Fixed a bug where an nginx exact root location "= /" is not recognized

tw17:
Calamari side fix for: https://github.com/OctopusDeploy/Issues/issues/6156

## How to test

1. Setup Server to use this build of Calamari
2. Add a `Deploy to NGINX` step
3. Define a location like `= /`
4. Perform a deployment
5. Ensure that the deployment is successful with no exception.
6. Try a deployment with both locations `= /` and `/` to ensure they are both configured


tw17:
[Nginx Step expects to have a root location, but does not parse the exact match prefix](https://trello.com/c/c3lh1xLk/2880-nginx-step-expects-to-have-a-root-location-but-does-not-parse-the-exact-match-prefix)

tw17:
> Is there a test matching this scenario?
Yea the unit test covers shit scenario

PR #526, opened 2020-03-10
Tweak gitversion to respect our tags on master when calculating versions

hnrkndrssn:
Because we have branch `release/2019.3` which has been merged into `master` GitVersion sometimes gets a little bit 😕 with which version it should use when building `master`. 

This PR "demotes" the `release` branch configuration from a _release_ branch which takes precedence when calculating versions.

### DIFF
--- a/GitVersion.yml
+++ b/GitVersion.yml
@@ -1,4 +1,7 @@
 mode: ContinuousDeployment
+branches:
+  release:
+    is-release-branch: false
 ignore:
   sha: [
     e823e5382ae16d433cbfda913f593f92aaaa462f,
--- a/build.cake
+++ b/build.cake
@@ -243,8 +243,10 @@ private void DoPackage(string project, string framework, string version, string 
 
 private void SignAndTimestampBinaries(string outputDirectory)
 {
+    // When building locally signing isn't really necessary and it could take up to 3-4 minutes to sign all the binaries 
+    // as we build for many, many different runtimes so disabling it locally means quicker turn around when doing local development.    
     if (BuildSystem.IsLocalBuild) return;
-    
+
     Information($"Signing binaries in {outputDirectory}");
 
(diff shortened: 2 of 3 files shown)


GitVersion.yml
```diff
@@ -1,4 +1,7 @@
 mode: ContinuousDeployment
+branches:
+  release:
+    is-release-branch: false
```
hnrkndrssn: This is the crux of the PR which stops GitVersion using `release/*` branches as part of calculating versions for builds.

build.cake
```diff
@@ -243,6 +243,8 @@ private void DoPackage(string project, string framework, string version, string
 
 private void SignAndTimestampBinaries(string outputDirectory)
 {
+    if (BuildSystem.IsLocalBuild) return;
```
hnrkndrssn: When working on the xplat pitch I ran **a lot** of 🍰 builds locally, when building locally signing isn't _really_ necessary and it could take up to 3-4 minutes to sign all the binaries as we build for many, many different runtimes so disabling it locally means quicker turn around when doing local development.

build.cake
```diff
@@ -1,7 +1,7 @@
 //////////////////////////////////////////////////////////////////////
 // TOOLS
 //////////////////////////////////////////////////////////////////////
-#tool "nuget:?package=GitVersion.CommandLine&version=4.0.0-beta0012"
+#tool "nuget:?package=GitVersion.CommandLine&version=5.2.0"
```
hnrkndrssn: We've had good success with this version of GitVersion so might as well update Calamari to it as well.

build.cake
matt-richardson: You may want to add a comment to the code explaining the why

hnrkndrssn:
>  build looks unhappy.

PR #528, opened 2020-03-19
Refactored `CommandLineRunner` so that it is injected everywhere

droyad:
This also makes it immutable 🙌. It now takes the command output options from the `CommandLineInvocation`

PR #529, opened 2020-03-20
Injected the log in all the things

droyad:
image

The main drive of this PR is so we can access the service message overloads from `ILog` for the extensions.

Refactored the Log to move the service message methods into `ILog`. Inverted the relationship between the static Log and the log instance. Also added a base class as abstract so it can be used in the tests as well.

I almost gave up on this until I found `ProxyLog` and decided to persist to get rid of it.

The core changes are in `Log.cs`. In the tests is some refactoring how they capture logging. The rest is mainly injecting and passing the log instead of using the static one.

There were 400 usages of the static log, this brings it down to 300 :/


source/Calamari.Shared/Log.cs
```diff
@@ -7,225 +7,206 @@
...
-        }
-
-        public void ErrorFormat(string message, params object[] args)
-        {
-            Log.ErrorFormat(message, args);
-        }
-
-    }
-
     public interface ILog
```
droyad: The key changes are in this file

source/Calamari.Azure.WebApps/Deployment/Conventions/AzureWebAppConvention.cs
```diff
@@ -76,7 +83,7 @@ public void Install(RunningDeployment deployment)
                             DeploymentSyncOptions(variables)
                         );
 
-                    Log.Info(
+                    log.InfoFormat(
```
droyad: For some reason `Info` didn't follow the patterns of all the others

source/Calamari.Shared/Log.cs
```diff
@@ -7,225 +7,206 @@
...
         void Info(string message);
-        void InfoFormat(string message, params object[] args);
+        void InfoFormat(string messageFormat, params object[] args);
         void Warn(string message);
-        void WarnFormat(string message, params object[] args);
+        void WarnFormat(string messageFormat, params object[] args);
         void Error(string message);
-        void ErrorFormat(string message, params object[] args);
+        void ErrorFormat(string messageFormat, params object[] args);
+        void SetOutputVariableButNotDoAddToVariables(string name, string value, bool isSensitive = false);
```
droyad: Renamed this overload to make it clear it doesn't add to the current variable dictionary. Looking at the callers most of them do anyway. Left for another day.

source/Calamari.Tests/Fixtures/Integration/Packages/PackageExtractorFixture.cs
```diff
@@ -25,7 +26,7 @@ public void ExtractPumpsFilesToFilesystem(Type extractorType, string extension,
             var fileName = GetFileName(extension);
             var timeBeforeExtraction = DateTime.Now.AddSeconds(-1);
 
-            var extractor = (IPackageExtractor)Activator.CreateInstance(extractorType);
+            var extractor = (IPackageExtractor) Activator.CreateInstance(extractorType, ConsoleLog.Instance);
```
droyad: Got to leave something to fix for another day

droyad:
The Helm tests are failing, but I think I need #528 before I can make it work properly

source/Calamari.Shared/Log.cs
```diff
@@ -7,225 +7,206 @@
...
         void Info(string message);
-        void InfoFormat(string message, params object[] args);
+        void InfoFormat(string messageFormat, params object[] args);
         void Warn(string message);
-        void WarnFormat(string message, params object[] args);
+        void WarnFormat(string messageFormat, params object[] args);
         void Error(string message);
-        void ErrorFormat(string message, params object[] args);
+        void ErrorFormat(string messageFormat, params object[] args);
+        void SetOutputVariableButDoNotAddToVariables(string name, string value, bool isSensitive = false);
```
droyad: Renamed this overload to make it clear it doesn't add to the current variable dictionary. Looking at the callers most of them do anyway. Left for another day.

PR #530, opened 2020-03-24
Add ICommand that does not take arguments and look package from the variables

droyad:
Each command parsing it's own variables seems unnecessary and adds extra code. For packages at least the path is already in the variables (https://github.com/OctopusDeploy/OctopusDeploy/blob/5c109575cd99157868b871acda6d517aeb78b640/source/Octopus.Core/Model/Projects/AcquiredPackageMap.cs#L72). Also with the advent of multiple packages per step, passing it at the command line no longer seems appropriate.

The setup of RunningDeployment and executing the conventions themselves seems to be repeated often, so this pulls that out of the command itself. If we do need a richer experience, we can add a Command type that allows that, but I think this simple interface makes it really easy to implement.

The Terraform step does it the "new" way, and if we agree to go down that path, we'll refactor it throughout the code base.

I haven't looked at other options yet, but if passing at the command line (instead of variables) is better for some reason we'd extend `ICommand` (`ICommandWithOptions<T>`?) to handle that.

There is an adapter for `ICommand` to `ICommandWithArguments` to keep the code simpler in `Program`.

### DIFF
--- a/source/Calamari.Shared/Commands/Support/ICommandWithArguments.cs
+++ b/source/Calamari.Shared/Commands/Support/ICommandWithArguments.cs
@@ -1,6 +1,4 @@
-﻿using System.Collections.Generic;
-using System.IO;
-using Calamari.Deployment.Conventions;
+﻿using System.IO;
 
 namespace Calamari.Commands.Support
 {
@@ -10,11 +8,5 @@ namespace Calamari.Commands.Support
     public interface ICommandWithArguments
     {
         int Execute(string[] commandLineArguments);
-    }
-
-    public interface ICommand
-    {
-        string PrimaryPackagePath { get; }
-        IEnumerable<IConvention> GetConventions();
     }
 }
--- a/source/Calamari.Shared/Commands/Support/ICommandWithArguments.cs
+++ b/source/Calamari.Shared/Commands/Support/ICommandWithArguments.cs
@@ -1,0 +1,17 @@
+﻿using System.IO;
+
+namespace Calamari.Commands.Support
+{
+    /// <summary>
+    /// A command that requires the command line arguments passed to it. We are transitioning away from this interface  
+    /// </summary>
+    public interface ICommandWithArguments
+    {
+        int Execute(string[] commandLineArguments);
+    }
+
+    public interface ICommand
+    {
+        int Execute();
+    }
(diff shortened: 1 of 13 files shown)


source/Calamari.Shared/Extensions/VariableExtensions.cs
```diff
@@ -0,0 +1,27 @@
+using Calamari.Commands.Support;
+using Calamari.Deployment;
+using Calamari.Integration.FileSystem;
+using Calamari.Util;
+
+namespace Calamari.Extensions
+{
+    public static class VariableExtensions
+    {
+        public static string GetPrimaryPackagePath(this IVariables variables, ICalamariFileSystem fileSystem, bool required)
```
droyad: Not sure whether to keep this as an extension. It didn't seem appropriate on `CalamariVariables` and a bit too slim for a standalone service. Happy to take advice.

source/Calamari.Shared/Commands/Support/ICommandWithArguments.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    /// <summary>
+    /// A command that requires the command line arguments passed to it. We are transitioning away from this interface  
+    /// </summary>
+    public interface ICommandWithArguments
+    {
+        int Execute(string[] commandLineArguments);
+    }
+
+    public interface ICommand
```
slewis74: Could we move this back out to its own file so it shows as a diff rather than a delete of `ICommand.cs`?

source/Calamari.Shared/Extensions/VariableExtensions.cs
slewis74: In here seems ok for now. There's a lot going on in here, I'm wondering if a name like `GetPathToPrimaryPackage` might be a bit easier for a caller to understand. The current name just says to me that it's going to give me a path back, why it wants the fileSystem isn't clear unless I look at the implementation. I also wouldn't expect, based on the name alone, that it'd throw if the package didn't exist at that path. I might expect it's my job as the caller to check that once it's given me the path.

droyad: I think the `required` parameter hints at that.

PR #533, opened 2020-03-25
Run a single packaged script based on OS preferences

adam-mccoy:
## Problem
If a package being deployed has multiple versions of the same packaged script in different languages (e.g., `PreDeploy.ps1` and `PreDeploy.sh`) they will all be run. This results in deployments failing if the requisite tools are not available (`bash.exe` on Windows, for example). The desired/expected behaviour is that, in this example, only `PreDeploy.ps1` will run on Windows targets and only `PreDeploy.sh` will run on Linux-y targets.

## Solution
Instead of running all matching scripts, a preferred script will be selected based on the OS being targeted. The orders of preference per OS are:
- _Linux_: Bash → Python → C# → F# → PowerShell
- _Windows_: PowerShell → Python → C# → F# → Bash

If multiple scripts are found for a particular hook a verbose message will be logged indicating which script was the selected and the preference order applied to do so.

See https://octopusdeploy.slack.com/archives/C01087M34HY for context and discussion.

Fixes https://github.com/OctopusDeploy/Issues/issues/6267

### Sample Log Message
```
Found 2 PreDeploy scripts. Selected /path/to/PreDeploy.sh based on OS preferential ordering: Bash -> Python -> CSharp -> FSharp -> PowerShell
```

### DIFF
--- a/source/Calamari.Shared/Deployment/Conventions/PackagedScriptConvention.cs
+++ b/source/Calamari.Shared/Deployment/Conventions/PackagedScriptConvention.cs
@@ -23,10 +23,10 @@ namespace Calamari.Deployment.Conventions
                 return;
             }
 
-            RunScripts(deployment);
+            RunPreferredScript(deployment);
             if (deployment.Variables.GetFlag(SpecialVariables.DeleteScriptsOnCleanup, true))
             {
-                DeleteScripts(deployment);
+                DeletePreferredScript(deployment);
             }
         }
     }
--- a/source/Calamari.Shared/Deployment/Conventions/PackagedScriptConvention.cs
+++ b/source/Calamari.Shared/Deployment/Conventions/PackagedScriptConvention.cs
@@ -26,7 +26,7 @@ namespace Calamari.Deployment.Conventions
             RunPreferredScript(deployment);
             if (deployment.Variables.GetFlag(SpecialVariables.DeleteScriptsOnCleanup, true))
             {
-                DeletePreferredScript(deployment);
+                DeleteScripts(deployment);
             }
         }
     }
--- a/source/Calamari.Shared/Integration/Scripting/CombinedScriptEngine.cs
+++ b/source/Calamari.Shared/Integration/Scripting/CombinedScriptEngine.cs
@@ -16,11 +16,7 @@ namespace Calamari.Integration.Scripting
 
         public ScriptSyntax[] GetSupportedTypes()
         {
-            var preferredScriptSyntax = new [] { ScriptSyntaxHelper.GetPreferredScriptSyntaxForEnvironment() };
-            var scriptSyntaxesSupportedOnAllPlatforms =  new[] { ScriptSyntax.Python, ScriptSyntax.CSharp, ScriptSyntax.FSharp, ScriptSyntax.PowerShell, ScriptSyntax.Bash };
-
-            //order is important, as we want to try the preferred syntax first
-            return preferredScriptSyntax.Concat(scriptSyntaxesSupportedOnAllPlatforms.Except(preferredScriptSyntax)).ToArray();
+            return ScriptSyntaxHelper.GetPreferenceOrderedScriptSyntaxesForEnvironment();
         }
 
         public CommandResult Execute(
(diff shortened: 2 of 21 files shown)


michaelnoonan:
There's a strange, seemingly unrelated test failure - might be worth a quick look since it may interrupt other people in the future.

michaelnoonan:
I suggest putting a link in this PR description to the channel/thread in Slack.

source/Calamari.Shared/Integration/Scripting/CombinedScriptEngine.cs
```diff
@@ -14,16 +13,16 @@ public CombinedScriptEngine(IEnumerable<IScriptWrapper> scriptWrapperHooks)
...
         }
-        
-        public ScriptSyntax[] GetSupportedTypes()	
-        {	
-            var preferredScriptSyntax = new [] { ScriptSyntaxHelper.GetPreferredScriptSyntaxForEnvironment() };	
-            var scriptSyntaxesSupportedOnAllPlatforms =  new[] { ScriptSyntax.PowerShell, ScriptSyntax.CSharp, ScriptSyntax.FSharp, ScriptSyntax.Python, ScriptSyntax.Bash };	
+
+        public ScriptSyntax[] GetSupportedTypes()
+        {
+            var preferredScriptSyntax = new [] { ScriptSyntaxHelper.GetPreferredScriptSyntaxForEnvironment() };
```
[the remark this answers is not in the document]
adam-mccoy: Fair point, it is a bit obscure. I'll address it.

source/Calamari.Shared/Integration/Scripting/PackagedScriptRunner.cs
```diff
@@ -45,25 +45,38 @@ protected void RunScripts(RunningDeployment deployment)
             }
         }
 
-        protected void DeleteScripts(RunningDeployment deployment)
+        protected void DeletePreferredScript(RunningDeployment deployment)
         {
-            var scripts = FindScripts(deployment);
+            var script = FindPreferredScript(deployment);
 
-            foreach (var script in scripts)
```
michaelnoonan: We should still delete all the scripts. It would feel strange if we left behind Octopus artifacts in an application folder.

source/Calamari.Shared/Deployment/Conventions/PackagedScriptConvention.cs
```diff
@@ -23,10 +23,10 @@ public void Install(RunningDeployment deployment)
                 return;
             }
 
-            RunScripts(deployment);
+            RunPreferredScript(deployment);
             if (deployment.Variables.GetFlag(SpecialVariables.DeleteScriptsOnCleanup, true))
             {
-                DeleteScripts(deployment);
+                DeletePreferredScript(deployment);
```
droyad: We should delete all scripts, this matches the previous behaviour when both bash and ps scripts were included

source/Calamari.Shared/Integration/Scripting/PackagedScriptRunner.cs
```diff
@@ -45,25 +45,38 @@ protected void RunScripts(RunningDeployment deployment)
...
+
+            var files = from file in fileSystem.EnumerateFiles(deployment.CurrentDirectory, searchPatterns)
+                        let nameWithoutExtension = Path.GetFileNameWithoutExtension(file)
+                        let preferenceOrdinal = Array.IndexOf(supportedScriptExtensions, file.ToScriptType())
+                        where nameWithoutExtension.Equals(scriptFilePrefix, StringComparison.OrdinalIgnoreCase)
+                        orderby preferenceOrdinal
+                        select file;
+
+            var numFiles = files.Count();
+            var selectedFile = files.FirstOrDefault();
```
droyad: You are enumerating the filesystem a second time here (lazy enumerable). I'd ToArray it

source/Calamari.Shared/Integration/Scripting/PackagedScriptRunner.cs
```diff
@@ -45,25 +45,38 @@ protected void RunScripts(RunningDeployment deployment)
...
+
+            if (numFiles > 1)
+            {
+                var preferenceOrderDisplay = string.Join(" -> ", supportedScriptExtensions);
```
droyad: Not sure an arrow is right here
```suggestion
                var preferenceOrderDisplay = string.Join(", ", supportedScriptExtensions);
```

PR #534, opened 2020-03-25
Ensured NGINX location files are processed in the correct order

mcasperson:
The order in which location files in NGINX are defined has significance. In particular any regular expression based location is processed from top to bottm. This is from the docs at http://nginx.org/en/docs/http/request_processing.html:

> Then nginx checks locations given by regular expression in the order listed in the configuration file.

Currently, location files are based on the location path, meaning locations will be imported in alphabetical order. This PR ensures the location configuration files have an index matching the order they were defined in the UI.


mcasperson:
Fixes 6274

PR #535, opened 2020-03-26
Cleanup of package extraction warnings

droyad:
Removed `SuppressNestedScriptWarning` as it's no longer relevant. (see https://octopusdeploy.slack.com/archives/C3KT1DFSM/p1585190441003100)
Created `PackageExtractionOptions` so we don't need to repeat the exact same `ExtractionOptions` all over the place

### DIFF
--- a/source/Calamari.Shared/Integration/Packages/NuGet/NupkgExtractor.cs
+++ b/source/Calamari.Shared/Integration/Packages/NuGet/NupkgExtractor.cs
@@ -41,7 +41,7 @@ namespace Calamari.Integration.Packages.NuGet
                         continue;
 
                     var targetFile = Path.Combine(targetDirectory, Path.GetFileName(unescapedKey));
-                    entry.WriteToFile(targetFile, new PackageExtractionOptions { ExtractFullPath = false });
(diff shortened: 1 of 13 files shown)


source/Calamari.Shared/Integration/Packages/NuGet/NupkgExtractor.cs
```diff
@@ -41,25 +41,17 @@ public int Extract(string packageFile, string directory, bool suppressNestedScri
                         continue;
 
                     var targetFile = Path.Combine(targetDirectory, Path.GetFileName(unescapedKey));
-                    entry.WriteToFile(targetFile, new ExtractionOptions { Overwrite = true, WriteSymbolicLink = WriteSymbolicLink });
+                    entry.WriteToFile(targetFile, new PackageExtractionOptions { ExtractFullPath = false });
```
droyad: I couldn't figure out what `ExtractFullPath` does, so set it to what it was before. The `PreserveFileTime` is set to `true` with this change which seems safe.

droyad: I was wrong https://build.octopushq.com/viewLog.html?buildId=1284990&buildTypeId=OctopusDeploy_CalamariNetCore_MonoTesting_TestAmazonLinux2Mono442

PR #536, opened 2020-03-26
Refactored the package extractors

droyad:
This PR decouples code and removed unneeded factories. Best reviewed commit by commit:

1. The `SimplePackageExtractor` did not spark joy as it just implemented the interface abstractly
2. The `createJavaGenericPackageExtractor` was a messy way to pass the supported extensions to `PackageStore`, so removed it and injected `PackageStore` into `FindPackageCommand` and injected `IPackageExtractor` directly. This commit also has a fixup as some `PackageExtractors` should have been `override`
3. Rename of `GenericPackageExtractor` to `CombinedPackageExtractor` as it describes it better. Removed `GenericPackageExtractorFactory` as it was unnecessary indirection, better to new it up directly. Replaced `JavaPackageExtractor` with `JarExtractor` (and renamed it) as it doesn't need the extension matching logic, might as well just try and let it 💥 if the command fails
4. Refactored `CombinedPackageExtractor`, removing the last exception throw as it never would get there due to the earlier Extension is valid checks. This meant `SupportLinkGenerator` was unused.


source/Calamari.Shared/Integration/Packages/CombinedPackageExtractor.cs
```diff
@@ -13,37 +13,33 @@ public interface ICombinedPackageExtractor : IPackageExtractor
...
-            }
-
-            var extractor = FindByExtension(file);
-            if (extractor != null)
-                return extractor;
+            var extractor = Extractors.FirstOrDefault(p => p.Extensions.Any(ext => file.Extension.Equals(ext, StringComparison.OrdinalIgnoreCase)));
+            if (extractor == null)
+                throw new CommandException($"Unsupported file extension `{extension}`");
             
-            throw new FileFormatException(supportLinkGenerator.GenerateSupportMessage(
```
droyad: I might be reading this wrong, but `extractor` should never be null as line `34` above already checked the extension exists. Therefore this error was never thrown, so might as well remove it.

source/Calamari.Aws/Commands/UploadAwsS3Command.cs
```diff
@@ -49,11 +49,10 @@ public override int Execute(string[] commandLineArguments)
             var substituter = new FileSubstituter(fileSystem);
             var bucketKeyProvider = new BucketKeyProvider();
             var targetType = GetTargetMode(targetMode);
-            var packageExtractor = new GenericPackageExtractorFactory().createStandardGenericPackageExtractor();
-
+            
             var conventions = new List<IConvention>
             {
-                new ExtractPackageToStagingDirectoryConvention(packageExtractor, fileSystem).When(_ => targetType == S3TargetMode.FileSelections),
+                new ExtractPackageToStagingDirectoryConvention(new CombinedPackageExtractor(), fileSystem).When(_ => targetType == S3TargetMode.FileSelections),
```
droyad: Was going to inject it, but when we introduce `IConventionFactory` this will be centralized there anyway.

PR #541, opened 2020-04-02
Removed the CommandAttribute

slewis74:
Changed the way commands are registered and resolved to simplify the model and pave the way for extensions to register things explicitly, rather than the magic that happens atm.

There are quite a few files in here, the crux of the change is the CommandAttribute for deleted and then the extensions register things through implementing `ICalamariExtension` (there are some well known Commands that Calamari registers for itself too).

The name that used to appear in the attribute is now the name the type is registered as in Autofac, so finding the command now relies on 

```
command = container.ResolveNamed<ICommand>(options.Command);
```

rather than only registering one with the right attribute value into the container.

The `Dictionary<string, Type> RegisterCommands()` method was used so the extensions didn't really have to get the named registration right for themselves, they just need to know to provide the name and the type and Calamari registers them the way it expects to use to find them.


source/Calamari.Aws/AwsExtension.cs
```diff
@@ -0,0 +1,27 @@
+using System;
+using System.Collections.Generic;
+using Autofac;
+using Calamari.Aws.Commands;
+
+namespace Calamari.Aws
+{
+    public class AwsExtension : ICalamariExtension
```
droyad, line 8: I'm not sure about the name `Extension`. `Plugin` perhaps?

source/Calamari.Aws/Calamari.Aws.csproj
```diff
@@ -27,6 +27,7 @@
     <TargetFrameworks>netstandard2.0</TargetFrameworks>
   </PropertyGroup>
   <ItemGroup>
+    <PackageReference Include="Autofac" Version="4.8.0" />
```
droyad, line 30: I would prefer to keep the `Autofac` reference out of the extensions/plugins. If we don't and we want to update it, it'll e a pain. I think it would be enough that the above class returned a list of types to register `.AsSingleInstance().AsImplementedInterfaces().AsSelf()`

source/Calamari.Aws/AwsExtension.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        readonly Dictionary<string, Type> commandTypes = new Dictionary<string, Type>
+        {
+            {"apply-aws-cloudformation-changeset", typeof(ApplyCloudFormationChangesetCommand)},
```
droyad, line 12: I'm not convinced this is better than the attribute. It separates the name of the command from the command itself.

source/Calamari/Program.cs
```diff
@@ -67,13 +68,17 @@ internal int Run(CommonOptions options)
             {
                 container.Resolve<VariableLogger>().LogVariables();
 
-                var command = container.Resolve<ICommand[]>();
-                if (command.Length == 0)
+                ICommand command;
+                try
+                {
+                    command = container.ResolveNamed<ICommand>(options.Command);
```
droyad, line 74: This should throw if there is more than one command with the same name

slewis74:
Closing as this has been implemented in a different way as part of the Sashimi changes

PR #548, opened 2020-04-21
Log failed get encrypted variables string in calamari verbose output

byrondelgado:
Add log entries to find variables that fail call Calamari.Integration.Scripting.WindowsPowerShell.PowerShellBootstrapper.GetEncryptedVariablesString(IVariables variables), as described in forum post  https://help.octopus.com/t/calamari-tentacle-throws-index-was-out-of-range-argumentoutofrangeexception-on-variables-parsing/24963


byrondelgado:
The CLA is signed, please recheck it

PR #549, opened 2020-04-21
[WIP] Bug azure cloud service deployment infinite loop

andrewabest:
Just pushing this up for visibility, will come back and add appropriate details


andrewabest:
Note this isn't ready for review / correct yet, it is just an initial view.

PR #564, opened 2020-05-25
Use full password argument because of Azure bug

mcasperson:
The Azure CLI bug at https://github.com/Azure/azure-cli/issues/12105 means that passwords starting with a dash are not processed correctly. This PR uses the full `--password` option to fix the issue.


mcasperson:
Fixes https://github.com/OctopusDeploy/Issues/issues/6385

PR #565, opened 2020-06-01
Spike to make it easier to refactor conventions

flin-8:
The approach I've taken here was to:

1. Extract a service out of each convention, and method inject anything that makes sense. 
2. The original convention then wraps this service so other command continue to work like before.
3. When all of the conventions for a given command have been converted, that command can call the services directly.

What's still missing is an abstract base command (or just a helper class) to replace `ConventionProcess` and do the error handling consistently.

### DIFF
--- a/source/Calamari.Azure.WebApps/Commands/DeployAzureWebCommand.cs
+++ b/source/Calamari.Azure.WebApps/Commands/DeployAzureWebCommand.cs
@@ -1,5 +1,6 @@
 ﻿using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using Calamari.Azure.WebApps.Deployment.Conventions;
 using Calamari.Commands;
 using Calamari.Commands.Support;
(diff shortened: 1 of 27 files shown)


source/Calamari.Azure.WebApps/Commands/DeployAzureWebCommand.cs
```diff
@@ -66,27 +67,24 @@ public override int Execute(string[] commandLineArguments)
...
-                new ConfiguredScriptConvention(DeploymentStages.PostDeploy, fileSystem, scriptEngine, commandLineRunner),
-            };
+            var packagedScriptService = new PackagedScriptService(log, fileSystem, scriptEngine, commandLineRunner);
+            var configuredScriptService = new ConfiguredScriptService(fileSystem, scriptEngine, commandLineRunner);
 
             var deployment = new RunningDeployment(pathToPackage, variables);
-            var conventionRunner = new ConventionProcessor(deployment, conventions);
 
-            conventionRunner.RunConventions();
+            extractPackage.ExtractToStagingDirectory(pathToPackage);
```
flin-8, line 75: This is an example of what a converted command looks like (the services will be injected instead of being new'ed up)

source/Calamari.Shared/Deployment/Conventions/ConfigurationTransformsConvention.cs
```diff
@@ -9,49 +9,65 @@
...
+
+        public ConfigurationTransformsConvention(ConfigurationTransformsService service)
+        {
+            this.service = service;
+        }
+
+        public void Install(RunningDeployment deployment)
+        {
+            var transformFilesApplied = service.Install(deployment);
+            deployment.Variables.SetStrings(SpecialVariables.AppliedXmlConfigTransforms, transformFilesApplied.ToList(), "|");
```
flin-8, line 23: Here we're setting the variable for a return value from the service

source/Calamari.Shared/Deployment/Conventions/ConfigurationVariablesConvention.cs
```diff
@@ -7,25 +7,38 @@
...
+        readonly ConfigurationVariablesService service;
+
+        public ConfigurationVariablesConvention(ConfigurationVariablesService service)
+        {
+            this.service = service;
+        }
+
+        public void Install(RunningDeployment deployment)
+        {
+            service.Install(deployment, deployment.Variables.GetStrings(SpecialVariables.AppliedXmlConfigTransforms, '|'));
```
flin-8, line 20: Here we're reading a variable and turning it into a method argument

source/Calamari.Shared/Deployment/Conventions/ConfigurationVariablesConvention.cs
```diff
@@ -7,25 +7,38 @@
...
 
-        public void Install(RunningDeployment deployment)
+        public void Install(RunningDeployment deployment, List<string> appliedAsTransforms)
         {
             if (deployment.Variables.GetFlag(SpecialVariables.Package.AutomaticallyUpdateAppSettingsAndConnectionStrings) == false)
             {
                 return;
             }
 
-            var appliedAsTransforms = deployment.Variables.GetStrings(SpecialVariables.AppliedXmlConfigTransforms, '|');
```
flin-8, line 27: There's a slight different in behaviour in that we now always read `SpecialVariables.AppliedXmlConfigTransforms` regardless of the `SpecialVariables.Package.AutomaticallyUpdateAppSettingsAndConnectionStrings` flag

PR #569, opened 2020-06-16
Add support for session duration during aws sts:assumerole

matt-richardson:
Customers were hitting time outs when deploying a long running cloud formation task.

This PR adds support for a new variable `Octopus.Action.Aws.AssumeRoleSessionDurationSeconds` which is passed to the AWS AssumeRole call.

This PR also adds nicer logging when it hits that timeout to prompt them to up the timeout. It also cleans up exception handling so that exception stack traces are logged as verbose and not duplicated.

Relates to https://github.com/OctopusDeploy/Issues/issues/6430

## Review

* please consider the name I've used the the variable. Is there a better one you can think of?
* please consider if there's any edge cases I've missed in the error handling changes

### DIFF
--- a/source/Calamari.CloudAccounts/AwsEnvironmentGeneration.cs
+++ b/source/Calamari.CloudAccounts/AwsEnvironmentGeneration.cs
@@ -25,6 +25,7 @@ namespace Calamari.CloudAccounts
         readonly string assumeRoleArn;
         readonly string assumeRoleExternalId;
         readonly string assumeRoleSession;
+        readonly string assumeRoleDurationSeconds;
 
         public static async Task<AwsEnvironmentGeneration> Create(ILog log, IVariables variables)
         {
@@ -61,6 +62,7 @@ namespace Calamari.CloudAccounts
             assumeRoleArn = variables.Get("Octopus.Action.Aws.AssumedRoleArn")?.Trim();
             assumeRoleExternalId = variables.Get("Octopus.Action.Aws.AssumeRoleExternalId")?.Trim();
             assumeRoleSession = variables.Get("Octopus.Action.Aws.AssumedRoleSession")?.Trim();
+            assumeRoleDurationSeconds = variables.Get("Octopus.Action.Aws.AssumeRoleSessionDurationSeconds")?.Trim();
         }
 
         /// <summary>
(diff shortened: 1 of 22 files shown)


source/Calamari.CloudAccounts/AwsEnvironmentGeneration.cs
```diff
@@ -61,6 +62,7 @@ private AwsEnvironmentGeneration(ILog log, IVariables variables)
             assumeRoleArn = variables.Get("Octopus.Action.Aws.AssumedRoleArn")?.Trim();
             assumeRoleExternalId = variables.Get("Octopus.Action.Aws.AssumeRoleExternalId")?.Trim();
             assumeRoleSession = variables.Get("Octopus.Action.Aws.AssumedRoleSession")?.Trim();
+            assumeRoleDurationSeconds = variables.Get("Octopus.Action.Aws.AssumeRoleSessionDurationSeconds")?.Trim();
```
mcasperson, line 65: I couldn't think of a better name for this variable (I would have called it the same thing).

source/Calamari.CloudAccounts/AwsEnvironmentGeneration.cs
```diff
@@ -187,7 +189,8 @@ async Task AssumeRole()
                    {
                        RoleArn = assumeRoleArn,
                        RoleSessionName = assumeRoleSession,
-                       ExternalId = string.IsNullOrWhiteSpace(assumeRoleExternalId) ? null : assumeRoleExternalId
+                       ExternalId = string.IsNullOrWhiteSpace(assumeRoleExternalId) ? null : assumeRoleExternalId,
+                       DurationSeconds = int.TryParse(assumeRoleDurationSeconds, out var durationSeconds) ? durationSeconds : 3600
```
mcasperson, line 193: Minor point, but I would assume passing null here lets the sdk assume the default value instead of hard coding 1 hour.

matt-richardson: You cant pass null, interestingly (it's an int).
Internally, it's a `int?`, but getting the value returns it as an int, so it ends up as 0.
I've changed it so it's more defensive and doesn't set the property unless we actually try and pass one in so its a bit more future proof.

PR #570, opened 2020-06-16
The private-alpine image was deleted (as we didnt know it was in use)

matt-richardson:
And we were running out of private repositories. This changes over to use a private image we know exists

### DIFF
--- a/source/Calamari.Tests/Fixtures/Integration/Packages/DockerImagePackageDownloaderFixture.cs
+++ b/source/Calamari.Tests/Fixtures/Integration/Packages/DockerImagePackageDownloaderFixture.cs
@@ -58,7 +58,7 @@ namespace Calamari.Tests.Fixtures.Integration.Packages
         public void DockerHubWithCredentials_Loads()
         {
             const string privateImage = "octopusdeploy/octo-prerelease";
-            var version =  new SemanticVersion("7.3.7");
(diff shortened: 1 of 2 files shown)


source/Calamari.Tests/Fixtures/Integration/Packages/DockerImagePackageDownloaderFixture.cs
```diff
@@ -57,8 +57,8 @@ public void PackageWithoutCredentials_Loads()
         [RequiresDockerInstalledAttribute]
         public void DockerHubWithCredentials_Loads()
         {
-            const string privateImage = "octopusdeploy/private-alpine";
-            var version =  new SemanticVersion("1.0.0");
+            const string privateImage = "octopusdeploy/octo-prerelease";
+            var version =  new SemanticVersion("7.3.7");
```
johnsimons, line 61: I think you mean `7.3.7-alpine`, `7.3.7` does not exist!

matt-richardson: Could've sworn we had a manifest for that... maybe only for prod builds..
```suggestion
            var version =  new SemanticVersion("7.3.7-alpine");
```

matt-richardson:
This also required the `readers` team in dockerhub to have `read-only` access to the `octo-prerelease` repo

PR #571, opened 2020-06-19
Attribute to not run tests on ARMx64 and applied to relevant failing tests

corey-underdown:
Our Ubuntu 18.04 ARMx64 agent is paused as there are some tests that don't run on it. This attribute will allow skipping tests on ARMx64 if applied to the tests which means we can skip the tests which don't run on ARMx64 and re-enable the agent.

### DIFF
--- a/source/Calamari.Tests/Calamari.Tests.csproj
+++ b/source/Calamari.Tests/Calamari.Tests.csproj
@@ -42,6 +42,7 @@
     <ProjectReference Include="..\Calamari.Aws\Calamari.Aws.csproj" />
   </ItemGroup>
   <ItemGroup Condition=" '$(TargetFrameworkIdentifier)' == '.NETFramework' ">
+    <PackageReference Include="NETStandard.Library" Version="2.0.3" />
     <ProjectReference Include="..\Calamari.Azure.ServiceFabric\Calamari.Azure.ServiceFabric.csproj" />
     <ProjectReference Include="..\Calamari.Azure.WebApps\Calamari.Azure.WebApps.csproj" />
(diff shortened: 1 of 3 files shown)


source/Calamari.Tests/Calamari.Tests.csproj
```diff
@@ -42,6 +42,7 @@
     <ProjectReference Include="..\Calamari.Aws\Calamari.Aws.csproj" />
   </ItemGroup>
   <ItemGroup Condition=" '$(TargetFrameworkIdentifier)' == '.NETFramework' ">
+    <PackageReference Include="NETStandard.Library" Version="2.0.3" />
```
Silcoish, line 45: Required for net framework to use System.Runtime.InteropServices.RuntimeInformation.ProcessArchitecture in RequiresNonARM64BitAttribute.cs

source/Calamari.Tests/Fixtures/Integration/Scripting/PowerShellScriptEngineFixture.cs
```diff
@@ -12,6 +12,7 @@ public class PowerShellScriptEngineFixture : ScriptEngineFixtureBase
     {
         [Test]
         [RequiresNonFreeBSDPlatform]
+        [RequiresNonARM64Bit]
```
matt-richardson, line 15: Would be good to have a description here
```suggestion
        [RequiresNonARM64Bit("As of June 2020, PowerShell Core for ARM is still an experimental release that is unsupported.")]
```
etc (so we know why we're excluding it)

Silcoish:
Not working with Mono: Will revise next sharpening

corey-underdown:
Old work, didn't work with Mono - If needed, can be revived but closing for now.

PR #574, opened 2020-07-08
Updated Json Configuration Variable Replacer Tests to use Assent

Silcoish:
The tests were previously using inline strings to set the expected output. I updated all the tests to use Assent instead to make it easier to compare what's wrong when issues occur and cleans up the test code.


andrew-moore-octo:
Looks like a good change 👍. Couple of things that stood out to me:

- The `source/Calamari.Tests/Fixtures/JsonVariables/JsonConfigurationVariableReplacerFixture.ShouldReplaceInSimpleFile.approved.txt` looks like it might be extraneous (there's a `.json` equivalent). Do we need that one?
- Looks like teamcity can't find the approved files. I haven't look into it too deeply, but I haven't seen that before. 🤔

Silcoish:
I think it's because I forgot to add to the approval files to the csproj. Updated, let's see if it's ok now.

PR #577, opened 2020-07-09
Add code to support the detection of different file formats.

andrew-moore-octo:
# Background
Calamari currently supports pushing variables into JSON files. This PR modifies this functionality so that Calamari can in future support more formats. 

This PR adds a stub for the YAML handling. The stub currently returns `false`, so nothing will ever be parsed as YAML.

This PR makes some assumptions about how we're going to detect file formats - that's still an area we're investigating, but the feel is that this PR brings us closer to our end goal regardless of how we end up detecting file formats.

# How to review this PR
- 👀 the code
- Consider if this will work for planned future functionality


source/Calamari.Azure.CloudServices/Commands/DeployAzureCloudServiceCommand.cs
```diff
@@ -74,7 +74,9 @@ public override int Execute(string[] commandLineArguments)
             var configurationTransformer = ConfigurationTransformer.FromVariables(variables);
             var transformFileLocator = new TransformFileLocator(fileSystem);
             var replacer = new ConfigurationVariablesReplacer(variables.GetFlag(SpecialVariables.Package.IgnoreVariableReplacementErrors));
-            var jsonVariablesReplacer = new JsonConfigurationVariableReplacer();
+            var jsonVariablesReplacer = new StructuredConfigVariableReplacer(
```
andrew-moore-octo, line 77: There will be changes to DI in a near-future PR. This is ugly but our DI will all be replaced soon.

PR #579, opened 2020-07-09
Implement path-based scalar value YamlFileFormatVariableReplacer

thedewi:
This includes an implementation of `YamlFileFormatVariableReplacer` which I believe is clarifying paths properly, although for now it only replaces scalar values present in the original document. Insertions and raw YAML replacements will come later, and will be bit a bit more complicated.

Most of the heavy lifting is done in `YamlEventStreamClassifier`. Many thanks to @Silcoish for a big pairing session where we figured out most of it.

This also includes `YamlTransformApproachComparisonDemoFixture`, which is an Explicit test that can aid experimentation with different parsers and different parsing approaches. We used it to conclude that YamlDotNet in streaming ParsingEvent mode applies the lowest amount of churn/canonicalization, and seems to handle advanced features the best, with the least amount of crashing.

Challenges to keep in mind:
- Both streaming parsers (which are the ones that handle advanced inputs a lot better) present mapping keys as scalars, which are tricky to differentiate from value scalars. We resolve this by paying attention to sequencing. For example, a flat key-value list looks like `[key, value, key, value, ...]`, but nested structures complicate things - we have to wait until they close and then set our expectation for the next key in the mapping.
- Both libraries we considered have some incomplete support for YAML 1.2, which seems to be **very** common with YAML parsers, despite the specification's age. The good news is that according to https://matrix.yaml.io/, YamlDotNet ranks highly, and the differences between 1.1 and 1.2 are mostly weird edge cases.

### DIFF
--- a/source/Calamari.Common/Features/StructuredVariables/YamlFormatVariableReplacer.cs
+++ b/source/Calamari.Common/Features/StructuredVariables/YamlFormatVariableReplacer.cs
@@ -1,4 +1,10 @@
-﻿namespace Calamari.Common.Features.StructuredVariables
+﻿using System.IO;
+using System.Linq;
+using Calamari.Integration.FileSystem;
+using YamlDotNet.Core;
+using YamlDotNet.Core.Events;
+
+namespace Calamari.Common.Features.StructuredVariables
 {
     public interface IYamlFormatVariableReplacer : IFileFormatVariableReplacer
     {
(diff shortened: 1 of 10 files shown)


source/Calamari.Common/Features/StructuredVariables/YamlFormatVariableReplacer.cs
```diff
@@ -7,11 +14,60 @@ public interface IYamlFormatVariableReplacer : IFileFormatVariableReplacer
...
+                            ev = scalar.ReplaceValue(newValue);
+                        }
+
+                        outputEvents.Add(ev);
+                    }
+                }
+            }
+            catch (SyntaxErrorException)
+            {
+                // TODO: Report where the problem was in the input file
```
andrew-moore-octo, line 51: Good point here, and I think another piece of evidence that having the user specify the file format is a good idea. Take this example:

1. User creates a package with a .properties file
2. User tries to deploy package
3. Calamari tries to parse the file as json, and shows an exception with details of the syntax error
4. Calamari tries to parse the file as yaml, and shows an exception with details of the syntax error
5. Calamari tries to parse the file as xml, and shows an exception with details of the syntax error
6. Calamari tries to parse the file as .properties, which succeeds
7. The deployment finishes.
8. User looks in logs to see 3 parsing errors even though the parsing succeeded.

If we know what the file _should_ be, then we can happily log the syntax error here and make life easy for the customer. But if we have to guess, we need to be a little more nuanced in how we handle syntax errors, and when to display them.

thedewi: We could show just one in cases where we're "confident" (such as a file extension match). But more and more I'm thinking that by the time a user enters a filename, correcting the file type isn't much effort, particularly if we auto-select the most likely one as soon as they enter the filename.

thedewi:

PR #581, opened 2020-07-13
Corrected the namespaces in the Calamari.Common project.

slewis74:
These files were moved out of Shared and the namespaces were supposed to have been updated. This got missed, so this PR is simply to correct all of the namespaces in the Common project.


source/Calamari.Common/Calamari.Common.csproj
```diff
@@ -50,7 +50,6 @@
     </ItemGroup>
 
     <ItemGroup>
-      <PackageReference Include="Octopus.CoreUtilities" Version="1.1.2" />
```
flin-8, line 53: Is this no longer needed?

slewis74: No, there was 1 extension method that was previously used, but it is no longer being used.

PR #584, opened 2020-07-14
Feature: replace structures (with strings for now)

thedewi:
This change adds

- Some refactorings. Might be easiest to review separate commits to disentangle.
- Implement replacing mappings (ie objects) and sequences (ie arrays) with string values
- Path matching is case-insensitive


thedewi:

PR #586, opened 2020-07-15
Remove IDoesDeploymentTargetTypeHealthChecks and HealthCheckCommand

flin-8:
Merging https://github.com/OctopusDeploy/Calamari/pull/585 to `master`

Also replacing `next-version: 14.0` in `Gitversion.yml` with `+semver: major` in the commit message instead


GitVersion.yml
```diff
@@ -1,5 +1,4 @@
 mode: ContinuousDeployment
-next-version: 14.0
```
flin-8, line 2: @michaelnoonan Adding the +semver:major works
https://github.com/OctopusDeploy/Calamari/pull/586/commits/b31ade9d0610e35140bdb965c9c926fae34d164e

michaelnoonan: That looks good. I can't comment on the rest. :)

PR #601, opened 2020-07-20
Andrewm/additional paths

andrew-moore-octo:
This PR adds some tests to cover usages of the `Octopus.Action.AdditionalPaths` variable and how it interacts with `StructuredConfigVariablesService`.


source/Calamari.Tests/Fixtures/StructuredVariables/StructuredConfigVariablesServiceFixture.cs
```diff
@@ -0,0 +1,103 @@
...
+            service.ReplaceVariables(deployment);
+
+            replacerAssertions?.Invoke(replacer);
+            logAssertions?.Invoke(log);
+        }
+
+        [Test]
+        public void ReplacesVariablesInAdditionalPathIfFileMatchedInWorkingDirectory()
+        {
+            void Assertions(IStructuredConfigVariableReplacer replacer)
```
andrew-moore-octo, line 62: I need to pass an `Action<...>` to `RunAdditionalPathsTest` on line 68, but given our current code formatting rules, this is the simplest way I can do it without the code being indented way over to the right.

PR #602, opened 2020-07-21
Expand Octopus variables during YAML replacement

thedewi:
This uses `IVariables.Get()` to resolve variable values, causing expansion of any references to other Octopus variables.

Includes a test for the new functionality, and also a test of the existing JSON replacer functionality.

### DIFF
--- a/source/Calamari.Common/Features/StructuredVariables/YamlFormatVariableReplacer.cs
+++ b/source/Calamari.Common/Features/StructuredVariables/YamlFormatVariableReplacer.cs
@@ -25,7 +25,9 @@ namespace Calamari.Common.Features.StructuredVariables
             var variablesByKey = variables
                                  .Where(v => !octopusReservedVariablePattern.IsMatch(v.Key))
                                  .DistinctBy(v => v.Key)
-                                 .ToDictionary(v => v.Key, v => v.Value, StringComparer.OrdinalIgnoreCase);
+                                 .ToDictionary<KeyValuePair<string, string>, string, Func<string>>(v => v.Key,
+                                                                                                   v => () => variables.Get(v.Key),
+                                                                                                   StringComparer.OrdinalIgnoreCase);
 
             // Read and transform the input file
             var outputEvents = new List<ParsingEvent>();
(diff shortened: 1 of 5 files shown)


source/Calamari.Tests/Fixtures/StructuredVariables/Approved/YamlVariableReplacerFixture.ShouldExpandOctopusVariableReferences.approved.yaml
```diff
@@ -0,0 +1,20 @@
+﻿server:
+  ports:
+  - "8080"
```
Waldo000000, line 3: Should this be a string, or int?

source/Calamari.Tests/Fixtures/StructuredVariables/JsonFormatVariableReplacerFixture.cs
```diff
@@ -211,5 +211,17 @@ public void ShouldFallBackToStringIfTypePreservationFails()
                                 "types-fall-back.json"),
                         TestEnvironment.AssentJsonDeepCompareConfiguration);
         }
+
+        [Test]
+        public void ShouldExpandOctopusVariableReferences()
```
Waldo000000, line 216: 💙 that we're improving the safety net for existing features as we go 👏

source/Calamari.Tests/Fixtures/StructuredVariables/Approved/YamlVariableReplacerFixture.ShouldExpandOctopusVariableReferences.approved.yaml
thedewi: The input is `"5555"`, so the replacer preserves the type at that position. I deliberately include a few ambiguous-looking inputs, to help confirm the implementation doesn't get thrown by them.

thedewi: I suppose it does come at the cost of not reading like a "normal" config file.

source/Calamari.Common/Features/StructuredVariables/YamlFormatVariableReplacer.cs
```diff
@@ -25,7 +25,9 @@ public bool TryModifyFile(string filePath, IVariables variables)
             var variablesByKey = variables
                                  .Where(v => !octopusReservedVariablePattern.IsMatch(v.Key))
                                  .DistinctBy(v => v.Key)
-                                 .ToDictionary(v => v.Key, v => v.Value, StringComparer.OrdinalIgnoreCase);
+                                 .ToDictionary<KeyValuePair<string, string>, string, Func<string>>(v => v.Key,
+                                                                                                   v => () => variables.Get(v.Key),
```
Waldo000000, line 29: This can just be `v => variables.Get(v.Key)` I think.

thedewi: Because variable expansion can be recursive, and that structure contains every variable in the project (whether they will apply to this document or not), I was hesitant to eagerly expand all of them. Do you think that's a premature optimization?

Waldo000000: Ah, right, nah all good 👍

PR #613, opened 2020-07-23
Change how we choose the file format for structured variable replacement

andrew-moore-octo:
This PR changes the logic in `StructuredConfigVariablesService` to the following pseudocode:

```
if !featureFlagSet
    if tryParse(json)
        success
    else
        failure
else
    if fileFormatSpecifiedByUser
        if tryParse(fileFormatSpecifiedByUser)
            success
        else
            failure
    else
        if tryParse(json)
            success
        else
            guessedFormat := tryGuessFormatFromFileName()
            if (guessedFormat)
                if tryParse(guessedFormat)
                    success
                else
                    failure
```

The key change is that when no format has been explicitly supplied, we try to parse as JSON before trying to guess the file format based on the file extension.

### DIFF
--- a/source/Calamari.Common/Features/StructuredVariables/StructuredConfigFileParseException.cs
+++ b/source/Calamari.Common/Features/StructuredVariables/StructuredConfigFileParseException.cs
@@ -1,0 +1,12 @@
+﻿using System;
+
+namespace Calamari.Common.Features.StructuredVariables
+{
+    public class StructuredConfigFileParseException : Exception
+    {
+        public StructuredConfigFileParseException(string message, Exception e) : base(message, e)
+        {
+            
+        }
+    }
+}
--- a/source/Calamari.Common/Features/StructuredVariables/StructuredConfigVariablesService.cs
+++ b/source/Calamari.Common/Features/StructuredVariables/StructuredConfigVariablesService.cs
@@ -16,22 +16,28 @@ namespace Calamari.Common.Features.StructuredVariables
 
     public class StructuredConfigVariablesService : IStructuredConfigVariablesService
     {
-        readonly IFileFormatVariableReplacer[] replacers;
         readonly IFileFormatVariableReplacer jsonReplacer;
+        readonly IFileFormatVariableReplacer[] nonJsonReplacers;
+        readonly IFileFormatVariableReplacer[] allReplacers;
         readonly ICalamariFileSystem fileSystem;
         readonly ILog log;
 
         public StructuredConfigVariablesService(
-            IEnumerable<IFileFormatVariableReplacer> replacers,
+            IFileFormatVariableReplacer[] replacers,
             ICalamariFileSystem fileSystem,
             ILog log)
         {
-            this.replacers = replacers.ToArray();
             this.fileSystem = fileSystem;
             this.log = log;
 
-            jsonReplacer = this.replacers.FirstOrDefault(r => r.FileFormatName == StructuredConfigVariablesFileFormats.Json)
+            allReplacers = replacers;
+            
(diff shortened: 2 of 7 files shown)


source/Calamari.Common/Features/StructuredVariables/StructuredConfigFileParseException.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace Calamari.Common.Features.StructuredVariables
+{
+    public class StructuredConfigFileParseException : Exception
```
andrew-moore-octo, line 5: This has been introduced in order to distinguish between parsing failures and other failures that could occur when dealing with files (disk space, permissions etc). The two existing replacers (json, yaml) have been updated to throw this exception when parsing has failed.

andrew-moore-octo:
[Auto-select default file format based on filename](https://trello.com/c/ADcUfGep/34-auto-select-default-file-format-based-on-filename)

source/Calamari.Common/Features/StructuredVariables/YamlFormatVariableReplacer.cs
```diff
@@ -18,89 +18,97 @@ public class YamlFormatVariableReplacer : IFileFormatVariableReplacer
...
-                                 .ToDictionary<KeyValuePair<string, string>, string, Func<string>>(v => v.Key,
-                                                                                                   v => () => variables.Get(v.Key),
-                                                                                                   StringComparer.OrdinalIgnoreCase);
-
-            // Read and transform the input file
-            var outputEvents = new List<ParsingEvent>();
-            (IYamlNode startEvent, string replacementValue)? structureWeAreReplacing = null;
-            
-            using (var reader = new StreamReader(filePath))
+            try
```
andrew-moore-octo, line 27: This is once again a messy diff, but I've just added a try/catch around this.

source/Calamari.Common/Features/StructuredVariables/StructuredConfigVariablesService.cs
```diff
@@ -78,26 +83,99 @@ List<string> MatchingFiles(RunningDeployment deployment, string target)
...
+                              + $"The available options are: {availableFileFormats}";
+                throw new Exception(message);
             }
+
+            return specifiedReplacer;
+        }
+
+        IFileFormatVariableReplacer? FindBestReplacerForFilePath(string filePath)
+        {
+            return nonJsonReplacers.FirstOrDefault(r => r.IsBestReplacerForFileName(filePath));
```
Silcoish, line 143: Rather than having the existance of nonJsonReplacers, maybe we could just do something like: 

`return allReplacers.FirstOrDefault(r => r.FileFormatName != StructuredConfigVariablesFileFormats.Json && r.IsBestReplacerForFileName(filePath))`? I was a little confused on the existance of the nonJsonReplacers to begin with.

Silcoish: If we did that, we could also change the Method to be something like `FindBestNonJsonReplacerForFilePath`.

source/Calamari.Common/Features/StructuredVariables/StructuredConfigVariablesService.cs
```diff
@@ -78,26 +83,99 @@ List<string> MatchingFiles(RunningDeployment deployment, string target)
...
-            if (replacer != null)
+            return new []
             {
-                log.Info($"The config file at '{filePath}' is being handled as format '{replacer.FileFormatName}' because of the filename.");
-                return replacer;
+                jsonReplacer
+            };
+        }
+
+        IFileFormatVariableReplacer FindReplacerForFormat(string specifiedFileFormat)
```
Silcoish, line 125: If we changed this to `TryFindReplacerForFormat` would that make it clearer from reading that it can throw an exception?

PR #614, opened 2020-07-23
Renaming Json Configuration features to Structured Configuration

Silcoish:
Renaming the files that had JsonConfiguration in them to be StructuredConfiguration.
Renamed the c# variables from JsonConfiguration* -> StructuredConfiguration and moved them to ActionsVariables. Kept their string values as Json*.


source/Calamari.Common/Plumbing/Variables/ActionVariables.cs
```diff
@@ -6,7 +6,12 @@ public static class ActionVariables
     {
         public const string Name = "Octopus.Action.Name";
         public const string AdditionalPaths = "Octopus.Action.AdditionalPaths";
-
+        
+        /* We've renamed these two variables from "Json*" to "Structured*" as they're being used with the renamed config feature
+         Structured Configuration Variables. There are issues around supporting two variables when re-naming so for the time being
```
Waldo000000, line 11:
```suggestion
         Structured Configuration Variables. These values are coupled to the API as they appear in deployment process resources (within the action's property list), so until we have support for API versioning, it is not feasible to change their values without causing a breaking change. For this reason, for the time being...
```

PR #616, opened 2020-07-24
Preserve YAML comments

thedewi:
This change adjusts the scanner to no longer discard comments, and restores some leading spaces using Mark information.

Includes a test demonstrating mapping, sequence, and mixed content indentation.

Unfortunately, there are some limitations here with YamlDotNet:
- the Emitter enforces a space after the output `#`
- Comments at the point of an indenting reduction lose that context, so they may look more indented than we'd prefer (as in line 10 of the test file).

### DIFF
--- a/source/Calamari.Common/Features/StructuredVariables/YamlEventStreamClassifier.cs
+++ b/source/Calamari.Common/Features/StructuredVariables/YamlEventStreamClassifier.cs
@@ -126,6 +126,19 @@ namespace Calamari.Common.Features.StructuredVariables
                              scalar.Start,
                              scalar.End);
         }
+
+        public static Comment RestoreLeadingSpaces(this Comment comment)
+        {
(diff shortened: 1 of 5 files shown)


source/Calamari.Common/Features/StructuredVariables/YamlEventStreamClassifier.cs
```diff
@@ -126,6 +126,19 @@ public static Scalar ReplaceValue(this Scalar scalar, string newValue)
                              scalar.Start,
                              scalar.End);
         }
+
+        public static Comment RestoreLeadingSpaces(this Comment comment)
+        {
+            var leadingSpaces = comment.Start.Line == comment.End.Line
+                ? comment.End.Column - comment.Value.Length - comment.Start.Column - 2
```
Silcoish, line 133: Would be nice to have a comment here about the emitter limitation causing a space after the `#`

thedewi:

PR #625, opened 2020-07-30
Upgrade output encoding if necessary to accomodate variables

thedewi:
This PR addresses a possible scenario where an encoding is being used that cannot accomodate characters added through an Octopus variable.

The approach here is to switch up to UTF-8 in such cases, to output something more useful than a question mark. Because UTF-8 is otherwise identical to most single-byte encodings, only the newly inserted characters are affected by the change.

Note: only covers the YAML replacer, but I think we'd be interested in adding this to our standard approach for other formats.


source/Calamari.Common/Features/StructuredVariables/YamlFormatVariableReplacer.cs
```diff
@@ -103,13 +104,19 @@ public void ModifyFile(string filePath, IVariables variables)
...
                 // Write the replacement file
-                using (var writer = new StreamWriter(filePath, false, encoding))
-                {
-                    writer.NewLine = lineEnding == StringExtensions.LineEnding.Dos ? "\r\n" : "\n";
-                    var emitter = new Emitter(writer, indentDetector.GetMostCommonIndent());
-                    foreach (var outputEvent in outputEvents)
-                        emitter.Emit(outputEvent);
-                }
+                if (EncodingAdaptiveFileWriter.EncoderDoesNotRaiseErrorsForUnsupportedCharacters(encoding))
+                    encoding = Encoding.GetEncoding(encoding.WebName, EncoderFallback.ExceptionFallback, DecoderFallback.ExceptionFallback);
```
thedewi, line 108: Note: I'm about to move this check & fix inside `EncodingAdaptiveFileWriter`, and incorporate `EncodingAdaptiveFileWriter` into `ICalamariFileSystem`

thedewi:
> LGTM 👍 Not sure if this is the place for it, but in cases where we have to change the encoding from the original file it'd be nice to output something to the logs.

Good point. Was on my TODO list but I forgot to mention it here. I'll make those adjustments and raise again before I merge this.

source/Calamari.Common/Features/StructuredVariables/YamlFormatVariableReplacer.cs
thedewi: Done.

thedewi:

PR #626, opened 2020-07-30
Calamari Testing moved here

johnsimons:


flin-8:
Looking good, but are we expecting to get a nuget package? I don't see it in Teamcity

johnsimons:
@flin-8 I got a nuget now 😃

PR #628, opened 2020-07-30
Support XML var replacement.

andrew-moore-octo:
This PR adds support for pushing variables into XML files using Structured Configuration Variables.


andrew-moore-octo:
[Calamari: Push variables into XML files](https://trello.com/c/FIY8xJC1/14-calamari-push-variables-into-xml-files)

source/Calamari.Common/Calamari.Common.csproj
```diff
@@ -63,6 +63,7 @@
       <PackageReference Include="Octopus.Versioning" Version="4.2.1" />
       <PackageReference Include="Octostache" Version="2.7.0" />
       <PackageReference Include="SharpCompress" Version="0.24.0" />
+      <PackageReference Include="XPath2" Version="1.0.11" />
```
andrew-moore-octo, line 66: .NET doesn't support XPath 2 OOTB.

source/Calamari.Tests/Fixtures/StructuredVariables/JsonFormatVariableReplacerFixture.cs
```diff
@@ -126,6 +126,17 @@ public void ShouldReplaceWholeObject()
                         TestEnvironment.AssentJsonDeepCompareConfiguration);
         }
 
+        [Test]
```
andrew-moore-octo, line 129: Not strictly related to this changeset, but we discovered that this test case was missing while implementing similar behaviour for XML.

source/Calamari.Tests/Fixtures/Integration/Process/Semaphores/SemaphoreFactoryFixture.cs
```diff
@@ -12,8 +12,6 @@ public class SemaphoreFactoryFixture
         [RequiresMono]
         public void ReturnsFileBasedSemaphoreManagerForMono()
         {
-            if (!CalamariEnvironment.IsRunningOnMono)
```
andrew-moore-octo, line 15: This is now correctly handled by `RequiresMonoAttribute`

source/Calamari.Tests/Fixtures/RequiresMonoAttribute.cs
```diff
@@ -1,11 +1,21 @@
-﻿namespace Calamari.Tests.Fixtures
+﻿using Calamari.Common.Plumbing;
+using NUnit.Framework;
+using NUnit.Framework.Interfaces;
+
+namespace Calamari.Tests.Fixtures
 {
-    public class RequiresMonoAttribute : RequiresMinimumMonoVersionAttribute
+    public class RequiresMonoAttribute : TestAttribute, ITestAction
```
andrew-moore-octo, line 7: This was previously not correctly ignoring tests in non-Mono runtimes.

thedewi:
Note: my review included testing and verifying all the test outputs.

source/Calamari.Common/Features/StructuredVariables/XmlFormatVariableReplacer.cs
```diff
@@ -0,0 +1,234 @@
...
+                            case XmlProcessingInstruction processingInstruction:
+                                processingInstruction.Data = variable.Value;
+                                break;
+                            
+                            case XmlNode node:
+                                log.Warn($"Node of type {node.NodeType} not supported");
+                                break;
+
+                            default:
+                                // TODO: consider whether to silently ignore
```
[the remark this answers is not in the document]
andrew-moore-octo: Agreed. I forgot to come back to this one. I've put in place some better handling for this scenario.

There's a bit of context for how this all works together:

1. XPaths in XPath 2.0 always return a sequence of things, but those things aren't always nodes (because XPath 2.0 supports things like calculations), and we can't perform replacement on something that's not a node.
2. On [this line](https://github.com/OctopusDeploy/Calamari/pull/628/files/0c45e986b9f83483b37119c48322ec09bb7bb437#diff-061eda38d8f1b947e74af6a65cb87c83R57), we specifically filter down the results to just nodes, because we're using `XPath2SelectNodes` and not `XPath2Select`.
3. So if I'm understanding all of that correctly, then the default case shouldn't get hit, but I like having handling for it in case of bugs.

I've also changed the handling for this case to warn rather than fail. Probably doesn't matter either way though to be fair.

PR #629, opened 2020-07-31
Added fallback flag for structured configuration to only do json replacement

Silcoish:
Added a fallback flag for Structured Configuration to turn off the YAML replacement - the idea behind this is when we roll out to production, if there's a P1 issue around structured configuration, we can tell the customer to enable this flag in their project and de-escalate the problem.


andrew-moore-octo:
[Add fallback flag](https://trello.com/c/OyXtEs3M/54-add-fallback-flag)

PR #631, opened 2020-08-02
Changing the packaging to split net40 and net452

slewis74:
This changes is so we can separate these two versions and have server utilise based on target machines OS.

The driver for this is that a lot of the newer features we are building rely on tooling the is net452 specific, net40 is not supported. So we will now be able to differentiate these and use the net452 version (server currently only ever uses the net40 version of Calamari, it uses `Calamari.Cloud` when it needs net452. So in essence this PR is really about renaming to match actual intent/usage.


andrewabest:
We will tackle this as a part of [simplifying Calamari](https://docs.google.com/document/d/1WK73Ep4E9FLqwrFYu7v8MTZ_q1Gk730rJH2g-L1pvhc), closing this one.

PR #633, opened 2020-08-03
Add JSON and YAML tests confirming raw structure replacement

thedewi:
These tests establish that for JSON and YAML replacers, replacements of objects/mappings and arrays/sequences are treated as raw markup.


source/Calamari.Tests/Fixtures/StructuredVariables/Approved/JsonFormatVariableReplacerFixture.ShouldReplaceWholeObjectWithStructure.approved.json
```diff
@@ -0,0 +1,12 @@
+{
```
Silcoish, line 1: Don't think we need this file - couldn't see a method that maps up to the approval file. Perhaps created, then changed method name, and re-created a new one?

thedewi: Good catch. I changed approach halfway through and didn't mean to commit that.

PR #634, opened 2020-08-03
Tests for pipeline

johnsimons:
Tests for the new pipeline


source/Calamari.Tests/NewPipeline/PipelineCommandFixture.cs
```diff
@@ -0,0 +1,451 @@
...
+                File.WriteAllText(Path.Combine(tempPath.DirectoryPath, "Hello.html"), "Hello World!");
+
+                await CommandTestBuilder.CreateAsync<MyCommand, MyProgram>()
+                                        .WithArrange(context =>
+                                                     {
+                                                         context.WithNewNugetPackage(tempPath.DirectoryPath, "MyPackage", "1.0.0");
+                                                     })
+                                        .WithAssert(result =>
+                                                    {
+                                                        File.Exists(Path.Combine(tempPath.DirectoryPath, "Hello.html")).Should().BeTrue();
```
flin-8, line 70: With all of the tests below, should we be making those assertions inside the `Deploy()` stage rather than at the end?

source/Calamari.Testing/CommandTestBuilder.cs
```diff
@@ -263,20 +262,20 @@ async Task<TestCalamariCommandResult> ExecuteActionHandler(List<string> args)
 
                     CopyFilesToWorkingFolder(workingPath);
 
-                    result = await ExecuteActionHandler(args);
+                    result = await ExecuteActionHandler(args, workingPath);
+
+                    if (assertWasSuccess)
+                    {
+                        result.WasSuccessful.Should().BeTrue($"{command} execute result was unsuccessful");
```
flin-8, line 269: Guessing this moved because you still needed the current directory ?

source/Calamari.Testing/TestCalamariCommandResult.cs
```diff
@@ -9,18 +9,20 @@ public class TestCalamariCommandResult
...
             OutputActions = outputActions;
             ServiceMessages = serviceMessages;
             ResultMessage = resultMessage;
             FullLog = fullLog;
+            WorkingPath = workingPath;
             Artifacts = artifacts;
         }
 
         public string FullLog { get; }
+        public string WorkingPath { get; }
```
flin-8, line 25: Where is this used?

source/Calamari.Common/Features/Behaviours/RollbackScriptBehaviour.cs
```diff
@@ -21,10 +21,7 @@ public class RollbackScriptBehaviour : PackagedScriptRunner, IBehaviour
 
         public bool IsEnabled(RunningDeployment context)
         {
-            var features = context.Variables.GetStrings(KnownVariables.Package.EnabledFeatures)
-                                  .Where(s => !string.IsNullOrWhiteSpace(s)).ToList();
-
-            return features.Contains(KnownVariables.Features.CustomScripts);
+            return true;
```
flin-8, line 24: Why doesn't this depend on the feature?

source/Calamari.Common/Features/Behaviours/StructuredConfigurationVariablesBehaviour.cs
```diff
@@ -18,7 +18,7 @@ public StructuredConfigurationVariablesBehaviour(IStructuredConfigVariablesServi
 
         public bool IsEnabled(RunningDeployment context)
         {
-            return context.Variables.GetFlag(KnownVariables.Package.JsonConfigurationVariablesEnabled);
+            return context.Variables.GetFlag(ActionVariables.StructuredConfigurationVariablesEnabled);
```
flin-8, line 21: I think you can delete `KnownVariables.Package.JsonConfigurationVariablesEnabled`

source/Calamari.Tests/NewPipeline/PipelineCommandFixture.cs
johnsimons: Not sure I understand the difference?

source/Calamari.Testing/CommandTestBuilder.cs
johnsimons: yes, so we can assert before the directory is deleted

source/Calamari.Testing/TestCalamariCommandResult.cs
johnsimons: I think originally i used, but then i changed things. That said it is useful when using real packages is we need to assert files on disk

source/Calamari.Common/Features/Behaviours/RollbackScriptBehaviour.cs
johnsimons: So I went back and reviewed how this rollback is done in https://github.com/OctopusDeploy/Calamari/blob/master/source/Calamari/Deployment/Conventions/RollbackScriptConvention.cs and it seems there is no predicate for it

source/Calamari.Tests/NewPipeline/PipelineCommandFixture.cs
flin-8: Strictly speaking these results need to be available by the time `Deploy()` is executed, so it would be more accurate to assert on that

flin-8:
I think you have some tests failing in Linux due to differences in line endings

PR #641, opened 2020-08-05
XML replacer: ignore Octopus Prefix and other unintentional variables

thedewi:
# Background

In the existing JSON and YAML replacers, we avoid treating variable names starting with `Octopus` as replacements unless they continue with `:` to indicate they probably are replacements.

That doesn't quite directly translate over to XML since colon `:` is not the separator for XPaths.

# Change

I think we can probably afford to avoid processing any variable that doesn't contain `/`, since such an XPath would only match the document node anyway, and if the user really does want that, they can prepend `/`, which is the conventional way to do it anyway.

This includes a test that demonstrates how we avoid accidentally treating `Octopus.Release.Id` as an XPath.

### DIFF
--- a/source/Calamari.Common/Features/StructuredVariables/XmlFormatVariableReplacer.cs
+++ b/source/Calamari.Common/Features/StructuredVariables/XmlFormatVariableReplacer.cs
@@ -181,6 +181,9 @@ namespace Calamari.Common.Features.StructuredVariables
 
         XPath2Expression TryGetXPathFromVariableKey(string variableKey, IXmlNamespaceResolver nsResolver)
         {
+            if (!variableKey.Contains("/")) // Prevent simple variable names being recognized as XPath unintentionally
+                return null;
+            
             try
             {
                 return XPath2Expression.Compile(variableKey, nsResolver);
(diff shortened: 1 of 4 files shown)


source/Calamari.Common/Features/StructuredVariables/XmlFormatVariableReplacer.cs
```diff
@@ -180,16 +179,18 @@ IEnumerable<NamespaceDeclaration> GetDeclaredNamespaceUris(XmlNode node)
                 }
         }
 
-        bool IsValidXPath(string xPath, IXmlNamespaceResolver nsResolver)
+        XPath2Expression TryGetXPathFromVariableKey(string variableKey, IXmlNamespaceResolver nsResolver)
```
slewis74, line 182: I'm surprised it's letting you get away without return type being `XPath2Expression?` in this method.

thedewi: @slewis74 I've been wondering about that. It currently doesn't let us use `?` on reference types. Have nullable reference types been completely enabled yet for Calamari?

thedewi:
Based on discussion with @andrew-moore-octo, I've adjusted this to also accept `:` as an indicator of a useful XPath that could traverse the document tree - because the user might be using axes.

PR #644, opened 2020-08-05
Change the test for Docker to look for the docker command only

benPearce1:
Passing test after installing following method provided by customer [here](https://secure.helpscout.net/conversation/1234854365/68018?folderId=3252735):
image

### DIFF
--- a/source/Calamari.Shared/Integration/Packages/Download/Scripts/DockerPull.ps1
+++ b/source/Calamari.Shared/Integration/Packages/Download/Scripts/DockerPull.ps1
@@ -1,10 +1,18 @@
+function IsDockerAvailable() {
+    $service = $(Get-Service -Name 'docker' -ErrorAction SilentlyContinue)
+    $serviceRunning = ($service -ne $null -and $service.Status -eq 'Running')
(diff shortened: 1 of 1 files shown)


emilol:
Would this still fail for a stopped docker instance? That's what the check was originally introduced for: not just whether or not docker was installed, but also running. More details [here](https://trello.com/c/c2IGI0HI/3542-add-check-for-dockerd-process-during-package-acquisition) (internal link)

source/Calamari.Shared/Integration/Packages/Download/Scripts/DockerPull.ps1
```diff
@@ -1,14 +1,20 @@
+function IsDockerAvailable() {
+    $service = $(Get-Service -Name 'docker' -ErrorAction SilentlyContinue)
+    $serviceRunning = ($service -ne $null -and $service.Status -eq 'Running')
+    $process = $(Get-Process 'com.docker.proxy' -ErrorAction SilentlyContinue)
```
uglybugger, line 4: This will fail if the Docker daemon is not running on the same machine. Remember, the `docker` binary can be configured to talk to a daemon anywhere, either via a pipe or TCP. It doesn't have to be local.

PR #649, opened 2020-08-09
Add new stage to PipelineCommand and allow implementations to specify auto registration on more than one assembly

flin-8:
Sashimi PR: https://github.com/OctopusDeploy/Sashimi/pull/87

This is to make it easier for Sashimi to share a common RunScript command

### DIFF
--- a/source/Calamari.Common/CalamariFlavourProgramAsync.cs
+++ b/source/Calamari.Common/CalamariFlavourProgramAsync.cs
@@ -85,9 +85,9 @@ namespace Calamari.Common
                 .Named<PipelineCommand>(t => t.GetCustomAttribute<CommandAttribute>().Name);
         }
 
-        Assembly GetProgramAssemblyToRegister()
+        protected virtual Assembly[] GetProgramAssembliesToRegister()
         {
-            return GetType().Assembly;
+            return new []{GetType().Assembly};
         }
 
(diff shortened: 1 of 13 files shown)


source/Calamari.Common/CalamariFlavourProgramAsync.cs
```diff
@@ -130,10 +132,11 @@ protected async Task<int> Run(string[] args)
 
         IEnumerable<Assembly> GetAllAssembliesToRegister()
         {
-            var programAssembly = GetProgramAssemblyToRegister();
+            var programAssembly = GetProgramAssembliesToRegister();
```
flin-8, line 135: This way Calamari.AzureServiceFabric will be able to just add Calamari.Scripting and have it registered automatically.

The alternative is that everyone registers manually and we create methods to make it easier to register a command etc.

johnsimons: why don't we just scan all assemblies loaded into the AppDomain?

flin-8: We can also do that, I'm not fussed. 

All I care about is that `Calamari.AzureServiceFabric` is treated the same way as `Calamari.Scripting`.

PR #650, opened 2020-08-10
Add support for .properties files to Calamari.

andrew-moore-octo:
This PR adds the ability to push variables into `.properties` files using the Structured Configuration Variables feature. This leverages the parser we have built [here](https://github.com/OctopusDeploy/JavaPropertiesParser/tree/master/source).


source/Calamari.Common/Features/StructuredVariables/PropertiesFormatVariableReplacer.cs
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        public void ModifyFile(string filePath, IVariables variables)
+        {
+            try
+            {
+                var (fileText, encoding) = EncodingDetectingFileReader.ReadToEnd(filePath);
+
+                var parsed = Parser.Parse(fileText);
+                var updated = parsed.Mutate(expr => TryReplaceValue(expr, variables));
```
andrew-moore-octo, line 40: Treating ASTs as immutable is generally the way to go. Our parser tries to treat them as immutable, but .net 4.0 doesn't have `IReadOnlyList<T>` or `ImmutableList`, so we just model the top level expressions in an array. While we could reach into the document and just change things, I've tried to keep this consistent with how I'd like the parser to evolve in future, once we don't have to support .net 4.0.

thedewi: I much prefer immutability. Much safer.

PR #651, opened 2020-08-10
Focus SubstitutionsFixture assertions on processing behavior

thedewi:
# Background

The existing SubstitutionsFixture tests focus a lot on detection, and base assertions on the same mechanisms that are under test, like `Encoding.Default` and `CalamariPhysicalFileSystem.ReadFile`.

# Change

This PR:
- retires two redundant tests, `ShouldDetectAscii` and `ShouldFallBackToDefaultCodePage`, which are covered by `RetainAscii` and `RetainAnsi`.
- reworks other encoding tests so they assert that outputs have the features the user is expecting, rather than emphasizing which encoding objects get detected.

_Note: this reveals a regression that occurred in the switch to .NET Core, so 1 of these tests only passes in .NET Framework. The next PR will resolve this, which is why this is a Draft._


.gitattributes
```diff
@@ -43,4 +43,5 @@
 *.sln text eol=crlf merge=union 
 
 *.approved.* -text
-**/*.Tests/**/Samples/enc-* -text
\ No newline at end of file
+**/*.Tests/**/Samples/enc-* -text
+**/*.Tests/**/Samples/* -text
```
andrew-moore-octo, line 47:
```suggestion
**/*.Tests/Fixtures/Substitutions/Samples/* -text
```

thedewi:

PR #655, opened 2020-08-10
JSON replacer: preserve encoding, EOL, BOM. Upgrade encoding if needed

thedewi:
This is similar to the previous PRs, but applies the now-familiar approach to JSON.


thedewi:

PR #658, opened 2020-08-10
Obey any XML encoding declaration, and update if necessary

thedewi:
This checks the input file for any XML declaration that includes an encoding, and tries to read, and later write, using that encoding.

If the introduction of exotic characters means that we write another encoding, the declared encoding in the output is updated.

### DIFF
--- a/source/Calamari.Common/Features/StructuredVariables/XmlFormatVariableReplacer.cs
+++ b/source/Calamari.Common/Features/StructuredVariables/XmlFormatVariableReplacer.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text;
 using System.Xml;
 using System.Xml.XPath;
 using Calamari.Common.Plumbing.Extensions;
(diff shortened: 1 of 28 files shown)


source/Calamari.Common/Features/StructuredVariables/XmlFormatVariableReplacer.cs
```diff
@@ -32,7 +33,13 @@ public bool IsBestReplacerForFileName(string fileName)
 
         public void ModifyFile(string filePath, IVariables variables)
         {
-            var fileText = fileSystem.ReadFile(filePath, out var encoding);
+            var encodingPrecedence = new List<Encoding>(CalamariPhysicalFileSystem.DefaultInputEncodingPrecedence);
+
+            var fileBytes = fileSystem.ReadAllBytes(filePath);
+            if (TryGetDeclaredEncoding(fileBytes) is {} declaredEncoding)
+                encodingPrecedence.Insert(0, declaredEncoding);
```
andrew-moore-octo, line 40: This is nice. Now that we've got all the infrastructure in place to handle falling back between encoding, I love how simple it is to handle XML declarations. This could have been so much more painful!

thedewi:
[Obey any XML encoding declaration, and update if necessary](https://trello.com/c/OweAGpDZ/79-obey-any-xml-encoding-declaration-and-update-if-necessary)

PR #660, opened 2020-08-12
Tidy up the logs output by structured config var replacement.

andrew-moore-octo:
This PR changes the logging in `StructuredConfigVariablesService` to be a bit less confusing. Previously, failing to parse a yaml file would produce:

- 4 warnings
- 1 error

This was confusing as the user would see both JSON and YAML parsing failure messages (with `Warning` severity) for a YAML file. This is because we always try JSON first for backwards compatibility, but reading the logs was confusing.

This PR changes the behaviour:

- The logs now explain how Calamari chooses the file formats to try
- If the file parses successfully, there will be a single `Info` level log saying `Structured variable replacement succeeded on file {path} with format {format}`
- If the file doesn't parse, there will be a single `Error` level log containing details of the parsing failure

# Before
```
VERBOSE Attempting structured variable replacement on file {path} with formats: Json, Yaml. 
VERBOSE Attempting structured variable replacement on file {path} with format 'Json' 
WARNING Attempting structured variable replacement on file {path} with format 'Yaml' 
WARNING Structured variable replacement failed on file {path}. 
WARNING Syntax error when parsing the file as Json: Additional text encountered after finished reading JSON content: ,. Path '', line 1, position 0. 
WARNING Syntax error when parsing the file as Yaml: (Line: 1, Col: 1, Idx: 0) - (Line: 1, Col: 2, Idx: 1): While parsing a node, did not find expected node content. 
ERROR   System.Exception: The file at {path} could not be parsed with any of the formats tried. See logs for more details.
```

# After
```
VERBOSE The file at {path} matches a known filename pattern, and will be treated as Yaml. The file will be tried as Json first for backwards compatibility.
VERBOSE Attempting structured variable replacement on file {path} with format Json
VERBOSE The file at {path} couldn't be parsed as Json: Additional text encountered after finished reading JSON content: ,. Path '', line 1, position 0.
VERBOSE Attempting structured variable replacement on file {path} with format Yaml
ERROR   System.Exception: Structured variable replacement failed on file {path}. The file could not be parsed as Yaml: (Line: 1, Col: 1, Idx: 0) - (Line: 1, Col: 2, Idx: 1): While parsing a node, did not find expected node content. See verbose logs for more details.
```

# Out of scope for this PR
Calamari still logs exceptions during "install" conventions twice, and it still logs the full stack trace. That behaviour has not been changed in this PR, but may be addressed elsewhere.

### DIFF
--- a/source/Calamari.Common/Features/StructuredVariables/StructuredConfigVariablesService.cs
+++ b/source/Calamari.Common/Features/StructuredVariables/StructuredConfigVariablesService.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using Calamari.Common.Commands;
+using Calamari.Common.Plumbing;
 using Calamari.Common.Plumbing.FileSystem;
 using Calamari.Common.Plumbing.Logging;
 using Calamari.Common.Plumbing.Variables;
@@ -58,8 +59,7 @@ namespace Calamari.Common.Features.StructuredVariables
 
                 foreach (var filePath in matchingFiles)
                 {
-                    // TODO: once we allow users to specify a file format, pass it through here.
-                    var replacersToTry = GetReplacersToTryForFile(filePath, null, onlyPerformJsonReplacement);
+                    var replacersToTry = GetReplacersToTryForFile(filePath, onlyPerformJsonReplacement);
                     DoReplacement(filePath, deployment.Variables, replacersToTry);
                 }
             }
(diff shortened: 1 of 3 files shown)


source/Calamari.Common/Features/StructuredVariables/StructuredConfigVariablesService.cs
```diff
@@ -58,8 +59,7 @@ public void ReplaceVariables(RunningDeployment deployment)
 
                 foreach (var filePath in matchingFiles)
                 {
-                    // TODO: once we allow users to specify a file format, pass it through here.
```
andrew-moore-octo, line 61: This work won't be done this cycle, so I'm removing this.

source/Calamari.Common/Features/StructuredVariables/StructuredConfigVariablesService.cs
```diff
@@ -78,29 +78,26 @@ List<string> MatchingFiles(RunningDeployment deployment, string target)
             return files;
         }
 
-        IFileFormatVariableReplacer[] GetReplacersToTryForFile(string filePath, string? specifiedFileFormat, bool onlyPerformJsonReplacement)
```
andrew-moore-octo, line 81: This work won't be done this cycle, so I'm removing this.

PR #667, opened 2020-08-19
Fix helm steps in runbooks

mcasperson:
`Octopus.Action.Package.PackageId` doesn't exist in runbooks, but `Octopus.Action.Package[].PackageId` does. This PR uses the newer variable type instead of the legacy one.

This fixes this error:

image

Here is the new variable in the runbooks:

image

### DIFF
--- a/source/Calamari/Kubernetes/Conventions/HelmUpgradeConvention.cs
+++ b/source/Calamari/Kubernetes/Conventions/HelmUpgradeConvention.cs
@@ -278,7 +278,7 @@ namespace Calamari.Kubernetes.Conventions
         {
             var packagePath = deployment.Variables.Get(PackageVariables.Output.InstallationDirectoryPath);
             
-            var packageId = deployment.Variables.Get(PackageVariables.PackageId);
+            var packageId = deployment.Variables.Get(PackageVariables.IndexedPackageId(string.Empty));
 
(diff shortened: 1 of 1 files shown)


MarkSiedle (commented):
Calamari seems to have quite a few places still using `PackageVariables.PackageId`

This looks like it will fix it for the Helm upgrade, but should we review and patch the other remaining places to follow suit while we're here?

It seems to happen in the nginx feature, legacy iis and a couple of other places (not many):

2020-08-19 14_37_07-Calamari - Microsoft Visual Studio (Administrator)

Alternatively, if you wanna keep this isolated to a fix in Helm for now, we can raise a card in QRF and we'll make sure it gets followed up at some point.

mcasperson:
For now fixing helm was my priority to get a workflow ready for the marketing launch we are preparing for. I do agree this should be a bigger investigation though given the other steps, but if we could do that through QRF that would be appricated.

MarkSiedle (approved):
Discussed with Matt. We'll make a separate card to action the other areas as part of QRF.

mcasperson:
Thanks @MarkSiedle. You're diligence in these reviews is much appreciated.

mcasperson:
Fixed https://github.com/OctopusDeploy/Issues/issues/6521

PR #668, opened 2020-08-20
Searched more locations for the Chart.yaml file

mcasperson:
This PR allows the helm upgrade step to search all sub-directories for the Helm Chart.yaml file. This means the package ID no longer needs to match the directory inside the helm archive.

The likely case where this happens is if someone uploads a chart to the built in feed, but uses a filename like `chart-1.1.1.tgz`. This results in a package id like `chart-1` and a version of `1.1`.

See https://octopususergroup.slack.com/archives/C6UGLUWMQ/p1597331156100800 for an example


mcasperson:
> this doesn't need to explicitly test the dash case as well, does it?

Ironically the package being tested is called `mychart-0.3.7.tgz`, but the test framework sets the package ID manually to the correct value so the filename isn't used here.

mcasperson:
Fixes https://github.com/OctopusDeploy/Issues/issues/6527

PR #670, opened 2020-09-01
Fix other steps that don't work in runbooks

adam-mccoy:
# Background

This change is the continuation of https://github.com/OctopusDeploy/Calamari/pull/667.

`Octopus.Action.Package.PackageId` doesn't exist in runbooks, but `Octopus.Action.Package[].PackageId` does. This PR updates the last remaining step types that reference the former variable to use the latter instead, making them function correctly in both deployments and runbooks.

# Testing
Manually tested the following steps in both deployment and runbook processes:
- Deploy Package with NGINX Feature
- Deploy Package with Legacy IIS (6+)

Also tested the AlreadyInstalledConvention in both above.


adam-mccoy:
[Fix the remaining Calamari references to the old Package variable - these are just waiting to bite anyone who uses Runbooks](https://trello.com/c/ryw3Qx1y/3629-fix-the-remaining-calamari-references-to-the-old-package-variable-these-are-just-waiting-to-bite-anyone-who-uses-runbooks)

andrewabest (approved):
Summary from :zoom:

- Tests are 💚 , have also manually tested the steps in question to ensure they execute correctly.
- Going to make a change to allow us to specify a PackageId key that does not have a step reference, to avoid all of the `.IndexedPackageId(string.empty)` around.

source/Calamari.Common/Plumbing/Variables/PackageVariables.cs
```diff
@@ -18,15 +19,12 @@ public static class PackageVariables
         public static readonly string SubstituteInFilesTargets = "Octopus.Action.SubstituteInFiles.TargetFiles";
         public static readonly string PackageCollection = "Octopus.Action.Package";
 
-        public static string IndexedPackageId(string packageReferenceName)
-        {
-            return $"Octopus.Action.Package[{packageReferenceName}].PackageId";
-        }
+        public static string PackageId => string.Format(PackageIdTemplate, string.Empty);
```
flin-8, line 22: Can these just call `IndexedPackageId("")` etc.

PR #671, opened 2020-09-01
Removing feature flags no longer in use

andyinaus:
This PR addresses the inconsistency among feature enable flag usage(s). 

Please refer to the relevant [Slack thread](https://octopusdeploy.slack.com/archives/CDANN5QLT/p1598843004074700) for clarification/direction.

TL;DR: We are moving towards using `Octopus.Action.EnabledFeatures` for all feature `enable` flag(s)

### DIFF
--- a/source/Calamari.Common/Features/Behaviours/ConfigurationVariablesBehaviour.cs
+++ b/source/Calamari.Common/Features/Behaviours/ConfigurationVariablesBehaviour.cs
@@ -25,7 +25,10 @@ namespace Calamari.Common.Features.Behaviours
 
         public bool IsEnabled(RunningDeployment context)
         {
-            return context.Variables.GetFlag(KnownVariables.Package.AutomaticallyUpdateAppSettingsAndConnectionStrings);
+            var features = context.Variables.GetStrings(KnownVariables.Package.EnabledFeatures)
+                                  .Where(s => !string.IsNullOrWhiteSpace(s)).ToList();
+
+            return features.Contains(KnownVariables.Features.ConfigurationVariables);
         }
 
         public Task Execute(RunningDeployment context)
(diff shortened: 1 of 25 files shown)


source/Calamari.Common/Features/Behaviours/ConfigurationVariablesBehaviour.cs
```diff
@@ -25,7 +25,10 @@ public ConfigurationVariablesBehaviour(ICalamariFileSystem fileSystem, IConfigur
 
         public bool IsEnabled(RunningDeployment context)
         {
-            return context.Variables.GetFlag(KnownVariables.Package.AutomaticallyUpdateAppSettingsAndConnectionStrings);
+            var features = context.Variables.GetStrings(KnownVariables.Package.EnabledFeatures)
```
flin-8, line 28: Looks like we would be calling `GetStrings()` and `features.Contains()` in a few places, how about making this an extension method so we can call `context.Variables.IsFeatureEnabled(KnownVariables.Features.ConfigurationVariables)`?

source/Calamari.Shared/Deployment/Conventions/SubstituteInFilesConvention.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using Calamari.Common.Commands;
+using Calamari.Common.Features.Behaviours;
+
+namespace Calamari.Deployment.Conventions
+{
+    public class SubstituteInFilesConvention : IInstallConvention
```
flin-8, line 7: Good to see this come back, we should get rid of `DelegateInstallConvention`

flin-8 (changes requested):
Looks good!, just the comment around `IsFeatureEnabled()`

PR #672, opened 2020-09-03
Ensure Tar extractor uses UTF8

johnsimons:
Fixes https://github.com/OctopusDeploy/Issues/issues/6472

I've decided to just use a small pool of random characters from different alphabets.
I am using Latin, Greek and Arabic.


flin-8 (approved):
Tests look good ✔️

PR #677, opened 2020-09-14
Upgrade Microsoft.Web.Xdt reference

johnsimons:
We forked Microsoft.Web.Xdt back in 2016 when the package did not target netstandard, but now the package does target netstandard so we updating to use it instead.
Fixes https://github.com/OctopusDeploy/Issues/issues/6569

The actual fix is [this line](https://github.com/dotnet/xdt/blob/cf071c1878487a728495acb3d233afb389edaa54/src/Microsoft.Web.XmlTransform/XmlAttributePreservationDict.cs#L70) which our fork does not contain https://github.com/OctopusDeploy/Octopus.Web.Xdt/blob/c14be275c8eabbc621908db1e9722072e34ac672/Octopus.Web.XmlTransform/XmlAttributePreservationDict.cs#L68-L69

### DIFF
--- a/source/Calamari.Tests/Fixtures/ConfigurationTransforms/ConfigurationTransformsFixture.cs
+++ b/source/Calamari.Tests/Fixtures/ConfigurationTransforms/ConfigurationTransformsFixture.cs
@@ -28,6 +28,12 @@ namespace Calamari.Tests.Fixtures.ConfigurationTransforms
         }
 
         [Test]
+        public void XmlWithNamespacesIsCorrectlyProcessed()
+        {
+            PerformTest(GetFixtureResource("Samples", "nlog.config"), GetFixtureResource("Samples", "nlog.Release.config"));
+        }
+
+        [Test]
         [RequiresMonoVersion423OrAbove] //Bug in mono < 4.2.3 https://bugzilla.xamarin.com/show_bug.cgi?id=19426
         public void WebReleaseConfig()
         {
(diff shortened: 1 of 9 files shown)


source/Calamari.Tests/Fixtures/ConfigurationTransforms/ConfigurationTransformsFixture.cs
```diff
@@ -27,6 +27,12 @@ public void SetUp()
             configurationTransformer = ConfigurationTransformer.FromVariables(variables, log);
         }
 
+        [Test]
+        public void XmlWithNamespacesIsCorrectlyProcessed()
+        {
+            PerformTest(GetFixtureResource("Samples", "nlog.config"), GetFixtureResource("Samples", "nlog.Release.config"));
```
flin-8, line 33: Can we make some assertions about the result of the transform?

flin-8 (approved):
Nice targeted change.

I think we just need to assert on the results of the transform (and there probably isn't much test coverage for this)

source/Calamari.Tests/Fixtures/ConfigurationTransforms/ConfigurationTransformsFixture.cs
johnsimons: we could, but before it was blowing up

johnsimons: Fixed

PR #680, opened 2020-10-07
"Upgrade a Helm Chart" command does not replace variables in values.yaml

johnsimons:
As stated in our docs, it should do this replacement automatically.
https://octopus.com/docs/deployment-examples/kubernetes-deployments/helm-update#template-values

Fixes https://github.com/OctopusDeploy/Issues/issues/6476


source/Calamari.Tests/KubernetesFixtures/HelmUpgradeFixture.cs
```diff
@@ -117,26 +116,27 @@ public virtual void SetUp()
...
-        
+
         [Test(Description = "Test the case where the package ID does not match the directory inside the helm archive.")]
         [RequiresNonFreeBSDPlatform]
         [RequiresNon32BitWindows]
         [RequiresNonMac]
         [Category(TestCategory.PlatformAgnostic)]
         public void MismatchPackageIDAndHelmArchivePathWorks()
         {
+            var packageName = $"{Variables.Get(PackageVariables.PackageId)}-{Variables.Get(PackageVariables.PackageVersion)}.tgz";
```
flin-8, line 135: This is the same as the default value in `DeployPackage()` right?

johnsimons: it is but it is executed before I change it in the next lines 😉

flin-8 (approved):
Looking ace 👍

PR #684, opened 2020-10-21
Initialises parameters string before use in PowerShell codegen

andrewabest:
This PR is the Calamari side of fixing https://github.com/OctopusDeploy/Issues/issues/6641

It ensures the parameters variable declared in our Powershell code gen is initialized before usage.

We will add an E2E test in Server to cover this case.


andrewabest:
I've built and tested this locally in Server, and it fixes the issue identified.

johnsimons (approved):
👍

paulegradie (approved):
👌

PR #686, opened 2020-10-23
Fix retentioin policy

andyinaus:
# Background

If users configure to keep N deployments, our target retention policy doesn't clean up files properly for unsuccessful deployments in between the successful deployments we would like to keep. 

e.g.
If users configure to keep 1 deployment, we currently only remove files for deployments occurred prior to deployment 1 (inclusive) instead of all deployments excluding deployment 99 and 2 at deployment 99.
**Deployment Journal:**
- Deployment 99 Success 06/09/2020 -> To Keep (Current)
- Deployment 98 Failed    05/09/2020 -> To Keep
- Deployment 97 Failed    04/09/2020 -> To Keep
- Deployment 96 Failed    03/09/2020 -> To Keep
- (Other failed deployments)                -> To Keep
- Deployment 2 Success   02/09/2020 -> To Keep
- Deployment 1 Failed      01/09/2020 -> To Remove
- ... (any deployments older than the previous one) -> To Remove

[Trello Card](https://trello.com/c/UPUpiTr7/3750-packages-associated-with-failed-steps-do-not-get-cleaned-up)

# Result

Only keeping Deployment 99 and 2 when retention is triggered at deployment 99

### DIFF
--- a/source/Calamari.Tests/Fixtures/Retention/RetentionPolicyFixture.cs
+++ b/source/Calamari.Tests/Fixtures/Retention/RetentionPolicyFixture.cs
@@ -4,7 +4,6 @@ using System.Linq;
 using Calamari.Common.Features.Deployment.Journal;
 using Calamari.Common.Plumbing.FileSystem;
 using Calamari.Deployment.Retention;
-using Calamari.Integration.FileSystem;
 using Calamari.Integration.Time;
 using NSubstitute;
 using NUnit.Framework;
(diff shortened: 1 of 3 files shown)


andrewabest (approved):
Andy took me through the changes and explained the scenario that we were fixing retention for. 

We looked at the code changes and made a small tweak to make reading the code a little easier.

We discussed a change that may make the tests have less coupling to the implementation detail of the retention policy - by returning out the list of items that we evaluated should be cleaned up. However we opted to keep with the existing style of tests that were already implemented to minimise churn on this particular PR.

You also noted that one of the input parameters was incorrectly named - it was "releases", but should be "deployments". Once again we agreed to not tackle this issue now as we are rolling of QRF and coordinating a calamari update in server would take more than our afternoon of time, but to leave a TODO for the next team who are working in this area to clean up.

Nice fix! :shipit:

PR #688, opened 2020-11-16
Add timeout argument to run-script command.

lilwa29:
This PR address the issue defined here, up to date with Tag version 16.4: https://github.com/OctopusDeploy/Calamari/pull/468#issue-362311331

**Issue**
Tentacles wait indefinitely on actions which run scripts. For the most part, this is desirable by design and handled here. There are many valid cases were a timeout on scripts should be supported to prevent tentacles from being indefinitely stuck processing action scripts (and thus preventing other operations such as clean up via Script Console, Runbooks, or other deployments).

**Pull Request**

This PR adds the special variable Octopus.Action.Script.Timeout to allow a millisecond timeout to be defined via a process's variables that will be used to limit the amount of time a process can wait. If no timeout is defined, the default behavior of waiting indefinitely will be used. 

Added tests for CommandLineRunner and ScriptExecutor to ensure timeout is applied to both the command line instruction, and is applied during execution.

### DIFF
--- a/source/Calamari.Common/Features/Processes/CommandLineInvocation.cs
+++ b/source/Calamari.Common/Features/Processes/CommandLineInvocation.cs
@@ -27,6 +27,12 @@ namespace Calamari.Common.Features.Processes
 
         public Dictionary<string, string>? EnvironmentVars { get; set; }
 
+
+        /// <summary>
+        /// Gets or sets a timeour in milliseconds for the command line process execution
+        /// </summary>
+        public TimeSpan Timeout { get; set; } = TimeSpan.Zero;
+
         /// <summary>
         /// Prevent this execution from starting if another execution is running that also has this set to true.
         /// It does not isolate from executions that have this set to false.
--- a/source/Calamari.Common/Features/Processes/SilentProcessRunner.cs
+++ b/source/Calamari.Common/Features/Processes/SilentProcessRunner.cs
@@ -51,7 +51,8 @@ namespace Calamari.Common.Features.Processes
                 null,
                 null,
                 output,
-                error);
+                error,
+                TimeSpan.Zero);
         }
 
         public static SilentProcessRunnerResult ExecuteCommand(
@@ -69,7 +70,8 @@ namespace Calamari.Common.Features.Processes
                 null,
                 null,
                 output,
-                error);
+                error,
+                TimeSpan.Zero);
         }
 
         public static SilentProcessRunnerResult ExecuteCommand(
@@ -80,7 +82,8 @@ namespace Calamari.Common.Features.Processes
             string? userName,
             SecureString? password,
             Action<string> output,
-            Action<string> error)
+            Action<string> error,
+            TimeSpan timeout)
         {
             try
             {
@@ -105,6 +108,7 @@ namespace Calamari.Common.Features.Processes
                     using (var outputWaitHandle = new AutoResetEvent(false))
                     using (var errorWaitHandle = new AutoResetEvent(false))
                     {
+                        var timedOut = false;
                         var errorData = new StringBuilder();
                         process.OutputDataReceived += (sender, e) =>
                         {
@@ -160,11 +164,30 @@ namespace Calamari.Common.Features.Processes
                         process.BeginOutputReadLine();
                         process.BeginErrorReadLine();
 
-                        process.WaitForExit();
+                        if (timeout > TimeSpan.Zero)
+                        {
+                            if(!process.WaitForExit((int)timeout.TotalMilliseconds))
+                            {
+                                timedOut = true;
+                                error($"Script execution timed out after {timeout.TotalMilliseconds}ms");
+
+                                // In certain cases the process may have exited already, if not we can kill it
+                                // and the wait handles will both be signaled as per normal
+                                if(!process.HasExited)
+                                {
+                                    process.Kill();
+                                }
+                            }
+                        }
+                        else
+                        {
+                            process.WaitForExit();
+                        }
+
                         outputWaitHandle.WaitOne();
                         errorWaitHandle.WaitOne();
 
-                        return new SilentProcessRunnerResult(process.ExitCode, errorData.ToString());
+                        return new SilentProcessRunnerResult(process.ExitCode, errorData.ToString(), timedOut);
                     }
                 }
             }
--- a/source/Calamari/Commands/RunScriptCommand.cs
+++ b/source/Calamari/Commands/RunScriptCommand.cs
@@ -29,6 +29,7 @@ namespace Calamari.Commands
         string scriptFileArg;
         string packageFile;
         string scriptParametersArg;
+        TimeSpan timeout;
         readonly ILog log;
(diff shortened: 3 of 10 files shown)


source/Calamari.Common/Features/Processes/SilentProcessRunner.cs
```diff
@@ -160,11 +164,30 @@ static SilentProcessRunner()
                         process.BeginOutputReadLine();
                         process.BeginErrorReadLine();
 
-                        process.WaitForExit();
+                        if (timeout > TimeSpan.Zero)
```
miguelelvir, line 167: Could probably replace this with a local function, but this seems readable.

miguelelvir (approved):
Thank you for updating this one to latest code. Works in my case as well.

miguelelvir:
Perhaps someone from the Octo team can review? @slewis74  has some history in this file.

miguelelvir:
@droyad Had reviewed the previous one as well.

source/Calamari.Common/Features/Processes/CommandLineInvocation.cs
```diff
@@ -27,6 +27,12 @@ public CommandLineInvocation(string executable, params string[] arguments)
 
         public Dictionary<string, string>? EnvironmentVars { get; set; }
 
+
+        /// <summary>
+        /// Gets or sets a timeour in milliseconds for the command line process execution
+        /// </summary>
+        public TimeSpan Timeout { get; set; } = TimeSpan.Zero;
```
droyad, line 34:
```suggestion
        public TimeSpan Timeout { get; set; } = Timeout.Infinte;
```

source/Calamari.Common/Features/Processes/SilentProcessRunner.cs
```diff
@@ -51,7 +51,8 @@ static SilentProcessRunner()
                 null,
                 null,
                 output,
-                error);
+                error,
+                TimeSpan.Zero);
```
droyad, line 55:
```suggestion
                Timeout.Infinte);
```

source/Calamari.Common/Features/Processes/SilentProcessRunner.cs
```diff
@@ -69,7 +70,8 @@ static SilentProcessRunner()
...
```
droyad, line 74:
```suggestion
                Timeout.Infinte);
```

source/Calamari/Commands/RunScriptCommand.cs
```diff
@@ -50,6 +51,7 @@ ISubstituteInFiles substituteInFiles
             Options.Add("package=", "Path to the package to extract that contains the script.", v => packageFile = Path.GetFullPath(v));
             Options.Add("script=", $"Path to the script to execute. If --package is used, it can be a script inside the package.", v => scriptFileArg = v);
             Options.Add("scriptParameters=", $"Parameters to pass to the script.", v => scriptParametersArg = v);
+            Options.Add("timeout=", $"Timeout for script in milliseconds.", v => timeout = int.TryParse(v, out var timeoutMs) && timeoutMs > 0 ? TimeSpan.FromMilliseconds(timeoutMs) : TimeSpan.Zero);
```
droyad, line 54: This isn't needed. The timeout will should always be set via the variable bag and `SpecialVariables.Action.Script.Timeout`

droyad (commented):
I'm on the fence a little about this. On one hand this is a simple change. However I get stuck trying to see how it fits into all the other things a step could do.

Should we perhaps create a cancellation token at the top level and pass that through as needed?

johnsimons:
Hi @lilwa29,

Thanks so much for this contribution.
I have reviewed your PR and I am not totally on board with this change, let me explain.

This would be a setting that is global for the whole Octopus Project, in other words, this would affect all script steps defined in a deployment process, which IMO is probably not what most customers desire.

If we were to support timeouts on script execution, I believe that should be part of the step UI, so the user can specify it or leave blank as part of authoring the step. Special variables are hard to discover.

The second reason why I am not comfortable with this PR is that I think we can solve this at the script level.
Example in Powershell:
```Powershell
$Job = Start-Job -ScriptBlock {
    # the code that I want to not execute for more than 30 seconds
}

$Job | Wait-Job -Timeout 30
$Job | Stop-Job
```
This way the user has total control what parts of their script require such timeouts.

So with all this in mind, I am going to close this PR.

If you strongly disagree with my assessment please let us know, I'll be happy to reopen and reevaluate the need for this request.
Once again thank you for taking the time to contribute, even though this time we are not taking the contribution.

Regards
John

leonio:
Thanks @johnsimons - we really appreciate the guidance and feedback here. As mentioned on the help forum, I think we could have explained the situation a little better - apologies for that, and hope to clear some of that up here.

The issue we've been encountering is due to a race condition issue with the IIS `ServerManager.CommitChanges` method. When execute this method on systems under high-load, the `CommitChanges` method returns too fast. While this bug has existed for some time, it seems the best way to copy with this when writing code is `Thead.Sleep` as pointed out in [this stackoverflow post](https://stackoverflow.com/questions/7446632/servermanager-commitchanges-makes-changes-with-a-slight-delay/7588790#7588790).

When used within Octopus and Calamari (for example [IISWebSite_BeforePostDeploy.ps1](https://github.com/OctopusDeploy/Calamari/blob/master/source/Calamari/Scripts/Octopus.Features.IISWebSite_BeforePostDeploy.ps1#L273) the issue becomes unavoidable (even with a `-Timeout` argument in the script) because `CommitChanges` exits before the `Set-ItemProperty` method has called `Wait` on it. The result is the our deployments get stuck while the script is waiting a void return from `CommitChanges` that has already occured.

If it possible to re-open this PR or open a new one with an approach to avoid these situations, that would help us eliminate the the need to use a custom modules setup, and with it the dependency on `Calamari.AzureScripting`. 

> If we were to support timeouts on script execution, I believe that should be part of the step UI, so the user can specify it or leave blank as part of authoring the step. Special variables are hard to discover.
That makes sense, and we appreciate that also creates another feature to support and maintain going forward. If there is an alternative approach that works with the issue we're experience please let us know.

> Should we perhaps create a cancellation token at the top level and pass that through as needed?
@droyad - we thought the same, only reason for not doing so was trying to not to introduce any large changes unnecessarily. We can add this in to assist with cancellation through the whole call stack.

The investigation below comes from an initial PR by @miguelelvir when we first started working on this issue  
**Calamari dump showing wait in `SilentProcessRunner`**  
image

**Dump from Powershell showing it's waiting on `Set-ItemProperty`**  
image

**ProcExp confirming that we're still stuck in the script (appcmd)**
image

johnsimons:
Hi @leonio,

Thanks so much for providing more context for this change (I should have read the previous PR).
So what you are describing is a bug in the IIS step, which I think we can tackle it from that context.
So what I am saying is that I believe this issue can be fixed for the IIS step in isolation.

I have raised a new issue in our [public facing repo](https://github.com/OctopusDeploy/Issues/issues/6740), with all the detail you have provided and we can then decide how to proceed over there.

PR #693, opened 2020-12-11
Skip tests on FreeBSD/Mono.

andrew-moore-octo:
This PR skips some tests on FreeBSD/Mono because:

- The case for supporting Mono is weakening
- The performance of `HttpListener` on Mono/FreeBSD seems to be significantly different to other platforms and runtimes.


matt-richardson (approved):
Nice one. :shipit:

PR #696, opened 2020-12-15
Apply structured configuration variables for files that don't use well known extensions

TomPeters:
Fixes https://github.com/OctopusDeploy/Issues/issues/6683

A common use case for structured configuration variables is substituting variables in files like `web.config`. This particular type of file is usually an xml file, but does not contain a `.xml` file extension. Our current structured configuration variables feature is unable to process this file because it relies heavily on the file extension to determine which parser it should use.

This PR changes the algorithm for determining which parser to use for substituting variables in structured configuration files, such that we can substitute variables in files that don't have well known extensions.

# Before

The algorithm used to look like this
- Always try to use a JSON parser first for backwards compatibility
- If the file contents can not be parsed as JSON, then:
- If we recognised the file extension (`.xml`, `.yaml`, `.yml` or `.properties`), then additionally try to parse using the single matching parser
- else if neither the JSON nor the parser matching the file extension works, we fail the process

# After

The algorithm now looks like this
- Always try to use a JSON parser first for backwards compatibility
- If the file contents can not be parsed as JSON then:
- If we recognised the file extension (`.xml`, `.yaml`, `.yml` or `.properties`), then additionally try to parse using the single matching parser
- else if the JSON parser did not work and we did not recognise the file extension, additionally try the remaining parsers in order (xml -> yaml -> properties) until we find one that works

If you want to understand the behaviour after this change, have a close look at the tests in this PR.

### DIFF
--- a/source/Calamari.Testing/Calamari.Testing.csproj
+++ b/source/Calamari.Testing/Calamari.Testing.csproj
@@ -1,11 +1,16 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
     <PropertyGroup>
-        <TargetFrameworks>net452;netstandard2.1</TargetFrameworks>
         <Nullable>enable</Nullable>
         <PackageProjectUrl>https://github.com/OctopusDeploy/Calamari/</PackageProjectUrl>
         <PackageLicenseExpression>Apache-2.0</PackageLicenseExpression>
         <LangVersion>default</LangVersion>
+    </PropertyGroup>
+    <PropertyGroup Condition="!$([MSBuild]::IsOSUnixLike())">
+        <TargetFrameworks>net452;netstandard2.1</TargetFrameworks>
+    </PropertyGroup>
+    <PropertyGroup Condition="$([MSBuild]::IsOSUnixLike())">
+        <TargetFramework>netstandard2.1</TargetFramework>
     </PropertyGroup>
 
     <ItemGroup>
--- a/source/Calamari.Tests/Fixtures/Deployment/DeployPackageWithStructuredConfigurationFixture.cs
+++ b/source/Calamari.Tests/Fixtures/Deployment/DeployPackageWithStructuredConfigurationFixture.cs
@@ -1,4 +1,4 @@
-﻿using System.IO;
+using System.IO;
 using Assent;
 using Calamari.Common.Plumbing.FileSystem;
 using Calamari.Common.Plumbing.Variables;
@@ -20,6 +20,7 @@ namespace Calamari.Tests.Fixtures.Deployment
         const string PropertiesFileName = "config.properties";
         const string MalformedFileName = "malformed.file";
         const string XmlFileNameWithNonXmlExtension = "xml.config";
+        const string XmlFileNameWithJsonExtension = "xml.json";
         const string YamlFileNameWithNonYamlExtension = "yaml.config";
         const string PropertiesFileNameWithNonPropertiesExtension = "properties.config";
         const string PropertiesFileNameWithYamlExtension = "properties.yaml";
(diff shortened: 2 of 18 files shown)


source/Calamari.Testing/Calamari.Testing.csproj
```diff
@@ -1,12 +1,17 @@
...
         <Nullable>enable</Nullable>
         <PackageProjectUrl>https://github.com/OctopusDeploy/Calamari/</PackageProjectUrl>
         <PackageLicenseExpression>Apache-2.0</PackageLicenseExpression>
         <LangVersion>default</LangVersion>
     </PropertyGroup>
+    <PropertyGroup Condition="!$([MSBuild]::IsOSUnixLike())">
+        <TargetFrameworks>net452;netstandard2.1</TargetFrameworks>
+    </PropertyGroup>
+    <PropertyGroup Condition="$([MSBuild]::IsOSUnixLike())">
+        <TargetFramework>netstandard2.1</TargetFramework>
```
TomPeters, line 13: Unrelated change, but this unblocks me from being able to develop on a linux environment

source/Calamari.Tests/Fixtures/Deployment/DeployPackageWithStructuredConfigurationFixture.cs
```diff
@@ -261,7 +250,7 @@ public void FailsIfAnyFileMatchingAGlobFailsToParse()
         }
 
         [Test]
-        public void FailsIfAFileFailsToParseWhenThereAreManyGlobs()
+        public void FailsIfAFileFailsToParseWhenThereAreMultipleTargetFiles()
```
TomPeters, line 253: Rename to better reflect the test scenario

source/Calamari.Tests/Fixtures/Deployment/DeployPackageWithStructuredConfigurationFixture.cs
```diff
@@ -311,6 +300,150 @@ public void FailsAndWarnsIfAFileCannotBeParsed()
...
+                result.AssertSuccess();
+
+                var extractedPackageUpdatedConfigFile = File.ReadAllText(Path.Combine(StagingDirectory, ServiceName, ServiceVersion, ConfigFileName));
+
+                this.Assent(extractedPackageUpdatedConfigFile, TestEnvironment.AssentJsonConfiguration);
+            }
+        }
+
+        [Test]
+        public void ShouldPerformReplacementInXmlFileWithANonXmlExtension()
```
TomPeters, line 322: This is the fundamental scenario we want to unblock in this PR

source/Calamari.Tests/Fixtures/Deployment/DeployPackageWithStructuredConfigurationFixture.cs
```diff
@@ -311,6 +300,150 @@ public void FailsAndWarnsIfAFileCannotBeParsed()
...
+
+                var extractedPackageUpdatedPropertiesFile = File.ReadAllText(Path.Combine(StagingDirectory, ServiceName, ServiceVersion, PropertiesFileNameWithNonPropertiesExtension));
+
+                // noship: assert that it tried json, xml and yaml first
+                this.Assent(extractedPackageUpdatedPropertiesFile, TestEnvironment.AssentPropertiesConfiguration);
+            }
+        }
+
+        [Test]
+        public void FailsAndWarnsIfFileWithAnXmlExtensionContainsValidNonXmlContent()
```
TomPeters, line 380: This (and similar tests below) test that if a file has a well known extension, for example `.xml`, then we _don't_ try other parsers (like yaml, properties, json) even if the content was not parseable as `xml` content

andrew-moore-octo (approved):
Nice one!

- New test coverage looks good.
- We were worried about some side effects (like confusing logs, or nearly every file being treated as .properties on invalid input) when we couldn't easily determine the file format, but I'm confident that we've sidestepped those issues.
- General code quality looks 👌

PR #697, opened 2020-12-16
More nullable reference type updates

slewis74:
As part of working on getting the newer versions (NetCore) of Calamari to run on newer versions of Windows a few more nullable reference type false positives came up. This PR is to address them.


source/Calamari.Common/Plumbing/Variables/IVariables.cs
```diff
@@ -10,13 +10,17 @@ public interface IVariables : IEnumerable<KeyValuePair<string, string>>
...
+        string? GetRaw(string variableName);
         [return: NotNullIfNotNull("defaultValue")]
         string? Get(string variableName, string? defaultValue = null);
         [return: NotNullIfNotNull("defaultValue")]
-        string? Get(string variableName, out string error, string? defaultValue = null);
-        string Evaluate(string expressionOrVariableOrText, out string error, bool haltOnError = true);
-        string Evaluate(string expressionOrVariableOrText);
+        string? Get(string variableName, out string? error, string? defaultValue = null);
+
+        [return: NotNullIfNotNull("expressionOrVariableOrText")]
```
benPearce1, line 19: This sucks a bit, looks like .Net Core will improve this in the future https://github.com/dotnet/csharplang/issues/373

source/Calamari.Common/Features/StructuredVariables/XmlFormatVariableReplacer.cs
```diff
@@ -61,6 +61,8 @@ public void ModifyFile(string filePath, IVariables variables)
                 {
                     var selectedNodes = navigator.XPath2SelectNodes(xPathExpression, nsManager);
                     var variableValue = variables.Get(variable.Key);
+                    if (variableValue == null)
+                        continue;
```
benPearce1, line 65: How do we want to handle empty strings here?

benPearce1 (approved):
👍 
no functional changes I can see. One question.

source/Calamari.Common/Plumbing/Variables/IVariables.cs
slewis74: yes, I did try `nameof`, just to be sure it didn't work :)

source/Calamari.Common/Features/StructuredVariables/XmlFormatVariableReplacer.cs
slewis74: good question. I've revisited and put it back to assigning the nulls, the Xml objects all seems to deal with that correctly and then we don't risk accidentally changing behavior

PR #699, opened 2021-01-12
Upgrade Helm Chart won't log warning if values.yaml not present

MJRichardson:
Relates to https://github.com/OctopusDeploy/Issues/issues/6728

### DIFF
--- a/source/Calamari.Common/Features/Substitutions/ISubstituteInFiles.cs
+++ b/source/Calamari.Common/Features/Substitutions/ISubstituteInFiles.cs
@@ -7,6 +7,6 @@ namespace Calamari.Common.Features.Substitutions
     public interface ISubstituteInFiles
     {
         void SubstituteBasedSettingsInSuppliedVariables(RunningDeployment deployment);
-        void Substitute(RunningDeployment deployment, IList<string> filesToTarget);
+        void Substitute(RunningDeployment deployment, IList<string> filesToTarget, bool warnIfFileNotFound = true);
     }
 }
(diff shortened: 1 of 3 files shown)


source/Calamari.Common/Features/Substitutions/ISubstituteInFiles.cs
```diff
@@ -7,6 +7,6 @@ namespace Calamari.Common.Features.Substitutions
     public interface ISubstituteInFiles
     {
         void SubstituteBasedSettingsInSuppliedVariables(RunningDeployment deployment);
-        void Substitute(RunningDeployment deployment, IList<string> filesToTarget);
+        void Substitute(RunningDeployment deployment, IList<string> filesToTarget, bool warnIfFileNotFound = true);
```
MJRichardson, line 10: Added additional parameter to indicate if warning should be logged

source/Calamari/Kubernetes/Commands/HelmUpgradeCommand.cs
```diff
@@ -71,7 +71,10 @@ public override int Execute(string[] commandLineArguments)
                 new DelegateInstallConvention(d => extractPackage.ExtractToStagingDirectory(pathToPackage)),
                 new StageScriptPackagesConvention(null, fileSystem, new CombinedPackageExtractor(log), true),
                 new ConfiguredScriptConvention(new PreDeployConfiguredScriptBehaviour(log, fileSystem, scriptEngine, commandLineRunner)),
-                new DelegateInstallConvention(d => substituteInFiles.Substitute(d, FileTargetFactory().ToList())),
+                // Any values.yaml files in any packages referenced by the step will automatically have variable substitution applied (we won't log a warning if these aren't present)
+                new DelegateInstallConvention(d => substituteInFiles.Substitute(d, DefaultValuesFiles().ToList(), false)),
```
mcasperson, line 75: Is this correct? DefaultValuesFiles().ToList() would generate the same list of files each time, so the two calls to `Substitute` do the same thing (except for the last param being true instead of false)?

mcasperson (changes requested):
I think the two calls to `substituteInFiles.Substitute` would need to be fixed.

source/Calamari/Kubernetes/Commands/HelmUpgradeCommand.cs
MJRichardson: Oops. No.

mcasperson (approved):
LGTM

PR #702, opened 2021-02-18
Adding test for execute-manifest command

johnsimons:


andrewabest (approved):
I pushed up a small refactor just to extract some of the bootstrap-type functionality from the test class itself and make the test itself a bit easier to read / understand. Test itself looks good, very happy with the novel use of dotnet new to create a "test node" to work with within the command.

source/Calamari.Tests/Fixtures/Manifest/ExecuteManifestCommandFixture.cs
```diff
@@ -56,8 +56,7 @@ public void WithInstructions()
             {
                 var generatedApplicationPath = CodeGenerator.GenerateConsoleApplication("node", temporaryDirectory.DirectoryPath);
 
-                var destinationPath =
-                    CalamariEnvironment.IsRunningOnWindows ? Path.Combine(temporaryDirectory.DirectoryPath, "app") : Path.Combine(temporaryDirectory.DirectoryPath, "app", "bin");
+                var destinationPath = Path.Combine(temporaryDirectory.DirectoryPath, "app");
```
johnsimons, line 59: @andrewabest  The paths are different in Linux vs Windows, did u test this in Linux?

PR #703, opened 2021-03-03
Support java package types via CombinedPackageExtractor

MJRichardson:
Today, java packages are only supported by the [DeployJavaArchiveCommand](https://github.com/OctopusDeploy/Calamari/blob/master/source/Calamari/Commands/Java/DeployJavaArchiveCommand.cs) which explicitly uses the `JarPackageExtractor` class.  

In the newer model of inheriting from `PipelineCommand`, this isn't an option.  

This PR extends the `CombinedPackageExtractor` to use the `JarPackageExtractor` for the appropriate java file types (`.jar`, `.war`, `.ear`)

The immediate need for this is to allow the [AzureAppService](https://github.com/OctopusDeploy/Sashimi.AzureAppService/blob/main/source/Calamari/Behaviors/AzureAppServiceBehaviour.cs) step to support deploying java apps.

### DIFF
--- a/source/Calamari.Common/Features/Packages/CombinedPackageExtractor.cs
+++ b/source/Calamari.Common/Features/Packages/CombinedPackageExtractor.cs
@@ -2,8 +2,11 @@
 using System.IO;
 using System.Linq;
 using Calamari.Common.Commands;
+using Calamari.Common.Features.Packages.Java;
 using Calamari.Common.Features.Packages.NuGet;
+using Calamari.Common.Features.Processes;
 using Calamari.Common.Plumbing.Logging;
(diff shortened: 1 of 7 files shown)


andrewabest (approved):
Outside of this PR we discussed testing the actual behaviour of the Jar extractor via the Combined Extractor - we aren't adding it here as not all of our test infrastructure has Java installed. You've agreed to cover it in the Azure App Service test suite. Aside from that, the change looks good. By adding the extractor at the end of the set, it won't conflict with any of the existing extractors (zip for example). Looks good to :shipit:

mcasperson (approved):
This looks like a nice simple change to me.

PR #704, opened 2021-03-11
Refactored logging infrastructure

slewis74:
Updates to the Calamari testing code, for the new server side logging infrastructure

### DIFF
--- a/source/Calamari.Testing/LogParser/ScriptOutputFilter.cs
+++ b/source/Calamari.Testing/LogParser/ScriptOutputFilter.cs
@@ -151,7 +151,7 @@ namespace Calamari.Testing.LogParser
 
     public class ScriptOutputFilter
     {
-        readonly ILogWithContext log;
+        readonly ITaskLog log;
         readonly ServiceMessageParser parser;
 
         readonly Action<string> nullTarget = s =>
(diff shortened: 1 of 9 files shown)


source/Calamari.Testing/LogParser/ScriptOutputFilter.cs
```diff
@@ -225,20 +223,17 @@ public void Write(ProcessOutputSource source, string text)
 
         void WritePlainText(ProcessOutputSource source, string text)
         {
-            using (log.WithinBlock(logContext))
```
benPearce1, line 228: I know this is only for test purposes, but I am not sure what impact removing this has?

benPearce1 (approved):
LGTM, one question.

source/Calamari.Testing/LogParser/ScriptOutputFilter.cs
slewis74: This is a good question, and what this line used to do, and why, was the reason it is going away :) What it used to do was change the thread static context to be the one with the correlationId (that's all ILogWithContext really was) before calling the callbacks. Those callback would have an ILog, most likely, that would depend on the thread context being right before they got called, and do terrible things if it wasn't. Now the tasklog instance has to be passed in, and it has the correlationId and there's no thread context involved, thus the callbacks can only write to the correct place.

PR #705, opened 2021-03-16
Fixing the context name with Azure admin login

mcasperson:
When logging in with an Azure admin account, the context name is the cluster name plus "-admin".

This PR updates an issue where the default namespace is not updated in the correct Kubernetes context for an admin login.


mcasperson:
Fixes https://github.com/OctopusDeploy/Issues/issues/6793

mcasperson:
This fix is valid but I'm closing this PR for now as I don't have time to push it through the builds.

PR #712, opened 2021-03-31
Backported: restrict tests to mono version 5.14.0 or lower

ray-nham:
# Background
[The Debian - Mono 5.14.0 Test](https://build.octopushq.com/buildConfiguration/OctopusDeploy_CalamariNetCore_MonoTesting_TestDebianMono5181/2180503) are only compatible with Mono version 5.14.0 or lower. When running on Mono version 5.14.0 or higher, we would like to ignore these tests.
This change have been implemented in `master`. This PR is to backport the change to release `2020.6` so that these tests will not fail when we apply patches to `2020.6` branch.
More info is in [this Slack thread](https://octopusdeploy.slack.com/archives/C012VMX1YMQ/p1617082094111100).


ray-nham:
No release tag as this change is only for our build pipeline.

shaunmarx (approved):
Changes look good to me 👍🏻

PR #714, opened 2021-04-09
Don't use a guid for the domain name

johnsimons:
This can cause security software to flag as potentially domain generated algorithm

Fixes https://github.com/OctopusDeploy/Issues/issues/6821

### DIFF
--- a/source/Calamari.Common/Plumbing/Proxies/ProxySettings.cs
+++ b/source/Calamari.Common/Plumbing/Proxies/ProxySettings.cs
@@ -27,7 +27,7 @@ namespace Calamari.Common.Plumbing.Proxies
 
     public class UseSystemProxySettings : IProxySettings
     {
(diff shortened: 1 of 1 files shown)


mcasperson:
There is another case of this issue at https://github.com/OctopusDeploy/OctopusDeploy/blob/a286afaf715ecbc083d2fce49698bcf6cd8a6f8a/source/Octopus.Tests/Core/Http/OctopusHttpClientFactoryFixture.cs#L23 that we should fix.

mcasperson (changes requested):
The infrastructure team has found another instance of a problematic domain that we should also resolve with this PR.

johnsimons:
@mcasperson that looks like it is in a different repo 😉

johnsimons:
and that is in a unit test, so not really a big deal

mcasperson (approved):
Ah, of course. I should have check. LGTM

PR #715, opened 2021-04-09
Adding transform commands

johnsimons:
This is so that we can use them individually for step packages.

I have converted the following behaviours to commands:
- `SubstituteInFilesBehaviour`
- `ConfigurationTransformsBehaviour`
- `ConfigurationVariablesBehaviour`
- `StructuredConfigurationVariablesBehaviour`
           
The majority of the work was to refactor the current implementations, so they do not depend on `RunningDeployment` which does not exist in raw `ICommand` implementations.

We are also honouring `SkipRemainingConventions` flag

### DIFF
--- a/source/Calamari.Common/Features/StructuredVariables/StructuredConfigVariablesService.cs
+++ b/source/Calamari.Common/Features/StructuredVariables/StructuredConfigVariablesService.cs
@@ -11,35 +11,38 @@ namespace Calamari.Common.Features.StructuredVariables
 {
     public interface IStructuredConfigVariablesService
     {
-        void ReplaceVariables(RunningDeployment deployment);
+        void ReplaceVariables(string currentDirectory);
     }
 
     public class StructuredConfigVariablesService : IStructuredConfigVariablesService
     {
         readonly IFileFormatVariableReplacer jsonReplacer;
         readonly IFileFormatVariableReplacer[] allReplacers;
+        readonly IVariables variables;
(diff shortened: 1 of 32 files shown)


source/Calamari.Common/Features/StructuredVariables/StructuredConfigVariablesService.cs
```diff
@@ -11,35 +11,38 @@ namespace Calamari.Common.Features.StructuredVariables
...
         readonly ICalamariFileSystem fileSystem;
         readonly ILog log;
 
         public StructuredConfigVariablesService(
             IFileFormatVariableReplacer[] replacers,
-            ICalamariFileSystem fileSystem,
-            ILog log)
+            IVariables variables,
+        ICalamariFileSystem fileSystem,
+                            ILog log)
```
andrewabest, line 29: formatting

andrewabest:
Aside from the red tests, this looks good. Let's :pear: on this tomorrow morning and get it merged.

source/Calamari.Common/Features/StructuredVariables/StructuredConfigVariablesService.cs
johnsimons: Fixed

andrewabest (approved):
Changes are inline with what we discussed. Build was 💚 before reverting a small commit, so let's ship this one.

PR #716, opened 2021-04-09
K8S: Add Pod Service Account authentication

ray-nham:
# Background

When running a worker as a container in a Kubernetes cluster, it is possible to connect back to the parent cluster using the service account token and cluster certificate files mounted as files in the pod. 

This allows Kubernetes workers to manage the cluster they are deployed to without sending down any additional credentials.

This PR is to implement the ability to use Pod Service Account as a mean of authentication when creating Kubernetes deployment targets. The details of this changes can be found in the pitch document below.

Pitch document: https://docs.google.com/document/d/1Psj5saZqhBDlEhbn6s361X_HXWk4jA9EEaqyGaeWyLI/edit#heading=h.eeq4erbzkzf8

Octopus Server PR: https://github.com/OctopusDeploy/OctopusDeploy/pull/8637

# Results

The script `KubectlPowershellContext.ps1` and `KubectlBashContext.sh` is updated to create a valid context using the information of the Pod Service Account.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites
- [x ] I have considered informing or consulting the right people, e.g.: 
  - For changes to the execution pipeline: [`#topic-execution`](https://octopusdeploy.slack.com/archives/C3KT1DFSM)
  - For other changes, consider the other `#topic-...` channels
- [x ] I have considered whether this should be a patch or wait for a minor.
- [x ] I have considered appropriate testing for my change.

### DIFF
--- a/source/Calamari/Kubernetes/Scripts/KubectlBashContext.sh
+++ b/source/Calamari/Kubernetes/Scripts/KubectlBashContext.sh
@@ -127,7 +127,7 @@ function setup_context {
     kubectl config set-cluster octocluster --server=$Octopus_K8S_ClusterUrl
     
     if [[ -z $Octopus_K8s_Server_Cert_From_Path ]]; then
-      kubectl config set-cluster octocluster --insecure-skip-tls-verify=$SkipTlsVerification
+      kubectl config set-cluster octocluster --insecure-skip-tls-verify=$Octopus_K8S_SkipTlsVerification
     else
       kubectl config set-cluster octocluster --certificate-authority=$Octopus_K8s_Server_Cert_Path
     fi
--- a/source/Calamari/Kubernetes/Scripts/KubectlBashContext.sh
+++ b/source/Calamari/Kubernetes/Scripts/KubectlBashContext.sh
@@ -12,6 +12,10 @@ Octopus_K8S_Client_Cert_Pem=$(get_octopusvariable "${Octopus_K8S_Client_Cert}.Ce
 Octopus_K8S_Client_Cert_Key=$(get_octopusvariable "${Octopus_K8S_Client_Cert}.PrivateKeyPem")
 Octopus_K8S_Server_Cert=$(get_octopusvariable "Octopus.Action.Kubernetes.CertificateAuthority")
 Octopus_K8S_Server_Cert_Pem=$(get_octopusvariable "${Octopus_K8S_Server_Cert}.CertificatePem")
+Octopus_K8s_Server_Cert_Path=$(get_octopusvariable "Octopus.Action.Kubernetes.PodServiceAccountTokenPath")
+Octopus_K8s_Pod_Service_Account_Token_Path=$(get_octopusvariable "Octopus.Action.Kubernetes.PodServiceAccountTokenPath")
+echo >&2 $Octopus_K8s_Server_Cert_Path
+echo >&2 $Octopus_K8s_Pod_Service_Account_Token_Path
 
 function check_app_exists {
 	command -v $1 > /dev/null 2>&1
@@ -71,7 +75,7 @@ function setup_context {
     exit 1
   fi
 
-  if [[ -z $Octopus_AccountType && -z $Octopus_K8S_Client_Cert && ${Octopus_EKS_Use_Instance_Role,,} != "true" ]]; then
+  if [[ -z $Octopus_AccountType && -z $Octopus_K8S_Client_Cert && -z $Octopus_K8s_Pod_Service_Account_Token_Path && ${Octopus_EKS_Use_Instance_Role,,} != "true" ]]; then
     echo >&2 "Kubernetes account type or certificate is missing"
     exit 1
   fi
--- a/source/Calamari/Kubernetes/Scripts/KubectlBashContext.sh
+++ b/source/Calamari/Kubernetes/Scripts/KubectlBashContext.sh
@@ -131,7 +131,7 @@ function setup_context {
       K8S_Azure_Cluster+="-admin"
     fi
     kubectl config set-context $K8S_Azure_Cluster --namespace=$Octopus_K8S_Namespace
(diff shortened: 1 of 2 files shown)


source/Calamari/Kubernetes/Scripts/KubectlBashContext.sh
```diff
@@ -70,10 +72,21 @@ function setup_context {
...
   if [[ -z $Octopus_AccountType && -z $Octopus_K8S_Client_Cert && ${Octopus_EKS_Use_Instance_Role,,} != "true" ]]; then
-    echo >&2 "Kubernetes account type or certificate is missing"
-    exit 1
+    if [[ -z $Octopus_K8s_Pod_Service_Account_Token_Path && -z $Octopus_K8s_Server_Cert_Path ]]; then
+      echo >&2 "Kubernetes account type or certificate is missing"
+      exit 1
+    fi
+    
+    Octopus_K8s_Pod_Service_Account_Token=$(cat ${Octopus_K8s_Pod_Service_Account_Token_Path})
+    Octopus_K8s_Server_Cert_From_Path=$(cat ${Octopus_K8s_Server_Cert_Path})
```
mcasperson, line 83: The cert path technically doesn't have to be defined. It may be possible for the K8s API to have a trusted CA certificate for example. It would be worth checking if `Octopus_K8s_Server_Cert_Path` is empty before trying to cat it. I'm not sure what `cat` will do with an empty argument.

source/Calamari/Kubernetes/Scripts/KubectlBashContext.sh
```diff
@@ -110,83 +123,97 @@ function setup_context {
       K8S_Azure_Cluster+="-admin"
     fi
     kubectl config set-context $K8S_Azure_Cluster --namespace=$Octopus_K8S_Namespace
+  elif [[ $IsUsingPodServiceAccount ]]; then
+    kubectl config set-cluster octocluster --server=$Octopus_K8S_ClusterUrl
+    
+    if [[ -z $Octopus_K8s_Server_Cert_From_Path ]]; then
+      kubectl config set-cluster octocluster --insecure-skip-tls-verify=$Octopus_K8S_SkipTlsVerification
+    else
+      kubectl config set-cluster octocluster --certificate-authority=$Octopus_K8s_Server_Cert_Path
```
mcasperson, line 132: If the cert path is not defined, this call can be skipped.

mcasperson (changes requested):
The cert path is technically not required - the k8s cluster could use a trusted certificate, or the endpoint could be http (not https). The PR will have to be updated slightly to allow the cert path to be optional.

source/Calamari/Kubernetes/Scripts/KubectlBashContext.sh
ray-nham: > I'm not sure what cat will do with an empty argument.
it returns empty :) but I'll just add a check just to be safe.

source/Calamari/Kubernetes/Scripts/KubectlBashContext.sh
ray-nham: I think the condition of the `if` statement has covered this case, hasn't it? it will be skipped if `$Octopus_K8s_Server_Cert_From_Path` is empty (due to the path `$Octopus_K8s_Server_Cert_Path` is invalid or empty)

mcasperson: Ah, that is try. Please ignore this comment.

source/Calamari/Kubernetes/Scripts/KubectlPowershellContext.ps1
```diff
@@ -134,17 +137,28 @@ function ConnectAzAccount {
...
 	if([string]::IsNullOrEmpty($K8S_AccountType) -and [string]::IsNullOrEmpty($K8S_Client_Cert) -and $EKS_Use_Instance_Role -ine "true"){
-		Write-Error "Kubernetes account type or certificate is missing"
-		Exit 1
+	  if([string]::IsNullOrEmpty($Octopus_K8s_Pod_Service_Account_Token_Path) -and [string]::IsNullOrEmpty($Octopus_K8s_Server_Cert_Path)){
+	    Write-Error "Kubernetes account type or certificate is missing"
+    	Exit 1
+	  }
+		
+		$Octopus_K8s_Pod_Service_Account_Token = Get-Content -Path $Octopus_K8s_Pod_Service_Account_Token_Path
+		$Octopus_K8s_Server_Cert = Get-Content -Path $Octopus_K8s_Server_Cert_Path
```
mcasperson, line 151: If `$Octopus_K8s_Server_Cert_Path` is an empty string, I suspect `Get-Content` will display an error. A check around this call to verify `$Octopus_K8s_Server_Cert_Path` is not empty would be a good idea.

ray-nham:
> The cert path is technically not required - the k8s cluster could use a trusted certificate, or the endpoint could be http (not https). The PR will have to be updated slightly to allow the cert path to be optional.

Thanks Matt, good to know this. I have add the checks for empty file path before retrieving the file values. 👍

source/Calamari/Kubernetes/Scripts/KubectlBashContext.sh
```diff
@@ -110,83 +131,97 @@ function setup_context {
       K8S_Azure_Cluster+="-admin"
     fi
     kubectl config set-context $K8S_Azure_Cluster --namespace=$Octopus_K8S_Namespace
+  elif [[ $IsUsingPodServiceAccount ]]; then
```
ray-nham, line 134: caught this during testing. This check will return `true` as long as `IsUsingPodServiceAccount ` has a value, regardless of `true` or `false`

mcasperson: Nice catch

mcasperson (approved):
LGTM

PR #717, opened 2021-04-13
Support stack-level tags as part of CloudFormation deployment

egorpavlikhin:
Pass-through tags value from action Properties to AWS SDK.


andrewabest (approved):
We've discussed improving the instantiation of the DeployAwsCloudFormationConvention - aside from that,  it looks good!

PR #720, opened 2021-05-04
Restrict permissions on kubectl-octo.yml config file

rhysparry:
When creating the config file, we can start with it being readable/writable only by the file owner.

This will suppress warnings from tools that expect this file to have these permissions, fixing OctopusDeploy/Issues#6858.

We also initially make sure the file is empty if the file already exists, which mirrors the behaviour of the PowerShell script. As this is a temporarily created file, the file shouldn't exist, but this should clarify the intent.

### DIFF
--- a/source/Calamari/Kubernetes/Scripts/KubectlBashContext.sh
+++ b/source/Calamari/Kubernetes/Scripts/KubectlBashContext.sh
@@ -229,7 +229,8 @@ function setup_context {
 function configure_kubectl_path {
   export KUBECONFIG=$(get_octopusvariable "Octopus.Action.Kubernetes.KubectlConfig")
   echo "Temporary kubectl config set to $KUBECONFIG"
-  echo "" >> $KUBECONFIG
+  # create an empty file, to suppress kubectl errors about the file missing
+  echo "" > $KUBECONFIG
   chmod 600 $KUBECONFIG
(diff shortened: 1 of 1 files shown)


source/Calamari/Kubernetes/Scripts/KubectlBashContext.sh
```diff
@@ -229,7 +229,9 @@ function setup_context {
 function configure_kubectl_path {
   export KUBECONFIG=$(get_octopusvariable "Octopus.Action.Kubernetes.KubectlConfig")
   echo "Temporary kubectl config set to $KUBECONFIG"
-  echo "" >> $KUBECONFIG
+  # create an empty file, to suppress kubectl errors about the file missing
+  echo "" > $KUBECONFIG
+  chmod 600 $KUBECONFIG
```
hnrkndrssn, line 234: small _(personal)_ thing _(not necessarily need to be changed): I like to use the `chmod o=rw <file>` format which _might_ make it easier to figure out what permissions we're adding (at least for me as I struggle to remember all the numbers in permissions 😁)

hnrkndrssn (approved):
looks good👍

Just added a minor note on the format for adding permissions to the file, not a blocker just a suggestion so feel free to ignore :)

PR #722, opened 2021-05-13
NGINX output is now captured as verbose messages

mcasperson:
Nginx writes many diagnostic messages to the error stream, which Octopus will highlight in red. This makes them look like errors when they are actually expected output.

This PR captures the nginx output as verbose and redirects std err to std out.

# Before

image

# After

image

image


mcasperson:
[Remove Nginx serror logging.](https://trello.com/c/8iQRCbjN/53-remove-nginx-serror-logging)

ray-nham (approved):
LGTM :)

PR #723, opened 2021-05-13
Use '=' in argument because of Azure CLI bug

geofflamrock:
There is an issue with the option parsing in the Azure CLI (https://github.com/Azure/azure-cli/issues/12105) that causes arguments starting with a '-' to be parsed incorrectly, this is fixed by using the full argument with a '=' e.g. --password=mypassword


mcasperson:
[Azure K8s login fails when password starts with a dash.](https://trello.com/c/OWYpuNGw/62-azure-k8s-login-fails-when-password-starts-with-a-dash)

johnsimons:
@geofflamrock just wondering if in bash we also need to do the same as windows and escape double quotes? See https://github.com/OctopusDeploy/Calamari/blob/6ed3be7de2160437d3e833952010c8825e77e7ba/source/Calamari.Common/Features/Scripting/WindowsPowerShell/Bootstrap.ps1#L531-L542

johnsimons (approved):
LGTM
I just added a small comment regarding escaping of double-quotes in bash

geofflamrock:
> @geofflamrock just wondering if in bash we also need to do the same as windows and escape double quotes?

Thanks, that's a good thought, I'll do some testing in bash to see what the behaviour is.

geofflamrock:
I'm planning on doing some more smoke testing in a proper test instance once some access issues with Azure sandboxes are resolved, my testing so far has been limited to running the script locally in powershell and bash.

johnsimons (changes requested):
I thought we agreed that we do not need `source/Calamari.Azure/` ? And we are going to delete it?

geofflamrock:
> I thought we agreed that we do not need `source/Calamari.Azure/` ? And we are going to delete it?

@johnsimons I was planning to do in a separate PR, but happy to do in this PR if that is easier (which it prob is), I'll double check with the rest of the team to see if anyone else knows of a reason we can't remove and then will action.

geofflamrock:
@johnsimons the Calamari.Azure and Calamari.AzureAccounts folders have been removed, I've re-done manual testing of the Azure and Kubectl script steps and they seem ok now, can you think of anything else that would be worth testing here? 

Also, do you have any thoughts on whether these additional changes should warrant a minor version change? It would go to 19.3.0 as there has been a 19.2.0 in the meantime since this was started.

johnsimons:
> @johnsimons the Calamari.Azure and Calamari.AzureAccounts folders have been removed, I've re-done manual testing of the Azure and Kubectl script steps and they seem ok now, can you think of anything else that would be worth testing here?

Nope



> do you have any thoughts on whether these additional changes should warrant a minor version change?

Yeah, we tend to bump the patch version

johnsimons (approved):
LGTM

PR #725, opened 2021-05-20
Fixes kubernetes powershell azure auth to work with Powershell Core

geofflamrock:
This PR changes the Kubernetes Azure authentication so it can work with Powershell Core. This is the same set of changes as https://github.com/OctopusDeploy/Sashimi.AzureScripting/pull/9 but for the kubernetes context. It initializes the Az context instead of the AzureRM context when running on Powershell Core and adds a warning if AzureRM is present (but Az isn't), so that scripts can run in powershell core if the worker has AzureRM installed.

### DIFF
--- a/source/Calamari.Tests/Fixtures/RequiresPowerShell5Attribute.cs
+++ b/source/Calamari.Tests/Fixtures/RequiresPowerShell5Attribute.cs
@@ -1,0 +1,19 @@
+﻿using Calamari.Common.Plumbing.Extensions;
+using NUnit.Framework;
+using NUnit.Framework.Interfaces;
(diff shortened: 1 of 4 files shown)


geofflamrock:
There are some "run a kubectl script" tests that are currently ignored, I wonder if its worth trying to get these running in CI?

geofflamrock:
[Using Bundled Tools with Azure Steps does not work when using PowerShell Core](https://trello.com/c/DTMxmwiE/29-using-bundled-tools-with-azure-steps-does-not-work-when-using-powershell-core)

johnsimons (approved):

PR #729, opened 2021-06-02
Improves AWS CloudFormation logging on rollback

geofflamrock:
This PR improves AWS CloudFormation logging when there is an error. All stack events from the current stack run will be logged, with any that indicate error logged as warnings, and any others logged as verbose. Additional improvements have been made to the logging of the current stack state as well.


source/Calamari.Tests/AWS/CloudFormation/CloudFormationFixture.cs
```diff
@@ -0,0 +1,182 @@
...
+using FluentAssertions;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
+using NUnit.Framework;
+using Octopus.CoreUtilities.Extensions;
+using Octostache;
+
+namespace Calamari.Tests.AWS
+{
+    [TestFixture, Explicit]
```
geofflamrock, line 34: These were based off existing S3 tests that are marked as Explicit and aren't run in CI.

To run these in CI would involve changing to the AWS OctopusAPITester credentials instead of the specific Calamari ones here, plus would require changing the access of those credentials to allow for CloudFormation deployments. We already have e2e tests for CloudFormation over in server, I'm undecided on whether we should:

- Keep these as is as they are helpful for debug style testing
- Throw these away altogether
- Move the e2e CloudFormation tests over to here in some form

geofflamrock:
[CloudFormation: Improve error messages](https://trello.com/c/eaCBIk07/84-cloudformation-improve-error-messages)

source/Calamari.Aws/Deployment/Conventions/DeleteCloudFormationStackConvention.cs
```diff
@@ -52,6 +52,8 @@ private async Task InstallAsync(RunningDeployment deployment)
         {
             Guard.NotNull(deployment, "deployment can not be null");
 
+            var deploymentStartTime = DateTime.Now;
```
johnsimons, line 55: This api does not use UTC?

source/Calamari.Aws/Integration/CloudFormation/CloudFormationObjectExtensions.cs
```diff
@@ -141,6 +142,64 @@ public static Task<CreateChangeSetResponse> CreateChangeSetAsync(this Func<IAmaz
...
+                    }
+                }
+
+                return stackEventResults.OrderBy(s => s.SelectValueOr(e => e.Timestamp, DateTime.MinValue)).ToList();
+            }
+            catch (AmazonCloudFormationException ex) when (ex.ErrorCode == "AccessDenied")
+            {
+                throw new PermissionException(
+                    "The AWS account used to perform the operation does not have the required permissions to query the current state of the CloudFormation stack. " +
+                    "This step will complete without waiting for the stack to complete, and will not fail if the stack finishes in an error state.\n " +
```
johnsimons, line 187: Why `\n`?

source/Calamari.Aws/Integration/CloudFormation/CloudFormationObjectExtensions.cs
```diff
@@ -141,6 +142,64 @@ public static Task<CreateChangeSetResponse> CreateChangeSetAsync(this Func<IAmaz
...
+                    "Please ensure the current account has permission to perform action 'cloudformation:DescribeStackEvents'" +
+                    ex.Message);
+            }
+            catch (AmazonCloudFormationException ex) when (ex.ErrorCode == "ExpiredToken")
+            {
+                throw new PermissionException(
+                    "Security token has expired. Please increase the session duration and/or check that the system date and time are set correctly. " +
+                    ex.Message);
```
johnsimons, line 195: Are we going to repeat the same error message here? I was looking to see what the original error message is and wondering if it is good enough?

source/Calamari.Aws/Deployment/Conventions/DeployAwsCloudFormationConvention.cs
```diff
@@ -203,6 +204,8 @@ private Task DeleteCloudFormation(StackArn stack)
             CloudFormationTemplate template)
         {
             Guard.NotNull(deployment, "deployment can not be null");
+            var deploymentStartTime = DateTime.Now;
```
johnsimons, line 207: Same as before

source/Calamari.Aws/Integration/CloudFormation/CloudFormationObjectExtensions.cs
```diff
@@ -276,13 +335,13 @@ public static Task<DeleteStackResponse> DeleteStackAsync(this Func<IAmazonCloudF
...
             {
-                var response =  await clientFactory().CreateStackAsync(request);
+                var response = await clientFactory().CreateStackAsync(request);
                 return response.StackId;
             }
             catch (AmazonCloudFormationException ex) when (ex.ErrorCode == "AccessDenied")
             {
                 throw new PermissionException(
-                    "The AWS account used to perform the operation does not have the required permissions to create the stack.\n"+
+                    "The AWS account used to perform the operation does not have the required permissions to create the stack.\n" +
```
johnsimons, line 344: Why `\n`? Can't we just let the UI deal with it?

source/Calamari.Tests/AWS/CloudFormation/CloudFormationFixture.cs
```diff
@@ -0,0 +1,197 @@
+﻿using Calamari.Common.Features.Packages;
+using Calamari.Common.Features.Processes;
+using Calamari.Common.Plumbing.FileSystem;
+using Calamari.Common.Plumbing.Variables;
+using Amazon.CloudFormation;
+using Amazon.CloudFormation.Model;
+using System.Threading.Tasks;
+using Calamari.Common.Plumbing.Logging;
+#if AWS
```
johnsimons, line 9: Should this surround the whole file?

source/Calamari.Tests/AWS/CloudFormation/CloudFormationFixture.cs
johnsimons: I would say remove the E2E tests in Server and test exclusively in Calamari

johnsimons (approved):
Added a few comments. I found the usage of the `Maybe` API really confusing and hard to reason. (Existing issue).
We should enable these tests to run all the time. And move the existing E2E test to here.
There is a lot of assumptions in this codebase regarding error codes and special strings. Makes me wonder how long it will work for!
I think it is worth asking @shaunmarx to review this as well.

source/Calamari.Aws/Deployment/Conventions/DeleteCloudFormationStackConvention.cs
geofflamrock: The filtering is done locally after obtaining the stack events from CloudFormation, and the timestamps that come back are in our local time. Do you think it would be good to do these comparisons using UTC?

source/Calamari.Aws/Integration/CloudFormation/CloudFormationObjectExtensions.cs
geofflamrock: No particular reason, the code was taken from some existing code that interacts with CloudFormation API. Looks like quite a few error messages and exception handling messages are deliberately including \n maybe so that the output in the logs is a bit easier to follow as it can be quite verbose?

source/Calamari.Aws/Deployment/Conventions/DeleteCloudFormationStackConvention.cs
johnsimons: I didn't realise the filter is dome client-side, why is that?

geofflamrock: Unfortunately the CloudFormation API for getting stack events only has parameters for stack name and a NextToken which appears to be used for paging, it doesn't allow any form of filtering on properties. I should probably check on the requirement around that NextToken to ensure we aren't forced to use it in this context if there are lots of events as we want to get all the relevant stack events for a run whereas previously we were only getting the latest, I'll check on this.

source/Calamari.Aws/Integration/CloudFormation/CloudFormationObjectExtensions.cs
geofflamrock: I'll check on the error message that comes from CloudFormation and see if it is good enough.

source/Calamari.Aws/Deployment/Conventions/DeleteCloudFormationStackConvention.cs
geofflamrock: Hmmm yeah looks like if the response size is > 1Mb it will send a NextToken and we'll need to request more events, will need to add support for that

johnsimons: So there is no way to filter these events server side?

geofflamrock: No, the CloudFormation API for getting stack events only allows StackName and NextToken as parameters to the request unfortunately

source/Calamari.Aws/Integration/CloudFormation/CloudFormationObjectExtensions.cs
geofflamrock: This handling was added as part of https://github.com/OctopusDeploy/OctopusDeploy/pull/6187 (and associated Calamari PRs) where users were getting timeouts with non-helpful error messages because IAM role assumption was expiring. Based on the screenshots from the linked PR the message that comes in from the CloudFormation exception looks to be "The security token included in the request is expired", so it is somewhat duplicated with the first part of the custom message here, the second part gives some guidance on what to do about it. Do you think we need/want to make any changes here?

source/Calamari.Tests/AWS/CloudFormation/CloudFormationFixture.cs
geofflamrock: To get these running in CI I've changed over to using the OctopusAPITester user, but I'll need to add some additional permissions to allow it to perform CloudFormation and S3 actions, do you foresee any issues with adding these @johnsimons ? Do I need to alert anyone else that I'm doing this?

source/Calamari.Tests/AWS/CloudFormation/CloudFormationFixture.cs
geofflamrock: Fixed

source/Calamari.Tests/AWS/CloudFormation/CloudFormationFixture.cs
johnsimons: I don't think that is a problem

source/Calamari.Aws/Integration/CloudFormation/CloudFormationObjectExtensions.cs
johnsimons: I think this is something we should clean up. I really don't see our job formatting errors at this level.

source/Calamari.Aws/Integration/CloudFormation/CloudFormationObjectExtensions.cs
johnsimons: Just remove "Security token has expired." ?

geofflamrock: We could probably put the message from AWS at the front, it would then read "The security token included in the request is expired. Please increase the session..." which seems to work pretty well. I wonder if this should also use the logging guidelines format and link to a known error code with docs, kind of like the "AWS-CLOUDFORMATION-ERROR-0001" error code.

source/Calamari.Aws/Integration/CloudFormation/CloudFormationObjectExtensions.cs
geofflamrock: I'll remove

source/Calamari.Aws/Integration/CloudFormation/CloudFormationObjectExtensions.cs
geofflamrock: This is done

source/Calamari.Tests/AWS/CloudFormation/CloudFormationFixture.cs
geofflamrock: These are now running successfully in CI :)

johnsimons:
@geofflamrock I noticed `#if AWS` is only set for netfull, any ideas why?

geofflamrock:
> @geofflamrock I noticed `#if AWS` is only set for netfull, any ideas why?

Hmmm yeah I'm not sure, it looks to have been added in https://github.com/OctopusDeploy/Calamari/pull/270 might have just been an oversight perhaps. I'll add it and check it all works ok, I don't think there is any reason we can't run these tests against .net 3.1.

geofflamrock:
Oh, looks like the S3 tests don't compile under netcore due to a difference in the AWSSDK, maybe that was a reason, weird. I'll make them work.

geofflamrock:
I've enabled the AWS tests to run in .Net Core 3.1 and fixed them up so they should run in CI now as well.

source/Calamari.Tests/AWS/S3Fixture.cs
```diff
@@ -49,7 +51,7 @@ private static JsonSerializerSettings GetEnrichedSerializerSettings()
 
         [Test]
         [Category(TestCategory.CompatibleOS.OnlyWindows)]
```
geofflamrock, line 53: @johnsimons do you have any knowledge on why these were marked to run on Windows only?

johnsimons: Nope, I would expect this to work on all plats

geofflamrock: Tests are now running successfully on all plats in CI

geofflamrock:
I think everything is done here except moving the e2e tests from server into here, I'm thinking that I'll tackle that in a follow up PR, is there anything else you think we would like to do here @johnsimons?

johnsimons:
> is there anything else you think we would like to do here

not right now. awesome work

PR #731, opened 2021-06-08
Fixes issues with YAML structured replacement and inline comments

geofflamrock:

Also adds a couple of more block comments to the test case to increase coverage.


geofflamrock:

geofflamrock:
Closing in favour of https://github.com/OctopusDeploy/Calamari/pull/732 which will then be merged through release branches to backport fix

PR #732, opened 2021-06-08
Fixes issues with YAML structure configuration replacement and inline comments

geofflamrock:

Also adds a couple more block comments to the test case to increase coverage.


geofflamrock:

johnsimons (approved):
The tests seem to give us the assurances the code is behaving correctly

PR #734, opened 2021-06-10
Fixes issues with YAML structure configuration replacement and inline comments

geofflamrock:

Also adds a couple more block comments to the test case to increase coverage.


geofflamrock:

geofflamrock:
I'm going to close this as will make a merge PR from release/2021.1 instead

PR #741, opened 2021-06-22
Adds a local build option to improve local dev experience

geofflamrock:
This PR adds some options to the Calamari build designed to help with the local dev experience, particularly the inner feedback loop of making changes in Calamari and then wanting to get them across to Server to test them locally.

Specifically this does:
- Adds a new script "build-local.ps1" which sets the various options
- Adds a new build target called Local that doesn't pack the *.Tests projects as these take quite a large chunk of time and aren't used locally
- Adds a timestamp to the back of the NuGet version for artifacts so that changes don't have to be committed locally in order to get a newer package version
- Adds an option to specify msbuild verbosity and sets the default for local builds to Minimal to reduce noise and improve performance. Default build target leaves this as Normal to match previous hardcoded setting.
- Creates NuGet packages in parallel to improve performance
- Sets the CalamariVersion property in the OctopusServer.csproj file if it exists

All up these changes took my local build time from ~5 minutes to <1 minute so hopefully they will be helpful.

There should be no changes to the normal build as all new options default to off or whatever they previously were.


source/Calamari.Tests/Fixtures/ConfigurationTransforms/TransformFileLocatorFixture.cs
```diff
@@ -352,7 +352,7 @@ public void When_TransformIsWildcardFileNameOnly_And_TargetIsFileNameOnly_2_ItSu
         }
 
         [Test]
-        public void When_TransformIsWildcardFileNameOnly_And_TargetIsWildCardInMiddleOfFileNameOnly_ItFails()
+        public void When_TransformIsWildcardFileNameOnly_And_TargetIsWildCardInsideFileNameOnly_ItFails()
```
geofflamrock, line 355: I renamed these as they were causing issues with too long path names for me locally during clean of tests, my path is `C:\src\OctopusDeploy\Calamari\...` so I don't think its unreasonably long, not sure if others have come across this.

johnsimons (approved):
Seems like a good addition

PR #743, opened 2021-06-25
Unset impersonate_service_account when not in use

ray-nham:
When running K8s deployment target health check using Octopus worker, I notice that sometimes the `auth/impersonate_service_account` value is not reset.
When we select an account to impersonate we set the impersonate email to `auth/impersonate_service_account`. This value maybe hanged around and may affect other processes which don't want to impersonate another account.

This PR is to ensure when `ImpersonateServiceAccount` is set to `false`, we will explicitly unset the `auth/impersonate_service_account` value.


andrewabest (approved):
The changes look good - as we discussed, using the environment variable is preferable as it allows us not to mess with global state on the worker. Ship it!

PR #746, opened 2021-07-02
Added: zone or region args when getting k8s credentials

ray-nham:
This PR is to add `zone` or `region` args when getting k8s credentials to avoid the edge case where the users might have `gcloud config` conflicts with the values defined in K8s script.
Octopus Server PR: https://github.com/OctopusDeploy/OctopusDeploy/pull/9394


mcasperson (approved):
LGTM

johnsimons (changes requested):
I think it is worth covering this case with more tests @ray-nham

ray-nham:
> I think it is worth covering this case with more tests @ray-nham

Good idea, John. Added unit tests for these scenarios:
- Zone is provided
- Region is provided
- Both Zone and Region are provided
- Either Zone or Region is provided

source/Calamari.Tests/KubernetesFixtures/KubernetesRunsScriptFixture.cs
```diff
@@ -245,6 +245,51 @@ public void ExecutionWithGoogleCloudAccount()
...
+            variables.Set($"{account}.JsonKey", Convert.ToBase64String(Encoding.UTF8.GetBytes(jsonKey)));
+            variables.Set("Octopus.Action.GoogleCloud.Project", "gke-project");
+            variables.Set("Octopus.Action.GoogleCloud.Region", "gke-region");
+            variables.Set("Octopus.Action.GoogleCloud.Zone", "gke-zone");
+            var wrapper = CreateWrapper();
+            TestScriptInReadOnlyMode(wrapper).AssertSuccess();
+        }
+        
+        [Test]
+        public void ExecutionWithGoogleCloudAccount_WhenEitherZoneAndRegionIsProvided()
```
johnsimons, line 281:
```suggestion
        public void ExecutionWithGoogleCloudAccount_WhenNeitherZoneOrRegionAreProvided()
```

johnsimons (approved):
LG @ray-nham, I've just added a small rename to one of your test names

PR #748, opened 2021-07-13
Use AzureSignTool for code signing

matt-richardson:
So that the cert is loaded from azure key vault rather than disk

outstanding tasks:
* [x] update teamcity to pass in new params (can we read from env?)
* [x] check we are still timestamping

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -4,6 +4,9 @@
 #tool "nuget:?package=GitVersion.CommandLine&version=5.2.0"
 #addin "nuget:?package=Cake.Incubator&version=5.0.1"
 #addin "nuget:?package=Cake.FileHelpers&version=4.0.1"
+// see https://www.gep13.co.uk/blog/introducing-cake.dotnettool.module
+#module nuget:?package=Cake.DotNetTool.Module&version=0.1.0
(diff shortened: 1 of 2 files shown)


build.cake
```diff
@@ -354,12 +371,43 @@ private bool HasAuthenticodeSignature(FilePath filePath)
...
+  {
+    signArguments
+      .Append("--description").AppendQuoted(display)
+      .Append("--description-url").AppendQuoted(displayUrl);
+  }
+
+  foreach (var file in files)
+    signArguments.AppendQuoted(file.FullPath);
+
+    var azureSignToolPath = MakeAbsolute(File("./tools/Cake/azuresigntool.exe"));
```
corey-underdown, line 394: Interestingly the path is `./tools/Cake/azuresigntool.exe` in octopus server but `./tools/azuresigntool.exe` in tentacle?

corey-underdown (approved):

build.cake
matt-richardson: Good catch... Eagle-Eye-Corey.
Looking locally... it should be `tools/azuresigntool.exe`
```suggestion
    var azureSignToolPath = MakeAbsolute(File("./tools/azuresigntool.exe"));
```

PR #752, opened 2021-07-21
Authenticated GitHub package download

tekerson:
# Problem

Resolves OctopusDeploy/Issues#6922

The GitHub API is returning a 404 for private repositories when the download is done from the target as opposed to the server. After some investigation the root cause appears to be that the `WebClient.Credentials` property only causes the `Authorization` header to be added when a 401 response is received but the GitHub API [returns a 404 for unauthenticated requests](https://docs.github.com/en/rest/overview/troubleshooting#404-error-for-an-existing-repository). So, the credentials are never used.

# Solution

The solution here involves explicitly setting the `Authorization` header instead of relying on the Credentials behaviour.

### DIFF
--- a/source/Calamari.Shared/Integration/Packages/Download/GitHubPackageDownloader.cs
+++ b/source/Calamari.Shared/Integration/Packages/Download/GitHubPackageDownloader.cs
@@ -37,15 +37,15 @@ namespace Calamari.Integration.Packages.Download
             this.freeSpaceChecker = freeSpaceChecker;
         }
 
-        public PackagePhysicalFileMetadata DownloadPackage(
-            string packageId,
-            IVersion version,
-            string feedId,
-            Uri feedUri,
-            ICredentials feedCredentials,
-            bool forcePackageDownload,
-            int maxDownloadAttempts,
-            TimeSpan downloadAttemptBackoff)
+        public PackagePhysicalFileMetadata DownloadPackage(string packageId,
+                                                           IVersion version,
+                                                           string feedId,
+                                                           Uri feedUri,
+                                                           string? feedUsername,
+                                                           string? feedPassword,
+                                                           bool forcePackageDownload,
+                                                           int maxDownloadAttempts,
+                                                           TimeSpan downloadAttemptBackoff)
         {
             var cacheDirectory = PackageDownloaderUtils.GetPackageRoot(feedId);
             fileSystem.EnsureDirectoryExists(cacheDirectory);
@@ -63,7 +63,7 @@ namespace Calamari.Integration.Packages.Download
             return DownloadPackage(packageId,
                                    version,
                                    feedUri,
-                                   feedCredentials,
+                                   feedPassword,
                                    cacheDirectory,
                                    maxDownloadAttempts,
                                    downloadAttemptBackoff);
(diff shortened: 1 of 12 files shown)


source/Calamari.Shared/Integration/Packages/Download/GitHubPackageDownloader.cs
```diff
@@ -238,6 +238,14 @@ JToken PerformRequest(ICredentials feedCredentials, string uri)
             }
         }
 
+        void AddAuthorization(WebClient client, string? authorizationToken)
+        {
+            if (!String.IsNullOrWhiteSpace(authorizationToken))
+            {
+                client.Headers.Set(HttpRequestHeader.Authorization, String.Concat("token ", authorizationToken));
```
tekerson, line 245: This is the actual fix - explicitly setting the `Authorization` header to use the OAuth/Personal Access token

source/Calamari/Commands/DownloadPackageCommand.cs
```diff
@@ -133,16 +134,6 @@ public override int Execute(string[] commandLineArguments)
             return 0;
         }
 
-        static ICredentials GetFeedCredentials(string feedUsername, string feedPassword)
```
tekerson, line 136: The requirements for authentication are different for each downloader implementation (e.g. GitHub doesn't require the username) so the checking and marshalling of the provided values into a valid credential is delegated to each implementation.

source/Calamari/Commands/DownloadPackageCommand.cs
```diff
@@ -62,10 +62,10 @@ public class DownloadPackageCommand : Command
                     }
                     versionFormat = format;
                 });
-            Options.Add("feedId=", "Id of the NuGet feed", v => feedId = v);
```
tekerson, line 65: This change is purely aesthetic - the CLI interface is not specific (any more) to NuGet feeds.

source/Calamari.Shared/Integration/Packages/Download/NuGetPackageDownloader.cs
```diff
@@ -140,5 +140,15 @@ void CheckWhetherThePackageHasDependencies(PackagePhysicalFileMetadata pkg)
                     string.Join(", ", dependencies.Select(dependency => $"{dependency.Id}")),
                     WhyAmINotAllowedToUseDependencies);
         }
+        
+        static ICredentials GetFeedCredentials(string? feedUsername, string? feedPassword)
```
johnsimons, line 144: This seems to be repeated a few times, wondering if it is worth sharing?

johnsimons (approved):
LGTM, I assume there will be a Server PR to go with it?

source/Calamari.Shared/Integration/Packages/Download/NuGetPackageDownloader.cs
tekerson: I did consider that, but I don't know if the behaviour **should** actually be the same or if it is only currently the same **because** it was shared (my guess is the latter).
I'm also not sure where the shared version would make sense? the `PackageDownloaderUtils`? But again, that seems to be too general when it doesn't apply to app downloaders and the `*Utils` could end up as a grab bag of dissociated "helpers".

PR #753, opened 2021-07-23
Code sign nuget packages for Calamari.Common and Calamari.CloudAccounts

mattocto:
Sashimi packages have a dependency on some Calamari nuget packages that contained unsigned binaries. This could potentially lead to antivirus flagging these files and causing deployment failures.

This PR updates the way we pack these binaries to ensure that they are signed first before being included in a nuget package. In order to do this, we first publish the files, then sign them, before finally packing them with `NuGetPack`. `NuGetPack` requires a nuspec file as an input, and so I've created nuspect files for Calamari.Common and Calamari.CloudAccounts accordingly.

We also update the associated build scripts with a new flag to allow us to turn on or off the signing of binaries when building locally. Due to other changes occurring while this was in flight, the local signing of binaries won't actually work yet, as this now references Key Vault. However the default is not to sign the binaries locally, and this would only ever be important if someone was trying to test the local signing. If that becomes necessary again in future then we can fix the local signing issues at that point.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -157,6 +157,7 @@ Task("PackBinaries")
     // make nuget packages out of Common and CloudAccounts so that they can be used by Sashimi packages
     actions.Add(() => DoPackage("Calamari.Common", "net40", nugetVersion));
     actions.Add(() => DoPackage("Calamari.Common", "net452", nugetVersion));
+    actions.Add(() => DoPackage("Calamari.Common", "netstandard2.1", nugetVersion));
     actions.Add(() => DoPackage("Calamari.CloudAccounts", "net452", nugetVersion));
 
     // Create a portable .NET Core package
--- a/build.cake
+++ b/build.cake
@@ -57,7 +57,6 @@ Setup(context =>
     if (appendTimestamp) 
         nugetVersion = nugetVersion + "-" + DateTime.Now.ToString("yyyyMMddHHmmss");
 
-    Information($"Started with SignFiles = {signFiles}");
     Information("Building Calamari v{0}", nugetVersion);
 });
 
@@ -273,7 +272,6 @@ private string DoPublish(string project, string framework, string version, strin
 
 private void DoPackage(string project, string framework, string version, string runtimeId = null)
 {
-
     var publishedTo = Path.Combine(publishDir, project, framework);
     var projectDir = Path.Combine("./source", project);
     var packageId = $"{project}";
--- a/build.cake
+++ b/build.cake
@@ -167,7 +167,9 @@ Task("PackBinaries")
     actions.Add(() => DoPackage("Calamari.Common", "net40", nugetVersion));
     actions.Add(() => DoPackage("Calamari.Common", "net452", nugetVersion));
     actions.Add(() => DoPackage("Calamari.Common", "netstandard2.1", nugetVersion));
-    actions.Add(() => DoPackage("Calamari.CloudAccounts", "net452", nugetVersion));
+    actions.Add(() => DoPackage("Calamari.CloudAccounts", "net452", nugetVersion));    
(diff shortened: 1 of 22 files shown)


build.cake
```diff
@@ -159,8 +160,13 @@ Task("PackBinaries")
 {
     var actions = new List<Action>();
 
-    actions.Add(() => DoPackage("Calamari", "net40", nugetVersion));
+    actions.Add(() => DoPackage("Calamari", "net40", nugetVersion));    
     actions.Add(() => DoPackage("Calamari", "net452", nugetVersion, "Cloud"));
+    
+    // make nuget packages out of Common and CloudAccounts so that they can be used by Sashimi packages
+    actions.Add(() => DoPackage("Calamari.Common", "net40", nugetVersion));
```
johnsimons, line 167: Do we need to do the same for `netcoreapp3.1`?

source/Calamari.CloudAccounts/Calamari.CloudAccounts.nuspec
```diff
@@ -0,0 +1,15 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
andrewabest, line 1: Why have we added nuspec files for these?

andrewabest: I suspect i may have something to do with

mattocto: Previously we had been using `DotNetCorePack` but that only allows you to specify a project, and not a published set of files. `NuGetPack` is the method that is used for the other packages that were already successfully signing the binaries so I thought for consistency I'd use the same here. And the `NuGetPack` method requires a nuspec file. 

The process seems to be publish -> sign -> pack, and `NuGetPack `was the only way I found that supported that - it could be that I'm missing something though, I found the Cake documentation to be pretty sparse and hard to navigate.

build.cake
mattocto: That's a good question. I haven't checked every Sashimi yet, only Terraform. I'll confirm, and update to include `netcoreapp3.1` if necessary.

johnsimons: no need, i know we need it 😉

mattocto: Righto! I'll add that in.

source/Calamari.CloudAccounts/Calamari.CloudAccounts.nuspec
mattocto: > Perhaps elaborating a bit more in the PR description might help clarify the changes made?

Missed this comment first time around. I've now updated the PR description to include the same details as in my comment above.

build.cake
mattocto: It's the netstandard2.1 build target that I've included in this build. (This is what the project was already targeting). 
.NET Core 3.1 supports .NET Standard 2.1 so this should be what you're after. Let me know if I've misunderstood!

andrewabest (approved):
Looks good to go!

build.cake
```diff
@@ -159,8 +160,14 @@ Task("PackBinaries")
...
+    actions.Add(() => DoPackage("Calamari.Common", "net452", nugetVersion));
+    actions.Add(() => DoPackage("Calamari.Common", "netstandard2.1", nugetVersion));
+    actions.Add(() => DoPackage("Calamari.CloudAccounts", "net452", nugetVersion));
```
johnsimons, line 170: Doesn't `Calamari.CloudAccounts` also need `netstandard2.1`?

johnsimons: Pretty sure this packages is referenced in Sashimi

mattocto: You're right @johnsimons. Sorry, I specifically went and checked this too, but it was last thing on a Friday afternoon so I must have overlooked something.

andrewabest: Good catch - yes it would.

PR #754, opened 2021-07-27
Fixes proxy handling for netcore builds of Calamari

andrewabest:
# Overview 

This PR addresses two bugs that have been identified in Calamari's proxy detection and handling code.

## Bug 1: `win-x64` Calamari does not support user default proxy

In Octopus 2021.1, a [change was made](https://github.com/OctopusDeploy/OctopusDeploy/pull/7884) that changed the way we ship Calamari to Windows targets. Previously we would ship the `netfx` version of Calamari to ALL Windows targets. From 2021.1 we have a more sophisticated way of determining whether we should send the `netfx` binary, or the `x64` netcore binary.

Calamari's proxy handling [was renovated](https://github.com/OctopusDeploy/Calamari/pull/420) mid-2019. Part of that renovation ([here](ecc9474640526e72606a26d05b3264bfbe7dccaf89680f5ee1427ff6903ba36cR13))  dealt with a limitation in dotnet core at the time - you couldn't retrieve the default user proxy if one was configured. If the `x64` version of Calamari was used, the user's default proxy would not be picked up and configured, and we would instead log a verbose message saying that no proxy was configured for that execution.

A subsequent [patch to dotnet core](https://github.com/dotnet/corefx/pull/41692) has addressed this limitation, and means you can retrieve the default system proxy in dotnet core in the same way you would from netfx.

### Fix

This PR removes the conditional compilation previously required due to the limitation in dotnet core, and ensures the default user proxy is utilised regardless of whether we are in the `netfx` binary or the `x64` binary.

## Bug 2: Calamari does not support `HTTP_PROXY` environment variable with hostname or IP address

In our Powershell Bootstrapper, there is a baked-in assumption that if the default `HTTP_PROXY` variable exists, it is a valid Uri that can be passed to the dot net `Uri` type constructor.

The value of `HTTP_PROXY` may be a hostname or IP address, optionally followed by a colon and port number, or it may be an http URL, optionally including a username and password for proxy authentication. [Source](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.defaultproxy?view=netcore-3.1).

When the value was a hostname or IP address, our bootstrapper would fail to construct a valid Uri and blow up.

### Fix

This PR makes the bootstrapper more flexible and will test the value to see if it is a valid URI, and if it is not, will instead treat it as a raw string.

It does NOT attempt to further validate the contents of the string.

## Reproducing / Testing

If you wish to reproduce or test this behaviour, on a Windows 10 x64 system:

- Install and configure a Tentacle on the machine, which will by default run under `Local System`
- Stop the tentacle service, and run it via the command line instead with `C:\Program Files\Octopus Deploy\Tentacle\Tentacle.exe" run --instance="YourTentacleInstanceName"` so that it is running in the context of you as the currently logged in user
- using [ccProxy](https://www.youngzsoft.net/ccproxy/proxy-server-download.htm) or your favourite Windows-supporting proxy, configure your default user proxy in Proxy Settings
image
- Create a deployment process in Octopus with a single script step with the following contents
```
Write-Warning "======"
Invoke-WebRequest -Uri "http://ident.me" -UseBasicParsing
Write-Warning "======"
```
- Run the script step with the proxy on - the step will pass
- Run the script step with the proxy off - the step will still pass, indicating it has not used the proxy to issue the request

### DIFF
--- a/source/Calamari.Common/Features/Scripting/WindowsPowerShell/Bootstrap.ps1
+++ b/source/Calamari.Common/Features/Scripting/WindowsPowerShell/Bootstrap.ps1
@@ -386,13 +386,18 @@ function Initialize-ProxySettings() {
 	if ($useDefaultProxy -and [string]::IsNullOrEmpty($proxyHost)) {
 		# Calamari ensure both http_proxy and HTTP_PROXY are set, so we don't need to worry about casing
 		if (![string]::IsNullOrEmpty($env:HTTP_PROXY)) {
-			$octopusProxyUri = New-Object System.Uri($env:HTTP_PROXY)
+
+			$octopusProxyUri = $null;
+			$isUri = [System.Uri]::TryCreate($env:HTTP_PROXY, [System.UriKind]::Absolute, [ref] $octopusProxyUri)
+			if (!$isUri) {
+				$octopusProxyUri = $env:HTTP_PROXY;
+			}
 
 			# The HTTP_PROXY env variable may also contain credentials.
 			# This is a common enough pattern, but we need to extract the credentials in order to use them
 
 			# But if credentials were explicitly provided, use those ones instead
-			if (-not $hasCredentials) {
+			if (-not $hasCredentials -and $octopusProxyUri.GetType() -eq [System.Uri]) {
 				$credentialsArray = $octopusProxyUri.UserInfo.Split(":")
 				$hasCredentials = $credentialsArray.length -gt 1;
 				if ($hasCredentials) {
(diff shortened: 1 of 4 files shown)


TomPeters (approved):
These changes look appropriate to me :+1:

I have not manually tested any of this, and have only looked at the code.

tekerson (approved):

andrewabest:
I think we might try and address the feedback on `master` once we've shipped 2021.1 and .2.

PR #756, opened 2021-08-05
Named packages support

ray-nham:
This PR added:
- A temporary change in `ExecuteManifestCommand` to enable package extraction for named packages.
- A temporary change in `NodeExecutor` to allow variables with square brackets to be evaluated by `Octostache`.

### DIFF
--- a/source/Calamari.Tests/Fixtures/Manifest/ExecuteManifestCommandFixture.cs
+++ b/source/Calamari.Tests/Fixtures/Manifest/ExecuteManifestCommandFixture.cs
@@ -81,6 +81,48 @@ namespace Calamari.Tests.Fixtures.Manifest
                 result.AssertOutput(nameof(NodeInstructions.DeploymentTargetInputsVariable));
             }
         }
+        
+        [Test]
+        public void WithNamedPackageInstructions()
+        {
+            var instructions =
+                InstructionBuilder
+                    .Create()
+                    .WithCalamariInstruction("test-calamari-instruction")
+                    .WithNodeInstruction()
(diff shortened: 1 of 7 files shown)


source/Calamari.Tests/Fixtures/Manifest/ExecuteManifestCommandFixture.cs
```diff
@@ -81,6 +81,48 @@ public void WithInstructions()
...
+                };
+
+                var result = ExecuteCommand(variables, "Calamari.Tests");
+
+                result.AssertSuccess();
+                result.AssertOutput("Hello from TestCommand");
+                result.AssertOutput("Hello from my custom node!");
+                result.AssertOutput(string.Join(Path.Combine("BootstrapperPathVariable_Value", "bootstrapper.js"),
+                                                Path.Combine("TargetPathVariable_Value", "executor.js")));
+                result.AssertOutput(nameof(NodeInstructions.DeploymentTargetInputsVariable));
```
egorpavlikhin, line 123: Let's change this to check that values are escaped correctly and see if we can reduce the test size, it feels that some parts of it are irrelevant to what we are trying to test.

ray-nham: I did a small refactoring: moving the method to escape json inputs to Calamari.Util and add a separate unit test for that method.

egorpavlikhin (approved):
Looks good, provided we refactor it next week to support the full range of Octostache variable expressions. 👍

PR #757, opened 2021-08-06
Revert deletion of `Calamari.CloudAccounts.AzureAccountVariables`

tekerson:
The `Calamari.CloudAccounts.AzureAccountVariables` class was cleaned up as part of fixing OctopusDeploy/Calamari#723 but it is still [being used in `Calamari.Terraform@10.x`](https://github.com/OctopusDeploy/Sashimi.Terraform/blob/10.2.2/source/Calamari/Behaviours/TerraformDeployBehaviour.cs#L135-L141). This PR reverts the deletion in lieu of a more principled solution.


geofflamrock (approved):
LGTM 👍

PR #758, opened 2021-08-09
#6965 Updated to latest version of NuGet.Commands.

Sellorio:
This PR is intended to resolve this issue: https://github.com/OctopusDeploy/Issues/issues/6965.

Please review this knowing that only a basic test was done in code in which the issue was reproduced, changes were made and then the issue was re-tested to ensure it was resolved. Therefore, no flow-on effects or regressions in other issues have been explored.

### DIFF
--- a/source/Calamari.Shared/Calamari.Shared.csproj
+++ b/source/Calamari.Shared/Calamari.Shared.csproj
@@ -64,7 +64,7 @@
     <PackageReference Include="Microsoft.Web.Xdt" Version="3.1.0" />
   </ItemGroup>
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.1'">
-    <PackageReference Include="NuGet.Commands" Version="3.5.0" />
+    <PackageReference Include="NuGet.Commands" Version="5.10.0" />
     <PackageReference Include="Markdown" Version="2.1.0" />
     <PackageReference Include="System.Diagnostics.Tracing" Version="4.3.0" />
     <PackageReference Include="System.Threading.AccessControl" Version="4.3.0" />
--- a/source/Calamari.Shared/Integration/Packages/NuGet/NuGetV3LibDownloader.cs
+++ b/source/Calamari.Shared/Integration/Packages/NuGet/NuGetV3LibDownloader.cs
@@ -8,57 +8,106 @@ using NuGet.Common;
 using NuGet.Protocol;
 using NuGet.Protocol.Core.Types;
 using System.IO;
-using Calamari.Common.Plumbing.Logging;
 using Calamari.Common.Plumbing.Extensions;
-using NuGet.DependencyResolver;
-using NuGet.LibraryModel;
 using Octopus.Versioning;
+using System.Threading.Tasks;
+using NuGet.Commands;
+using PackageIdentity = NuGet.Packaging.Core.PackageIdentity;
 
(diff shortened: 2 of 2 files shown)


source/Calamari.Tests/Fixtures/PackageDownload/PackageDownloadFixture.cs
```diff
@@ -75,16 +74,71 @@ public void TearDown()
...
+        {
+            UnsuccessfulPackageDownloadTest("artifactory-nuget", "Artifactory.Test.NuGet", packageVersion, ArtifactoryUri);
+        }
+
+        [TestCase("1.0.1", "6ff3395878566eba2b8bef307bd8845be5cb43ff")]
+        [TestCase("1.0.2", "d5ed35df167a8d635f43c890bcba6dd124ef77fb")] // actual package is 1.0.02
+        [TestCase("1.0.02", "d5ed35df167a8d635f43c890bcba6dd124ef77fb")]
+        [TestCase("1.0.0-alpha.2", "eb0ffcc9f614e83d3a69f0963b3da9a6460e75df")]
+        [TestCase("1.0.0-alpha.3", "d148ab7b4caa577b3f4b84960fef13a649182353")]
+        [TestCase("1.0.0-alpha.3+metadata", "d148ab7b4caa577b3f4b84960fef13a649182353")]
```
geofflamrock, line 128: The fact that these package version scenarios work for feedz.io but don't for artifactory even with the V2 libs (presumably running under netframework) suggest to me that we probably need to do a bit more looking into why this is happening rather than just applying a tactical fix of upgrading the NuGet lib for netcore only as we haven't really solved the underlying problem.

source/Calamari.Tests/Fixtures/PackageDownload/PackageDownloadFixture.cs
```diff
@@ -75,16 +74,71 @@ public void TearDown()
...
+        [TestCase("1.0.1", "8950c54441b67dcd8ae54b4dbb0fe3735be85a9b")]
+        [TestCase("1.0.2", "3967afe4c54e5f0626677d2f11606d491a453bca")] // actual package is 1.0.02
+        [TestCase("1.0.02", "3967afe4c54e5f0626677d2f11606d491a453bca")]
+        [TestCase("1.0.0-alpha.2", "991a595d780593acf75b891c8b24edbfaa3505f2")]
+        [TestCase("1.0.0-alpha.3", "bd22572257d5ed956b0dd1716e290fba2b4f407c")]
+        [TestCase("1.0.0-alpha.3+metadata", "bd22572257d5ed956b0dd1716e290fba2b4f407c")]
+#endif
+        [RequiresMonoVersion480OrAboveForTls12]
+        [RequiresMinimumMonoVersion(5, 12, 0, Description = "HttpClient 4.3.2 broken on Mono - https://xamarin.github.io/bugzilla-archives/60/60315/bug.html#c7")]
+        public void ShouldDownloadArtifactoryNuGetPackage(string packageVersion, string hash)
```
geofflamrock, line 103: I feel like it would be good to have these tests for the NuGet feed too

mjhilton:
An alternative solution was implemented in #779 against the Release Branch for 2021.1, and then merged forward to 2021.2 in #780, and to the main branch in #781. 

Closing this PR as no longer required.

PR #759, opened 2021-08-10
Correctly parse NodeExecutor inputs

egorpavlikhin:
### Background
We would like to JSON escape all evaluated variables inside of our inputs so that they can be safely passed to steps executing through Node.js. Our previous implementation did not support the many ways Octostache expressions can be written, this is an attempt to resolve this.

The way it works is quite simple:

1. Parse the Octostache template.
2. Convert the parsed template back to a string, so that there is a deterministic representation of the inputs.
3. Evaluate the variables inside.
4. `string.Replace` the variable expressions with values.

I've put in tests for most available variable expression types and they all appear to work, except one, [see here](https://github.com/OctopusDeploy/Octostache/pull/61).

```csharp
var template = TemplateParser.ParseTemplate(jsonInputs);
jsonInputs = template.ToString();
foreach (var templateToken in template.Tokens)
{
    var arguments = templateToken.GetArguments();
    if (!arguments.Any()) continue; // to avoid TextToken, which doesn't need to be evaluated
    string evaluated = variables.Evaluate(templateToken.ToString());
    jsonInputs = jsonInputs.Replace($"\"{templateToken.ToString()}\"", JsonConvert.ToString(evaluated));
}
```

### Dependencies
Octostache: https://github.com/OctopusDeploy/Octostache/pull/61

### DIFF
--- a/source/Calamari.Shared/Util/InputSubstitution.cs
+++ b/source/Calamari.Shared/Util/InputSubstitution.cs
@@ -12,7 +12,11 @@ namespace Calamari.Util
     {
         public static string SubstituteAndEscapeAllVariablesInJson(string jsonInputs, IVariables variables, ILog log)
         {
-            var template = TemplateParser.ParseTemplate(jsonInputs);
+            if (!TemplateParser.TryParseTemplate(jsonInputs, out var template, out string error))
+            {
+                throw new CommandException($"Variable expression could not be parsed. Error: {error}");
+            }
+            
             jsonInputs = template.ToString(); // we parse the template back to string to have a consistent representation of Octostache expressions
             foreach (var templateToken in template.Tokens)
             {
--- a/source/Calamari.Shared/Util/InputSubstitution.cs
+++ b/source/Calamari.Shared/Util/InputSubstitution.cs
@@ -1,5 +1,6 @@
 ﻿using System.Linq;
 using Calamari.Common.Commands;
+using Calamari.Common.Plumbing.Logging;
 using Calamari.Common.Plumbing.Variables;
 using Newtonsoft.Json;
 using Octostache;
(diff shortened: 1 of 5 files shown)


source/Calamari.Shared/Util/InputSubstitution.cs
```diff
@@ -9,22 +11,19 @@ public static class InputSubstitution
...
                 if (!arguments.Any()) continue; // to avoid TextToken, which doesn't need to be evaluated
-                var variableName = templateToken.ToString()
-                                                .Replace("#{", string.Empty)
-                                                .Replace("}", string.Empty);
-                var expanded = variables.Evaluate($"#{{{variableName} | JsonEscape}}");
-                tempVariableDictionaryToUseForExpandedVariables.Add(variableName, expanded);
+                string evaluated = variables.Evaluate(templateToken.ToString());
+                if (evaluated == templateToken.ToString())
+                    throw new CommandException($"Variable {templateToken.ToString()} could not be evaluated, check that the referenced variables exist.");
+                jsonInputs = jsonInputs.Replace($"\"{templateToken.ToString()}\"", JsonConvert.ToString(evaluated));
```
ray-nham, line 23: I don't quite get it here. Why do we include double quotes in the replacement? 
For example, inputs = `{ "ExtractedPath": "#{Package[package].ExtractedPath}" }` will be evaluated to  `{ "ExtractedPath": C:\\some-path }` => double quotes are replaced. Is this what would happen?

egorpavlikhin: `JsonConvert.ToString()` adds extra quotes around the result, I couldn't find a way to avoid that.

source/Calamari.Shared/Util/InputSubstitution.cs
```diff
@@ -9,22 +11,19 @@ public static class InputSubstitution
...
                 var arguments = templateToken.GetArguments();
                 if (!arguments.Any()) continue; // to avoid TextToken, which doesn't need to be evaluated
-                var variableName = templateToken.ToString()
-                                                .Replace("#{", string.Empty)
-                                                .Replace("}", string.Empty);
-                var expanded = variables.Evaluate($"#{{{variableName} | JsonEscape}}");
-                tempVariableDictionaryToUseForExpandedVariables.Add(variableName, expanded);
+                string evaluated = variables.Evaluate(templateToken.ToString());
+                if (evaluated == templateToken.ToString())
+                    throw new CommandException($"Variable {templateToken.ToString()} could not be evaluated, check that the referenced variables exist.");
```
ray-nham, line 22: Should we stop the execution here or just log the error message?

ray-nham (approved):
LGTM

PR #760, opened 2021-08-10
Fix nuget package structure for signed Calamari packages

mattocto:
Create SignAndPack method to sign files before using the DotNetPack command to build a correctly structured nuget package.


build.cake
```diff
@@ -180,6 +172,16 @@ Task("PackBinaries")
         actions.Add(() => DoPackage("Calamari", "netcoreapp3.1", nugetVersion, rid));
     }
 
+    var dotNetCorePackSettings = GetDotNetCorePackSettings();
+
+    var commonProjects = GetFiles("./source/**/*.Common.csproj");
```
johnsimons, line 177: Do we need wildcards here, isn't there really only `Calamari.Common.csproj`?

johnsimons (approved):
looks ok to me, but the tests seem to be failing

build.cake
mattocto: When I went back to the way we were constructing packages before, I chose to go back to using the wildcard because that was done before too. It doesn't look to me like it's needed, but it also doesn't appear to be doing any harm.

mattocto:
> the tests seem to be failing
I'll take a look and fix those up.

PR #761, opened 2021-08-11
Pass step folder path instead of executor path as parameter to bootstrapper invocation.

gb-8:
The step bootstrapper now needs multiple files from the step package as it is validating as well as executing it, so this is a small change to point the bootstrapper at the whole step package folder, rather than just the executable file.


shaunmarx (approved):
Changes look good to me 👍🏻

gb-8:
[Execution time validation](https://trello.com/c/910Ws1zg/51-execution-time-validation)

PR #763, opened 2021-08-18
Introduces commands with inputs for package-related features

andrewabest:
As a part of Step Package step execution, we have promoted [package extraction and manipulation](https://github.com/OctopusDeploy/Architecture/blob/main/Steps/Concepts/Execution.md#solution-the-execution-manifest) as first class features that can be coordinated by Octopus Server.

Octopus Server coordinates these features by sending through an appropriate Execution Manifest to the calamari `execute-manifest` command. 

This PR introduces the structure required to invoke package extraction and manipulation commands via the `execute-manifest` command.

The primary component that enables this is a new `Command<TInputs>` base class which allows us to take a set of raw inputs supplied on the execution manifest and invoke the named command with the inputs supplied by deserializing them into `TInputs` and passing them to the commands `Execute` function for it to do its job with.

This PR builds on https://github.com/OctopusDeploy/ServerExtensibility/pull/198

This PR is a dependency of https://github.com/OctopusDeploy/OctopusDeploy/pull/9846


source/Calamari.Common/Features/Substitutions/ISubstituteInFiles.cs
```diff
@@ -6,6 +6,7 @@ namespace Calamari.Common.Features.Substitutions
     public interface ISubstituteInFiles
     {
         void SubstituteBasedSettingsInSuppliedVariables(string currentDirectory);
+        void SubstituteBasedSettingsInSuppliedVariables(string currentDirectory, string[] filesToTarget);
```
johnsimons, line 9: Do we really need another one, this one is pretty much the same as `Substitute` method below?
If the issue is the name, we could just have a extension method?

source/Calamari/Commands/ExecuteManifestCommand.cs
```diff
@@ -58,24 +58,11 @@ public override int Execute(string[] commandLineArguments)
...
-
-            conventionRunner.RunConventions();
-
             foreach (var instruction in instructions)
             {
                 var tool = executionTools.First(x => x.Metadata.Tool == instruction.Launcher);
 
-                var result = tool.Value.Execute(instruction.LauncherInstructionsRaw,
-                    commandLineArguments.Skip(1).ToArray());
+                var result = tool.Value.Execute(instruction.LauncherInstructionsRaw);
```
johnsimons, line 65: Much nicer

source/Calamari/Commands/ICommandWithInputs.cs
```diff
@@ -0,0 +1,21 @@
+using System;
+using Calamari.Serialization;
+using Newtonsoft.Json;
+
+namespace Calamari.Commands
+{
+    public interface ICommandWithInputs
+    {
+        int Execute(string inputs);
```
johnsimons, line 9: Given all implementation here either `throw` to `return 0` it makes me wonder if this signature should return `void`?

johnsimons (approved):
Added a few small comments.
Wondering if there are enough tests that cover all these changes?

source/Calamari.Common/Features/Substitutions/ISubstituteInFiles.cs
andrewabest: Good call.

andrewabest: Fixed

source/Calamari/Commands/ICommandWithInputs.cs
andrewabest: I agree, seems a bit pointless here.

andrewabest: Fixed

andrewabest:
> Added a few small comments.
> Wondering if there are enough tests that cover all these changes?

I feel like the `ExecuteManifestCommandFixture` does a good job in covering the changes made. I didn't really want to test the new commands specifically, as they are just very thin wrappers over existing behaviour that is well tested.

PR #764, opened 2021-08-19
Support AWS IMDSv2

mcasperson:
IMDSv2 is a new security layer added to the EC2 metadata endpoint. It is documented here: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html.

Essentially all it requires is the generation of a token to be used with any subsequent IMDS API call. The only API calls we make are to get the EC2 IAM role.


mcasperson:
[Add support for IMDSv2](https://trello.com/c/agkPIab1/100-add-support-for-imdsv2)

andrewabest (approved):
I assume you've tested this appropriately :) change looks good.

mcasperson:
> I assume you've tested this appropriately

👍 The work John did to add new tests caught this use case nicely.

PR #767, opened 2021-08-24
Added WaitForDebugger variable, and code to do just that.

MissedTheMark:
This PR makes debugging Calamari a nicer experience out of the box.

If you add an `Octopus.Calamari.WaitForDebugger` variable to a project and set its value to 'true', then point the `Octopus.Calamari.Executable` variable to an appropriate exe, Calamari will wait for a debugger to attach before any commands are run:

image

image

This should work cross platform.

Thanks to @TomPeters for basically all of the code, if I'm being honest 😄

### DIFF
--- a/source/Calamari.Common/CalamariFlavourProgram.cs
+++ b/source/Calamari.Common/CalamariFlavourProgram.cs
@@ -21,7 +21,6 @@ using Calamari.Common.Plumbing.FileSystem;
 using Calamari.Common.Plumbing.Logging;
 using Calamari.Common.Plumbing.Proxies;
 using Calamari.Common.Plumbing.Variables;
-using Octopus.Versioning;
 
 namespace Calamari.Common
 {
--- a/source/Calamari.Common/CalamariFlavourProgram.cs
+++ b/source/Calamari.Common/CalamariFlavourProgram.cs
@@ -59,8 +59,7 @@ namespace Calamari.Common
                 using var container = builder.Build();
                 container.Resolve<VariableLogger>().LogVariables();
 
-                #if DEBUG
-
+#if DEBUG
                 var waitForDebugger = container.Resolve<IVariables>().Get(KnownVariables.Calamari.WaitForDebugger);
 
                 if (string.Equals(waitForDebugger, "true", StringComparison.CurrentCultureIgnoreCase))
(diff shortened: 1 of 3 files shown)


matt-richardson:
:sobeautiful:

Many many moons ago, I did something similar on a project, where it used the Visual Studio libraries to automatically attach to a newly launched process. Googling led me to [this gist](https://gist.github.com/atruskie/3813175) (if you happened to want to follow that idea).

That said, no idea if/how we could make it work for Rider... We'd also need to consider if adding this would end up with us shipping visualsudio libs with calamari (obviously a horrible idea).

source/Calamari.Common/CalamariFlavourProgram.cs
```diff
@@ -53,12 +55,26 @@ protected virtual int Run(string[] args)
...
+                if (string.Equals(waitForDebugger, "true", StringComparison.CurrentCultureIgnoreCase))
                 {
-                    container.Resolve<VariableLogger>().LogVariables();
+                    using var proc = Process.GetCurrentProcess();
+                    Log.Info($"Waiting for debugger to attach... (PID: {proc.Id})");
 
-                    return ResolveAndExecuteCommand(container, options);
+                    while (!Debugger.IsAttached)
+                    {
+                        Thread.Sleep(100);
```
andrewabest, line 72: 100ms feels like a very aggressive poll - let's change this to 1000ms - users won't notice the difference.

andrewabest (approved):
Changes look good - I've made one small suggestion. I considered whether we would want it to time out, or to continue to prompt the developer during polling, but I don't think either of these are required - the developer would have set the variable, and so would be expecting to wait for the prompt and attach a debugger. Not timing out gives them the luxury of coming back to it, and not repeating the prompt keeps the log spam down. Looks good 👍👍

MissedTheMark:
> Many many moons ago, I did something similar on a project, where it used the Visual Studio libraries to automatically attach to a newly launched process. Googling led me to [this gist](https://gist.github.com/atruskie/3813175) (if you happened to want to follow that idea).
> 
> That said, no idea if/how we could make it work for Rider... We'd also need to consider if adding this would end up with us shipping visualsudio libs with calamari (obviously a horrible idea).

@matt-richardson That's a really neat bit of code.  I may keep that in mind as a sharpening idea...

PR #773, opened 2021-09-15
fix: shift encryption key out of scope of user bash scripts

mjhilton:
# Background
The "Run a Script" step generates a number of supporting scripts on the target machine in the leadup to step execution. These include a top-level script that orchestrates, and a couple of scripts that provide octopus-specific functions that step authors can use, for example to access variables, publish artefacts etc. 

There is a top-level bootstrapper script generated, which looks something like this:
```
#!/bin/bash
source "$(pwd)/Bootstrap.57cb81bc-5.script.sh"
source "$(pwd)/script.sh" "command" "arguments" "from" "step" "definition"
```

When Calamari executes the step, a decryption key is the single argument to the bootstrapper script. The command executed on the target looks something like this, assuming the decryption key is `AE001BC55GAC29800B`
`/bin/bash "top-level-script.sh" "AE001BC55GAC29800B"`

The resulting bash process has a single argument, $1 = "AE001BC55GAC29800B"

# Problem
If no script arguments are defined by the user to be passed to their script, `source`ing `script.sh` results in the decryption key being available to the user script as `$1`, which will be masked if echo'd as console output.

This was discovered by a user accidentally, who was attempting to pass a first argument of `#{Octopus.Release.Id}` to a script in a Runbook. ReleaseId doesn't exist in a runbook, and the leading `#` resulted in all further arguments being bash-commented out. This effectively leads to a bootstrapping script that looks like this:

```
#!/bin/bash
source "$(pwd)/Bootstrap.57cb81bc-5.script.sh"
source "$(pwd)/script.sh" #{Octopus.Release.Id} "anything following commented out from a Bash perspective"
```

The user's script was then echo'ing `$1`, which they thought should be empty, but was printing out masked characters unexpectedly.

# Solution
Without making too much potentially dangerous change to how scripts get executed, which could have pretty far-reaching consequences, this change uses the `shift` bash command to "shift" the decryption key out of the scope in which the user's script is executed. 

Because the user script's arguments are set directly in the bootstrapper script when it's generated, there should be no impact on things when arguments are explicitly provided.

### DIFF
--- a/source/Calamari.Common/Features/Scripting/Bash/BashScriptBootstrapper.cs
+++ b/source/Calamari.Common/Features/Scripting/Bash/BashScriptBootstrapper.cs
@@ -116,7 +116,7 @@ namespace Calamari.Common.Features.Scripting.Bash
                 writer.NewLine = LinuxNewLine;
                 writer.WriteLine("#!/bin/bash");
                 writer.WriteLine("source \"$(pwd)/" + Path.GetFileName(configurationFile) + "\"");
-                writer.WriteLine("shift");
+                writer.WriteLine("shift"); // Shift the variable decryption key out of scope of the user script (see: https://github.com/OctopusDeploy/Calamari/pull/773)
                 writer.WriteLine("source \"$(pwd)/" + Path.GetFileName(script.File) + "\" " + script.Parameters);
                 writer.Flush();
             }
--- a/source/Calamari.Common/Features/Scripting/Bash/BashScriptBootstrapper.cs
+++ b/source/Calamari.Common/Features/Scripting/Bash/BashScriptBootstrapper.cs
@@ -116,6 +116,7 @@ namespace Calamari.Common.Features.Scripting.Bash
                 writer.NewLine = LinuxNewLine;
                 writer.WriteLine("#!/bin/bash");
                 writer.WriteLine("source \"$(pwd)/" + Path.GetFileName(configurationFile) + "\"");
+                writer.WriteLine("shift");
                 writer.WriteLine("source \"$(pwd)/" + Path.GetFileName(script.File) + "\" " + script.Parameters);
                 writer.Flush();
             }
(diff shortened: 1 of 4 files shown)


source/Calamari.Common/Features/Scripting/Bash/BashScriptBootstrapper.cs
```diff
@@ -116,6 +116,7 @@ public static (string bootstrapFile, string[] temporaryFiles) PrepareBootstrapFi
                 writer.NewLine = LinuxNewLine;
                 writer.WriteLine("#!/bin/bash");
                 writer.WriteLine("source \"$(pwd)/" + Path.GetFileName(configurationFile) + "\"");
+                writer.WriteLine("shift");
```
shaun-od, line 119: Maybe a comment (if it is what we do here) about why shift is needed to be added?

source/Calamari.Tests/Fixtures/Bash/BashFixture.cs
```diff
@@ -58,6 +59,19 @@ public void ShouldConsumeParametersWithQuotes()
             output.AssertSuccess();
             output.AssertOutput("Parameters Para meter0Para meter1");
         }
+        
+        [Test]
+        [RequiresBashDotExeIfOnWindows]
+        public void ShouldNotHaveDecryptionKeyInScopeOfUserScript()
```
shaun-od, line 65: Maybe a little context about why this test is actually testing we don't get the decryption key in the out when no parameters are available.

source/Calamari.Common/Features/Scripting/Bash/BashScriptBootstrapper.cs
mjhilton: ✅ Comment added

source/Calamari.Tests/Fixtures/Bash/BashFixture.cs
mjhilton: ✅ Simplified test and made it more generic/clearer in intent

shaun-od (approved):
LGTM

PR #774, opened 2021-09-20
adding code to suppress write-progress breaking output on linux

shaun-od:
# Background

When the "Run a Script" is running a PowerShell script that contains a call to a cmdlet or contains a `Write-Progress` call breaks the following line if it contains an Octopus marker like ##octopus[stdout-warning]. It appears that `Write-Progress` does not put a trailing newline character which means the marker is not treated correctly.

This only happens on Linux-based OS (including macOS). You can see below this script when runs on a Windows worker always parses the output correctly.

You can read more about [Write-Progess](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/write-progress?view=powershell-7.1) and [$ProgressPreference](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_preference_variables?view=powershell-7.1#progresspreference) in the Microsoft documentation.

# Problem

The following PowerShell script is a small repro of the issue mention above.

```PowerShell

Write-Warning "This should log as a warning"
Write-Progress -Activity "Progress" -PercentComplete 100
Write-Warning "This should also log as a warning"

```

## Before

Prior to the fix, this is what is shown in the output in Octopus when the step that ran the code block on a Linux worker is below.

Screen Shot 2021-09-21 at 11 34 48 am

## After

The output in Octopus is now parsing the second warning correctly

Screen Shot 2021-09-21 at 11 35 28 am

# Solution

This has been resolved by adding the following line to the PowerShell bootstrap script so that the provided user script is run `"$ProgressPreference = 'SilentlyContinue';"` by default.

The PR also adds a new Octopus Variable that can be set `Octopus.Action.PowerShell.OutputPowerShellProgress` to true and this will set `"$ProgressPreference = 'Continue';"` instead of `'SilentlyContinue'` this would allow any script that actually requires `$ProgressPreference` to be run with the default value of `'Continue'`.

### DIFF
--- a/source/Calamari.Common/Features/Scripting/WindowsPowerShell/PowerShellBootstrapper.cs
+++ b/source/Calamari.Common/Features/Scripting/WindowsPowerShell/PowerShellBootstrapper.cs
@@ -201,13 +201,13 @@ namespace Calamari.Common.Features.Scripting.WindowsPowerShell
             return commandArguments.ToString();
         }
 
-        string GetDisplayProgressCommand(IVariables variables)
+        static string GetDisplayProgressCommand(IVariables variables)
         {
             var powerShellProgressPreferenceArg = variables[PowerShellVariables.OutputPowerShellProgress];
             
-            int.TryParse(powerShellProgressPreferenceArg, out var processArgAsInt);
-            bool.TryParse(powerShellProgressPreferenceArg, out var processArgAsBool);
-            if (processArgAsInt > 0 || processArgAsBool)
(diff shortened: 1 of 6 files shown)


shaun-od:
This PR is being abandoned as it is decided that this issue should be raised with PowerShell and wait for it to be fixed.

See https://github.com/OctopusDeploy/Issues/issues/7062

and https://github.com/PowerShell/PowerShell/issues/16139 

for details.

PR #775, opened 2021-09-21
Fixed race condition with stack creation and event processing

mcasperson:
A race condition exists in the deployment of a CF template where a stack may complete in an error state after the last event was inspected and before the check to determine if the stack is completed. This in turn results in the step succeeding even if the stack was not deployed correctly.

This PR adds a final event check once the stack is complete to catch any error:

image

Fixes https://github.com/OctopusDeploy/Issues/issues/7058


andrewabest (approved):
Changes look sensible - comment for clarity is good too.

PR #778, opened 2021-10-07
Ensure all file format replacers are present for all commands

andrewabest:
As Calamari evolved it underwent a few changes. One was in supporting more file formats for our structured configuration replacement feature. Another was it gained DI, as it originally did not use it.

Originally our non-DI solution for the file format replacers was a simple factory class that returned the set of replacers available: https://github.com/OctopusDeploy/Calamari/commit/b5e6b0d20e09449f4ab302e32c3d5dd8b6a8dba2#diff-5c599a99254cd6918f74120d10a7bf101d40d189e7078cc500d83c6cfd80a07a

This eventually grew to support four replacers - json, xml, yaml, and properties formats.

When the modularity refactor was undertaken, it seems that we didn't register all of the available formats. My suspicion is that we copied the available formats from the factory, and then some work was done to implement new formats that added them to the factory but didn't add DI registrations for them. Which now leads to some steps having all format replacers available for this feature, and some only having json and yaml replacers available.

This PR corrects this to ensure the same, central set of registered format replacers is used everywhere. This change will need to be propagated through Sashimi and release branches as appropriate.


source/Calamari.Common/CalamariFlavourProgramAsync.cs
```diff
@@ -56,15 +56,14 @@ protected virtual void ConfigureContainer(ContainerBuilder builder, CommonOption
...
-            builder.RegisterType<JsonFormatVariableReplacer>().As<IFileFormatVariableReplacer>();
-            builder.RegisterType<YamlFormatVariableReplacer>().As<IFileFormatVariableReplacer>();
-            builder.RegisterType<StructuredConfigVariablesService>().As<IStructuredConfigVariablesService>();
             builder.Register(context => ConfigurationTransformer.FromVariables(context.Resolve<IVariables>(), context.Resolve<ILog>())).As<IConfigurationTransformer>();
             builder.RegisterType<DeploymentJournalWriter>().As<IDeploymentJournalWriter>().SingleInstance();
             builder.RegisterType<CodeGenFunctionsRegistry>().SingleInstance();
 
             var assemblies = GetAllAssembliesToRegister().ToArray();
 
+            builder.RegisterAssemblyTypes(assemblies).AssignableTo<IFileFormatVariableReplacer>().As<IFileFormatVariableReplacer>();
```
mjhilton, line 65: I'm not sure why this is here. I think that the `StructuredConfigVariablesModule` ([below](https://github.com/OctopusDeploy/Calamari/pull/778/files#diff-04f72119f47dbc7a899939a63d263c455c3aa8cd09de9dd04dca3fa713f6e400R92)) should be handling this, in a way that respects the ordering of services.

andrewabest: I _suspect_ this caters for the fact that we might want register file format variable replacers from an assembly that implements `CalamariFlavourProgramAsync` and overrides `GetProgramAssembliesToRegister`.

Caveat: I don't think we do this, nor do I think we really want to either. 

Also, I think the code here probably leads to double registrations, which I don't like. If we want to correct it for now and leave this capability in tact, I'd change this to

`builder.RegisterAssemblyTypes(GetProgramAssembliesToRegister())...`

mjhilton:
Replaced by #805

PR #779, opened 2021-10-07
adding nuget metadata version fix for NET full framework V3 downloader

shaun-od:
# Background

When Calamari is instructed to download NuGet packages with a version that contains Semver metadata (+somestring) and example of this `1.0.0-alpha.3+metadata` it fails to find the package when it exists.

This relates to OctopusDeploy/Issues#6965 and is a partial fix

# Problem

There was a number of issues found while investigating this.

- With NET Full Calamari's it was choosing the wrong NuGet downloader e.g V2 downloader when it was V3 feed.
- The downloading code in Calamari does not semver packages.

## Before

Screen Shot 2021-10-07 at 3 46 21 pm

## After

Screen Shot 2021-10-07 at 3 48 19 pm

# Solution

- This PR improves detecting if the NuGet feed is V3 or V2 protocol.
- It also resolves a deadlock that was appearing in the NET 4.0 of Calamari.
- It improves the NuGet V3 downloader for NET Framework when feeds do not implement all of the protocol

### DIFF
--- a/source/Calamari.Shared/Integration/Packages/NuGet/NuGetV3Downloader.cs
+++ b/source/Calamari.Shared/Integration/Packages/NuGet/NuGetV3Downloader.cs
@@ -4,6 +4,7 @@
 #if USE_NUGET_V2_LIBS
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Net;
@@ -11,6 +12,7 @@ using System.Net.Http;
 using System.Web;
 using Calamari.Common.Commands;
 using Calamari.Common.Plumbing.Logging;
+using Calamari.Deployment;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using Octopus.CoreUtilities.Extensions;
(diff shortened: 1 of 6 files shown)


source/Calamari.Shared/Integration/Packages/NuGet/NuGetV3Downloader.cs
```diff
@@ -8,23 +8,71 @@
...
+            if (feedUri.ToString().EndsWith(".json", StringComparison.OrdinalIgnoreCase))
+            {
+                return true;
+            }
+            
+            return IsJsonEndpoint(feedUri, feedCredentials, httpTimeout);
+        }
+
+        static bool IsJsonEndpoint(Uri feedUri, ICredentials feedCredentials, TimeSpan httpTimeout)
+        {
```
andrewabest, line 35: It feels like we've duplicated most of the code from `GetHttp` below. Is there a reasonable way to avoid this duplication?

andrewabest (approved):
@shaun-od and I paired on this implementation and it is good to :shipit:

PR #781, opened 2021-10-15
Merging NuGet partial fix from release 2021.2 to master

shaun-od:
This PR is merging the following fix forward from 2021.2 release to master, original details are below.

# Background

When Calamari is instructed to download NuGet packages with a version that contains Semver metadata (+somestring) and example of this `1.0.0-alpha.3+metadata` it fails to find the package when it exists.

This relates to OctopusDeploy/Issues#6965 and is a partial fix

# Problem

There was a number of issues found while investigating this.

- With NET Full Calamari's it was choosing the wrong NuGet downloader e.g V2 downloader when it was V3 feed.
- The downloading code in Calamari does not semver packages.

## Before

Screen Shot 2021-10-07 at 3 46 21 pm

## After

Screen Shot 2021-10-07 at 3 48 19 pm

# Solution

- This PR improves detecting if the NuGet feed is V3 or V2 protocol.
- It also resolves a deadlock that was appearing in the NET 4.0 of Calamari.
- It improves the NuGet V3 downloader for NET Framework when feeds do not implement all of the protocol

### DIFF
--- a/source/Calamari.Tests/KubernetesFixtures/Terraform/Clusters/providers.tf
+++ b/source/Calamari.Tests/KubernetesFixtures/Terraform/Clusters/providers.tf
@@ -2,7 +2,7 @@ terraform {
   required_providers {
     azurerm = {
       source  = "hashicorp/azurerm"
-      version = ">= 2.63.0"
+      version = "= 2.78.0" #Due to a bug with 2.79.x we are pinning to this version for now. https://github.com/Azure/AKS/issues/2584 
     }
     aws = {
       source  = "hashicorp/aws"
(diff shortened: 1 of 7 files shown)


source/Calamari.Tests/KubernetesFixtures/Terraform/Clusters/providers.tf
```diff
@@ -2,7 +2,7 @@ terraform {
   required_providers {
     azurerm = {
       source  = "hashicorp/azurerm"
-      version = ">= 2.63.0"
```
andrewabest, line 5: This looks like it has crept in from another commit? Probably all good to drag it forward if the tests are green

andrewabest (approved):
Looks like the changes have come through as expected, with a hitchhiker being picked up as well. Good to ship on 🟢

source/Calamari.Tests/KubernetesFixtures/Terraform/Clusters/providers.tf
shaun-od: Yeah, that was added in the last PR to fix the failing Kubernetes tests 😄

PR #784, opened 2021-10-20
Ensure tests fail fast and give a clear reason, when EnvVars not present

mjhilton:
# Background
Our Calamari tests comprise a large number of integration-style tests that interact with external infrastructure to do their work. 

These often require credentials which we don't want to store in the source code, so they get pushed into the test context at runtime using environment variables. This works nicely for our build pipeline, as TeamCity has all the configuration and sensitive values stored appropriately. However, it means that the local test-run experience is quite poor.

There is actually some infrastructure already in the `Calamari.Testing` test infrastructure library which handles this in a pretty graceful way. Rather than directly reading the environment variables in each test, we can use the `ExternalVariables.Get` method along with a centralised register of variables our tests rely on as defined in `ExternalVariable`. Then, if a required variable is not defined, the test fails as soon as the variable is read, with a nice message that guides the (Octopus internal) user to success, like this:

```
Environment Variable `AWS_OctopusAPITester_Access` could not be found. The value can be found in the password store under `AWS - OctopusAPITester`
```

However, a number of our integration-style tests don't use these helpers, and read the variables directly with `Environment.GetEnvironmentVariable`. The result is that these tests don't fail until the variable is used, often with confusing messages such as authentication failures that take a while to reverse-work why the failure occurred. It also means the test has to wait for the call to an external service to fail (which each has their own timeouts) before the test itself fails.

# Changes

* Converts all usages of `Environment.GetEnvironmentVariable` to instead use `ExternalVariables.Get`
* Adds missing `ExternalVariable` entries

# Results
Previous failure messages (which are not always as expedient in failing as these ones, some of them take up to 4 minutes before timing out... it's all dependent on what the test is attempting to interact with)
image

Updated failure messages:
image

### DIFF
--- a/source/Calamari.Testing/EnvironmentVariables.cs
+++ b/source/Calamari.Testing/EnvironmentVariables.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Linq;
 using Calamari.Common.Plumbing.Logging;
+using NUnit.Framework;
 
 namespace Calamari.Testing
 {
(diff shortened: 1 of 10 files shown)


mjhilton:
> Heh. I did some similar stuff for Server this morning - [OctopusDeploy/OctopusDeploy#10516](https://github.com/OctopusDeploy/OctopusDeploy/pull/10516). (I don't think it's directly useful to you, but sharing on the off chance it is...)

Definitely a good alternate approach to think about, and could be another iterative step to further improve the experience. Thanks @matt-richardson.

mjhilton:
@matt-richardson I've had a look into your changes now. I couldn't quite figure out how the dependency resolution is hooked up for those tests, but I've got a hunch it's done in XUnit?

I'm wondering if that dependency resolution approach is something that could reasonably be done in NUnit or not.

(I definitely wouldn't be averse to getting off NUnit for Calamari, when we get a chance)

matt-richardson:
> @matt-richardson I've had a look into your changes now. I couldn't quite figure out how the dependency resolution is hooked up for those tests, but I've got a hunch it's done in XUnit?
> 
> I'm wondering if that dependency resolution approach is something that could reasonably be done in NUnit or not.
> 
> (I definitely wouldn't be averse to getting off NUnit for Calamari, when we get a chance)

it's all part of the magic of the E2E tests - it uses a custom framework that uses Autofac under the hood. So... not the easist thing to copy, but as you say, maybe with XUnit there's nicer future.

andrewabest (approved):
Looks like a nice improvement 👍

PR #785, opened 2021-10-31
Renames the script modules responsible for creating dynamic targets

andrewabest:
The existing naming for the script modules responsible for creating dynamic targets were somewhat confusing for users.

This PR updates the names of the modules to:

PS: `New-OctopusTarget`
Bash: `new_octopustarget()`

Which should be easier for people to understand the intent of.

### DIFF
--- a/source/Calamari.Common/Features/Scripting/Bash/Bootstrap.sh
+++ b/source/Calamari.Common/Features/Scripting/Bash/Bootstrap.sh
@@ -145,7 +145,7 @@ function remove-octopustarget {
 	echo "##octopus[delete-target machine='$(encode_servicemessagevalue "$1")']"
 }
 
-function create_steppackagetarget() (
+function new_octopustarget() (
   parameters=""
 
   while :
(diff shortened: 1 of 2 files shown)


andrewabest:
@johnsimons do you think this rename is OK? I understand the risk in declaring a function in global scope with a reasonably generic name, but is there anything else I may need to worry about here?

andrewabest:
Related:

https://github.com/OctopusDeploy/OctopusDeploy/pull/9774
https://github.com/OctopusDeploy/Calamari/commit/2413e18f6f6fb409f2c58dbe3ada53b339bd199f

johnsimons:
@andrewabest the change looks good, but whether the new name is accurate, that is open to interpretation.
IMO I find it a bit too generic, and hence a bit confusing because I feel like I should be able to create any type of target not just step package ones.
I hope this makes sense.

johnsimons (approved):

johnsimons:
Also, I forgot, you intentionally did not make it backwards compatible correct?

andrewabest:
> Also, I forgot, you intentionally did not make it backwards compatible correct?

Correct, we don't anticipate anyone to be using this out in the wild yet, so making a breaking change here. Thanks for taking a look!

> but whether the new name is accurate, that is open to interpretation.

Yes I get what you are saying, there is a trade-off here. I've added some documentation [here](https://github.com/OctopusDeploy/docs/pull/1380) that hopefully goes a way to showing what is available via that function. I'll have a think about what we can do in Server to provide an informative failure back if someone tries to create a target that isn't supported via this function, and implement that when I pull these changes in.

PR #786, opened 2021-11-09
Use environment variable as path for package retention journal

N-lson:
This PR updates the location of the journal used for package retention from a hard-coded one to an environment variable `CalamariPackageRetentionJournalPath` provided by Tentacle. The journal will now be stored at the tentacle's root directory along with the existing `DeploymentJournal.xml`.

## How to test

1. Start Octopus server
2. Create a Tentacle using the latest `main` branch from `OctopusTentacle` (which has the new environment variable)
4. Set up project variables to debug Calamari (thanks Mark for the [guide here](https://trello.com/c/jS95BzZ3/10-how-do-i-debug-calamari))

## Older Tentacle versions
~~If a Tentacle is created using a version prior to one that has the environment variable, then the deployment will blow up `System.ArgumentNullException: Value cannot be null. (Parameter 'path')`. How should we deal with versions that don't have the environment variable? Have a default value? Do we need to deal with older versions?~~ If you use an older Tentacle version it will default to using `TentacleHome` as the base path and `PackageRetentionJournal.json` as the name of the file (which is what the environment variable is set to in Tentacle anyway).


source/Calamari/Deployment/PackageRetention/Repositories/JournalJsonRepositoryFactory.cs
```diff
@@ -9,11 +11,11 @@ public class JournalJsonRepositoryFactory : IJournalRepositoryFactory
         readonly ISemaphoreFactory semaphoreFactory;
         readonly string journalPath;
 
-        public JournalJsonRepositoryFactory(ICalamariFileSystem fileSystem, ISemaphoreFactory semaphoreFactory)//, string journalPath)
+        public JournalJsonRepositoryFactory(IVariables variables, ICalamariFileSystem fileSystem, ISemaphoreFactory semaphoreFactory)//, string journalPath)
```
tothegills, line 14: Think that comment at the end of the line can go now

MissedTheMark:
Yes, we'll need to handle the scenario where we don't have the environment variable.  What do we do with the current journal when it uses the other env variable?

N-lson:
```
            var canWrite = deployment.Variables.Get(TentacleVariables.Agent.JournalPath) != null;

            if(canWrite && hasPackages && !deployment.SkipJournal)
                journal.AddJournalEntry(new JournalEntry(deployment, wasSuccessful));
```
We could add it as a condition to `if (variables.IsPackageRetentionEnabled())`? Something like
```
if (variables.IsPackageRetentionEnabled() && canWrite)
```

MissedTheMark:
> We could add it as a condition to `if (variables.IsPackageRetentionEnabled())`? Something like
> 
> ```
> if (variables.IsPackageRetentionEnabled() && canWrite)
> ```
I think that works.  @tothegills what are your thoughts?   Realistically, given that Calamari is updated by the server, they should mostly match, so this scenario will hopefully not occur too often.

tothegills:
If we intend for this to replace the current package cleanup I think it will need to work for old versions of Tentacle.

tothegills:
(Also in cases where Tentacle is not being used, eg SSH)

MissedTheMark:
That makes sense.  What about this:
1. If our new env variable exists, use that location
2. If it doesn't but the old journal variable exists, strip the filename from that path, and replace it with packageJournal.json or something similar.  I.e. if old variable is 'c:\octopus\journal.xml', convert it to 'c:\octopus\packageJournal.json'
3. If it doesn't have either of those, then they're trying pretty hard not to have a journal, and maybe we should listen to them, and not have a journal.

MissedTheMark:
@tothegills do you know off the top of your head how the current journal works for SSH targets?  Like do they have to manually set the env variable, or is it done via scripts or something?

tothegills:
I think the Octopus server set the journal path variable for SSH targets. IIRC there's a Tentacle home directory variable either in the variable bag or as an environment variable which should be the directory that the old journal is in. Maybe could use that as a default?

MissedTheMark:
Sounds good to me.  @N-lson does that make sense?  Happy to Zoom or 🍐 if you want a hand.

N-lson:
Yeah sounds good. I did find that home variable before - it's `TentacleHome` and it's used in some other places already:
```
var tentacleHome = Environment.GetEnvironmentVariable("TentacleHome");
if (tentacleHome == null)
                throw new Exception("Environment variable 'TentacleHome' has not been set.");
```
I wasn't sure whether it would be set on SSH targets but sounds like it would be set?

tothegills:
https://github.com/OctopusDeploy/OctopusDeploy/blob/master/source/Octopus.Server/Orchestration/Targets/Ssh/SshBashPaths.cs

source/Calamari.Common/Plumbing/Variables/TentacleVariables.cs
```diff
@@ -14,6 +14,7 @@ public static class Agent
             public static readonly string InstanceName = "Octopus.Tentacle.Agent.InstanceName";
             public static readonly string ApplicationDirectoryPath = "Octopus.Tentacle.Agent.ApplicationDirectoryPath";
             public static readonly string JournalPath = "env:TentacleJournal";
+            public static readonly string TentacleHome = "env:TentacleHome";
```
N-lson, line 17: The existing usages of `TentacleHome` do something like `Environment.GetEnvironmentVariable("TentacleHome")`. But since we have the variables bag I think it makes sense to just grab it from there in the same way we're trying to get the journal path variable.

source/Calamari/Deployment/PackageRetention/Repositories/JsonJournalRepository.cs
```diff
@@ -21,11 +22,18 @@ public class JsonJournalRepository
...
+        public JsonJournalRepository(ICalamariFileSystem fileSystem, ISemaphoreFactory semaphoreFactory, IVariables variables)
         {
             this.fileSystem = fileSystem;
             this.semaphoreFactory = semaphoreFactory;
-            this.journalPath = @"C:\Octopus\PackageJournal.json";//journalPath;
+
+            var packageRetentionJournalPath = variables.Get(KnownVariables.Calamari.PackageRetentionJournalPath);
+            if (packageRetentionJournalPath == null)
+            {
+                var tentacleHome = variables.Get(TentacleVariables.Agent.TentacleHome) ?? throw new Exception("Environment variable 'TentacleHome' has not been set.");
```
N-lson, line 33: Other usages of `TentacleHome` throw an exception if it's null. Not sure what the alternative would be here.

source/Calamari/Deployment/PackageRetention/Repositories/JsonJournalRepository.cs
```diff
@@ -16,16 +17,24 @@ public class JsonJournalRepository
     {
         Dictionary<PackageIdentity, JournalEntry> journalEntries;
         const string SemaphoreName = "Octopus.Calamari.PackageJournal";
+        internal const string DefaultJournalName = "PackageRetentionJournal.json";
 
         readonly ICalamariFileSystem fileSystem;
         readonly ISemaphoreFactory semaphoreFactory;
-        readonly string journalPath;
+        internal string JournalPath { get; }
```
N-lson, line 24: Thinking this isn't the best way to do this (but not sure on how else). I've added a unit test to check that `JournalPath` is being set correctly based on the environment variables available which means this needs to be accessible by the test.

The alternative would be to have a more integration-test style test, with saving the file and checking it exists. But it doesn't look like we have an existing mock of `ICalamariFileSystem` which has this capability (closest thing is `TestCalamariPhysicalFileSystem`).

PR #787, opened 2021-11-10
Rename package retention journal if contents are invalid

N-lson:
This PR adds a check when reading from the journal file on whether the contents are valid json or not. If not, the old file is renamed to `<OLDNAME>_<DATETIME>.json` where `<DATETIME>` is the current datetime.

## How to test

1. Start Octopus server
2. Add a Tentacle
4. Set up project variables to debug Calamari (Mark's [guide here](https://trello.com/c/jS95BzZ3/10-how-do-i-debug-calamari))
7. Open the file location of the tentacle and locate `PackageRetentionJournal.json`
8. Edit it to be invalid json eg. remove the closing `]`
9. Re-deploy the release w/ attaching Calamari
10. A backup `.json` file should be created and the `PackageRetentionJournal.json` will be restored to valid json. The log should look something like this:
image

I'll probably tear this PR down and make a new one when the filepath changes are merged in, since this change relies on some functionality from that change (location of the file, unit test file)

### DIFF
--- a/source/Calamari.Tests/Fixtures/PackageRetention/JsonJournalRepositoryFixture.cs
+++ b/source/Calamari.Tests/Fixtures/PackageRetention/JsonJournalRepositoryFixture.cs
@@ -1,6 +1,7 @@
-﻿using System.IO;
+﻿using System;
+using System.IO;
 using Calamari.Common.Features.Processes.Semaphores;
-using Calamari.Common.Plumbing.FileSystem;
+using Calamari.Common.Plumbing.Logging;
 using Calamari.Common.Plumbing.Variables;
 using Calamari.Deployment.PackageRetention.Repositories;
 using Calamari.Tests.Fixtures.Integration.FileSystem;
@@ -38,7 +39,7 @@ namespace Calamari.Tests.Fixtures.PackageRetention
             var variables = Substitute.For<IVariables>();
             variables.Get(KnownVariables.Calamari.PackageRetentionJournalPath).Returns(journalPath);
 
-            var repository = new JsonJournalRepository(TestCalamariPhysicalFileSystem.GetPhysicalFileSystem(), Substitute.For<ISemaphoreFactory>(), variables);
+            var repository = new JsonJournalRepository(TestCalamariPhysicalFileSystem.GetPhysicalFileSystem(), Substitute.For<ISemaphoreFactory>(), variables, Substitute.For<ILog>());
 
             repository.Commit();
             Assert.IsTrue(File.Exists(journalPath));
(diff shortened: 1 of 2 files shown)


source/Calamari/Deployment/PackageRetention/Repositories/JsonJournalRepository.cs
```diff
@@ -68,8 +71,21 @@ void Load()
...
+                if (TryParseJournal(json, out var journalContents))
+                {
+                    journalEntries = journalContents.ToDictionary(entry => entry.Package, entry => entry);
+                }
+                else
+                {
+                    var journalFileName = journalPath.Substring(0, journalPath.LastIndexOf(".", StringComparison.Ordinal));
+                    var backupJournalPath = $"{journalFileName}_{DateTimeOffset.UtcNow:yyyyMMddTHHmmss}.json"; // eg. PackageRetentionJournal_20210101T120000.json
+
+                    log.Warn($"The existing package retention journal file {journalPath} could not be read. The file will be renamed to {backupJournalPath}. A new journal will be created.");
```
MissedTheMark, line 84: I like this message 👍

source/Calamari/Deployment/PackageRetention/Repositories/JsonJournalRepository.cs
```diff
@@ -68,8 +71,21 @@ void Load()
...
+                    File.Move(journalPath, backupJournalPath);
```
MissedTheMark, line 85: There's an overload for File.Move that accepts a boolean to overwrite the file if it already exists, which would be good to use - otherwise it will throw an exception.  The odds of this happening aren't high, but they're not zero either.

source/Calamari/Deployment/PackageRetention/Repositories/JsonJournalRepository.cs
```diff
@@ -92,5 +108,20 @@ void Save()
...
+        static bool TryParseJournal(string json, out List<JournalEntry> result)
+        {
+            try
+            {
+                result = JsonConvert.DeserializeObject<List<JournalEntry>>(json);
+                return true;
+            }
+            catch (Exception e)
+            {
+                Log.Error($"Unable to parse the package retention journal file. Error message: {e.Message}");
```
MissedTheMark, line 121: Given that we're telling the customer about this issue as a warning on line 84, lets make this Log.Verbose, so this info is recorded, but not shown up front.  Changing e.Message to e.ToString() will give more detail for this too, including the exception type, and a stack trace.

source/Calamari.Tests/Fixtures/PackageRetention/JsonJournalRepositoryFixture.cs
```diff
@@ -53,12 +54,29 @@ public void WhenCalamariPackageRetentionJournalPathDoesNotExist_ThenTheJournalIs
...
+        {
+            const string invalidJson = @"[{""a"",}]";
+            var journalPath = Path.Combine(TentacleHome, "PackageRetentionJournal.json");
+
+            File.WriteAllText(journalPath, invalidJson);
+
+            var variables = Substitute.For<IVariables>();
+            variables.Get(KnownVariables.Calamari.PackageRetentionJournalPath).Returns(journalPath);
+
+            // Load()
```
MissedTheMark, line 76: Do we need this comment?

source/Calamari/Deployment/PackageRetention/Repositories/JsonJournalRepository.cs
```diff
@@ -68,8 +71,21 @@ void Load()
...
-                journalEntries = JsonConvert.DeserializeObject<List<JournalEntry>>(json)
-                                            .ToDictionary(entry => entry.Package, entry => entry);
+
+                if (TryParseJournal(json, out var journalContents))
+                {
+                    journalEntries = journalContents.ToDictionary(entry => entry.Package, entry => entry);
+                }
+                else
+                {
+                    var journalFileName = journalPath.Substring(0, journalPath.LastIndexOf(".", StringComparison.Ordinal));
```
MissedTheMark, line 81: Path.GetFileNameWithoutExtension() is your friend here.  Nice implementation though!

MissedTheMark (changes requested):

source/Calamari/Deployment/PackageRetention/Repositories/JsonJournalRepository.cs
N-lson: Oh of course there's a thing for that 😅

source/Calamari/Deployment/PackageRetention/Repositories/JsonJournalRepository.cs
N-lson: Doesn't exist in 4.0 😢 https://docs.microsoft.com/en-us/dotnet/api/system.io.file.move?view=netframework-4.0

I'll use the alternative they suggest seeing as we should have a small file

> In all .NET versions, you can call Copy(String, String, Boolean) to copy with overwrite, then call Delete to remove the excess source file. This strategy is advisable if the file being copied is small, and you are looking for an "atomic" file operation. If you Delete the file first, and the system or program crashes, the destination file will no longer exist.

source/Calamari.Tests/Fixtures/PackageRetention/JsonJournalRepositoryFixture.cs
N-lson: Don't think so. Initially had it as a "this is where the load happens" but the test is pretty small and I think the name of it explains what it's doing well enough.

source/Calamari/Deployment/PackageRetention/Repositories/JsonJournalRepository.cs
MissedTheMark: Ah, Net40 is such a pain!

MissedTheMark (approved):
LGTM! 👍

PR #788, opened 2021-11-10
Use environment variable as path for package retention journal

N-lson:
Older conversations can be found on this PR https://github.com/OctopusDeploy/Calamari/pull/786

This PR updates the location of the journal used for package retention from a hard-coded one to an environment variable `CalamariPackageRetentionJournalPath` provided by Tentacle. The journal will now be stored at the tentacle's root directory along with the existing `DeploymentJournal.xml`.

## How to test

1. Start Octopus server
2. Create a Tentacle using the latest `main` branch from `OctopusTentacle` (which has the new environment variable)
4. Set up project variables to debug Calamari (thanks Mark for the [guide here](https://trello.com/c/jS95BzZ3/10-how-do-i-debug-calamari))

### DIFF
--- a/source/Calamari.Tests/Fixtures/PackageRetention/JsonJournalRepositoryFixture.cs
+++ b/source/Calamari.Tests/Fixtures/PackageRetention/JsonJournalRepositoryFixture.cs
@@ -32,8 +32,8 @@ namespace Calamari.Tests.Fixtures.PackageRetention
                 Directory.Delete(testDir, true);
         }
 
-        [TestCase("PackageRetentionJournal.json")]
-        [TestCase(null)]
+        [TestCase("PackageRetentionJournal.json", TestName = "CalamariPackageRetentionJournalPath is set to a non-null value")]
+        [TestCase(null, TestName = "CalamariPackageRetentionJournalPath is null; default to TentacleHome")]
         public void WhenTheJournalIsLoadedAndCommittedTo_ThenTheJournalContainsTheCorrectContents(string packageRetentionJournalPath)
         {
             var journalPath = packageRetentionJournalPath == null ? null : Path.Combine(testDir, packageRetentionJournalPath);
--- a/source/Calamari.Tests/Fixtures/PackageRetention/JsonJournalRepositoryFixture.cs
+++ b/source/Calamari.Tests/Fixtures/PackageRetention/JsonJournalRepositoryFixture.cs
@@ -16,52 +16,34 @@ namespace Calamari.Tests.Fixtures.PackageRetention
     [TestFixture]
     public class JsonJournalRepositoryFixture
     {
-        readonly string tentacleHome = TestEnvironment.GetTestPath("Fixtures", "JsonJournalRepository");
+        readonly string testDir = TestEnvironment.GetTestPath("Fixtures", "JsonJournalRepository");
 
         [SetUp]
         public void SetUp()
         {
-            if (!Directory.Exists(tentacleHome))
-                Directory.CreateDirectory(tentacleHome);
+            if (!Directory.Exists(testDir))
+                Directory.CreateDirectory(testDir);
         }
 
         [TearDown]
         public void TearDown()
         {
-            if (Directory.Exists(tentacleHome))
-                Directory.Delete(tentacleHome, true);
+            if (Directory.Exists(testDir))
+                Directory.Delete(testDir, true);
         }
 
-        [Test]
-        public void WhenCalamariPackageRetentionJournalPathExists_ThenTheJournalIsCreatedAtTheGivenPath()
+        [TestCase("PackageRetentionJournal.json")]
+        [TestCase(null)]
+        public void WhenTheJournalIsLoadedAndCommittedTo_ThenTheJournalContainsTheCorrectContents(string packageRetentionJournalPath)
         {
-            var journalPath = Path.Combine(tentacleHome, "PackageRetentionJournal.json");
-            var thePackage = new PackageIdentity("TestPackageWithJournalPath", "0.0.1");
-            var journalEntry = new JournalEntry(thePackage);
+            var journalPath = packageRetentionJournalPath == null ? null : Path.Combine(testDir, packageRetentionJournalPath);
 
             var variables = Substitute.For<IVariables>();
             variables.Get(KnownVariables.Calamari.PackageRetentionJournalPath).Returns(journalPath);
+            variables.Get(TentacleVariables.Agent.TentacleHome).Returns(testDir);
 
-            var repositoryFactory = new JsonJournalRepositoryFactory(TestCalamariPhysicalFileSystem.GetPhysicalFileSystem(), Substitute.For<ISemaphoreFactory>(), variables);
-
-            var repository = repositoryFactory.CreateJournalRepository();
-            repository.AddJournalEntry(journalEntry);
-            repository.Commit();
-
-            var updated = repositoryFactory.CreateJournalRepository();
-            updated.TryGetJournalEntry(thePackage, out var journalEntryFromFile).Should().BeTrue();
-            journalEntryFromFile.Package.Should().BeEquivalentTo(thePackage);
-        }
-
-        [Test]
-        public void WhenCalamariPackageRetentionJournalPathDoesNotExist_ThenTheJournalIsCreatedAtTheDefaultPath()
-        {
-            var thePackage = new PackageIdentity("TestPackageWithTentacleHome", "0.0.1");
+            var thePackage = new PackageIdentity("TestPackage", "0.0.1");
             var journalEntry = new JournalEntry(thePackage);
-
-            var variables = Substitute.For<IVariables>();
-            variables.Get(KnownVariables.Calamari.PackageRetentionJournalPath).Returns((string) null);
-            variables.Get(TentacleVariables.Agent.TentacleHome).Returns(tentacleHome);
 
             var repositoryFactory = new JsonJournalRepositoryFactory(TestCalamariPhysicalFileSystem.GetPhysicalFileSystem(), Substitute.For<ISemaphoreFactory>(), variables);
 
--- a/source/Calamari.Tests/Fixtures/PackageRetention/JsonJournalRepositoryFixture.cs
+++ b/source/Calamari.Tests/Fixtures/PackageRetention/JsonJournalRepositoryFixture.cs
@@ -1,0 +1,63 @@
+﻿using System;
+using System.IO;
+using Calamari.Common.Features.Processes.Semaphores;
+using Calamari.Common.Plumbing.Variables;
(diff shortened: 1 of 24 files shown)


source/Calamari/Deployment/PackageRetention/Repositories/JsonJournalRepository.cs
```diff
@@ -16,16 +17,24 @@ public class JsonJournalRepository
     {
         Dictionary<PackageIdentity, JournalEntry> journalEntries;
         const string SemaphoreName = "Octopus.Calamari.PackageJournal";
+        internal const string DefaultJournalName = "PackageRetentionJournal.json";
```
MissedTheMark, line 20: Does this need to be exposed outside this class as `internal`? Can we make this `private` like the other variables?

source/Calamari/Deployment/PackageRetention/Repositories/JsonJournalRepository.cs
```diff
@@ -16,16 +17,24 @@ public class JsonJournalRepository
...
+        public JsonJournalRepository(ICalamariFileSystem fileSystem, ISemaphoreFactory semaphoreFactory, IVariables variables)
         {
             this.fileSystem = fileSystem;
             this.semaphoreFactory = semaphoreFactory;
-            this.journalPath = @"C:\Octopus\PackageJournal.json";//journalPath;
+
+            var packageRetentionJournalPath = variables.Get(KnownVariables.Calamari.PackageRetentionJournalPath);
+            if (packageRetentionJournalPath == null)
+            {
+                var tentacleHome = variables.Get(TentacleVariables.Agent.TentacleHome) ?? throw new Exception("Environment variable 'TentacleHome' has not been set.");
```
MissedTheMark, line 34: I think throwing is reasonable here 👍 If we don't know what the tentacle home path is, then I imagine we're probably in trouble...

source/Calamari.Tests/Fixtures/PackageRetention/JsonJournalRepositoryFixture.cs
```diff
@@ -0,0 +1,63 @@
...
+using Calamari.Tests.Helpers;
+using NSubstitute;
+using NUnit.Framework;
+
+namespace Calamari.Tests.Fixtures.PackageRetention
+{
+    [TestFixture]
+    public class JsonJournalRepositoryFixture
+    {
+        static readonly string TentacleHome = TestEnvironment.GetTestPath("Fixtures", "JsonJournalRepository");
```
MissedTheMark, line 16: Does this need to be static?

source/Calamari.Tests/Fixtures/PackageRetention/JsonJournalRepositoryFixture.cs
```diff
@@ -0,0 +1,63 @@
...
+                Directory.Delete(TentacleHome, true);
+        }
+
+        [Test]
+        public void WhenCalamariPackageRetentionJournalPathExists_ThenTheJournalIsCreatedAtTheGivenPath()
+        {
+            var journalPath = Path.Combine(TentacleHome, "PackageRetentionJournal.json");
+
+            var variables = Substitute.For<IVariables>();
+            variables.Get(KnownVariables.Calamari.PackageRetentionJournalPath).Returns(journalPath);
```
MissedTheMark, line 38: Nice 👍

MissedTheMark (changes requested):
Just two minor things, but looks good generally!

source/Calamari/Deployment/PackageRetention/Repositories/JsonJournalRepository.cs
N-lson: Rider says it can be made private but I use it in the test class which needs it to be exposed 🤔 Otherwise we'd have to duplicate the journal name in the test class.

source/Calamari.Tests/Fixtures/PackageRetention/JsonJournalRepositoryFixture.cs
N-lson: Nope 👍

source/Calamari.Tests/Fixtures/PackageRetention/JsonJournalRepositoryFixture.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        [TearDown]
+        public void TearDown()
+        {
+            if (Directory.Exists(TentacleHome))
+                Directory.Delete(TentacleHome, true);
+        }
+
+        [Test]
+        public void WhenCalamariPackageRetentionJournalPathExists_ThenTheJournalIsCreatedAtTheGivenPath()
```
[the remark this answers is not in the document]
MissedTheMark: I think these tests should actually be revised. We don't really care _how_ the repository implements things, just that it does what it should.   So if we pass the variables like you have, then write something and read it back correctly, then we know that it works, without worrying about the implementation.

source/Calamari/Deployment/PackageRetention/Repositories/JsonJournalRepository.cs
MissedTheMark: See my point above with the tests.

MissedTheMark (changes requested):

source/Calamari.Tests/Fixtures/PackageRetention/JsonJournalRepositoryFixture.cs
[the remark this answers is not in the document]
N-lson: That makes sense. I've updated the test to load, write something, then re-load the journal to check that what's there is what we expect. Also removed the `internal` from that default journal name since we no longer need it for the test.

source/Calamari/Deployment/PackageRetention/Repositories/JsonJournalRepositoryFactory.cs
```diff
@@ -1,23 +1,31 @@
...
         {
             this.fileSystem = fileSystem;
             this.semaphoreFactory = semaphoreFactory;
 
-            this.journalPath = @"C:\Octopus\PackageJournal.json";//journalPath;
+            var packageRetentionJournalPath = variables.Get(KnownVariables.Calamari.PackageRetentionJournalPath);
+            if (packageRetentionJournalPath == null)
+            {
+                var tentacleHome = variables.Get(TentacleVariables.Agent.TentacleHome) ?? string.Empty; // Retention is only enabled when TentacleHome exists
+                packageRetentionJournalPath = Path.Combine(tentacleHome, DefaultJournalName);
```
N-lson, line 26: `Path.Combine` requires a non-null string so if we don't set `tentacleHome` to something here (at the moment I went with an empty string) then it would fail in the integration tests again.

source/Calamari/Commands/FindPackageCommand.cs
```diff
@@ -78,7 +78,8 @@ public override int Execute(string[] commandLineArguments)
             //Exact package found, so we need to register use and lock it.
             // This command can't use the PackageJournalCommandDecorator because we don't lock on partial finds, which the decorator would include.  We don't lock on partials because there may be too many packages, blocking retention later,
             //  and we will lock them on apply delta anyway.
-            if (variables.IsPackageRetentionEnabled()) packageJournal.RegisterPackageUse(variables);
+            var tentacleHome = Environment.GetEnvironmentVariable("TentacleHome");
+            if (variables.IsPackageRetentionEnabled() && tentacleHome != null) packageJournal.RegisterPackageUse(variables);
```
N-lson, line 82: Do we also need to add the `tentacleHome` check to `PackageJournalCommandDecorator`?

source/Calamari.Tests/Fixtures/PackageRetention/JsonJournalRepositoryFixture.cs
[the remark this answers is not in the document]
MissedTheMark: I think these are a lot better.   Given that these are almost completely the same code-wise, it could be worth using [test cases](https://docs.nunit.org/articles/nunit/writing-tests/attributes/testcase.html) for the differences, and specify the name with the TestName parameter on the test case attributes?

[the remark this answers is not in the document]
N-lson: Oh yeah true 😄 looks like a good use case for that.

source/Calamari/Commands/FindPackageCommand.cs
N-lson: I think that sounds better

N-lson: I went with a public method which includes `variables.IsPackageRetentionEnabled() && tentacleHome != null`. The checks are still in the same places (`PackageJournalCommandDecorator` and `FindPackageCommand`) but they're not concerned with what the check is doing anymore.

source/Calamari.Tests/Fixtures/PackageRetention/JsonJournalRepositoryFixture.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        [TearDown]
+        public void TearDown()
+        {
+            if (Directory.Exists(testDir))
+                Directory.Delete(testDir, true);
+        }
+
+        [TestCase("PackageRetentionJournal.json")]
+        [TestCase(null)]
```
MissedTheMark, line 36: I think it would be good to use the TestName or Description parameters of the TestCase attribute here, just so we know what we're actually testing here with these values.

N-lson: Oh I completely missed that from the page you linked 🤦 yeah that would be nice haha

N-lson: Added a TestName for both

source/Calamari/Commands/FindPackageCommand.cs
MissedTheMark: I'm a bit concerned about the commands having to worry about whether or not retention is enabled - We can pass the variables to the Journal (even though that creates coupling between the journal and the variables collection coming from server), and set a "IsRetentionEnabled" property on the journal, and check that at the start of the registerPackageUse() etc variables.

This also means the journal knows about the package info (because of the variables collection), and we can modify the RegisterPackageUse() methods to not take in any parameters - though that part should be in another PR....

N-lson: I agree with you on the commands still having to check for retention being enabled. Looking at it again I'm not completely happy with how it is. Adding the variables to the journal sounds like a nicer alternative. For now I can make those changes and just leave the parameters of `RegisterPackageUse` how they are for the time being.

MissedTheMark (approved):

PR #789, opened 2021-11-11
Handle concurrent usages of the package retention journal

N-lson:
This PR re-adds the repository factory which is used to restrict access to `JsonJournalRepository` which is where the reads/writes to the journal actually occur. This implementation may change in the future but we're going with this as an initial way of handling concurrency.

`JsonJournalRepository` now implements `IDisposable` and `Journal` now retrieves a `JsonJournalRepository` via `JsonJournalRepositoryFactory.CreateJournalRepository`. Currently I'm using `Semaphore` from `System.Threading`, wondering if we have another one which we should use instead? `ISemaphoreFactory` only has `Acquire` whereas we need to release the semaphore in `Dispose`.

(This PR is based off the file path PR branch again)


source/Calamari.Tests/Fixtures/PackageRetention/JournalFixture.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            if (Directory.Exists(TentacleHome))
+                Directory.Delete(TentacleHome, true);
+        }
+
+        [Test]
+        public void WhenMultipleThreadsTryToWriteToTheJournalAtTheSameTime_ThenTheWritesAreHandledCorrectly()
+        {
+            //TODO: Do this using Autofac
+            TypeDescriptor.AddAttributes(typeof(ServerTaskId), new TypeConverterAttribute(typeof(TinyTypeTypeConverter<ServerTaskId>)));
```
N-lson, line 43: The test was failing on converting `ServerTaskId` so I copied this over from `Program` for now.

source/Calamari/Deployment/PackageRetention/Repositories/JsonJournalRepository.cs
```diff
@@ -15,26 +15,22 @@ namespace Calamari.Deployment.PackageRetention.Repositories
     public class JsonJournalRepository
         : IJournalRepository
     {
+        static readonly Semaphore JournalSemaphore = new Semaphore(1, 1);
```
N-lson, line 18: Is there a different class we should use here?

source/Calamari/Deployment/PackageRetention/Repositories/JsonJournalRepository.cs
```diff
@@ -92,5 +85,10 @@ void Save()
                 fileSystem.OverwriteAndDelete(journalPath, tempFilePath);
```
N-lson, line 85: Do we still need the semaphore in `Save`?

MissedTheMark: Nope!

MissedTheMark:
I wonder if this PR should be split into two - one that updates the .Net version and adds convention test capabilities, and one that does the concurrency/factory change?

N-lson:
> 
> 
> I wonder if this PR should be split into two - one that updates the .Net version and adds convention test capabilities, and one that does the concurrency/factory change?

That works

source/Calamari/Deployment/PackageRetention/Model/Journal.cs
```diff
@@ -28,24 +29,28 @@ public void RegisterPackageUse(PackageIdentity package, ServerTaskId serverTaskI
         {
             try
             {
-                if (repository.TryGetJournalEntry(package, out var entry))
+                using (var repository = repositoryFactory.CreateJournalRepository())
```
MissedTheMark, line 32: You can use the alternative `using` syntax here, and save some braces, i.e:
`using var repository = repositoryFactory.CreateJournalRepository();`
The repository will be disposed of at the end of the current block.
See info here: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-statement

N-lson: I'm getting this in Rider:
image

N-lson:
New PR without convention test changes: https://github.com/OctopusDeploy/Calamari/pull/792

source/Calamari/Deployment/PackageRetention/Model/Journal.cs
MissedTheMark: That's weird, Calamari is c# 8.0 project for me.  Feel free to go back to the old syntax, it's no big deal.
image

N-lson: 🤔 Some of the projects like `Calamari.Common` are C# 8.0, no idea why the `Calamari` one is stuck on C# 7.3 for me.

PR #791, opened 2021-11-15
Add convention test capability to Calamari

N-lson:
As part of the package retention changes #team-fire-and-motion are making, we want to add convention test(s) to Calamari. For this we're taking a similar approach to OctopusServer, which uses `Best.Conventional`. However the current `Calamari.Tests` project isn't compatible as it's running `net452`. This PR updates `Calamari.Tests` to `net461` along with moving the current `Fixtures/Conventions` tests into `Fixtures/Deployment/Conventions`. This is done because:

1. This better matches the directory structure in the `Calamari` project which also has `Fixtures/Deployment/Conventions`
2. We can put our actual convention tests in `Fixtures/Conventions` which aligns with OctopusServer's structure

# How to review
The main change in this PR is in `Calamari.Tests.csproj`. All `.cs` files _should_ just be namespace changes with the exception of test fixtures which have been excluded from running on Mono <5.12.0 due to HttpClient issues (I've commented these on the PR to make them easier to find amongst the mass of namespace changes).

This PR also adds a `global.json` file which is required to build the solution on our TeamCity agents using .NET SDK 5 instead of .NET SDK 3.1 which leads to incompatibility issues with `net461`.

To go along with this PR we have a docs change to update our earliest recommended version of Mono https://github.com/OctopusDeploy/docs/pull/1390


source/Calamari.Tests/Fixtures/PackageDownload/PackageDownloadFixture.cs
```diff
@@ -142,6 +142,7 @@ public void ShouldDownloadMavenSnapshotPackage()
 
         [Test]
         [RequiresMonoVersion480OrAboveForTls12]
+        [RequiresMinimumMonoVersion(5, 12, 0, Description = "HttpClient 4.3.2 broken on Mono - https://xamarin.github.io/bugzilla-archives/60/60315/bug.html#c7")]
         public void ShouldDownloadPackageWithRepositoryMetadata()
```
N-lson, line 146: This test now requires Mono >= 5.12.0

source/Calamari.Tests/Fixtures/Integration/Packages/NuGetPackageDownloaderFixture.cs
```diff
@@ -91,6 +93,7 @@ public void IgnoresTheTimeoutIfAnInvalidTimeoutIsDefinedInVariables()
         [Test]
         [NonParallelizable]
         [RequiresNonFreeBSDPlatform(SkipFreeBsdBecause)]
+        [RequiresMinimumMonoVersion(5, 12, 0, Description = "HttpClient 4.3.2 broken on Mono - https://xamarin.github.io/bugzilla-archives/60/60315/bug.html#c7")]
         public void DoesNotTimeOutIfTheServerRespondsBeforeTheTimeout()
```
N-lson, line 97: This test now requires Mono >= 5.12.0

source/Calamari.Tests/Fixtures/Integration/Packages/NuGetPackageDownloaderFixture.cs
```diff
@@ -83,6 +84,7 @@ public void TimesOutIfAValidTimeoutIsDefinedInVariables()
         [Test]
         [NonParallelizable]
         [RequiresNonFreeBSDPlatform(SkipFreeBsdBecause)]
+        [RequiresMinimumMonoVersion(5, 12, 0, Description = "HttpClient 4.3.2 broken on Mono - https://xamarin.github.io/bugzilla-archives/60/60315/bug.html#c7")]
         public void IgnoresTheTimeoutIfAnInvalidTimeoutIsDefinedInVariables()
```
N-lson, line 88: This test now requires Mono >= 5.12.0

source/Calamari.Tests/Fixtures/Integration/Packages/NuGetPackageDownloaderFixture.cs
```diff
@@ -75,6 +75,7 @@ public int AttemptsTheRightNumberOfTimesOnError(int maxDownloadAttempts)
         [Test]
         [NonParallelizable]
         [RequiresNonFreeBSDPlatform(SkipFreeBsdBecause)]
+        [RequiresMinimumMonoVersion(5, 12, 0, Description = "HttpClient 4.3.2 broken on Mono - https://xamarin.github.io/bugzilla-archives/60/60315/bug.html#c7")]
         public void TimesOutIfAValidTimeoutIsDefinedInVariables()
```
N-lson, line 79: This test now requires Mono >= 5.12.0

global.json
```diff
@@ -0,0 +1,6 @@
+{
```
andrewabest, line 1: Do we need this for this PR?

source/Calamari.Tests/Fixtures/Deployment/Conventions/AlreadyInstalledConventionFixture.cs
```diff
@@ -8,7 +8,7 @@
 using NSubstitute;
 using NUnit.Framework;
 
-namespace Calamari.Tests.Fixtures.Conventions
+namespace Calamari.Tests.Fixtures.Deployment.Conventions
```
andrewabest, line 11: This namespace shuffle has added quite a lot of churn / noise to the PR. Perhaps we could separate this into two PRs: one that introduces the framework changes, and another that then does the "tidy up" of namespace reshuffling? It will make them easier to understand later on if people need to return to them.

andrewabest (changes requested):
I've made a suggestion to reorganise this work into two PRs so that they are a bit smaller / more focussed. When that is done I think they will both be good to go :)

global.json
N-lson: Hmm we _did_ need this when jumping to `net480` since the 3.1 SDK wasn't compatible with building the tests. But for `net461` I'm not certain. I can check when I split the PR up.

source/Calamari.Tests/Fixtures/Deployment/Conventions/AlreadyInstalledConventionFixture.cs
N-lson: That sounds good to me

global.json
N-lson: Update: The tests built fine without this change on `net461`, so looks like no it's not needed.

N-lson:
The changes have been split into the above PRs

PR #792, opened 2021-11-16
Handle concurrent usages of the package retention journal

N-lson:
This PR re-adds the repository factory which is used to restrict access to `JsonJournalRepository` which is where the reads/writes to the journal actually occur. This implementation may change in the future but we're going with this as an initial way of handling concurrency.

`JsonJournalRepository` now implements `IDisposable` and `Journal` now retrieves a `JsonJournalRepository` via `JsonJournalRepositoryFactory.CreateJournalRepository`.

This PR doesn't include the convention test changes, and also doesn't include `JournalFixture` (I was having some issues with this on TeamCity so need to look further into it).

### DIFF
--- a/source/Calamari/Deployment/PackageRetention/Repositories/JsonJournalRepository.cs
+++ b/source/Calamari/Deployment/PackageRetention/Repositories/JsonJournalRepository.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Text;
+using System.Threading;
 using Calamari.Common.Features.Processes.Semaphores;
 using Calamari.Common.Plumbing.Deployment.PackageRetention;
 using Calamari.Common.Plumbing.FileSystem;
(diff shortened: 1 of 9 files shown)


source/Calamari/Deployment/PackageRetention/Repositories/JsonJournalRepository.cs
```diff
@@ -14,18 +15,19 @@ namespace Calamari.Deployment.PackageRetention.Repositories
     public class JsonJournalRepository
         : IJournalRepository
     {
+        static readonly SemaphoreSlim JournalSemaphore = new SemaphoreSlim(1, 1);
```
MissedTheMark, line 18: We need to use a named semaphore here. `SemaphoreSlim` is only for a single process (including multiple threads), whereas we need this across multiple processes -  Calamari can be run multiple times simultaneously on a single machine.
Use ISemaphoreFactory, because the implementation uses a file-base semaphore when running on mono, and a system one for the rest.

MissedTheMark (changes requested):

MissedTheMark:
For anyone looking at this in the future, we're ok not including the tests for now because this isn't actually enabled yet.

source/Calamari/Deployment/PackageRetention/Repositories/JsonJournalRepository.cs
N-lson: Ahh right that makes sense. I added `ISemaphoreFactory` back in, does that look like the right way to dispose of the semaphore?

source/Calamari/Deployment/PackageRetention/Repositories/JsonJournalRepository.cs
```diff
@@ -14,18 +15,19 @@ namespace Calamari.Deployment.PackageRetention.Repositories
...
+        readonly IDisposable semaphore;
 
-        public JsonJournalRepository(ICalamariFileSystem fileSystem, ISemaphoreFactory semaphoreFactory)
+        public JsonJournalRepository(ICalamariFileSystem fileSystem, ISemaphoreFactory semaphoreFactory, string journalPath)
         {
             this.fileSystem = fileSystem;
-            this.semaphoreFactory = semaphoreFactory;
-            this.journalPath = @"C:\Octopus\PackageJournal.json";//journalPath;
+            this.journalPath = journalPath;
+            this.semaphore = semaphoreFactory.Acquire(SemaphoreName, "Another process is using the package retention journal");
```
MissedTheMark, line 30: I was going to suggest that we should move the semaphore factory to the JsonJournalRespository, but given that this repository disposes the semaphore object, I think this is the correct implementation 👍

MissedTheMark (approved):
LGTM 👍

PR #794, opened 2021-11-17
Update Calamari.Tests to net461

N-lson:
As part of the package retention changes #team-fire-and-motion are making, we want to add convention test(s) to Calamari. For this we're taking a similar approach to OctopusServer, which uses `Best.Conventional`. However the current `Calamari.Tests` project isn't compatible as it's running `net452`. This PR updates `Calamari.Tests` to `net461`.

## How to review
The main change in this PR is in `Calamari.Tests.csproj`. All `.cs` files are test fixture changes which have been excluded from running on Mono <5.12.0 due to HttpClient issues

(Old PR which this has been split from: https://github.com/OctopusDeploy/Calamari/pull/791)

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -191,7 +191,7 @@ Task("PackTests")
 {
     var actions = new List<Action>();
 
-    actions.Add(() => Zip("./source/Calamari.Tests/bin/Release/net452/", Path.Combine(artifactsDir, "Binaries.zip")));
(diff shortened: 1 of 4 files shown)


andrewabest (approved):
Looks good @N-lson - thanks for breaking the PRs down, it makes it much easier to understand the core change :)

PR #795, opened 2021-11-17
Move Fixtures/Conventions to Fixtures/Deployment/Conventions

N-lson:
The second half of https://github.com/OctopusDeploy/Calamari/pull/791

This PR moves the current tests in `Fixtures/Conventions` to `Fixtures/Deployment/Conventions` so that we can put convention tests related to the code itself in `Fixtures/Conventions` like OctopusServer.


source/Calamari.Tests/Calamari.Tests.csproj
```diff
@@ -163,6 +163,75 @@
     <None Update="AWS\S3\Package3\file.json">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
+    <None Update="Fixtures\Deployment\Conventions\ConfigTransforms\config\fizz.buzz.config">
```
andrewabest, line 166: Something like `Fixtures\**\ConfigTransforms\**` might work better here, as it will mean it will scoop up all of these under a single entry (and capture new ones) as opposed to having to list each fo them independently

andrewabest (approved):
I've made a small suggestion, but looks good aside from that 👍

source/Calamari.Tests/Calamari.Tests.csproj
N-lson: Argh.. something that Rider generated. Agreed it should be like that.

PR #797, opened 2021-11-21
Don't delete unextracted package files as part of lifecycle retention

N-lson:
## Background
Since #team-fire-and-motion are adding package retention to workers we shouldn't need this functionality when the lifecycle retention policy gets applied as it will be moved to package retention. However, we do still want to clean up the `ExtractedTo` directory where appropriate. This PR removes the functionality which removes package files stored on a tentacle/worker (ie. the package files now remain on the machine).

## How to review

1. Start Octopus server
2. Create a Tentacle using the latest `main` branch from `OctopusTentacle` (or any version really)
3. Set up an external package feed `Library -> External Feeds`. I use `https://www.nuget.org/api/v2`.
4. Add a "Deploy a package" step, for the package specify anything with multiple versions. I used `Newtonsoft.Json`.
5. Set the retention policy of the default lifecycle to keep 1 release.
6. Create a release - for this release pick an older version of the package eg. `12.0.1`. Deploy the release.
7. Create another release - for this release pick the next version up eg. `12.0.2`. Deploy the release.
8. Create another release - for this release pick the next version up eg. `12.0.3`. Deploy the release. This release should contain a log that looks like the following:
image
  Note this is using an old version of Calamari. As you can see it removes both the `ExtractedTo` at `C:\Octopus\Applications\<tentaclename>\Development\Newtonsoft.Json\12.0.1` and the `DeployedFrom` at `C:\Octopus\<tentaclename>\Files\Newtonsoft.Json@S12.0.1...`.
9. Pull this branch and build Calamari
10. Run Calamari in your IDE with the following arguments and environment variables:
```
TentacleJournal=C:\Octopus\<tentaclename>\DeploymentJournal.xml
TentacleName=C:\Octopus\<tentaclename>
clean --retentionPolicySet "Environments-1/Projects-1/Step-Deploy a Package/Machines-61/<default>" --releases 1
```
You can find the `retentionPolicySet` in `DeploymentJournal.xml` at the home directory of your machine. It should look something like `Environments-1/Projects-1/Step-Deploy a Package/Machines-1/<default>`.
11. The log should only say that it removed the `ExtractedTo` contents with no mention of the `DeployedFrom`. Check the `Files` directory of your tentacle and `Newtonsoft.Json@S12.0.2` should still be there.

The tests have been changed as well to account for the fact that packages aren't being deleted anymore. I think this makes sense for `RetentionPolicyFixture` since removing packages won't be its responsibility anymore. Not as sure about `CleanFixture` though but the same logic applies - it won't be dealing with packages anymore.

### DIFF
--- a/source/Calamari.Tests/Fixtures/Retention/RetentionPolicyFixture.cs
+++ b/source/Calamari.Tests/Fixtures/Retention/RetentionPolicyFixture.cs
@@ -215,12 +215,10 @@ namespace Calamari.Tests.Fixtures.Retention
 
             // The older artifacts should have been removed
             fileSystem.Received().DeleteDirectory(fourDayOldSuccessfulDeployment.ExtractedTo);
-            fileSystem.Received().DeleteFile(fourDayOldSuccessfulDeployment.Package.DeployedFrom, Arg.Any<FailureOptions>());
 
             // The newer artifacts, and those from the non-matching policy-set, should have been kept
             // In other words, nothing but the matching deployment should have been removed
             fileSystem.DidNotReceive().DeleteDirectory(Arg.Is<string>(s => !s.Equals(fourDayOldSuccessfulDeployment.ExtractedTo)));
-            fileSystem.DidNotReceive().DeleteFile(Arg.Is<string>(s => !s.Equals(fourDayOldSuccessfulDeployment.Package.DeployedFrom)), Arg.Any<FailureOptions>());
 
             // The older entry should have been removed from the journal
             deploymentJournal.Received().RemoveJournalEntries(Arg.Is<IEnumerable<string>>(ids => ids.Count() == 1 && ids.Contains(fourDayOldSuccessfulDeployment.Id)));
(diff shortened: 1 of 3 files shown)


source/Calamari.Tests/Fixtures/Retention/RetentionPolicyFixture.cs
```diff
@@ -307,23 +302,6 @@ public void ShouldNotDeleteAnythingWhenSpecifiedNumberOfSuccessfulDeploymentsNot
             deploymentJournal.Received().RemoveJournalEntries(Arg.Is<IEnumerable<string>>(ids => !ids.Any()));
         }
 
-        [Test]
-        public void ShouldApplyRetentionToMultiplePackages()
```
N-lson, line 311: The purpose of this test was to check that if the same package was in separate deployments, that it wouldn't be deleted as part of one deployment getting cleaned up. Since lifecycle retention policy should no longer be dealing with packages at all I've removed this test. The last part where it checks for an older entry being removed is already part of some of the above tests.

tothegills:
When do we want to remove the existing retention logic?

N-lson:
Not sure, maybe this one should be merged when everything else package retention related has been merged? @MissedTheMark

MissedTheMark:
Yeah, definitely not yet 😅 
We'll hold off on this until the rest is done.

ankithkonda:
@N-lson does this still need reviewing?

N-lson:
Not sure. @tothegills do you know?

tothegills (approved):

PR #803, opened 2021-12-01
Rename package retention journal if contents are invalid

N-lson:
Updated from https://github.com/OctopusDeploy/Calamari/pull/787 copy-pasted description:

This PR adds a check when reading from the journal file on whether the contents are valid json or not. If not, the old file is renamed to `<OLDNAME>_<DATETIME>.json` where `<DATETIME>` is the current datetime.

## How to test

1. Start Octopus server
2. Add a Tentacle
4. Set up project variables to debug Calamari (Mark's [guide here](https://trello.com/c/jS95BzZ3/10-how-do-i-debug-calamari))
7. Open the file location of the tentacle and locate `PackageRetentionJournal.json`
8. Edit it to be invalid json eg. remove the closing `]`
9. Re-deploy the release w/ attaching Calamari
10. A backup `.json` file should be created and the `PackageRetentionJournal.json` will be restored to valid json. The log should look something like this:
image

### DIFF
--- a/source/Calamari.Tests/Fixtures/PackageRetention/JsonJournalRepositoryFixture.cs
+++ b/source/Calamari.Tests/Fixtures/PackageRetention/JsonJournalRepositoryFixture.cs
@@ -2,6 +2,7 @@
 using System.IO;
 using Calamari.Common.Features.Processes.Semaphores;
 using Calamari.Common.Plumbing.Deployment.PackageRetention;
+using Calamari.Common.Plumbing.Logging;
 using Calamari.Common.Plumbing.Variables;
 using Calamari.Deployment.PackageRetention.Model;
 using Calamari.Deployment.PackageRetention.Repositories;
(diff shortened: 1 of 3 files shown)


source/Calamari.Tests/Fixtures/PackageRetention/JsonJournalRepositoryFixture.cs
```diff
@@ -38,14 +39,14 @@ public void WhenTheJournalIsLoadedAndCommittedTo_ThenTheJournalContainsTheCorrec
         {
             var journalPath = packageRetentionJournalPath == null ? null : Path.Combine(testDir, packageRetentionJournalPath);
 
-            var variables = Substitute.For<IVariables>();
-            variables.Get(KnownVariables.Calamari.PackageRetentionJournalPath).Returns(journalPath);
-            variables.Get(TentacleVariables.Agent.TentacleHome).Returns(testDir);
+            var variables = new CalamariVariables();
+            variables.Set(KnownVariables.Calamari.PackageRetentionJournalPath, journalPath);
+            variables.Set(TentacleVariables.Agent.TentacleHome, testDir);
```
N-lson, line 44: I decided to use `CalamariVariables` instead of a mock since this is what the other tests do. It's also more realistic since it's a proper variable bag rather than a fake one.

source/Calamari/Deployment/PackageRetention/Repositories/JsonJournalRepository.cs
```diff
@@ -58,8 +61,25 @@ void Load()
...
+                    journalEntries = journalContents.ToDictionary(entry => entry.Package, entry => entry);
+                }
+                else
+                {
+                    var journalFileName = Path.GetFileNameWithoutExtension(journalPath);
+                    var backupJournalFileName = $"{journalFileName}_{DateTimeOffset.UtcNow:yyyyMMddTHHmmss}.json"; // eg. PackageRetentionJournal_20210101T120000.json
+
+                    log.Warn($"The existing package retention journal file {journalPath} could not be read. The file will be renamed to {backupJournalFileName}. A new journal will be created.");
+
+                    var backupJournalPath = Path.GetDirectoryName(journalPath) + Path.DirectorySeparatorChar + backupJournalFileName;
```
N-lson, line 76: This is updated since I found that the invalid journal was being copied to the directory where the exe was running which seems not very ideal.

MissedTheMark: Ooooh good pick up.

Use Path.Combine to join these though.

N-lson: That's much nicer!

source/Calamari/Deployment/PackageRetention/Repositories/JsonJournalRepository.cs
```diff
@@ -84,5 +104,20 @@ public void Dispose()
         {
             semaphore.Dispose();
         }
+
+        static bool TryParseJournal(string json, out List<JournalEntry> result)
```
MissedTheMark, line 107: I think this would be more accurately named if we called it `TryDeserializeJournalEntries` (just based off the JsonConvert method call)
You'll also need to change log message too.

MissedTheMark (changes requested):
Minor changes, see my comments.

MissedTheMark (approved):
LGTM 👍

PR #804, opened 2021-12-06
Add package cache replacement/retention process

desruc:
This PR add the package retention process.

- `ApplyRetention` function added to the package journal 
  - Deletes package files if there is less than 500mb available in the given directory
  - Removes packages from journal
- `ApplyRetention` runs whenever a Download or Apply Delta command is executed
  - This replaces `EnsureDiskHasEnoughFreeSpace` (which just threw an error if there wasn't enough space)
- `OctopusPackageRetentionCacheSizeInMegaBytes` variable support added to allow for a maximum cache size to be specified.

Previously, all package retention was handled by the `Clean` command. This is removed in this [PR](https://github.com/OctopusDeploy/Calamari/pull/797).

### DIFF
--- a/source/Calamari.Shared/Commands/ApplyDeltaCommand.cs
+++ b/source/Calamari.Shared/Commands/ApplyDeltaCommand.cs
@@ -23,17 +23,15 @@ namespace Calamari.Commands
         string? newFileName;
         bool showProgress;
         bool skipVerification;
-        
-        readonly IFreeSpaceChecker freeSpaceChecker;
+
         readonly ICalamariFileSystem fileSystem;
         readonly ICommandLineRunner commandLineRunner;
         readonly IManagePackageUse packageJournal;
         readonly ILog log;
 
-        public ApplyDeltaCommand(ILog log, IFreeSpaceChecker freeSpaceChecker, ICalamariFileSystem fileSystem, ICommandLineRunner commandLineRunner,
+        public ApplyDeltaCommand(ILog log, ICalamariFileSystem fileSystem, ICommandLineRunner commandLineRunner,
                                  IManagePackageUse packageJournal)
         {
-            this.freeSpaceChecker = freeSpaceChecker;
             this.fileSystem = fileSystem;
             this.commandLineRunner = commandLineRunner;
             this.packageJournal = packageJournal;
--- a/source/Calamari.Shared/Commands/ApplyDeltaCommand.cs
+++ b/source/Calamari.Shared/Commands/ApplyDeltaCommand.cs
@@ -5,6 +5,7 @@ using Calamari.Common.Commands;
 using Calamari.Common.Features.Packages;
 using Calamari.Common.Features.Processes;
 using Calamari.Common.Plumbing;
+using Calamari.Common.Plumbing.Deployment.PackageRetention;
 using Calamari.Common.Plumbing.Extensions;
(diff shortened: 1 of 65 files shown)


desruc:
Tests will fail at the moment. Waiting on finalized retention algorithm and paths to be set on packages.

source/Calamari.Tests/Fixtures/PackageRetention/JournalEntryFixture.cs
```diff
@@ -128,12 +139,30 @@ public void WhenPackageIsRegisteredAndDeregistered_ThenUsageIsStillRecorded()
...
+            var deploymentOne = new ServerTaskId("Deployment-1");
+    
+            var journal = GetPackageJournal();
+            journal.RegisterPackageUse(thePackage, deploymentOne);
+
+            journal.ApplyRetention(PackageDirectory);
+
+            // TODO: Fix these assertions
+            // How do we ensure the given package will be marked for removal?
+            // How do we simulate needing more disk space
```
MissedTheMark, line 162: You can use a real `IFreeSpaceChecker`, with a mock `IFileSystem` passed to it - and make the mock `fileSystem.GetDiskFreeSpace` call return the space you want it to.
Let me know if you want me to walk you through it.

source/Calamari.Tests/Fixtures/PackageRetention/JournalEntryFixture.cs
```diff
@@ -128,12 +139,30 @@ public void WhenPackageIsRegisteredAndDeregistered_ThenUsageIsStillRecorded()
...
+            var thePackage = new PackageIdentity("Package", "1.0");
+            var deploymentOne = new ServerTaskId("Deployment-1");
+    
+            var journal = GetPackageJournal();
+            journal.RegisterPackageUse(thePackage, deploymentOne);
+
+            journal.ApplyRetention(PackageDirectory);
+
+            // TODO: Fix these assertions
+            // How do we ensure the given package will be marked for removal?
```
MissedTheMark, line 161: Mock the retention algorithm to return the package you want to have removed.

MissedTheMark (changes requested):
Couple of things to change, nothing huge.  Added some notes on how to get the free space tests to do what you need 👍

source/Calamari.Shared/Commands/ApplyDeltaCommand.cs
```diff
@@ -54,7 +56,7 @@ public override int Execute(string[] commandLineArguments)
             try
             {
                 ValidateParameters(out basisFilePath, out deltaFilePath, out newFilePath);
-                freeSpaceChecker.EnsureDiskHasEnoughFreeSpace(PackageStore.GetPackagesDirectory());
+                packageJournal.ApplyRetention(PackageStore.GetPackagesDirectory());
```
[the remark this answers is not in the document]
desruc: > Given that we already have a check in `ApplyRetention`, perhaps we should add a dependency on the old FreeSpaceChecker there. Would make the calling code cleaner too if it's inside `ApplyRetention`

I've taken this approach.

`ApplyRetention` calls the old `EnsureDiskHasEnoughFreeSpace` method if package retention is disabled. I've also wrapped it in a try/catch.

MissedTheMark (approved):
👍

PR #805, opened 2021-12-08
Ensure all file format replacers are present for implementors of CalamariFlavourProgram and CalamariFlavourProgramAsync

mjhilton:
# Overview
This PR fixes the root cause of two (and probably more) bugs in the Structured Configuration Variables functionality of our Sashimi packages:

* OctopusDeploy/Issues/issues/6794
* OctopusDeploy/Issues/issues/7177

It is a revival of @andrewabest's original #778. It's now been slightly updated, and successfully tested for both issues on a full local Server chain down into some of the impacted Sashimi.

>As Calamari evolved it underwent a few changes. One was in supporting more file formats for our structured configuration replacement feature. Another was it gained DI, as it originally did not use it.
>
> Originally our non-DI solution for the file format replacers was a simple factory class that returned the set of replacers available: [`b5e6b0d`#diff-5c599a99254cd6918f74120d10a7bf101d40d189e7078cc500d83c6cfd80a07a](https://github.com/OctopusDeploy/Calamari/commit/b5e6b0d20e09449f4ab302e32c3d5dd8b6a8dba2#diff-5c599a99254cd6918f74120d10a7bf101d40d189e7078cc500d83c6cfd80a07a)
>
> This eventually grew to support four replacers - json, xml, yaml, and properties formats.
>
> When the modularity refactor was undertaken, it seems that we didn't register all of the available formats. My suspicion is that we copied the available formats from the factory, and then some work was done to implement new formats that added them to the factory but didn't add DI registrations for them. Which now leads to some steps having all format replacers available for this feature, and some only having json and yaml replacers available.
>
> This PR corrects this to ensure the same, central set of registered format replacers is used everywhere. This change will need to be propagated through Sashimi and release branches as appropriate.

# Reproduction/testing
This is a `Calamari.Common` change which will need to be propagated down the whole Calamari/Sashimi graph, so this PR can't be tested on its own (unless you feel like building and updating the packages all locally from this branch, which trust me, you don't)


geofflamrock (approved):
Changes look good, I like the additional logging, that will help anyone trying to understand what replacement is happening 👍

PR #807, opened 2021-12-20
Fixed null cache age issue

MissedTheMark:
In certain circumstances, the package cache age could end up being null.

This PR fixes that.  There was also an issue with JSON deserialization of the cache age too.


desruc (approved):
LGTM 👍

PR #816, opened 2022-01-19
Changed all retention log levels to be info or lower, and added extra check for IsRetentionEnabled 

MissedTheMark:
A call to Journal.RegisterPackageUse(out bool packageRegistered) wasn't protected with a call to IsRetentionEnabled().  This PR adds that call.

I've also changed retention log levels to be info or lower - they don't need to be higher than that at the moment.

A new try/catch was added to the apply delta command, where it was maybe potentially _possible_ for an exception to be thrown by retention.

### DIFF
--- a/source/Calamari.Shared/Commands/ApplyDeltaCommand.cs
+++ b/source/Calamari.Shared/Commands/ApplyDeltaCommand.cs
@@ -56,7 +56,14 @@ namespace Calamari.Commands
             try
             {
                 ValidateParameters(out basisFilePath, out deltaFilePath, out newFilePath);
-                packageJournal.ApplyRetention(PackageStore.GetPackagesDirectory());
+                try
+                {
+                    packageJournal.ApplyRetention(PackageStore.GetPackagesDirectory());
+                }
+                catch (Exception ex)
+                {
+                    log.Info($"Unable to apply retention to package.  Error message: {ex.ToString()}");
+                }
 
                 var tempNewFilePath = newFilePath + ".partial";
 #if USE_OCTODIFF_EXE
--- a/source/Calamari/Deployment/PackageRetention/Model/Journal.cs
+++ b/source/Calamari/Deployment/PackageRetention/Model/Journal.cs
@@ -44,6 +44,12 @@ namespace Calamari.Deployment.PackageRetention.Model
 
         public void RegisterPackageUse(out bool packageRegistered)
         {
+            if (!IsRetentionEnabled())
+            {
+                packageRegistered = false;
+                return;
+            }
+
             try
             {
                 RegisterPackageUse(new PackageIdentity(variables), new ServerTaskId(variables), out packageRegistered);
@@ -51,7 +57,7 @@ namespace Calamari.Deployment.PackageRetention.Model
             catch (Exception ex)
             {
                 packageRegistered = false;
-                log.Error($"Unable to register package use for retention.{Environment.NewLine}{ex.ToString()}");
+                log.Info($"Unable to register package use for retention.{Environment.NewLine}{ex.ToString()}");
             }
         }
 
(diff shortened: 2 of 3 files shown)


source/Calamari.Shared/Commands/ApplyDeltaCommand.cs
```diff
@@ -56,7 +56,14 @@ public override int Execute(string[] commandLineArguments)
             try
             {
                 ValidateParameters(out basisFilePath, out deltaFilePath, out newFilePath);
-                packageJournal.ApplyRetention(PackageStore.GetPackagesDirectory());
+                try
+                {
+                    packageJournal.ApplyRetention(PackageStore.GetPackagesDirectory());
+                }
+                catch (Exception ex)
```
mjhilton, line 63: I don't understand the reason behind the addition of this `try/catch`. There are other places that `packageJournal.ApplyRetention` is called (eg in the [MavenPackageDownloader](https://github.com/OctopusDeploy/Calamari/blob/master/source/Calamari.Shared/Integration/Packages/Download/MavenPackageDownloader.cs#L137))... would they need a similar guard?

mjhilton:
Overall seems okay, I'm just wondering if there's anything else we can do to get more confidence we haven't missed any other areas we need the `IsRetentionEnabled` check?

mjhilton (commented):
Overall seems okay, I'm just wondering if there's anything else we can do to get more confidence we haven't missed any other areas we need the `IsRetentionEnabled` check?

source/Calamari.Shared/Commands/ApplyDeltaCommand.cs
MissedTheMark: The other usages use a packages directory variable which has already been determined - this is the only one which calls `PackageStore.GetPackagesDirectory()`, which can throw an exception if an environment variable is missing.

mjhilton: Ah okay, it's the `GetPackagesDirectory()` you're guarding rather than `ApplyRetention()`?

I assume you're talking about the `TentacleHome` variable... was that not set in the E2E tests or something?

MissedTheMark: No, that didn't actually cause the failure for the E2E test - this is just me being a bit paranoid in making sure there's _nothing_ in retention that could fail TBH.  It may not be necessary.

The actual failure was caused by `new PackageIdentity(variables)` [here](https://github.com/OctopusDeploy/Calamari/blob/2dcc0c53b32f89f22f5213812166dba645cea387/source/Calamari/Deployment/PackageRetention/Model/Journal.cs#L49).

MissedTheMark:
> I'm just wondering if there's anything else we can do to get more confidence we haven't missed any other areas we need the IsRetentionEnabled check?

@mjhilton - That's a really fair point. I'm not sure, really.  
The only entry points to retention are via the [Journal ](https://github.com/OctopusDeploy/Calamari/blob/2dcc0c53b32f89f22f5213812166dba645cea387/source/Calamari/Deployment/PackageRetention/Model/Journal.cs#L15)(as [IManagePackageUse](https://github.com/OctopusDeploy/Calamari/blob/0406e554938d4155b243aa304bd9bd44ec0d35da/source/Calamari.Common/Plumbing/Deployment/PackageRetention/IManagePackageUse.cs#L9)), which is called by the command decorator, ApplyDeltaCommand, FindPackageCommand, and ReleasePackageLockCommand (which hasn't been merged into server yet).  I've just done a manual check of the methods on the Journal, and there's no other places that call it that I can see.

mjhilton:
> @mjhilton - That's a really fair point. I'm not sure, really. The only entry points to retention are via the [Journal ](https://github.com/OctopusDeploy/Calamari/blob/2dcc0c53b32f89f22f5213812166dba645cea387/source/Calamari/Deployment/PackageRetention/Model/Journal.cs#L15)(as [IManagePackageUse](https://github.com/OctopusDeploy/Calamari/blob/0406e554938d4155b243aa304bd9bd44ec0d35da/source/Calamari.Common/Plumbing/Deployment/PackageRetention/IManagePackageUse.cs#L9)), which is called by the command decorator, ApplyDeltaCommand, FindPackageCommand, and ReleasePackageLockCommand (which hasn't been merged into server yet). I've just done a manual check of the methods on the Journal, and there's no other places that call it that I can see.

Great, thankyou for double checking. At this point I think we trust our safety nets and roll forward, if anything else crops up then we expand the net :)

mjhilton (approved):

PR #818, opened 2022-02-09
Package retention

tothegills:
Relates to https://github.com/OctopusDeploy/Issues/issues/6664
Relates to https://github.com/OctopusDeploy/OctopusDeploy/pull/11515

This PR refactors package retention to make it explicit commands rather than a side-effect of other commands.

In addition to the existing `release-package-lock` command, this PR adds:
- `register-package`, which locks a package and registers a new usage of the package
- `clean-packages`, which runs retention

The Octopus Server orchestrates these commands during deployments:
- `clean-packages` to make enough space for package acquisition
- `register-package` whenever a package has been acquired
- `release-package-lock` when a deployment is done with packages


source/Calamari/Deployment/PackageRetention/Model/PackageJournal.cs
```diff
@@ -0,0 +1,139 @@
...
+            this.fileSystem = fileSystem;
+            this.retentionAlgorithm = retentionAlgorithm;
+            this.semaphoreFactory = semaphoreFactory;
+        }
+
+        public void RegisterPackageUse(PackageIdentity package, ServerTaskId deploymentTaskId, ulong packageSizeBytes)
+        {
+            try
+            {
+                using (AcquireSemaphore())
```
MissedTheMark, line 37: I know it's never really going to be an issue, but I did like the idea of the repository handling the semaphore - if we were to change to (say) using a SQL database-backed repository that handles concurrent operations ok, the semaphore wouldn't be needed.
Bit of a [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) situation though 😀

source/Calamari/Deployment/PackageRetention/Caching/PercentFreeDiskSpacePackageCleaner.cs
```diff
@@ -0,0 +1,60 @@
...
+            {
+                log.Info("Unable to determine disk space. Skipping free space package retention.");
+                return new PackageIdentity[0];
+            }
+
+            var percentFreeDiskSpaceDesired = variables.GetInt32(PackageRetentionPercentFreeDiskSpace) ?? DefaultPercentFreeDiskSpace;
+            var desiredSpaceInBytes = totalNumberOfBytes * (ulong) percentFreeDiskSpaceDesired / 100;
+            if (totalNumberOfFreeBytes > desiredSpaceInBytes)
+            {
+                log.VerboseFormat("Detected enough space for new packages. ({0}/{1})", totalNumberOfBytes.ToFileSizeString(), totalNumberOfFreeBytes.ToFileSizeString());
```
MissedTheMark, line 45: If we're logging that we have enough space, should we log when we don't and need to remove packages? (or did I miss that in another file?)

MissedTheMark (approved):
This is _so_ much cleaner with the server sending the commands for usage and locking.  Definitely like the changes overall, and the improved separation of concerns.  Couple of nit-picks scattered throughout.

I assume that we're going to hope that packages don't get removed between the find package command and their use?

source/Calamari/Deployment/PackageRetention/Model/PackageJournal.cs
tothegills: Where would you put the semaphores in the repository to provide thread-safety? I think to ensure operations are atomic, the semaphore needs to span from the Load() to Commit().

MissedTheMark: The pattern I used was to have the semaphore creation and repo `Load()` occur within the constructor (which was called using the factory `CreateJournalRepository()` method), and the semaphore closed in the `Dispose()` method .
That way the semaphore is transparent, like here: https://github.com/OctopusDeploy/Calamari/blob/e2d4794bda1b919fc1cea84b1b6463d853a28d85/source/Calamari/Deployment/PackageRetention/Model/Journal.cs#L78 and you don't have to worry about calling `Load()` (I guess similarly to how a repo over a DB might work).

MissedTheMark: BTW don't feel like you have to change it or anything - it was more a point of interest than anything else 😃

PR #825, opened 2022-02-16
Ensures structured config replacers are always in correct order

mjhilton:
Back-ports and cleans up the fix for OctopusDeploy/Issues#7371

* Adds an explicit prioritisation mechanism for our registered services that works for Autofac3+
* Adds a test to validate the behaviour


andrewabest (approved):
Looks good!

PR #828, opened 2022-02-16
Merge forward from 2021.3 release

mjhilton:
Merges forward https://github.com/OctopusDeploy/Calamari/pull/825, which addresses https://github.com/OctopusDeploy/Issues/issues/7371, replacing the short term fix we applied in https://github.com/OctopusDeploy/Calamari/pull/820


mattocto (approved):
LGTM

PR #831, opened 2022-02-22
Add extra output variables for S3 Upload step

raynhamdev:
# Background

When uploading an entire package file you can choose to use the filename of the package as the key for the file (and optionally add prefix). The resulting filename is not available as an output variable.

# Solution

When uploading an entire package file the resulting key should be available in an output variable. The following output variables should be added to support this scenario:

- `Package.Key` should return the key to the uploaded file, including any prefix which may have been specified
- `Package.S3Uri` should return the S3 Uri to the uploaded file, of the form `s3://bucket-name/key`
- `Package.Uri` should return the virtual-hosted-style Uri to the uploaded file, of the form `https://bucket-name.s3.region.amazonaws.com/key`
- `Package.Arn` should return the [ARN](https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-arn-format.html) for the uploaded file, of the form `arn:aws:s3:::bucket-name/key` (Note: China (Beijing) region uses `aws-cn` instead of `aws`)
- `Package.ObjectVersion` should return the version of the object, if applicable.

This should cover key scenarios allowing the package to be referenced in subsequent steps.

# Result

When uploading a single file to S3 Bucket, these output variables will be created:
- Octopus.Action[`step-name`].Output.Package.Key: The key to the uploaded file, including any prefix which may have been specified.
- Octopus.Action[`step-name`].Output.Package.S3Uri: The S3 Uri to the uploaded file, of the form `s3://bucket-name/key`                                              
- Octopus.Action[`step-name`].Output.Package.Uri: The virtual-hosted-style Uri to the uploaded file, of the form `https://bucket-name.s3.region.amazonaws.com/key` 
- Octopus.Action[`step-name`].Output.Package.Arn: The ARN for the uploaded file, of the form `arn:aws:s3:::bucket-name/key` 
- Octopus.Action[`step-name`].Output.Package.ObjectVersion: The version of the object, if applicable. (`Not versioned` is returned if the object doesn't have a version)

image

When uploading multiple files to S3 Bucker, these output variables will be created:
- Octopus.Action[`step-name`].Output.Package.Key[`file-name`]: The key to the uploaded file, including any prefix which may have been specified.
- Octopus.Action[`step-name`].Output.Package.S3Uri[`file-name`]: The S3 Uri to the uploaded file, of the form `s3://bucket-name/key`                                              
- Octopus.Action[`step-name`].Output.Package.Uri[`file-name`]: The virtual-hosted-style Uri to the uploaded file, of the form `https://bucket-name.s3.region.amazonaws.com/key` 
- Octopus.Action[`step-name`].Output.Package.Arn[`file-name`]: The ARN for the uploaded file, of the form `arn:aws:s3:::bucket-name/key` 
- Octopus.Action[`step-name`].Output.Package.ObjectVersion[`file-name`]: The version of the object, if applicable. (`Not versioned` is returned if the object doesn't have a version)

screencapture-localhost-5000-octopus-html-2022-02-23-15_03_32

# Testing
- Smoke testing using a script step to print out the output variables after the "Upload to AWS S3 Bucket" step shows that the variables are created correctly (see screenshots above).
- ⚠️  Unit testing: I've have attempted to have unit test coverage. Unfortunately, with the current implementation of the Calamari command, there is no way to access the values of the output variables in unit tests. To enable this, we will need to refactor the command class. Given the time constraint of this cycle, I think we should leave this for now.

### DIFF
--- a/source/Calamari.Aws/Deployment/Conventions/UploadAwsS3Convention.cs
+++ b/source/Calamari.Aws/Deployment/Conventions/UploadAwsS3Convention.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
@@ -115,7 +115,7 @@ namespace Calamari.Aws.Deployment.Conventions
             {
                 (await UploadAll(options, Factory, deployment)).Tee(responses =>
                 {
-                    SetOutputVariables(deployment, responses);
+                    SetOutputVariables(Factory, deployment, responses);
                 });
             }
             catch (AmazonS3Exception exception)
@@ -136,7 +136,7 @@ namespace Calamari.Aws.Deployment.Conventions
             }
         }
 
-        private void SetOutputVariables(RunningDeployment deployment, IEnumerable<S3UploadResult> results)
+        private void SetOutputVariables(Func<AmazonS3Client> factory, RunningDeployment deployment, IEnumerable<S3UploadResult> results)
         {
             log.SetOutputVariableButDoNotAddToVariables(PackageVariables.Output.FileName, Path.GetFileName(deployment.PackageFilePath));
             log.SetOutputVariableButDoNotAddToVariables(PackageVariables.Output.FilePath, deployment.PackageFilePath);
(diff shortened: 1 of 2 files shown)


source/Calamari.Aws/Deployment/Conventions/UploadAwsS3Convention.cs
```diff
@@ -140,11 +140,43 @@ private void SetOutputVariables(RunningDeployment deployment, IEnumerable<S3Uplo
...
+                    //  ARN format: `arn:aws:s3:::bucket-name/key` (Note: China (Beijing region (cn-north-1) uses `aws-cn` instead of `aws`)
+                    log.Info($"Saving object ARN to variable \"Octopus.Action[{actionName}].Output.Package.Arn[{fileName}]\"");
+                    log.SetOutputVariableButDoNotAddToVariables($"Package.Arn[{fileName}]", $"arn:{(awsEnvironmentGeneration.AwsRegion.SystemName.Equals("cn-north-1") ? "aws-cn" : "aws")}:s3:::{result.BucketName}/{result.BucketKey}");
+                    log.Info($"Saving object version id to variable \"Octopus.Action[{actionName}].Output.Package.ObjectVersion[{fileName}]\"");
+                    log.SetOutputVariableButDoNotAddToVariables($"Package.ObjectVersion[{fileName}]", result.Version);
+                    log.Info($"Saving object version id to variable \"Octopus.Action[{actionName}].Output.Files[{result.BucketKey}]\"");
+                    log.SetOutputVariableButDoNotAddToVariables($"Files[{result.BucketKey}]", result.Version);
+                }
```
egorpavlikhin, line 179: Could you join these two blocks? I think they differ only by the filename key you add at the end, so you could always append that postfix, but for the case where there is only one element leave the postfix as an empty string.

source/Calamari.Aws/Deployment/Conventions/UploadAwsS3Convention.cs
```diff
@@ -140,11 +140,43 @@ private void SetOutputVariables(RunningDeployment deployment, IEnumerable<S3Uplo
         {
             log.SetOutputVariableButDoNotAddToVariables(PackageVariables.Output.FileName, Path.GetFileName(deployment.PackageFilePath));
             log.SetOutputVariableButDoNotAddToVariables(PackageVariables.Output.FilePath, deployment.PackageFilePath);
-            foreach (var result in results)
+            var successfulResults = results.Where(z => z.IsSuccess()).ToArray();
+            var actionName = deployment.Variables["Octopus.Action.Name"];
+            if (successfulResults.Length == 1)
```
tekerson, line 145: Should this be based on whether they have selected "upload whole package" or not as opposed to the number of files successfully uploaded? If they've selected to "upload some files" they won't necessarily know at configuration time how many files will match the glob pattern (it could be 1) and so won't know which output variables to use.

egorpavlikhin (approved):

PR #832, opened 2022-03-01
Allow variable substitution when uploading full package to AWS S3

tekerson:
# Background

Users can currently perform Structured Variable Replacement and/or Variable Substitution to be run against the files from their package if they are only uploading specific files from the package. This feature should be available when uploading the entire package.

# Results

The package will now be extracted to allow for variable substitution and repackaged (if required) before being uploaded.

Release_with_extraction_and_variable_replacement

### DIFF
--- a/source/Calamari.Aws/Commands/UploadAwsS3Command.cs
+++ b/source/Calamari.Aws/Commands/UploadAwsS3Command.cs
@@ -70,7 +70,7 @@ namespace Calamari.Aws.Commands
 
             var conventions = new List<IConvention>
             {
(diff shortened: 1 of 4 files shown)


egorpavlikhin:
Could we add a test to verify repackaging and substitution happen correctly?

tekerson:
> Could we add a test to verify repackaging and substitution happen correctly?

Yes. Added a test that verifies variable substitution within the uploaded package.

egorpavlikhin (approved):

PR #833, opened 2022-03-03
Add option to appended package checksum when uploading to AWS S3

raynhamdev:
# Background

Users of AWS Lambda may need to upload their zipped code into an S3 bucket. The uploaded file may then be referenced in a subsequent step (e.g. Cloud Formation) to activate the deployment. There are some limitations to the current step that require working around to use in this way. For CloudFormation style deployments if the CodeUri is unchanged it won’t trigger a re-deployment. If simply replacing a file, the subsequent Cloud Formation step won’t update the reference to the new file.

# Results

- Add an extra option for the filename to allow a hash of the contents of the package to be included in the resulting key. The hash should appear before the extension.
- Hash should be based on the contents of the zip package, not the zip file itself and should be deterministic based on on the contents of the files and not consider modified dates. The hash should be preceded with an `@` character to support current package name parsing.
- The hash must be calculated **_after_** any variable substitution/replacement.

image

image

# Review

Quality


source/Calamari.Aws/Deployment/Conventions/BucketKeyProvider.cs
```diff
@@ -1,26 +1,93 @@
...
+        }
+
+        (string filename, string extension) GetFileNameParts(string fileNameWithExtensions)
+        {
+            var fileNameParts = fileNameWithExtensions.Split('.');
+            var fileName = fileNameParts[0];
+            var extension = string.Empty;
+            for (var i = 1; i < fileNameParts.Length; i++)
+            {
+                if (int.TryParse(fileNameParts[i], out _))
```
egorpavlikhin, line 79: I am not sure if this is correct. Consider filename `helloworld.1.0.0-beta.zip` where version `1.0.0` is not an integer. Perhaps you could first use a built-in method to get the filename (with version) and extension? And we have probably implemented extracting the version somewhere else already.

source/Calamari.Aws/Deployment/Conventions/BucketKeyProvider.cs
```diff
@@ -1,26 +1,93 @@
...
+        string SubstitutePackageHashVariable(string input, string contentHash)
+        {
+            return input.Replace($"#{{Octopus.Action.Package.PackageContentHash}}", contentHash);
+        }
+        
+        public string EncodeBucketKeyForUrl(string bucketKey)
+        {
+            var prefix = Path.GetDirectoryName(bucketKey)?.Replace('\\','/') ?? string.Empty;
+            var (fileName, extension) = GetFileNameParts(Path.GetFileName(bucketKey));
+            return $"{prefix}/{Uri.EscapeDataString(fileName)}{extension}";
```
egorpavlikhin, line 69: If `prefix` is empty the bucket key will be `/{filename}`, if `prefix` is a folder the bucket key will be `folder//{filename}` ?

source/Calamari.Tests/AWS/S3/BucketKeyProviderFixture.cs
```diff
@@ -61,14 +67,136 @@ public void SingleSelectionFilenameBehaviorUsesDefaultWithPrefix()
...
+                                              BucketKeyBehaviour = BucketKeyBehaviourType.Custom,
+                                              BucketKeyPrefix = "test/",
+                                              BucketKey = "something"
+                                          });
 
             result.Should().Be("something");
         }
+
+        [Test]
+        public void EncodeBucketKeyForUrl_ShouldEncodeFileName()
```
egorpavlikhin, line 81: Great to see tests that answer my questions 👍

source/Calamari.Aws/Deployment/Conventions/BucketKeyProvider.cs
raynhamdev: nom it will be `folder/{filename}`. `Path.GetDirectoryName()` return the path without the / at the end.

source/Calamari.Aws/Deployment/Conventions/BucketKeyProvider.cs
raynhamdev: thanks for picking this up. I added a test `GetBucketKey_PackageOptions_ShouldAppendContentHash_WhenFileNameHasVersionNumbersAndReleaseTag` to cover this case.

I tried to use the PackageName parser but for some reason, it doesn't work well with BucketKey. I looked at [the list of extensions that we supported](https://octopus.com/docs/packaging-applications) and see the only multiple-parts extensions are `.tar.something`. Thus, I changed my implementation a bit to handled only `.tar.something` if the extensions may have multiple parts.

egorpavlikhin (approved):

PR #834, opened 2022-03-09
Applying target discovery changes

scottmerchant:
Adds functionality so that Calamari can execute step package discovery scripts.

Using patch created via diff between:
2708dcfcb732b7aa01f7fe2eb9f554e5227ba54d (master)
f5fa34f545f9488525e0793d20e618761d102ea0 (geoffb/step-package-target-discovery)

### DIFF
--- a/source/Calamari.Tests/Fixtures/Manifest/InstructionBuilder.cs
+++ b/source/Calamari.Tests/Fixtures/Manifest/InstructionBuilder.cs
@@ -25,7 +25,7 @@ namespace Calamari.Tests.Fixtures.Manifest
             instructions.Add(
                              new Instruction
                              {
-                                 Launcher = LaunchTools.LaunchTools.Calamari,
+                                 Launcher = Calamari.LaunchTools.LaunchTools.Calamari,
                                  LauncherInstructions = JsonConvert.SerializeObject(new CalamariInstructions
                                                                                     {
                                                                                         Command = commandName,
@@ -41,7 +41,7 @@ namespace Calamari.Tests.Fixtures.Manifest
         {
             instructions.Add(new Instruction
             {
-                Launcher = LaunchTools.LaunchTools.Node,
+                Launcher = Calamari.LaunchTools.LaunchTools.Node,
                 LauncherInstructions = JsonConvert.SerializeObject(new NodeInstructions
                                                                    {
                                                                        BootstrapperPathVariable = nameof(NodeInstructions.BootstrapperPathVariable),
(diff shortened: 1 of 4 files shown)


source/Calamari.Tests/Fixtures/Manifest/InstructionBuilder.cs
```diff
@@ -25,7 +25,7 @@ public InstructionBuilder WithCalamariInstruction(string commandName, object inp
             instructions.Add(
                              new Instruction
                              {
-                                 Launcher = LaunchTools.LaunchTools.Calamari,
+                                 Launcher = Calamari.LaunchTools.LaunchTools.Calamari,
```
andrewabest, line 28: Why was this change required?

andrewabest, line 28: Why was this change required?

source/Calamari/LaunchTools/NodeExecutor.cs
```diff
@@ -34,63 +35,70 @@ public NodeExecutor(CommonOptions options, IVariables variables, ICommandLineRun
...
+            var pathToBootstrapperFolder = variables.Get(instructions.BootstrapperPathVariable);
+            var pathToBootstrapper = Path.Combine(pathToBootstrapperFolder, "bootstrapper.js");
+            parameters.Add(pathToBootstrapper);
+            parameters.Add(instructions.BootstrapperInvocationCommand);
+            var pathToStepPackage = variables.Get(instructions.TargetPathVariable);
+            parameters.Add(pathToStepPackage);
+            parameters.Add(sensitiveVariablesFilePath);
+            parameters.Add(options.InputVariables.SensitiveVariablesPassword);
+            parameters.Add(AesEncryption.SaltRaw);
+            if (instructions.BootstrapperInvocationCommand == "Execute")
```
andrewabest, line 80: Should we be explicit here instead of falling back to discovery? i.e. pattern match "Execute" and "Discover" (or whatever the discovery command is), or throw an exception if we don't recognise the command?

andrewabest, line 80: Should we be explicit here instead of falling back to discovery? i.e. pattern match "Execute" and "Discover" (or whatever the discovery command is), or throw an exception if we don't recognise the command?

source/Calamari.Tests/Fixtures/Manifest/InstructionBuilder.cs
gb-8: There is now a `Calamari.Tests.LaunchTools` namespace, and that is causing problems for resolving `LaunchTools` (probably compounded by the fact that that type shares a name with its enclosing namespace).

Most of the existing tests in the project are integration tests that actually invoke Calamari, and they are organized into fixtures covering different use cases.

The new tests in this PR are unit tests of the NodeExecutor, and so I put them in a location that mirror's the NodeExecutor's location in the production project, which resulted in the above namespace resolution problem.

I could move them somewhere else and avoid this change, but don't know that there is a really good place to put them, so I just went with fully scoping the types here instead.

andrewabest: Gotcha, just wanted to make sure it wasn't just an arbitrary formatting change.

source/Calamari/LaunchTools/NodeExecutor.cs
gb-8: Thanks. Done.

andrewabest (approved):
Looks good 👍

PR #836, opened 2022-03-10
Geoffb/aws iam authenticator version fix

gb-8:
There is a [failing test.](https://build.octopushq.com/buildConfiguration/OctopusDeploy_CalamariNetCore_PlatformAgnosticTesting_Windows/3844604?showRootCauses=true&expandBuildChangesSection=true&expandPull+Request+Details=true&expandBuildTestsSection=true)

Here's an explanation:
> Yes, it's broken and legit that the test is failing.
> We're rolling a change to the Dynamic Workers which pins to version 0.5.3 of the `aws-iam-authenticator` tool.
> We need to fix the tests to reflect that this is our expectation in all environments we control [...]
> Those tests currently pull the latest of that tool, which doesn't align to the environment in our Dynamic Workers nor our Worker Tools

This PR fixes the version of the  `aws-iam-authenticator` to 0.5.3.


gb-8:
Wrong branch!

PR #839, opened 2022-03-16
Adds support for S3 feed

egorpavlikhin:
### Background

Adds support for basic S3 feed with supplied credentials. This package feed expects packages in the format of `{bucketName}/{packageName}` and will attempt to find matching files in the bucket by going through the list of know extensions (starting with `zip`). Having access to an S3 feed will enable serverless deployment scenarios, which are typically uploaded to an S3 bucket as a single file.

Related Server PR https://github.com/OctopusDeploy/OctopusDeploy/pull/11606

### How to review

I have done smoke testing and everything appears to work correctly. The one thing that might be surprising is the packageId format, but on the other hand we use the same format for github and docker feeds, so maybe that's ok.

I've only done a happy path test, as it's not clear what exactly needs testing, as the feed mostly calls to AWS SDK and we probably don't need to test that the SDK works.

### DIFF
--- a/source/Calamari.Shared/Calamari.Shared.csproj
+++ b/source/Calamari.Shared/Calamari.Shared.csproj
@@ -23,7 +23,7 @@
     <RootNamespace>Calamari</RootNamespace>
     <LangVersion>8</LangVersion>
     <Nullable>enable</Nullable>
-    <TargetFrameworks>net40;net452;netstandard2.1</TargetFrameworks>
+    <TargetFrameworks>net452;netstandard2.1;net40</TargetFrameworks>
   </PropertyGroup>
   <PropertyGroup Condition="$([MSBuild]::IsOSUnixLike())">
     <TargetFramework>netstandard2.1</TargetFramework>
--- a/source/Calamari.Shared/Calamari.Shared.csproj
+++ b/source/Calamari.Shared/Calamari.Shared.csproj
@@ -48,6 +48,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Calamari.Common\Calamari.Common.csproj" />
+    <PackageReference Include="AWSSDK.S3" Version="3.7.8.8" />
     <PackageReference Include="FSharp.Compiler.Tools" Version="4.0.0.1" />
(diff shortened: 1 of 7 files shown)


source/Calamari.Shared/Calamari.Shared.csproj
```diff
@@ -23,10 +23,7 @@
     <RootNamespace>Calamari</RootNamespace>
     <LangVersion>8</LangVersion>
     <Nullable>enable</Nullable>
-  </PropertyGroup>
-
-  <PropertyGroup Condition="!$([MSBuild]::IsOSUnixLike())">
-    <TargetFrameworks>net452;netstandard2.1;net40</TargetFrameworks>
+    <TargetFrameworks>net40;net452;netstandard2.1</TargetFrameworks>
```
andrewabest, line 26: Does the latter `TargetFramework` declaration overwrite this one if the condition is correct?

source/Calamari.Shared/Integration/Packages/Download/S3PackageDownloader.cs
```diff
@@ -0,0 +1,131 @@
...
+                                                           string? feedUsername,
+                                                           string? feedPassword,
+                                                           bool forcePackageDownload,
+                                                           int maxDownloadAttempts,
+                                                           TimeSpan downloadAttemptBackoff)
+        {
+            var (bucketName, prefix) = GetBucketAndKey(packageId);
+            if (string.IsNullOrWhiteSpace(bucketName) || string.IsNullOrWhiteSpace(prefix))
+            {
+                throw new InvalidOperationException("Invalid PackageId for S3 feed. Expecting format `<bucketName>/<packageId>`");
```
andrewabest, line 62: Should we put what was received in the exception too? i.e. `received {packageId}, expected format ...`

source/Calamari.Shared/Calamari.Shared.csproj
egorpavlikhin: This is just an artifact of manual testing, I'll revert.

andrewabest (approved):
All looks good, just a couple of small comments.

PR #842, opened 2022-04-01
Port Calamari's Cake Build script to Nuke

scottmerchant:
This change has been added to facilitate cross platform building of Calamari. Previously the build scripts for mac/linux were out of date and did not work.

GitVersion has also been replaced by OctoVersion as part of this change.

This update adds the Nuke project and updates to build.ps1/.sh scripts for parameterised builds and build-local.ps1/.sh scripts for basic local builds of calamari on Windows and OSX (it should also work for Linux but that has not been tested).

The original build files have been removed from the project as they will no longer be required once merged.


scottmerchant:
Hey team, just a note that the contents of the build folder does not show in the new PR file tree view because a file named "build" was removed which has the same name as the new folder. I've made a comment [here](https://github.com/github/feedback/discussions/12341#discussioncomment-2504280) to raise the issue.

build/Build.cs
```diff
@@ -0,0 +1,616 @@
+﻿using System;
```
scottmerchant, line 1: This file is the analogue of "build.cake". Most steps (nuke "targets") should be similarly named to their equivalent build.cake step.

build/Build.cs
```diff
@@ -0,0 +1,616 @@
...
+        readonly bool SetOctopusServerVersion;
+
+        [Parameter("AzureKeyVaultUrl")] 
+        readonly string? AzureKeyVaultUrl;
+
+        [Parameter("AzureKeyVaultAppId")] 
+        readonly string? AzureKeyVaultAppId;
+
+        [Parameter("AzureKeyVaultAppSecret")] 
+        readonly string? AzureKeyVaultAppSecret;
```
corey-underdown, line 77: Probably should be `[Secret]` too - like server: https://github.com/OctopusDeploy/OctopusDeploy/blob/master/nuke-build/Build.cs#L76

build/Build.cs
```diff
@@ -0,0 +1,616 @@
...
+        [Parameter("SigningCertificatePassword")] 
+        readonly string SigningCertificatePassword = "Password01!";
+
+        [Required]
+        [OctoVersion(BranchParameter = nameof(BranchName), AutoDetectBranchParameter = nameof(AutoDetectBranch))]
+        readonly OctoVersionInfo? OctoVersionInfo;
+
+        public Build()
+        {
+            NugetVersion = new Lazy<string>(() => AppendTimestamp
```
corey-underdown, line 94: The formatting here is a bit wonky, and I'm not sure I quite understand why it's a Lazy in this situation.

corey-underdown:
I've had a read through it and it's looking like it's on the right track, but I'd be keen to work with you a bit tomorrow (If that works for you?) and we can try and refactor this a bit and pair review it together. Would you be down for that @scottmerchant ?

build/Build.cs
scottmerchant: Ah yep good catch.

build/Build.cs
scottmerchant: I think I did that because during construction the OctoVersionInfo property is not set? Or maybe I just thought it wouldn't be set... I'll try making it not a lazy and see how it goes.

scottmerchant: > I've had a read through it and it's looking like it's on the right track, but I'd be keen to work with you a bit tomorrow (If that works for you?) and we can try and refactor this a bit and pair review it together. Would you be down for that @scottmerchant ?

Yes that would be great, thanks!

scottmerchant:
> I've had a read through it and it's looking like it's on the right track, but I'd be keen to work with you a bit tomorrow (If that works for you?) and we can try and refactor this a bit and pair review it together. Would you be down for that @scottmerchant ?

Yes that would be great, thanks!

mjhilton (changes requested):
Overall it's looking really good, thankyou for putting in this effort @scottmerchant :)

A couple of notes:
* The `build-local` scripts don't currently pass the `-IsLocalBuild` parameter, which is one of the requirements for the `SetOctopusServerVersion` stage to run. Should we add that parameter into those local scripts?
* When I run `build-local.ps1` on my windows machine twice consecutively, the first time passes, but the second time fails at the `CopyToLocalPackages` stage with an error `IOException: The file 'C:\code\LocalPackages\Calamari.22.3.36-scme-port-cake-buil.nupkg' already exists.`. Did I do something wrong?
* The README.md could do with some love to reflect the changes to the build architecture (it's out of date anyway, but if you could clean it up with this PR that'd be greatly appreciated).

I have a niggling concern about the complexity of the different framework versions we target and build, in particular, the .NET 4.0 target. This one doesn't have any direct tests in the Calamari solution - it's only tested based on those steps' usage in the Server integration tests (if any). And from what I understand, these targets aren't built when running on Mac, which I believe is where most of the dev work for this PR has happened... It would be good if we can get a little bit of comfort here to make sure these critical steps (which include Deploy to IIS and Deploy a Package) are still all good.

scottmerchant:
> Overall it's looking really good, thankyou for putting in this effort @scottmerchant :)

No problems at all, it was good to learn about the build process etc. :)

> * The `build-local` scripts don't currently pass the `-IsLocalBuild` parameter, which is one of the requirements for the `SetOctopusServerVersion` stage to run. Should we add that parameter into those local scripts?

In Nuke this parameter is set automatically so we don't need to specify it as a parameter. 👍 

> * When I run build-local.ps1 on my windows machine twice consecutively, the first time passes, but the second time fails at the CopyToLocalPackages stage with an error IOException: The file 'C:\code\LocalPackages\Calamari.22.3.36-scme-port-cake-buil.nupkg' already exists.. Did I do something wrong?

No that's a good catch, thanks! Previously I had been calling Build.ps1 directly with my own parameters which included `-AppendTimestamp` which meant that I never had a clash like this. Firstly I'll add this parameter to the BuildLocal scripts but also I'll update the Nuke build to copy the new file over any existing files so we don't get this failure.

> * The README.md could do with some love to reflect the changes to the build architecture (it's out of date anyway, but if you could clean it up with this PR that'd be greatly appreciated).

No worries I can do that. I'll probably focus on that side of things tomorrow or Friday.

> I have a niggling concern about the complexity of the different framework versions we target and build, in particular, the .NET 4.0 target. This one doesn't have any direct tests in the Calamari solution - it's only tested based on those steps' usage in the Server integration tests (if any). And from what I understand, these targets aren't built when running on Mac, which I believe is where most of the dev work for this PR has happened... It would be good if we can get a little bit of comfort here to make sure these critical steps (which include Deploy to IIS and Deploy a Package) are still all good.

Ok that's good to know, I should have asked you if there were things that required manual testing. Do you know the process for spinning up a branch instance? I feel like that would be the best way to test it.

build/Build.cs
scottmerchant: Checking it in the debugger - in the constructor the OctoVersionInfo is still null so I need to delay the evaluation until later.

corey-underdown:
We talked about refactoring the `Build.cs` file a bit into it's own classes, similar to how Octopus Server does it. A few things can be done like pulling signing out into it's own class, but we're not going to go over the top here - we'd rather get the Nuke build in and in a state that can be refactored later, rather than leave it as Cake. 

Few things we talked about to just double check before merging:
* The files are being signed correctly
* The signing on TC is being done via Azure Sign Tool rather than SignTool.exe
* Build numbers are lining up

Thanks Scott for taking this on, Calamari is a pretty huge one to start with so great effort!

scottmerchant:
@corey-underdown I have completed the refactoring we talked about. Can you please take another quick look? I'm waiting for the newest build to complete before testing that the files are signed correctly and comparing them to the cake build nuget package. Once that is done I'll update my Server branch and get a branch instance of Octopus to test against.

build/Build.cs
scottmerchant: Done.

build/Signing.cs
```diff
@@ -0,0 +1,185 @@
+using System;
```
corey-underdown, line 1: I find the formatting of this file to be very odd - I'm not going to say change it, but I'd consider the indenting here to be too excessive.

corey-underdown (commented):
Thanks heaps for this Scott, it's looking really good.

There's a `Test` target in Cake - is this target not used in the build process at all as it doesn't appear to have been converted over to Calamari? 

If you can confirm that for me please, I'd be happy to approve after.

build/Build.cs
```diff
@@ -0,0 +1,440 @@
...
+                  });
+
+        Target PackBinaries =>
+            _ => _.DependsOn(Publish)
+                  .Executes(async () =>
+                  {
+                      var nugetVersion = NugetVersion.Value;
+                      var packageActions = new List<Action>
+                      {
+                          () => DoPackage(RootProjectName,
```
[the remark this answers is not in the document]
scottmerchant: I basically did this because Octopus Server has references to all these Calamari packages and so if you don't include them when you build the solution on non windows machines it will complain that the packages don't exist. At first I was going to update the Server .csproj to exclude them if running on non-windows but I was rightfully corrected by @shaun-od because we want to include those packages when using standard Calamari build (not a local build). So i've included them as netcore3.1 versions just to keep local builds of server happy.

scottmerchant:
> Thanks heaps for this Scott, it's looking really good.
> 
> There's a `Test` target in Cake - is this target not used in the build process at all as it doesn't appear to have been converted over to Calamari?
> 
> If you can confirm that for me please, I'd be happy to approve after.

Oh good catch, I must have just missed it when I was porting it over. Probably because it's not really used (other than specifying it manually when calling build). Do you think I should put it in?

build/Signing.cs
scottmerchant: Sorry about that - I've been fighting Rider to give me good auto formatting. I think it looks better now.

build/Build.cs
```diff
@@ -0,0 +1,440 @@
...
+        readonly string? AzureKeyVaultAppId;
+
+        [Parameter("AzureKeyVaultAppSecret")]
+        [Secret]
+        readonly string? AzureKeyVaultAppSecret;
+
+        [Parameter("AzureKeyVaultCertificateName")] 
+        readonly string? AzureKeyVaultCertificateName;
+
+        [Parameter("SigningCertificatePath")]
```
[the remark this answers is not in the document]
scottmerchant: Good point, fixed.

I don't think these are really specified anywhere because the defaults are used but you're right, I should keep them the same as the previous parameter names.

build/Build.cs
```diff
@@ -0,0 +1,440 @@
...
+        // when doing local development.
+        bool WillSignBinaries => !IsLocalBuild || SignBinaries;
+
+        [Parameter("Append Timestamp")] 
+        readonly bool AppendTimestamp;
+
+        [Parameter("Set Calamari Version on OctopusServer")] 
+        readonly bool SetOctopusServerVersion;
+
+        [Parameter("AzureKeyVaultUrl")]
```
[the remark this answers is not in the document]
scottmerchant: Done.

corey-underdown:
I can see the build is using Nuke, and none of the other targets seem to use the build scripts, they just call the test dlls manually so it does seem like it was never used - plus the build is green so it looks good to go 👍

corey-underdown (approved):
If you've managed to confirm that the signing is all identical and using azure sign tool for Nuke still, then I'm happy for you to run it through server one last time and merge it.

Thanks for taking this huge task on, it's looking really good 👍

scottmerchant:
> If you've managed to confirm that the signing is all identical and using azure sign tool for Nuke still, then I'm happy for you to run it through server one last time and merge it.
> 
> Thanks for taking this huge task on, it's looking really good 👍

Thanks Corey. I've compared the outputs of version `22.3.3-scme-port-cake-b0066` to `22.3.2` and it contains the same files. The only differences are what I would expect given the different signing timestamps etc.

For the signing, I noticed that the Name of signer is all caps for the new one... I don't think it's related to my changes because on a build on my branch from before I had switched it over to Nuke `22.3.3-scme-port-cake-b0025` it also had all Caps which makes me think that it's potentially a new cert or something. What do you think?
image

scottmerchant:
@mjhilton 
Hi Matt, I have now tested the `Deploy to IIS` and `Deploy a Package` steps where the target is running Windows Server 2016 with my Branch instance using the latest build of Calamari from this branch. It all seems to be working correctly. Are there any other deployment steps which would be good picks for spot testing Calamari?

I've also updated the README.md - hopefully it's what you were imagining. Please let me know if you need anything else before approving.

scottmerchant:
> I'm sure you would anyway, but if you could do a squash merge for this PR as there's quite a lot of commits, that would keep the history nice and clean.

No problems, can do 👍

PR #843, opened 2022-04-08
Add message when downloading  a docker image that is not cached

YihaoOct:
This PR is part closes [this story](https://app.shortcut.com/octopusdeploy/story/863/add-a-message-in-deployment-log-when-a-customer-is-using-a-non-cached-image).

A information-level log message is added to task log when a user tries to download a docker image that is not cached on the executing machine. This log message informs the user that the downloading can be slow when the image is not cached.

Screen Shot 2022-04-08 at 3 41 02 PM

Whether the image is cached is determined by checking whether the local full image ID matches the digest returned from `docker manifest inspect` command. We can't simply use the tag to determine because the same docker image can be associated with multiple tags.


IsaacCalligeros95:
LGTM

IsaacCalligeros95 (commented):
LGTM

IsaacCalligeros95 (approved):

source/Calamari.Shared/Integration/Packages/Download/Scripts/DockerLogin.ps1
```diff
@@ -0,0 +1,51 @@
+function IsDockerAvailable() {
```
geofflamrock, line 1: Do these new scripts automatically get included in any built version of Calamari? Just want to be sure we don't need to explicitly add them to the project file or something along those lines.

source/Calamari.Shared/Integration/Packages/Download/DockerImagePackageDownloader.cs
```diff
@@ -122,25 +169,71 @@ void PerformPull(string? username, string? password, string fullImageName, strin
...
+            var result = SilentProcessRunner.ExecuteCommand("docker",
+                                                            $"manifest inspect --verbose {fullImageName}",
+                                                            ".", 
+                                                            environmentVariables, 
+                                                            (stdout) => { output += stdout; },
+                                                            (error) => { });
+
+            if (result.ExitCode != 0)
+            {
+                return null;
```
geofflamrock, line 205: Would it helpful to know that something went wrong with trying to get image digests? The end message saying "image may not be cached" may not really need to change depending on whether we were able to query digests successfully or not, so potentially it may not really matter that much to us.

geofflamrock (approved):
Changes look good 👍 I left a couple of small thoughts/questions.

source/Calamari.Shared/Integration/Packages/Download/Scripts/DockerLogin.ps1
YihaoOct: Yep, these scripts are auto included because previously the scripts under this folder are added as embedded resources using the wildcard.

source/Calamari.Shared/Integration/Packages/Download/DockerImagePackageDownloader.cs
YihaoOct: Basically I think we would swallow the errors encountered when trying to get the digest information. This is because the aim of this change is to add a friendly information-level log to help users understand what is potentially causing the deploy to run longer than expected. Therefore, if we report the errors encountered in making these checks, that is likely to pollute the logs with irrelevant information to the users. So here if there are any errors happened during checking whether the digest is cached, we simply treat the image as being cached and don't write the log. (But when we really need to do debug something, we can always know if the image is being cached by checking the logs from docker pull).

PR #846, opened 2022-04-29
Consolidate duplicate test helper classes

scottmerchant:
There were a number of duplicate classes in `Calamari.Tests.Helpers` and `Calamari.Testing.Helpers` which were practically the same with a few minor differences. This PR consolidates the classes by merging the differences and updating all references in Calamari.Tests.

If all tests pass I think this means the merged helper classes are working as expected.

There are quite a few touched files but fortunately the majority just have some adjusted `using` statements.


source/Calamari.Tests/Helpers/TestEnvironmentExtended.cs
```diff
@@ -0,0 +1,35 @@
+using Assent;
+using Assent.Namers;
+using Calamari.Testing.Helpers;
+
+namespace Calamari.Tests.Helpers
+{
+    public static class TestEnvironmentExtended
+    {
+        static bool IsCi => TestEnvironment.IsCI;
```
gb-8, line 9: This can be `private`, but I think it is better to just remove it and use `TestEnvironment.IsCI` directly.

source/Calamari.Tests/Helpers/TestEnvironmentExtended.cs
```diff
@@ -0,0 +1,35 @@
+using Assent;
+using Assent.Namers;
+using Calamari.Testing.Helpers;
+
+namespace Calamari.Tests.Helpers
+{
+    public static class TestEnvironmentExtended
```
gb-8, line 7: A class better name would be `AssentConfigurations`, with the properties renamed to `Default`, `Json`, `Yaml`, etc.

source/Calamari.Tests/Helpers/TestEnvironmentExtended.cs
scottmerchant: Good point, fixed.

source/Calamari.Tests/Helpers/TestEnvironmentExtended.cs
scottmerchant: Good idea, done.

scottmerchant:
[Turn K8s spike into MMVP](https://trello.com/c/GNmxkV27/104-turn-k8s-spike-into-mmvp)

PR #850, opened 2022-05-11
Fix kubectl test bug by ensuring executable from known version is used in generated script

scottmerchant:
The first attempt to fix this issue was done in [this PR](https://github.com/OctopusDeploy/Calamari/pull/848) but that didn't completely fix the issue due to the fact that the generated scripts in the tests were using which ever version of `kubectl` was on the path for the agent.

In this PR I've updated the script generation to point specifically to the version of `kubectl` which was downloaded by the test setup code so that this issue doesn't occur.

### DIFF
--- a/source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixture.cs
+++ b/source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixture.cs
@@ -28,6 +28,8 @@ namespace Calamari.Tests.KubernetesFixtures
     [Category(TestCategory.RunOnceOnWindowsAndLinux)]
     public class KubernetesContextScriptWrapperLiveFixture: KubernetesContextScriptWrapperLiveFixtureBase
     {
+        const string KubeCtlExecutableVariableName = "Octopus.Action.Kubernetes.CustomKubectlExecutable";
+        
         InstallTools installTools;
 
         string eksClientID;
@@ -75,7 +77,7 @@ namespace Calamari.Tests.KubernetesFixtures
         [SetUp]
         public void SetExtraVariables()
         {
-            variables.Set("Octopus.Action.Kubernetes.CustomKubectlExecutable", installTools.KubectlExecutable);
+            variables.Set(KubeCtlExecutableVariableName, installTools.KubectlExecutable);
         }
 
         protected override Dictionary<string, string> GetEnvironments()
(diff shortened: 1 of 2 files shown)


source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixture.cs
```diff
@@ -271,7 +271,7 @@ public void AuthorisingWithAmazonAccount()
             variables.Set("Octopus.Action.Kubernetes.CertificateAuthority", certificateAuthority);
             variables.Set($"{certificateAuthority}.CertificatePem", eksClusterCaCertificate);
             var wrapper = CreateWrapper();
-            TestScript(wrapper, "Test-Script");
+            TestScript(wrapper, "Test-Script", variables.Get("Octopus.Action.Kubernetes.CustomKubectlExecutable"));
```
shaun-od, line 274: I think we need to be checking if this exists, otherwise what would it do?

scottmerchant: I don't think that it will ever not be set because it's set in the `[SetUp]` method (line 78). If it's not set then there is something wrong with the test and we would expect it to fail, wouldn't we?

scottmerchant: I guess I could make it throw if it's not set with a nicer message.

shaun-od: I agree with what you are saying but it is about helping someone in the future in my mind 😄 Maybe a comment about why we are doing this wouldn't hurt either

scottmerchant: Yep good point - I've updated it now 👍

source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixture.cs
```diff
@@ -271,7 +273,14 @@ public void AuthorisingWithAmazonAccount()
             variables.Set("Octopus.Action.Kubernetes.CertificateAuthority", certificateAuthority);
             variables.Set($"{certificateAuthority}.CertificatePem", eksClusterCaCertificate);
             var wrapper = CreateWrapper();
-            TestScript(wrapper, "Test-Script");
+            
+            // When authorising via AWS, We need to make sure we are using the correct version of
+            // kubectl for the test script as newer versions may cause kubectl to fail with an error like:
+            // 'error: exec plugin: invalid apiVersion "client.authentication.k8s.io/v1alpha1"'
+            var kubectlExecutable = variables.Get(KubeCtlExecutableVariableName) ??
+                throw new Exception($"Unable to find required kubectl executable in variable '{KubeCtlExecutableVariableName}'");
```
shaun-od, line 281: I am in two minds about this exception. I know why you have done this but should we just set it to "kubectl"?

scottmerchant: Yeah I guess we can

shaun-od: I just didn't want to try and set it null if that is `variables.Get` does if the variable doesn't exist

scottmerchant: It feels weird though because we don't expect using `kubectl` to work in this case (well not always)... isn't it better to be explicit in the failure? I'm in two minds too haha

shaun-od: Ok leave as it is with the exception

shaun-od (approved):
LGTM

PR #851, opened 2022-05-11
Calamari - Add EKS support for Kubernetes Target Discovery

scottmerchant:
In [a previous PR](https://github.com/OctopusDeploy/Calamari/pull/847), Target Discovery for Kubernetes Clusters was added for Azure. This change extends that initial change to allow for target discovery using AWS credentials.

This coincides with [a server PR](https://github.com/OctopusDeploy/OctopusDeploy/pull/12177) which adds the ability for AWS IAM Role credentials to be added to Kubernetes Cluster targets via Service Messages.


source/Calamari.Aws/Kubernetes/Discovery/AwsKubernetesDiscoverer.cs
```diff
@@ -0,0 +1,108 @@
...
+            }
+            
+            foreach (var region in authenticationDetails.Regions)
+            {
+                var client = new AmazonEKSClient(authenticationDetails.ToCredentials(),
+                    RegionEndpoint.GetBySystemName(region));
+
+                var clusters = client.ListClustersAsync(new ListClustersRequest()).GetAwaiter().GetResult();
+                
+                log.Info("Discovery has discovered");
```
gb-8, line 66: Is this logging supposed to be here? If so, it probably needs rewording.

source/Calamari.Aws/Kubernetes/Discovery/AwsKubernetesDiscoverer.cs
```diff
@@ -0,0 +1,108 @@
...
+            this.log = log;
+        }
+
+        public string Name => "Aws";
+        
+        public IEnumerable<KubernetesCluster> DiscoverClusters(string contextJson, IVariables variables)
+        {
+            if (!TryGetAuthenticationDetails(contextJson, out var authenticationDetails))
+                yield break;
+            var workerPool = variables.Get("Octopus.Aws.WorkerPool");
```
gb-8, line 31: We can't rely on using variables inside target discovery.

While this will work at present, there will be a change soon whereby the discovery action no longer be using the same ID as the action which triggered target discovery.

That means that any variables which are scoped to the triggering action will not be available during target discovery, and we should therefore avoid using any variables inside discovery.

The exception is the variable which holds the target discovery scope, which will still be explicitly be scoped to the discovery action.

The worker pool ID will be available on the discovery scope, so if we want to log it here, then that's where it can be fetched from.

source/Calamari.Aws/Kubernetes/Discovery/AwsKubernetesDiscoverer.cs
scottmerchant: Oops, good spot!

scottmerchant: I've removed this because it isn't required. We log how many clusters have been discovered in `KubernetesDiscoveryCommand.cs`

source/Calamari.Aws/Kubernetes/Discovery/AwsKubernetesDiscoverer.cs
```diff
@@ -0,0 +1,108 @@
...
+        {
+            this.log = log;
+        }
+
+        public string Name => "Aws";
+        
+        public IEnumerable<KubernetesCluster> DiscoverClusters(string contextJson, IVariables variables)
+        {
+            if (!TryGetAuthenticationDetails(contextJson, out var authenticationDetails))
+                yield break;
```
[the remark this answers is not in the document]
scottmerchant: I've updated it to log a warning, is that ok? I've also updated the Azure discoverer to log a warning in the same spot as I missed it for that one too.

scottmerchant:
[Support discovery on EKS k8s clusters.](https://trello.com/c/2iGngBbV/109-support-discovery-on-eks-k8s-clusters)

source/Calamari.Aws/Kubernetes/Discovery/AwsAuthenticationDetails.cs
```diff
@@ -0,0 +1,63 @@
...
+namespace Calamari.Kubernetes.Aws
+{
+    public class AwsAuthenticationDetails : ITargetDiscoveryAuthenticationDetails
+    {
+        const string DefaultSessionName = "OctopusKubernetesClusterDiscovery";
+        public AWSCredentials ToCredentials()
+        {
+            var account = Credentials.Type == "account"
+                ? new BasicAWSCredentials(Credentials.Account.AccessKey, Credentials.Account.SecretKey)
+                : (AWSCredentials)new EnvironmentVariablesAWSCredentials();
```
scottmerchant, line 14: I'm a bit worried that this might not handle the E2C Instance case (after talking in the Cloud Credentials meeting). We don't have a test using an E2C instance worker to test and se we're only using EnvironmentVariables in tests where `Credentials.Type` is not account...

scottmerchant: I've updated this now but we don't have any tests which exercise the `InstanceProfileAwsCredentials`... Should I add that in?

source/Calamari.Aws/Kubernetes/Discovery/AwsAuthenticationDetails.cs
```diff
@@ -0,0 +1,80 @@
...
+                try
+                {
+                    // If not currently running on an EC2 instance,
+                    // this will throw an exception.
+                    account = new InstanceProfileAWSCredentials();
+                }
+                catch
+                {
+                    // The last attempt is trying to use Environment Variables.
+                    account = new EnvironmentVariablesAWSCredentials();
```
gb-8, line 29: As per our discussion, if this constructor can throw we should catch the possible exceptions here, and make sure appropriate error messages are generated.

Otherwise, let's make sure that the error case where no env vars exist is handled at the point at which it surfaces.

source/Calamari.Aws/Kubernetes/Discovery/AwsKubernetesDiscoverer.cs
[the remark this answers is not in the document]
gb-8: Our philosophy here is that target discovery errors should not prevent a deployment from proceeding, so from that point of view logging a warning would be appropriate.

It would be nice to actually get an error in the server logs if we think think we have detected a code defect, but since this is in Calamari, that is not trivial to achieve, and not worth investing in just for this unlikely scenario.

source/Calamari.Aws/Kubernetes/Discovery/AwsKubernetesDiscoverer.cs
```diff
@@ -0,0 +1,116 @@
...
+        
+        public IEnumerable<KubernetesCluster> DiscoverClusters(string contextJson)
+        {
+            if (!TryGetDiscoveryContext(contextJson, out var discoveryContext))
+                yield break;
+
+            var authenticationDetails = discoveryContext.Authentication;
+            if (authenticationDetails == null)
+            {
+                Log.Warn("Target Discovery Context is in the wrong format: Unable to serialise authentication details");
```
gb-8, line 35: It's deserialization that has failed here, not serialization.
Maybe just "read" or "access" is a friendlier message here.
Since these are user facing messages we should make it clear that this is a problem with the application, not a user error.

@rhysparry Are there any guidelines on consistent use of language to inform users of server/application errors which the user cannot do anything about vs errors users can take corrective action for?

rhysparry: The [Logging](https://github.com/OctopusDeploy/Architecture/blob/main/Steps/Logging.md) document in the Architecture repo is the one I know about.

I think it's worth considering what a user is expected to do with the information. The "Build succeeded with warnings" is something we'd like users to be able to ensure they can achieve (users don't want to have to inspect the logs to see warnings they can't do anything about).

source/Calamari.Aws/Kubernetes/Discovery/AwsKubernetesDiscoverer.cs
```diff
@@ -0,0 +1,129 @@
...
+            discoveryContext = null;
+            try
+            {
+                discoveryContext =
+                    JsonConvert.DeserializeObject<TargetDiscoveryContext<AwsAuthenticationDetails>>(json);
+                
+                if (discoveryContext != null)
+                    return true;
+                
+                log.Warn("Target discovery context is in the wrong format.");
```
gb-8, line 119: I think these warnings need changing too.
The way it is handled above looks good, in that there is a warning that lets the user know it is not something they can solve themselves, and need to contact Octopus support, and additional details in the verbose log.
The same approach should be taken here and in the catch statement below.
All three places could call a single method to write the log output to guaranteed consistency.

The same approach should be followed in the Azure discovery code too.

gb-8:
Nice job getting the EC2 tests working.

Looks good now, except for a bit more quibbling I've added about the logging.

source/Calamari.Aws/Kubernetes/Discovery/AwsAuthenticationDetails.cs
scottmerchant: Done 👍

source/Calamari.Aws/Kubernetes/Discovery/AwsKubernetesDiscoverer.cs
scottmerchant: Thanks for the link @rhysparry Good to be reminded of those principles. I've updated the code to better align with those I think 👍

source/Calamari.Aws/Kubernetes/Discovery/AwsKubernetesDiscoverer.cs
scottmerchant: Fixed - I ended up created a `KubernetesDiscovererBase` which implements the basic json parsing code and error logging. This keeps everything in one place and works for both Azure and AWS 👍

source/Calamari.Aws/Kubernetes/Discovery/AwsAuthenticationDetails.cs
```diff
@@ -0,0 +1,92 @@
...
+            }
+            else
+            {
+                try
+                {
+                    // If not currently running on an EC2 instance,
+                    // this will throw an exception.
+                    account = new InstanceProfileAWSCredentials();
+                }
+                catch (Exception instanceProfileException)
```
gb-8, line 31: We really should be catching a more specific exception here.
It looks like this constructor and `EnvironmentVariablesAWSCredentials` below will both throw an `InvalidOperationException` if they fail, so we should just catch that.

source/Calamari.Aws/Kubernetes/Discovery/AwsAuthenticationDetails.cs
```diff
@@ -0,0 +1,92 @@
...
+                {
+                    try
+                    {
+                        // The last attempt is trying to use Environment Variables.
+                        account = new EnvironmentVariablesAWSCredentials();
+                    }
+                    catch (Exception environmentVariablesException)
+                    {
+                        throw new AggregateException(instanceProfileException, environmentVariablesException);
```
gb-8, line 40: Rather than throwing this general aggregate exception here wrapping two `Exception`s, we should either be throwing a more semantically informative exception.
In this case, I would prefer modifying the return type to use IResult, and return success or failure objects.

source/Calamari.Aws/Kubernetes/Discovery/AwsAuthenticationDetails.cs
scottmerchant: As discussed it's not clear what exception we might receive from inspecting the decompiled code so I've added a comment and left the exception handlers catching the base Exception.

source/Calamari.Aws/Kubernetes/Discovery/AwsAuthenticationDetails.cs
scottmerchant: As discussed, we've updated the method to be a "TryGet" style method and return a bool, it also log it's own exceptions.

gb-8 (approved):
Looks good now.

PR #852, opened 2022-05-24
Fix Helm Upgrade Command to work with Kubernetes Targets which don't have a ClusterUrl set

scottmerchant:
# Background

If Cluster URL is null, the Upgrade Helm step will fail, even if the Cluster URL is not required (eg: AKS (Azure) K8s Clusters).
This has been an issue when creating Kubernetes Cluster targets for Azure via service messages because the ClusterURL is set to null in that case. Rather than correct the service message change, I've opted to remove the redundant and unnecessary check from Calamari which means all Clusters which currently have this issue will now work with the Helm step.

**Note:** Add clusters via the UI will result in Cluster URL being set to an empty string which bypasses this issue.

Here is a screen cap of two clusters with URL null, one AKS (url not required) and one EKS (url required) you can see that the error message for the EKS cluster is still informative without this check.
image

### DIFF
--- a/source/Calamari/Kubernetes/Commands/HelmUpgradeCommand.cs
+++ b/source/Calamari/Kubernetes/Commands/HelmUpgradeCommand.cs
@@ -64,8 +64,6 @@ namespace Calamari.Kubernetes.Commands
             if (!File.Exists(pathToPackage))
                 throw new CommandException("Could not find package file: " + pathToPackage);
 
-            ValidateRequiredVariables();
-
             var conventions = new List<IConvention>
             {
                 new DelegateInstallConvention(d => extractPackage.ExtractToStagingDirectory(pathToPackage)),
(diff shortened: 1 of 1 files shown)


scottmerchant:
[Investigate Helm k8s problem](https://trello.com/c/dqOzUbw6/152-investigate-helm-k8s-problem)

geofflamrock (approved):
LGTM. Only thing you might consider is whether it's worth getting a test in there that validates the cluster url for auth scenarios that require it e.g. EKS.

scottmerchant:
> LGTM. Only thing you might consider is whether it's worth getting a test in there that validates the cluster url for auth scenarios that require it e.g. EKS.

I was considering this but in the end I realised there is actually handling for this situation in `Calamari/Kubernetes/KubernetesContextScriptWrapper.cs` which is where the error in the screen cap above comes from. So I think we're safe to just remove it here as it was a double-up anyway.

PR #853, opened 2022-05-27
Separate kubernetes tests into separate cloud providers

scottmerchant:
It's very annoying to run these tests locally as if you don't comment out the cloud providers you don't care about, it will spin up infrastructure which you don't need which takes a long time. I've separated them out so that you can just run the tests you care about which will only create the required infrastructure. This change is dependent on [this PR](https://github.com/OctopusDeploy/Calamari/pull/851) being merged first.


source/Calamari.Tests/KubernetesFixtures/Terraform/Clusters/EKS/kubeconfig.tpl
```diff
@@ -0,0 +1,28 @@
+apiVersion: v1
```
scottmerchant, line 1: @gb-8 has suggested sharing this file between each provider as all three files are the same.

scottmerchant: @gb-8 I managed to do this but I had to update the code to copy it from the new location.

gb-8: Turns out we only need this for EKS, so the other ones have now been removed.

gb-8 (approved):

scottmerchant:
[Support discovery on EKS k8s clusters.](https://trello.com/c/2iGngBbV/109-support-discovery-on-eks-k8s-clusters)

PR #859, opened 2022-06-10
Create AKS Target with name to include subscription id and resource group

scottmerchant:
This is guaranteed to be unique even across different subscriptions etc.

Previously we were just using the cluster name which is obviously not unique across different subscriptions of resource groups.

This PR has pivoted to use the format `aks/{subscriptionid}/{resourcegroup}/{clustername}`
eg:
image


scottmerchant:
[Fix issue where Discovered AKS Clusters use the Cluster Name as the Target name](https://trello.com/c/rNZN6BUf/194-fix-issue-where-discovered-aks-clusters-use-the-cluster-name-as-the-target-name)

geofflamrock (approved):
LGTM

PR #861, opened 2022-06-15
Prevent exception thrown when executing create namespace for k8 conte…

mik-ky:
**Background:**
When the EKS cluster is unreachable when Calamari tries to setup a k8 authentication context, the current state of Calamari throws an exception, causing the step to fail. This PR changes the behaviour to continue executing in the above scenario.

While the current behaviour make sense for a built-in k8 step, we would want to proceed execution of a target script as this may be a valid scenario where the script may not be directly working on the target.

[Fixes https://github.com/OctopusDeploy/Issues/issues/7506](https://github.com/OctopusDeploy/Issues/issues/7603)

Results
---
Before:
Screen Shot 2022-06-17 at 11 54 25 AM

After (+logs):
[ServerTasks-2429.log.txt](https://github.com/OctopusDeploy/Calamari/files/8953468/ServerTasks-2429.log.txt)
image

---


mjhilton:
Hey @mik-ky, nice simple fix! I think it gets the job done, but have a couple of suggestions on how we might make it feel a little nicer.

It feels a little bit funny that the Task succeeds (with Warnings), but has error-level entries in the task log. Given running one of these scripts on behalf of an unhealthy k8s cluster is something we allow, it would probably be nicer if we were somehow able to prevent the output from `kubectl create namespace` from showing in our logs as an error. 

If it were easy enough to add, it might also be helpful if we could add some kind of log message that informs the users that "we couldn't contact the cluster, but we're running your script anyway".

mjhilton:
In terms of testing, I think you should probably be able to write a reasonable test for this scenario as an addition to either the `Calamari.Tests.KubernetesFixtures.KubernetesContextScriptWrapperFixture`, or the `Calamari.Tests.KubernetesFixtures.KubernetesContextScriptWrapperLiveFixtureBase` derivatives.

I'd be looking at configuring the test scenario with a non-existent cluster as the target, having a simple script that echoes some value to the log, and then checking:

- Script did not fail
- (Maybe) script did not log errors
- Expected echo value appeared in logs

Configure it with a non-existent cluster,

PR #868, opened 2022-06-28
Merge forward changes in 2021.3 to 2022.1 for #866

mik-ky:
Merge forward the changes for updating AzureCodeSignTool #866 introduced to `release/2021.3` branch

PR #869, opened 2022-06-29
Add keyVaultTenantId when signing using azure signtool

mik-ky:
This is effectively the same as #866 and #868 but slightly different - as cake has been nuked.

* Add keyVaultTenantId when signing with Azure Sign Tool
* Updates `Nuke.Common` version to latest - from `6.0.1` to  `6.1.1`


IsaacCalligeros95 (approved):
LGTM 👍

PR #873, opened 2022-07-04
Add logging around file access when replace or copy file fails

IsaacCalligeros95:
Adds logging when copying or replacing a file fails. Logs the file names, along with whether or not that file can be opened by the calamari process. This lets us know which of the files is inaccessible and the exception gives further details as to why.


mjhilton (approved):
LGTM

PR #877, opened 2022-07-24
Port existing ConsolidateCalamariPackagesTask to Calamari

mik-ky:
**Background**
We are pulling the Calamari halves from Sashimi back into Calamari repository. In order for us to do that, we need to modify how our Calamaris** are being packaged. This change is intended as a step towards making Calamari responsible for creating a bundled zip, rather than server doing the unzipping, re-zipping and creating a Calamari package bundle for it to consume.

**Changes**

- Move existing classes/files that lived inside [ConsolidateCalamariPackagesTask Repo](https://github.com/OctopusDeploy/ConsolidateCalamariPackagesTask) into Calamari, with a few changes to work against nuke build instead of a MSBuildTask.
- Calamari and Sashimi flavours are read from a .nupkg file to make things simple for the time being to accomodate for "unmigrated" calamari flavours to be packaged up in an expected manner.
- Created a new, temporary project called "Calamari.UnmigratedCalamariFlavours" which:
    - Grabs packages for flavours that have not been migrated over yet, for consolidation and to keep track of things easily
    - Copies sashimi nuget packages to `./artifacts` by using `GeneratePathProperty` for each package
        - Which, at the moment the file is hardcoded(!) which means when a version changes, this copy task needs to be updated. I could not find a way to do this with wildcards, so an alternative suggestion would be appreciated. This is a temporary process, so maybe it's okay for the time being.

**Output**
Calamari.<hash>.zip contains a bunch of folder with the hash of the files and the `index.json` has values in the format that we expect:

image


.nuke/build.schema.json
```diff
@@ -14,11 +14,15 @@
         },
         "AzureKeyVaultAppSecret": {
           "type": "string",
-          "default": "Secrets must be entered via 'nuke :secret [profile]'"
+          "default": "Secrets must be entered via 'nuke :secrets [profile]'"
```
mik-ky, line 17: Not sure why this has changed yet. :shrug:

source/Calamari.UnmigratedCalamariFlavours/Calamari.UnmigratedCalamariFlavours.csproj
```diff
@@ -0,0 +1,39 @@
...
+        <PackageReference Include="Sashimi.GoogleCloud.Accounts" Version="2.1.2" GeneratePathProperty="true" />
+        <PackageReference Include="Sashimi.GoogleCloud.Scripting" Version="2.1.2" GeneratePathProperty="true" />
+        <PackageReference Include="Sashimi.Terraform" Version="11.1.0" GeneratePathProperty="true" />
+        <PackageReference Include="Sashimi.AzureServiceFabric" Version="12.1.2" GeneratePathProperty="true" />
+        <PackageReference Include="Sashimi.AzureScripting" Version="14.2.1" GeneratePathProperty="true" />
+    </ItemGroup>
+
+    <!-- Copy over sashimi nuget packages from the packages directory. -->
+    <Target Name="CopyPackages" BeforeTargets="PrepareForBuild">
+        <Copy SourceFiles="$(PkgSashimi_Aws_Accounts)\sashimi.aws.accounts.14.2.0.nupkg" DestinationFolder="$(ProjectDir)\artifacts"></Copy>
```
mik-ky, line 26: This is hardcoded(!) as the filename contains the version. I couldn't find a way to do a wildcard on these, but if anyone knows a better alternative to this, would be appreciated. (although this process is temporary)

build/ConsolidateCalamariPackages/CalamariPackageReference.cs
```diff
@@ -0,0 +1,56 @@
...
+            PackagePath = packageReference.PackagePath;
+        }
+
+        public IReadOnlyList<SourceFile> GetSourceFiles(ILogger log)
+        {
+            var isNetFx = Name == "Calamari";
+            var isCloud = Name == "Calamari.Cloud";
+            var platform = isNetFx || isCloud
+                ? "netfx"
+                : Name.Split('.')[1]; //this needs to change to account for sashimi's calamari halves later; TODO.
```
mik-ky, line 31: This may potentially need to be modified, depending on how we package things up with lift&shift.

build-local.ps1
```diff
@@ -18,7 +18,7 @@ Write-Host "
 ##################################################################################################
 " -ForegroundColor Cyan
 
-./build.ps1 -BuildVerbosity Minimal -Verbosity Minimal -PackInParallel -AppendTimestamp -SetOctopusServerVersion 
+./build.ps1 -BuildVerbosity Minimal -Verbosity Normal -PackInParallel -AppendTimestamp -SetOctopusServerVersion
```
mik-ky, line 21: Changing this to Normal Verbosity. It's for local builds, but if we want to keep it Minimal, I'm happy to change this back.

build/ConsolidateCalamariPackages/SashimiPackageReference.cs
```diff
@@ -0,0 +1,74 @@
...
+            this.packageReference = packageReference;
+        }
+
+        public string Name => packageReference.Name;
+        public string Version => packageReference.Version;
+        public string PackagePath => packageReference.PackagePath;
+
+        public IReadOnlyList<SourceFile> GetSourceFiles(ILogger log)
+        {
+            var extractPath = $"{Path.GetDirectoryName(PackagePath)}/extract/{Name}";
```
mik-ky, line 27: I've decided to unzip the packages instead of working directly off a zipped package;
This seemed cleaner, and was less "disruptive" to the original logic. Happy to change this back to working with a zipped package if we don't want to unzip.

mjhilton (commented):
I've been having a review and a think about the overall approach.

Currently, we produce everything as NuGet packages because we ship them across to Server, and it consumes and consolidates them itself.

Once we take the leap to Server consuming the consolidated Calamari directly, it will no longer need to reference the Calamari NuGet packages we produce today. 

So, I think there's a not-far-off future where we don't need to publish NuGet packages from the Calamari repo at the granularity that we do today.

To support our first migration of `Sashimi.AzureAppService` -> `Calamari.AzureAppService`, we could choose one of two paths:

* Update the pipeline in the Calamari repo to produce framework-specific NuGet packages for Calamari.AzureAppService like this:
  - Calamari.AzureAppService.win-x64.nupkg
  - Calamari.AzureAppService.linux-x64.nupkg
  - .. etc

or

* For Calamari projects, stop publishing NuGet packages at all, and update the Consolidate logic for CalamariPackageReferences to be based off the folders in publish rather than NuGet packages.

I think in the longer-term, the second option would be more efficient: we can reduce the amount of steps in the overall build process, and remove a reasonable bit of unnecessary packaging, file copying, unzipping etc. But, it's not something we **need** to do now.

What do you think?

source/Calamari.UnmigratedCalamariFlavours/Calamari.UnmigratedCalamariFlavours.csproj
mjhilton: I think given the process is temporary, we can live with it :)

build-local.ps1
mjhilton: I prefer being able to see some detail. The `Minimal` was a bit useless when things went wrong.

mjhilton:
(As a final goal, I think we should remove as much complexity as possible, which would include not generating NuGet files that we don't publish. But we don't necessarily need to go straight to that right now.)

mik-ky:
> I've been having a review and a think about the overall approach.
> 
> Currently, we produce everything as NuGet packages because we ship them across to Server, and it consumes and consolidates them itself.
> 
> Once we take the leap to Server consuming the consolidated Calamari directly, it will no longer need to reference the Calamari NuGet packages we produce today.
> 
> So, I think there's a not-far-off future where we don't need to publish NuGet packages from the Calamari repo at the granularity that we do today.
> 
> To support our first migration of `Sashimi.AzureAppService` -> `Calamari.AzureAppService`, we could choose one of two paths:
> 
> * Update the pipeline in the Calamari repo to produce framework-specific NuGet packages for Calamari.AzureAppService like this:
>   
>   * Calamari.AzureAppService.win-x64.nupkg
>   * Calamari.AzureAppService.linux-x64.nupkg
>   * .. etc
> 
> or
> 
> * For Calamari projects, stop publishing NuGet packages at all, and update the Consolidate logic for CalamariPackageReferences to be based off the folders in publish rather than NuGet packages.
> 
> I think in the longer-term, the second option would be more efficient: we can reduce the amount of steps in the overall build process, and remove a reasonable bit of unnecessary packaging, file copying, unzipping etc. But, it's not something we **need** to do now.
> 
> What do you think?

Yes! I 100% agree with the second option being the approach forward. I think, for what it's worth, the effort required to do option 2 is, I'm assuming, marginally higher than option 1 compared to the benefits it bring as you've mentioned. I haven't really considered option 1 before - but depending how it goes, we can do this in the interim while option 2 is implemented?

As you've mentioned, yeah I thought we wouldn't need this now and deferred the implementation change for Calamari* cnosolidation to when we were doing the lift & shift as my next steps in mind was to work with someone doing the lift & shift to implement option 2, and to keep the scope to migrating the logic.

build/Build.cs
```diff
@@ -291,6 +296,45 @@ public Build()
...
+                               foreach (var file in Directory.GetFiles(SashimiPackagesDirectory))
+                                   CopyFile(file, ArtifactsDirectory / Path.GetFileName(file), FileExistsPolicy.Overwrite);
+                           });
+
+        Target PackageConsolidatedCalamariZip =>
+            _ => _.DependsOn(PackBinaries)
+                  .DependsOn(CopySashimiPackagesForConsolidation)
+                  .Executes(() =>
+                            {
+                                var artifacts = Directory.GetFiles(ArtifactsDirectory, "*.nupkg");
```
mjhilton, line 312: I noticed that right now, _everything_ is being published as a NuPkg, including things like `Calamari.Common`, which don't contain a Calamari executable. The Consolidate task is picking these up, and adding them unnecessarily to the consolidated package. 

Could we use an inclusion-list to only consider a subset of packages known to contain Calamari executables here?

build/Build.cs
```diff
@@ -291,6 +296,45 @@ public Build()
...
+        Target CopySashimiPackagesForConsolidation =>
+            _ => _.DependsOn(Compile)
+                 .Executes(() =>
+                           {
+                               foreach (var file in Directory.GetFiles(SashimiPackagesDirectory))
+                                   CopyFile(file, ArtifactsDirectory / Path.GetFileName(file), FileExistsPolicy.Overwrite);
+                           });
+
+        Target PackageConsolidatedCalamariZip =>
+            _ => _.DependsOn(PackBinaries)
```
mjhilton, line 308: The dependency tree here runs the `Clean` target, which removes everything from `**/obj`, `**/bin`, `./artifacts` and `./publish`.

We may want to look at keeping the Consolidated Calamari package around, because the Consolidate task firstly looks to see _whether_ it needs to rebuild the package, and if not, skips it. By clearing it away every time, we lose this potential efficiency.

mjhilton (changes requested):
Looking good. In addition to the above discussion, I've added a few specific comments here.

A final consideration is that in the previous approach (`ConsolidateCalamariPackagesTask` as an MSBuild target), we had at least one test which validated that the functionality of the Build Task produced an approved output that matched expectations. We've lost that here.

How could we add a safety net to ensure that we don't accidentally break the way the packaging works?

build/Build.cs
mik-ky: Yup that seems more appropriate - I'll follow this up in another PR to have optimisations and tests in place.

build/Build.cs
mik-ky: Yep. This is simple to do - I'll add this too, in the follow-up with cleanups and optimisations PR.

mik-ky:
> Looking good. In addition to the above discussion, I've added a few specific comments here.
> 
> A final consideration is that in the previous approach (`ConsolidateCalamariPackagesTask` as an MSBuild target), we had at least one test which validated that the functionality of the Build Task produced an approved output that matched expectations. We've lost that here.
> 
> How could we add a safety net to ensure that we don't accidentally break the way the packaging works?

I have some ideas to do this, but I don't want to be blocking the way forward with dependency management work - I'll include this and the other good suggestions you've raised in the PR if you're okay with it.

mjhilton (approved):
Other changes to be addressed in a follow-up PR. We'll assume that the resulting ConsolidateCalamariPackage is not yet ready for consumption in Server until then.

PR #878, opened 2022-07-28
Add step to create nupkg from zip

YihaoOct:
This PR generates a `Calamari.Consolidated.<version>.nupkg` file that contains the consolidated `Calamari.*.zip` file.

This is achieved by adding a `Calamari.Consolidated` project to the solution, which does not contain any source file, but just includes the consolidated zip file when targeted by `dotnet pack`. (I can't find a better solution to generate this .nupkg file that doesn't require using an empty project. Let me know if there is).

The package is currently named `Calamari.Consolidated` rather than `Octopus.Dependencies.Calamari`, because there is some magic done in the Server build process that deals with things started with `Octopus.Dependencies.*`. I'm not sure if they will work together if we use `Octopus.Dependencies.Calamari`.

The zip file is put inside the .nupkg file under the `contentFiles/any/any` directory. This is to follow the [folder structure convention](https://docs.microsoft.com/en-us/nuget/reference/nuspec#package-folder-structure), which is also adopted by [`Octopus.Dependencies.Java`](https://feedz.io/org/octopus-deploy/repository/dependencies/packages/Octopus.Dependencies.Java).

A `Calamari.Consolidated.<version>.nupkg` file should be generated in the artifacts folder, which will then be pushed to Feedz.io together with other `Calamari.*.<version>.nupkg`s from TC. (We can remove these unnecessary nupkgs in the future)


mjhilton:
> This is achieved by adding a Calamari.Consolidated project to the solution, which does not contain any source file, but just includes the consolidated zip file when targeted by dotnet pack. (I can't find a better solution to generate this .nupkg file that doesn't require using an empty project. Let me know if there is).

Thanks for calling this out - I also felt that this smells a little bit. We're trying to make everything as obvious as possible, so stuff like this probably deserves some close interrogation.

What's most weird about it, is that the packaging is not done as part of the solution build - ie if I just do `dotnet build` at a shell, or if I do "rebuild all" in my IDE, I won't end up with a Consolidated package. It's only if I run the Nuke script using `build.ps1` that I'll end up with a Consolidated package. So mixing in an empty project into the Solution is a bit strange.

The other thing that we've added using a `csproj` which is a little strange here is the `Calamari.UnmigratedCalamariFlavours`, but that's intended to be removed as soon as we've migrated all the Calamari, so I'm fine with that sticking around for now.

There is another way we could achieve this, which would be to use `nuget.exe` directly with a `.nuspec` file rather than trying to do it through `dotnet pack`, which always needs a `csproj` to operate against. It would look like this:

```
NuGetPack(s => s.SetTargetPath(BuildDirectory / "Calamari.Consolidated.nuspec")
                                                .SetVersion(NugetVersion.Value)
                                                .SetOutputDirectory(ArtifactsDirectory));
```

and the `build\Calamari.Consolidated.nuspec` file would contain:

```nuspec
<?xml version="1.0" encoding="utf-8"?>
<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
    <metadata minClientVersion="100.0.0.1">
        <id>Calamari.Consolidated</id>
        <version>$version$</version>
        <title />
        <authors>Octopus Deploy</authors>
        <requireLicenseAcceptance>false</requireLicenseAcceptance>
        <description>My package description.</description>
    </metadata>
    <files>
        <file src="..\artifacts\Calamari.*.zip" target="content\any\any\" />
    </files>
</package>
```

mjhilton:
If you haven't used NuGet like this before, one thing to note is that it has convention-based replacement of certain `$symbol$`s, as documented under [Replacement Tokens](https://docs.microsoft.com/en-us/nuget/reference/nuspec#replacement-tokens).

mjhilton (changes requested):
Let's change this over to use a `nuspec` file and keep the packaging concerns encapsulate din the Nuke project, rather than bleeding it over to the main solution. Details in comments above.

YihaoOct:
Thanks Matt, I did try using a `.nuspec` only but as you mentioned `dotnet pack` always requires a `.csproj`. And once I added that `.csproj`, it wasn't really necessary to add a `.nuspec`.
`nuget pack` would be a better solution here, I'll change to it.

mjhilton (approved):
LGTM

PR #879, opened 2022-08-01
Add Calamari flavour from project referenceto consolidate package task

IsaacCalligeros95:
Reference Calamari.AzureAppService in consolidate packages task from a project reference instead of a Nuget package.

### DIFF
--- a/build/Build.cs
+++ b/build/Build.cs
@@ -84,6 +84,8 @@ namespace Calamari.Build
         [Required]
         [GitVersion]
         readonly GitVersion? GitVersionInfo;
+        
+        static List<string> CalamariProjectsToSkipConsolidation = new List<string> { "Calamari.CloudAccounts", "Calamari.Common" };
 
         public Build()
         {
@@ -134,6 +136,7 @@ namespace Calamari.Build
                 SourceDirectory.GlobDirectories("**/bin", "**/obj", "**/TestResults").ForEach(DeleteDirectory);
                 EnsureCleanDirectory(ArtifactsDirectory);
                 EnsureCleanDirectory(PublishDirectory);
+                EnsureCleanDirectory(SashimiPackagesDirectory);
             });
 
         Target Restore =>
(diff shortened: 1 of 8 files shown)


build/Build.cs
```diff
@@ -184,15 +186,68 @@ public Build()
...
+                                                                                var frameworks = XmlTasks.XmlPeekSingle(project, "Project/PropertyGroup/TargetFrameworks") ?? XmlTasks.XmlPeekSingle(project, "Project/PropertyGroup/TargetFramework");
+
+                                                                                foreach (var framework in frameworks.Split(';'))
+                                                                                {
+                                                                                    void RunPublish(string runtime, string platform)
+                                                                                    {
+                                                                                        DotNetPublish(s => s
+                                                                                                           .SetProject(project)
+                                                                                                           .SetConfiguration(Configuration)
+                                                                                                           .SetFramework(framework)
```
mjhilton, line 214: In the calls below to `RunPublish`, we're doing things like:

`RunPublish("win-x64","win-x64") // for net5.0 windows`
`RunPublish("linux-x64","linux-x64") // for net5.0 linux`
`RunPublish("", "netfx") // for net452`

This tells me there's something not quite right with the parameters to this method. The mixed use of some variables from the outter scope and some as parameters is a little messy/confusing.

build/Build.cs
```diff
@@ -184,15 +186,68 @@ public Build()
...
+        Target PublishCalamariFlavourProjects => _ => _
+                                                          .DependsOn(Compile)
+                                                          .Executes(() =>
+                                                                    {
+                                                                        var calamariFlavourGroup = MigratedCalamariFlavours.Flavours.Select(flavour => SourceDirectory.GlobFiles($"**/{flavour}*.csproj")); //We need Calamari & Calamari.Tests
+                                                                        foreach (var calamariFlavourProjects in calamariFlavourGroup)
+                                                                        {
+                                                                            foreach (var project in calamariFlavourProjects)
+                                                                            {
+                                                                                var calamariFlavour = XmlTasks.XmlPeekSingle(project, "Project/PropertyGroup/AssemblyName");
```
mjhilton, line 203: Reading the `csproj` using the XML reader is different to what the other Targets [such as `Publish`](https://github.com/OctopusDeploy/Calamari/blob/d2ae9b171fb6119331ab02afd0793773f2a89985/build/Build.cs#L167-L192) do. It would probably be a bit cleaner and more consistent if it were possible to [use the Solution object](https://github.com/OctopusDeploy/Calamari/blob/d2ae9b171fb6119331ab02afd0793773f2a89985/build/Build.cs#L190) instead.

build/Build.cs
```diff
@@ -184,15 +186,68 @@ public Build()
...
 
-                      DoPublish(RootProjectName, Frameworks.NetCoreApp31, nugetVersion, FixedRuntimes.Portable);
-
                       // Create the self-contained Calamari packages for each runtime ID defined in Calamari.csproj
                       foreach (var rid in Solution?.GetProject(RootProjectName).GetRuntimeIdentifiers()!)
                           DoPublish(RootProjectName, Frameworks.NetCoreApp31, nugetVersion, rid);
                   });
 
+        Target PublishCalamariFlavourProjects => _ => _
+                                                          .DependsOn(Compile)
```
mjhilton, line 195: The indentation here is a bit excessive, it could be improved by aligning it to the way the other Targets have been indented.

build/Build.cs
```diff
@@ -184,15 +186,68 @@ public Build()
...
+                                                          .Executes(() =>
+                                                                    {
+                                                                        var calamariFlavourGroup = MigratedCalamariFlavours.Flavours.Select(flavour => SourceDirectory.GlobFiles($"**/{flavour}*.csproj")); //We need Calamari & Calamari.Tests
+                                                                        foreach (var calamariFlavourProjects in calamariFlavourGroup)
```
mjhilton, line 199: I wonder if we could make this a little easier to maintain with some Linq instead of nested loops and conditionals?

mjhilton (changes requested):
I think overall we've got this to "make it work", but we've got a bit of clean-up to do in the PublishCalamariFlavourProjects target to ensure our future selves are able to understand and maintain it.

build/Build.cs
IsaacCalligeros95: I've updated is a bit let me know what you think

build/Build.cs
IsaacCalligeros95: I've refactored this method, the original code came directly from the [Sashimi.AzureAppService publish](https://github.com/OctopusDeploy/Sashimi.AzureAppService/blob/d5e9e0e8ccb3daf44f241e232b68d014c06392e3/build/Build.cs#L111).

mjhilton:
Hey @IsaacCalligeros95, I was thinking of a more substantial change :) 

I've done a refactor in a branch off this one, and raised a PR for your consideration:

* https://github.com/OctopusDeploy/Calamari/pull/880

It's targeting _your_ branch, so if you're happy with it (and you can test/validate that I haven't completely broken what you already had working), then you can merge this PR into your branch, and then we can merge that into `master`.

build/Build.cs
```diff
@@ -329,6 +391,16 @@ public Build()
                                     }
                                 }
 
+                                foreach (var flavour in MigratedCalamariFlavours.Flavours)
+                                {
+                                    packageReferences.Add(new BuildPackageReference
+                                    {
+                                        Name = flavour,
+                                        Version = NugetVersion.Value,
+                                        PackagePath = $"{ArtifactsDirectory}\\{flavour}.zip"
```
mik-ky, line 400: Is this a typo or is it supposed to be `\\`?

mjhilton (approved):
LGTM from a code perspective :)

build/Build.cs
IsaacCalligeros95: Supposed to be \ escapes the next character so \\ is required to use a \. Is this correct for Mac aswell?

mik-ky: Yeah duh. Ignore this, dumb comment in the morning 🐵 
Wasn't apparent to me that we had to escape when creating a path as we generally use forwardslashes instead of a backslash

PR #881, opened 2022-08-02
Remove min nuget version

YihaoOct:
There was a previous error in `Calamari.Consolidated.nuspec` that prevents server from consuming the generated package. This is fixed in this PR.


mjhilton (approved):
No reason we should need a specific NuGet client version, we're doing nothing funky, so LGTM.

mjhilton (approved):
LGTM

PR #883, opened 2022-08-03
Transfer over existing blackbox test from existing ConsolidateCalamariPackagesTask repository

mik-ky:
Transfers existing tests inside [ConsolidateCalamariPackagesTask repo](https://github.com/OctopusDeploy/ConsolidateCalamariPackagesTask/blob/master/source/Tests/IntegrationTests.cs).


source/Calamari.ConsolidateCalamariPackages.Tests/IntegrationTests.cs
```diff
@@ -0,0 +1,109 @@
...
+            Directory.Delete(temp, true);
+            Directory.Delete(TestPackagesDirectory, true);
+        }
+
+        public void GivenABunchOfPackageReferences()
+        {
+            var artifacts = Directory.GetFiles(TestPackagesDirectory, "*.nupkg");
+            foreach (var artifact in artifacts)
+            {
+                using (var zip = ZipFile.OpenRead(artifact))
```
IsaacCalligeros95, line 49: I'm a little torn on whether it's worthwhile adding a Calamari.Flavour.Zip to the TestPackagesDirectory and making sure the migrated flavours handle pathing correctly. Thinking about it more it seems a bit overkill but just wanted to raise it.

IsaacCalligeros95 (approved):
LGTM, only thought was whether it was worthwhile adding a test to handle Calamari Flavour zips, it seems redundant to me but just wanted to flag it.

source/Calamari.ConsolidateCalamariPackages.Tests/IntegrationTests.cs
mik-ky: I'd like to leave it as-is for the time being, as this is the only certain input that we know of for sure, that produces what we expect. I would not be confident about modifying the expected result(s) based on the new packages we provide (unless we've thoroughly tested it in server etc which we haven't yet done).

PR #895, opened 2022-08-12
Migrate tests for Azure App Service

YihaoOct:
Migrate the integration tests for Sashimi to Calamari.

### DIFF
--- a/source/Calamari.AzureAppService.Tests/AzureWebAppHealthCheckActionHandlerFixtures.cs
+++ b/source/Calamari.AzureAppService.Tests/AzureWebAppHealthCheckActionHandlerFixtures.cs
@@ -16,6 +16,7 @@ using OperatingSystem = Microsoft.Azure.Management.AppService.Fluent.OperatingSy
 namespace Calamari.AzureAppService.Tests
 {
     [TestFixture]
+    [NonParallelizable]
     class AzureWebAppHealthCheckActionHandlerFixtures
     {
         const string NonExistentProxyHostname = "non-existent-proxy.local";
@@ -25,6 +26,7 @@ namespace Calamari.AzureAppService.Tests
         readonly string originalProxyPort = Environment.GetEnvironmentVariable(EnvironmentVariables.TentacleProxyPort);
         
         [Test]
+        [NonParallelizable]
         public async Task WebAppIsFound_WithAndWithoutProxy()
         {
             IAzure azure = null;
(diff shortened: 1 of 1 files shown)


mjhilton (commented):
Nice.

If I've read it correctly, the proxy test is now testing the same thing as the `AzureWebAppHealthCheckActionHandlerFixtures.WebApp_Is_Found` test, so I think we can consolidate the `ProxyFixture` down into the `ActionHandlerFixture`.

What if we replace `WebApp_Is_Found` with `WebAppIsFound_WithAndWithoutProxy`, and remove the `ProxyFixture`?

source/Calamari.AzureAppService.Tests/AzureWebAppHealthCheckActionHandlerFixtures.cs
```diff
@@ -0,0 +1,173 @@
...
+                
+                await CommandTestBuilder.CreateAsync<HealthCheckCommand, Program>()
+                                        .WithArrange(context => SetUpVariables(context, resourceGroup.Name, webApp.Name))
+                                        .WithAssert(result => result.WasSuccessful.Should().BeTrue())
+                                        .Execute();
+                
+                // Here we verify whether the proxy is correctly picked up
+                // Since the proxy we use here is non-existent, health check to the same Web App should fail due this this proxy setting
+                SetLocalEnvironmentProxySettings(NonExistentProxyHostname, NonExistentProxyPort);
+                SetCiEnvironmentProxySettings(NonExistentProxyHostname, NonExistentProxyPort);
```
mjhilton, line 46: Do we need to un-set these in the clean-up of the test? I'm wondering if we could end up with these settings impacting our `WebAppIsNotFound` now that we've moved them to be in the same fixture 😬

mjhilton (approved):
Just one thing to check, other than that it looks good.

source/Calamari.AzureAppService.Tests/AzureWebAppHealthCheckActionHandlerFixtures.cs
YihaoOct: Thanks Matt. Good catch. I've modified the set up and clean up a bit, so now it should work correctly in both tests:
- it keeps a record of the original settings before any test runs
- for the IsFound test, the proxy settings will be changed in test, and will be restored in the `TearDown` method after the test is finished ([even if the test terminates with an exception](https://docs.nunit.org/articles/nunit/writing-tests/attributes/teardown.html))
- for the IsNotFound test, the proxy settings will not change, but the `TearDown` method will run in the end anyway. This takes no effect since it just changes the settings to the original values (without the modification, this could change settings to `null` and `""` which may cause problems in other tests)

PR #898, opened 2022-08-18
Calamari package reference to project reference updates

IsaacCalligeros95:
# Background

As a part of dependency management work, we're migrating the Calamari half of Sashimi.Flavour projects to this Calamari repository. In this process a number of migrated projects reference nuget packages produces from this solution. This PR addresses this, namespace issues and a few QOL issues in the build process (pathing and updating server with the new Consolidated package).

# How to review this PR
* build.cs changes to fix hard coded paths
* SetOctopusServerCalamariVersion in build.cs should update the project reference in the local Octopus.Server project
* Package References for Calamari projects should be upgraded to Project References
* Calamari.AzureAppService namespaces should updates to reflect their project namespace and prevent overlapping with the Calamari.Azure project

# Results
Target discovery using AKS: 
image

Target discovery for an AzureWebApp:
image


source/Calamari.Common/Features/Discovery/AccountAuthenticationDetails.cs
```diff
@@ -17,14 +16,6 @@ public AccountAuthenticationDetails
             AccountDetails = accountDetails;
         }
 
-        public AccountAuthenticationDetails(
```
IsaacCalligeros95, line 20: Overall I don't have a good understanding of target discovery and this change is more of a spike/showing what I think this should look like. Please let me know if there are any issues with this change.

Deserializing with both constructors was throwing errors:
`Deserialization of types without a parameterless constructor, a singular parameterized constructor, or a parameterized constructor annotated with 'JsonConstructorAttribute' is not supported. Type 'Calamari.Common.Features.Discovery.AccountAuthenticationDetails`1[Calamari.AzureAppService.Azure.ServicePrincipalAccount]'. Path: $.authentication | LineNumber: 8 | BytePositionInLine: 23.`

You cannot have two JsonConstructor attributes to a class and I'm assuming the original constructor with a null type was maintained for Calamari Flavours that aren't updated to include the Type property. 
I've added this property in the test config and am assuming that targetDiscoveryContext coming in from server will have a Type attribute for the AzureAppService, is this the case?

source/Calamari.AzureAppService.Tests/TargetDiscoveryBehaviourIntegrationTestFixture.cs
```diff
@@ -288,6 +289,7 @@ private void CreateVariables(RunningDeployment context)
         ""roles"": [""{Role}""]
     }},
     ""authentication"": {{
+        ""type"": ""{Type}"",
```
IsaacCalligeros95, line 292: I've added this type to the authenticationDetails, is this a valid assumption that it'll come from server with a type?

scottmerchant: I think I added the `Type` property in server is just used by Telemetry but it probably doesn't hurt to add it in.

build/Build.cs
```diff
@@ -379,7 +379,8 @@ public Build()
                            });
 
         Target PackageConsolidatedCalamariZip =>
-            _ => _.DependsOn(CalamariConsolidationTests)
+            _ => _
+                 //.DependsOn(CalamariConsolidationTests)
```
scottmerchant, line 383: Was this on purpose?

build/Build.cs
```diff
@@ -564,12 +564,12 @@ void DoPackage(string project, string framework, string version, string? runtime
...
 
-        // Sets the Octopus.Server.csproj <CalamariVersion> property
+        // Sets the Octopus.Server.csproj Calamari.Consolidated package version
         void SetOctopusServerCalamariVersion(string projectFile)
         {
             var text = File.ReadAllText(projectFile);
-            text = Regex.Replace(text, @"<CalamariVersion>([\S])+<\/CalamariVersion>",
-                                 $"<CalamariVersion>{NugetVersion.Value}</CalamariVersion>");
+            text = Regex.Replace(text, @"<PackageReference Include=""Calamari.Consolidated"" Version=""([\S])+"" />",
+                                 $"<PackageReference Include=\"Calamari.Consolidated\" Version=\"{NugetVersion.Value}\" />");
```
scottmerchant, line 572: Nice 👍

source/Calamari.Common/Features/Discovery/AccountAuthenticationDetails.cs
scottmerchant: It's weird that it threw that error as the top constructor is marked with `[JsonConstructor]`... Either way I think you're right about why the original constructor still exists. Are all those flavours merged into this solution now?

scottmerchant (commented):
I think the changes look good from a Target Discovery perspective. I'm just a bit wary of the removed constructor in `AccountAuthenticationDetails`.

source/Calamari.Common/Features/Discovery/AccountAuthenticationDetails.cs
IsaacCalligeros95: Yeah I was surprised by this behaviour, it really didn't make sense to me but this way works... Currently we only have AzureAppService, AzureResourceGroup and soon to be AzureWebApp migrated. All other packages would be using a pinned package reference for Calamari.Common so even if this deprecated support for the standalone flavour projects I don't think that's an issue. We'll be updating to project references when migrating and having to make sure tests pass/this doesn't cause an issue anyway.

build/Build.cs
IsaacCalligeros95: Re-enabled

source/Calamari.Common/Features/Discovery/AccountAuthenticationDetails.cs
scottmerchant: I think the only other flavour that would use this type would be AzureWebApp as it's the only other step which has target discovery. (AWS ECS also supported target discovery but it's in a step package which means it's done differently).
As long as AzureWebApp is happy I think this change should be ok.

source/Calamari.AzureAppService/Behaviors/TargetDiscoveryBehaviour.cs
```diff
@@ -175,6 +174,12 @@ private IEnumerable<IDeploymentSlot> ListDeploymentSlots(IWebApp webApp)
             TargetDiscoveryContext<AccountAuthenticationDetails<ServicePrincipalAccount>> context,
             TargetMatchResult matchResult)
         {
+            if (context.Authentication?.AccountId == null)
```
mjhilton, line 177: I think these changes make sense, but not sure what the impetus for them was. Something about nullables?

mjhilton (approved):
Overall this looks good.

We have comprehensive tests around `TargetDiscovery`, so although there are some changes there that I wouldn't have expected, it looks like it's necessary clean-up which would have already happened had there not been the NuGet-reference separation in the first place.

I think it would be worth a smoke-test of Target Discovery in each of these Calamari-provided Steps to ensure things are working; we might be able to get some help from @scottmerchant or @geofflamrock there, given they know the functionality deeply.

source/Calamari.Common/Features/Discovery/AccountAuthenticationDetails.cs
geofflamrock: I believe kubernetes discovery will also use this class for both AKS and EKS, so might be worth checking that too.

scottmerchant: Hrm yes you're right. I'm happy to run through doing discovery with you @IsaacCalligeros95 on AzureWebApp, K8s EKS/AKS to smoke test it, if you want. I'll try and get my PR to Calamar in first so that AKS is fixed.

geofflamrock: @gb-8 might also remember if there were any reasons this extra constructor was included previously. One thought is that older versions of Server may not have sent the `Type` property as part of the authentication property on the discovery context, assuming these changes are only going to the latest Server release and not back to LTS branches then removing the constructor I imagine would be fine.

source/Calamari.AzureAppService/Behaviors/TargetDiscoveryBehaviour.cs
geofflamrock: The `AccountId` property is marked as non-nullable and should always be provided by Server, so I'm also wondering why this check needs to be done, given we are already checking up above if the `Authentication` property is null.

source/Calamari.Common/Features/Discovery/AccountAuthenticationDetails.cs
```diff
@@ -17,14 +16,6 @@ public AccountAuthenticationDetails
...
-            string accountId, TAccountDetails accountDetails)
-        {
-            Type = null;
-            AccountId = accountId;
-            AccountDetails = accountDetails;
-        }
-        
         public string? Type { get; set; }
```
geofflamrock, line 19: If the `Type` property is always going to be supplied by Server now (and we only have one constructor with a non-nullable parameter), then we might consider making this non-nullable?

source/Calamari.Common/Features/Discovery/AccountAuthenticationDetails.cs
IsaacCalligeros95: @scottmerchant A walk-through target discovery would be good, just let me know when it suits you

source/Calamari.AzureAppService/Behaviors/TargetDiscoveryBehaviour.cs
IsaacCalligeros95: @mjhilton This change came in because Authentication and Scope became nullable on `TargetDiscoveryContext` between the Calamari.Common project AzureAppService was referencing and now. Type is used for logging the authentication type I believe, i.e. for AzureServicePrincipal authentication it's just Azure. There's a bit more context on this PR https://github.com/OctopusDeploy/Calamari/pull/901/files

IsaacCalligeros95: @geofflamrock The earlier null check means this should never be the case but `CreateWebAppTargetCreationServiceMessage` method expects a string for `AccountId` and with Authentication being nullable I added this for compilation. I've updated this to just check `context.Authentication` for null instead. Is there a better way to handle this?

geofflamrock: The methods we've added these to are private and are only called from with a code path that happens after we've check that `context.Authentication` is null, is that right? We could consider just telling the compiler that we know it isn't null at this point with `context.Authentication!.AccountId`. I'm not sure if we have any guidance on whether we should/shouldn't be doing that within the codebase, @mjhilton do you know of anything in this regard?

mjhilton: It's not an area I know about or have thought about enough to have an opinion on. I don't believe we have any explicit guidance.

If you're happy with this approach, then I trust that, and am too.

mjhilton (approved):
LGTM again :)

PR #900, opened 2022-08-18
Consolidate Sashimi.AzureWebApp's Calamari.* projects

mik-ky:
This PR consolidates the `Calamari.*` projects into Calamari from the existing Sashimi.AzureWebApp repository as part of dependency consolidation work.


IsaacCalligeros95 (commented):
LICENSE.txt is coming from Calamari this should stay right?

source/Calamari.AzureWebApp/Calamari.AzureWebApp.csproj
```diff
@@ -0,0 +1,24 @@
...
+    <AssemblyName>Calamari.AzureWebApp</AssemblyName>
+    <RootNamespace>Calamari.AzureWebApp</RootNamespace>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <IsPackable>false</IsPackable>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net452</TargetFramework>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="Calamari.AzureScripting" Version="14.2.0" />
+    <PackageReference Include="Calamari.Common" Version="21.1.0" />
```
IsaacCalligeros95, line 12: Project Reference for Calamari.Common

IsaacCalligeros95 (changes requested):
Nice! just a couple of comments, you'll also need to add the Calamari.AzureWebApp flavour to MigratedCalamariFlavours.cs file

source/Calamari.AzureWebApp/Calamari.AzureWebApp.csproj
mik-ky: As Calamari.Common targets .netstandard on unix machines and won't be compatible for AzureWebApp, let's keep this as is for now.

source/Calamari.AzureWebApp.Tests/Calamari.AzureWebApp.Tests.csproj
```diff
@@ -0,0 +1,24 @@
...
+  <PropertyGroup>
+    <RootNamespace>Calamari.AzureWebApp.Tests</RootNamespace>
+    <AssemblyName>Calamari.AzureWebApp.Tests</AssemblyName>
+    <TargetFramework>net461</TargetFramework>
+    <LangVersion>8.0</LangVersion>
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\Calamari.AzureWebApp\Calamari.AzureWebApp.csproj" />
+    <PackageReference Include="Calamari.Tests.Shared" Version="14.2.0" />
```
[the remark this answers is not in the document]
mik-ky: Keeping this as is as it's in the same veins as https://github.com/OctopusDeploy/Calamari/pull/900#discussion_r949864968

IsaacCalligeros95 (approved):
LGTM

build/Build.cs
```diff
@@ -219,22 +219,43 @@ public Build()
...
-                        .SelectMany(packageToBuild => packageToBuild.Project.GetRuntimeIdentifiers(), (packageToBuild, runtimeIdentifier) => new
-                        {
-                            packageToBuild.Project,
-                            packageToBuild.Framework,
-                            Architecture = packageToBuild.CrossPlatform ? runtimeIdentifier : null,
-                            packageToBuild.CrossPlatform
-                        })
-                        .Distinct(t => new { t.Project.Name, t.Architecture, t.Framework });
+                    // for NetFx target frameworks, we use "netfx" as the architecture, and ignore defined runtime identifiers
+                    var netFxPackages = calamariFlavours
```
IsaacCalligeros95, line 223: Can we pull the below snippet into it's own variable and share between the netfx and cross-platform packages?
`calamariFlavours
                      .SelectMany(project => project.GetTargetFrameworks()!, (p, f) => new
                      {
                          Project = p,
                          Framework = f,
                          CrossPlatform = IsCrossPlatform(f)
                      })`

build/Build.cs
```diff
@@ -219,22 +219,43 @@ public Build()
...
+                          Project = p,
+                          Framework = f,
+                          CrossPlatform = IsCrossPlatform(f)
+                      })
+                      .Where(p => !p.CrossPlatform)
+                      .Select(packageToBuild => new
+                      {
+                          packageToBuild.Project,
+                          packageToBuild.Framework,
+                          Architecture = (string)null!,
```
IsaacCalligeros95, line 235: should this just be null?

build/Build.cs
```diff
@@ -262,9 +283,15 @@ public Build()
                     foreach (var flavour in calamariFlavours)
                     {
                         Log.Verbose($"Compressing Calamari flavour {PublishDirectory}/{flavour.Name}");
-                        CompressionTasks.CompressZip(PublishDirectory / flavour.Name, $"{ArtifactsDirectory / flavour.Name}.zip");
+                        var compressionSource = PublishDirectory / flavour.Name;
+                        if (!Directory.Exists(compressionSource))
```
IsaacCalligeros95, line 287: I think this should error out If the directory doesn't exist and we're not copying the flavour across, am I missing something here? If not should we just throw an exception or let the CompressZip method throw an exception?

build/Build.cs
YihaoOct: Just use null here give a warning in the IDE, it suggests me to write it like so

build/Build.cs
YihaoOct: I added this to suppress a local error. For flavours like this that only has netfx (windows only) targets, when building it on Mac, since every package was simply skipped, the output directory ends up not being created. So at the compression step it will give an error saying that the directory does not exist.

IsaacCalligeros95: Ahh that makes sense, I'm assuming building on mac skipped these packages previously aswell?

YihaoOct: Yes, window-only builds were skipped on Mac. In CI they should work fine.

IsaacCalligeros95: All good by me then 🙂

PR #901, opened 2022-08-19
Fix AzureKubernetesDiscoverer Name to match what server sends

scottmerchant:
# Background

After looking at @IsaacCalligeros95's [PR](https://github.com/OctopusDeploy/Calamari/pull/898) I realised I had introduced a bug in K8s target discovery when I [added Target Discovery Telemetry](https://github.com/OctopusDeploy/OctopusDeploy/pull/12707). It broke because I forgot that K8s target discovery uses the Type property on the Authentication Context payload to figure out which type of discoverer to use (AWS or Azure). Previously the `Type` value for the Azure payload would be `"AzureServicePrinciple"` but now it is `"Azure"` as the authentication method eg: `"ServicePrinciple"` is stored on a separate property (which isn't including in Calamari, as it's only used in Telemetry on Server). 

# Results
I've updated the value of the `Name` property on `AzureKubernetesDiscoverer` to `"Azure"` to match what the server sends but I've also updated the `Name` property to be called `Type` to match the property name on the Authentication Context.

Story: [sc-9442]

### DIFF
--- a/source/Calamari.Azure/Kubernetes/Discovery/AzureKubernetesDiscoverer.cs
+++ b/source/Calamari.Azure/Kubernetes/Discovery/AzureKubernetesDiscoverer.cs
@@ -14,7 +14,7 @@ namespace Calamari.Azure.Kubernetes.Discovery
         {
         }
 
-        public override string Name => KubernetesAuthenticationContextTypes.AzureServicePrincipal;
+        public override string Name => "Azure";
 
         public override IEnumerable<KubernetesCluster> DiscoverClusters(string contextJson)
         {
--- a/source/Calamari.Azure/Kubernetes/Discovery/AzureKubernetesDiscoverer.cs
+++ b/source/Calamari.Azure/Kubernetes/Discovery/AzureKubernetesDiscoverer.cs
@@ -14,7 +14,7 @@ namespace Calamari.Azure.Kubernetes.Discovery
         {
         }
 
-        public override string Name => "Azure";
+        public override string Type => "Azure";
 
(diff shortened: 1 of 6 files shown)


source/Calamari.Azure/Kubernetes/Discovery/AzureKubernetesDiscoverer.cs
```diff
@@ -14,7 +14,7 @@ public AzureKubernetesDiscoverer(ILog log) : base(log)
         {
         }
 
-        public override string Name => KubernetesAuthenticationContextTypes.AzureServicePrincipal;
```
gb-8, line 17: Can we reduce the chance of a similar problem in future by defining these  in constant strings that get referenced, instead of using string literals in multiple places? Might be hard because of the Calamari/Server split?
Whatever we end up doing here, we could also re-visit it when Calamari is all back in the server solution.

scottmerchant: At this stage I don't think we have any shared projects across server and Calamari so I think the best option is to revisit this when calamari is merged into the Server Solution.

gb-8: Ok. As per our conversation, let's just get some comments here and for the equivalent `"Aws"` type, indicating which server code they need to keep in sync with, and noting that we can revisit to move to shared code after the mono-repo changes are done.

gb-8 (approved):
Pre-approved, pending those code comments we discussed.

PR #905, opened 2022-08-26
Make nuke work on Mac

YihaoOct:
This PR makes sure the nuke build can work on Mac as well as Windows:

- Added conditional compile checks to migrated flavours that have targeted frameworks other than .net framework
- Added a `Calamari.NonWindows.sln` file that excludes projects that can only be compiled on Windows
- Updated nuke to pick up the non windows solution file on non-windows environments
- Updated AzureWebApp to use project reference, which we did not update previously
- Fixed an error in one of the previous build target which prevents the build to be executed (even on Windows I think)

### DIFF
--- a/build/Build.cs
+++ b/build/Build.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.IO.Compression;
 using System.Linq;
+using System.Runtime.InteropServices;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Calamari.ConsolidateCalamariPackages;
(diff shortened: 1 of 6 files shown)


source/Calamari.AzureWebApp/Calamari.AzureWebApp.csproj
```diff
@@ -9,7 +9,6 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Calamari.AzureScripting" Version="14.2.0" />
-    <PackageReference Include="Calamari.Common" Version="21.1.0" />
```
YihaoOct, line 12: We did not change this in the previous migration

build/Build.cs
```diff
@@ -456,7 +458,7 @@ public Build()
         
         Target UpdateCalamariVersionOnOctopusServer =>
             _ =>
-                _.Requires(() => SetOctopusServerVersion)
+                _.OnlyWhenStatic(() => SetOctopusServerVersion)
```
YihaoOct, line 461: I think the intention of this line is to only run this target when we set `SetOctopuServerVersion` to true. `Requires` is the wrong method to use here, since if `SetOctopuServerVersion` is false, it just prevents the whole build from running, rather than skipping this particular target.

mjhilton (approved):
LGTM, thankyou for putting effort into making this work for all platforms :)

PR #909, opened 2022-09-12
Azure App Sevice - improve ability to understand what's going on deployments fail

mjhilton:
The Azure App Service step frequently generates support requests, because there is minimal logging, and when it fails, it does so in a very generic way with little supporting information to help customers reason about what might be going wrong.

This PR improves the debuggability of the Azure App Service step by adding much more logging, and failing fast when there are unrecoverable conditions:

* Expands exception message for the Zip Upload step, which is where the step currently fails most often
* Adds verbose logging to each step of the process to help pinpoint failures
* Adds earlier checks for the following unrecoverable conditions:
    * Fail fast if Resource Group can't be found
    * Fail fast if App Service can't be found

### DIFF
--- a/source/Calamari.AzureAppService/Behaviors/AzureAppServiceBehaviour.cs
+++ b/source/Calamari.AzureAppService/Behaviors/AzureAppServiceBehaviour.cs
@@ -37,15 +37,50 @@ namespace Calamari.AzureAppService.Behaviors
 
         public async Task Execute(RunningDeployment context)
         {
+            Log.Verbose("Starting Azure App Service deployment.");
+            
             var variables = context.Variables;
             var servicePrincipal = ServicePrincipalAccount.CreateFromKnownVariables(variables);
-            string? webAppName = variables.Get(SpecialVariables.Action.Azure.WebAppName);
-            if (webAppName == null)
-                throw new Exception("Web App Name must be specified");
+            Log.Verbose($"Using Azure Tenant '{servicePrincipal.TenantId}'");
+            Log.Verbose($"Using Azure Subscription '{servicePrincipal.SubscriptionNumber}'");
+            Log.Verbose($"Using Azure ServicePrincipal AppId/ClientId '{servicePrincipal.ClientId}'");
+            Log.Verbose($"Using Azure Cloud '{servicePrincipal.AzureEnvironment}'");
+            
             string? resourceGroupName = variables.Get(SpecialVariables.Action.Azure.ResourceGroupName);
             if (resourceGroupName == null)
                 throw new Exception("resource group name must be specified");
+            Log.Verbose($"Using Azure Resource Group '{resourceGroupName}'.");
+            
+            string? webAppName = variables.Get(SpecialVariables.Action.Azure.WebAppName);
+            if (webAppName == null)
+                throw new Exception("Web App Name must be specified");
+            Log.Verbose($"Using App Service Name '{webAppName}'.");
+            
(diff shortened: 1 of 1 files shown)


source/Calamari.AzureAppService/Behaviors/AzureAppServiceBehaviour.cs
```diff
@@ -37,15 +37,50 @@ public bool IsEnabled(RunningDeployment context)
...
+            Log.Verbose($"Checking existence of Resource Group '{resourceGroupName}'.");
+            if (!(await azureClient.ResourceGroups.ContainAsync(resourceGroupName)))
+            {
+                Log.Error($"Resource Group '{resourceGroupName}' could not be found. Either it does not exist, or the Azure Account in use may not have permissions to access it.");
+                throw new Exception("Resource Group not found.");
+            }
+            Log.Verbose($"Resource Group '{resourceGroupName}' found.");
+            
+            Log.Verbose($"Checking existence of App Service '{targetSite.Site}'.");
+            var webApp = await azureClient.WebApps.GetByResourceGroupAsync(resourceGroupName, targetSite.Site);
```
IsaacCalligeros95, line 76: I'm assuming the WebAppName -> targetSite.Site here is a non-change and targetSite.Site and webAppName are equivalent. Are there any cases where this would be true?

IsaacCalligeros95 (approved):
LGTM, one question on the targetSite.Site usage. I'm assuming this is intentional and correct but don't have an in depth understanding.

source/Calamari.AzureAppService/Behaviors/AzureAppServiceBehaviour.cs
mjhilton: `webAppName` is raw parameter passed into the step. Before deployment is done, there's some potential manipulation that goes on. From the code's perspective, it's possible for `webAppName` to be either just the name of the App Service, or some kind of legacy naming format like `webappname/slotname` or `webappname(slotname)`.

The point of [the `AzureWebAppHelper.GetTargetSite` code](https://github.com/OctopusDeploy/Calamari/pull/909/files#diff-c3b5187131bd38d91701a36b8fa99b9f90254dadf07d689bc0d756d4e38f5914L68) is to extract a separate AppService and Slot name from those potential formats. So, it doesn't make a great deal of sense in the initial checks for us to be using that raw `webAppName`.

mjhilton: Hmmm thinking about it now though, there's a little too much risk for unexpected breakages here given the scope I intended for this change (basically logging-only).

I'm going to roll back the part where I changed from WebAppName -> TargetSite.

mjhilton: Ooooh I remember why I did it now. 

That Azure SDK method which retrieves the WebApp doesn't throw an exception if it doesn't exist; it just returns `null`. So if there was no slot specified, previously, it would just silently fail there as we did no null-checking on the result. It wouldn't fail until we got down later to trying to do the Zip Upload.

So I thought that we should be consistent with what we're using to call Azure when we provide the WebApp Name. The existing deployment code uses `targetSite.Site`, so we should use that everywhere.

I changed my mind back again: this was intentional, and it's part of the "fail fast" change. I'd like to keep it.

mik-ky (approved):

PR #916, opened 2022-09-14
Improve performance of azure web app discovery

scottmerchant:
# Background

When implementing Parallel target discovery, we noticed that Azure Web App Target Discovery was extremely slow especially when the azure subscription being searched was full on Azure Web Apps with numerous slots.

# Result

We have decided to move away from the old fluent Azure .NET Client `IAzure` and use the new one `ArmClient` to improve performance. In this way we can search for all Azure Web Apps and Web App Slots in two (paged) queries and then we only need to get the details of apps and slots which have matching tags.

This method is a lot faster and requires far less web requests overall.

As a plus, this also works with Azure Functions Apps so we also get that win for free.

Relates to https://github.com/OctopusDeploy/ResearchAndDevelopment/issues/596

stories: [[sc-10048](https://app.shortcut.com/octopusdeploy/story/10048/azure-web-app-target-discovery-does-not-seem-to-scale-well-when-used-in-an-azure-environment-with-dozens-of-azure-web-apps)], [[sc-10688](https://app.shortcut.com/octopusdeploy/story/10688/add-discovery-of-azure-function-apps-to-azure-web-app-target-discovery)]


source/Calamari.AzureAppService/Behaviors/TargetDiscoveryBehaviour.cs
```diff
@@ -102,101 +85,46 @@ public async Task Execute(RunningDeployment runningDeployment)
...
+            // so the two required elements can be extracted.
+            if (resource.Type.Equals(AzureRestClient.WebAppSlotsType, StringComparison.InvariantCultureIgnoreCase))
+            {
+                var indexOfSlash = resourceName.LastIndexOf("/", StringComparison.InvariantCulture);
+                if (indexOfSlash >= 0)
+                {
+                    slotName = resourceName[(indexOfSlash + 1)..];
+                    resourceName = resourceName[..indexOfSlash];
+                }
+            }
```
scottmerchant, line 118: This was the best way I could get the Web App Name and the Slot Name from a Slot Resource. I'm not sure if this is too brittle though. I don't think there is really a better option though unfortunately.

geofflamrock: Given we need to make an additional api call to get the resource group on each discovered resource anyway, I wonder if its worth seeing if we could use the client to query the webapp/slot resource properties in a more typed way, that way we could have more safety around getting the slot name as well as the other resource properties?

scottmerchant: I just went on another voyage of discovery with this library. I can actually get all the info I need from the `Id` property - of type `ResourceIdentifier`. So we don't actually need a second call at all, and I don't need to parse the name into web app name and slot name! 👍

source/Calamari.AzureAppService/Behaviors/TargetDiscoveryBehaviour.cs
```diff
@@ -1,25 +1,35 @@
...
+using JsonSerializer = System.Text.Json.JsonSerializer;
 
 #nullable enable
 namespace Calamari.AzureAppService.Behaviors
 {
     public class TargetDiscoveryBehaviour : IDeployBehaviour
     {
+        private const string WebAppSlotsType = "sites/slots";
+        private const string WebAppType = "sites";
+        private const int PageSize = 500;
```
geofflamrock, line 32: If there are more than 500 resources are we handling that anywhere? Or is it handled by the client?

geofflamrock (approved):
Looks good, just had one question about paging

source/Calamari.AzureAppService/Behaviors/TargetDiscoveryBehaviour.cs
scottmerchant: I tested this, the PageSize we pass in doesn't stop us from getting all the resources if we simply iterate the async collection, it will actually get us all the resources but it will request them in batches of PageSize when it does it. It's really nice and clean for us 👍

geofflamrock: Ah nice one 👍 how did we settle on a page size of 500?

scottmerchant: To be honest I just kinda napkin mathed it - knowing we had ~250 results coming back from a single request and that took ~1-1.5 seconds I thought double that would be the max I would want to query for - maybe that's too long? Paging too low does seem to slow down the process overall. I did some tests where I set it to 10, 100, 200 and 10 and 100 were slower, 200 was pretty good, a single request was the fastest. I guess that makes sense. But for someone with 1000 web apps/slots I don't know what would be best.

geofflamrock: Hmmm yeah I think that seems reasonable given what you've described there, we want to balance the number of requests with the size of each request and try to find something nice in the middle. Do the azure docs for the sdk have any guidance?

geofflamrock: We have also started to use the sdk in other places too right? Do we use paged queries anywhere there that could give us a guide on what a good number could be?

scottmerchant: > Do the azure docs for the sdk have any guidance?

I've looked around a bit and I couldn't find any recommendations on page size. I guess it depends on the query you're making. For instance the performance is a lot worse if you're filtering by tags so you probably want a lower page size. When filtering by resource type, it's quite quick. 

> We have also started to use the sdk in other places too right? Do we use paged queries anywhere there that could give us a guide on what a good number could be?

It's only used in one other place and unfortunately for us, no paging is used.

geofflamrock: > I've looked around a bit and I couldn't find any recommendations on page size. I guess it depends on the query you're making. For instance the performance is a lot worse if you're filtering by tags so you probably want a lower page size. When filtering by resource type, it's quite quick.

Yeah makes sense, I think based on your testing so far 500 seems reasonable, if it starts causing problems we can revisit.

PR #921, opened 2022-09-21
Added RetryTracker to file move. 

voronoipotato:
There appear to be one or two other uses of File.Move instead of the filesystem.MoveFile, however this PR should fix https://github.com/OctopusDeploy/Issues/issues/7775 . The ApplyDeltaCommand already leverages the CalamariPhysicalFileSystem for other operations so it appears File.Move was overlooked because it previously was just a wrapper around File.Move. 

Hope it helps!

### DIFF
--- a/source/Calamari.Common/Plumbing/FileSystem/CalamariPhysicalFileSystem.cs
+++ b/source/Calamari.Common/Plumbing/FileSystem/CalamariPhysicalFileSystem.cs
@@ -564,7 +564,7 @@ namespace Calamari.Common.Plumbing.FileSystem
 
         public void MoveFile(string sourceFile, string destinationFile)
         {
-            File.Move(sourceFile, destinationFile);
+             RetryTrackerFileAction(() => File.Move(sourceFile, destinationFile), destinationFile, "move");
         }
 
         public void EnsureDirectoryExists(string directoryPath)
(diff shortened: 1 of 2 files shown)


voronoipotato:
@mik-ky I think your build runner is hanging.

voronoipotato:
@YihaoOct  hello :) I noticed you've been able to merge PR's. Is there something needed to move this forward or some way I could help?

voronoipotato:
Ah I see, you got approval from @IsaacCalligeros95. Sorry I pinged the wrong person! Issac, is there anything I should do to better move this along?

YihaoOct:
Hi @voronoipotato thanks for the PR! Actually, it is very strange to me why your PR does not kick off our pipeline in TC. And I couldn't find this PR in our TeamCity instance to start it manually. Initially I thought it was because it was created from a forked repo (our internal work flow is to use branches in the same repo). But I just created a [test PR](https://github.com/OctopusDeploy/Calamari/pull/936) from a forked repo, looks like that one triggers the pipeline without problem. So I think this might be a permission issue. Do you have any idea on this? @mjhilton

IsaacCalligeros95:
Hi @voronoipotato following up on Yihao's response we don't automatically trigger PR builds as the build scripts are exposed as a part of this repository. Additional security changes have prevented us from manually triggering these builds and I'm waiting on feedback to see if I'm allowed to override these settings. In the mean time I've created a [PR](https://github.com/OctopusDeploy/Calamari/pull/937) here. Apologies that it's a fresh branch, we really appreciate the contribution, and I'll have a deeper look into the change this afternoon/early next week. These sorts of issues are hard to diagnose and find the underlying cause of so thanks again! I'll keep you updated probably next week on the progress of this one.

voronoipotato:
Thanks so much!

voronoipotato:
Resolved by #938

PR #922, opened 2022-09-22
Consolidate Calamari.Terraform

YihaoOct:
This PR is part of Dependency Management work, it consolidates Calamari.Terraform into this repository.

Some other contents included in this PR:

- Migrated the requirement attributes from `Calamari.Tests` to `Calamari.Testing`
- References to `Calamari.Tests` from flavours were removed
- Changed the line-endings of AzureCloudService


source/Calamari.Terraform/TerraformSpecialVariables.cs
```diff
@@ -1,35 +1,72 @@
...
+            public static readonly string ScriptSource = "Octopus.Action.Script.ScriptSource";
+
+            public static string ScriptBodyBySyntax(ScriptSyntax syntax) => "Octopus.Action.Script.ScriptBody[" + syntax.ToString() + "]";
+
+            public static class ScriptSourceOptions
+            {
+                public const string Package = "Package";
+                public const string Inline = "Inline";
+            }
+        }
```
mik-ky, line 60: Inside Calamari.Common, we defined some of these constants (if not all) inside `ScriptVariables.cs`. As these are well-known Octopus variables used across flavours, let's just use those instead, adding in additional things we are missing from there.

mik-ky (changes requested):
Looks good, nice one! Just a small change request.

mik-ky (approved):

PR #925, opened 2022-10-07
fix Python updateprogress()

smlng:
According to the [documentation](https://octopus.com/docs/deployments/custom-scripts/logging-messages-in-scripts) the message for updateprogress in Python scripts is optional. However, when a message is not provided the script fails because the default for message is None which is then passed to the inner encode() function that does not like None. 

In this fix the default value for message is changed from None to "" (empty string). Further this PR provided necessary test fixtures to assert the correct behavior.

### DIFF
--- a/source/Calamari.Common/Features/Scripting/Python/Configuration.py
+++ b/source/Calamari.Common/Features/Scripting/Python/Configuration.py
@@ -50,7 +50,7 @@ def createartifact(path, fileName = None):
     print("##octopus[stdout-default]");
     print("##octopus[createArtifact path='{0}' name='{1}' length='{2}']".format(servicepath, serviceFileName, length))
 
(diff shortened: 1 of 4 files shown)


smlng:
Ping, just some minor changes. Do I need to do something to move this forward.

IsaacCalligeros95:
Thanks for the PR @smlng. I've created a fresh branch and PR, as our build scripts are exposed within this repository we do not automatically build community pull requests and have to trigger them manually. Some new security measures internally have prevented us from manually triggering these, and I'm waiting for confirmation whether I can. In the meantime, I've branched off of your changes and created a [fresh pull request](https://github.com/OctopusDeploy/Calamari/pull/939). I'll take a look at the changes early next week and see if we can get them merged in. Thanks again for the contribution!

PR #927, opened 2022-10-10
Allow all types of worker credentials for EKS discovery

scottmerchant:
# Background

Originally worker credentials were discovery for EKS discovery by initially trying to retrieve EC2 credentials and then attempt to use Environment variable based account credentials. This means that other valid credentials that could be used (such as a k8s pod IAM role available to all containers in that pod).

# Result

I've updated the code so that it now uses a "FallbackFactory" which will go through all the different ways AWS can authenticate and this should (pending tests) cover the case above.

Note: that AWS has a specific order in which it attempts to retrieve credentials with EC2 instance being the last one. This is an AWS factory so I think that order should be sufficient for most users and it should match other AWS SDKs (like the one used for step-package-ecs target discovery)

[sc-25801]

### DIFF
--- a/source/Calamari.Aws/Kubernetes/Discovery/AwsAuthenticationDetails.cs
+++ b/source/Calamari.Aws/Kubernetes/Discovery/AwsAuthenticationDetails.cs
@@ -9,102 +9,64 @@ namespace Calamari.Aws.Kubernetes.Discovery
     public class AwsAuthenticationDetails : ITargetDiscoveryAuthenticationDetails
     {
         const string DefaultSessionName = "OctopusKubernetesClusterDiscovery";
-        
+
         /// <exception cref="AggregateException">
         /// If both InstanceProfile and EnvironmentVariable Credentials fail.
         /// Contains AmazonClientExceptions for both InstanceProfile and EnvironmentVariable failures</exception>
         /// <exception cref="AmazonClientException">If Basic (Account) Credentials fail</exception>
-        public bool TryGetCredentials(ILog log, out AWSCredentials credentials)
+        public AWSCredentials GetCredentials()
         {
-            credentials = null;
-            if (Credentials.Type == "account")
-            {
-                try
-                {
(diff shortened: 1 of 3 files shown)


source/Calamari.Tests/KubernetesFixtures/Terraform/Clusters/EKS/eks.kubernetes.tf
```diff
@@ -13,8 +13,8 @@ provider "kubernetes" {
   token                  = data.aws_eks_cluster_auth.default.token
 }
 
-resource "local_file" "kubeconfig" {
-  sensitive_content = templatefile("${path.module}/kubeconfig.tpl", {
+resource "local_sensitive_file" "kubeconfig" {
```
scottmerchant, line 16: I was getting a test failure because `sensitive_content` is not supported in the `local_file` resource anymore. I've updated it to use the new `local_sensitive_file` instead.

source/Calamari.Aws/Kubernetes/Discovery/AwsAuthenticationDetails.cs
```diff
@@ -9,102 +9,64 @@ namespace Calamari.Aws.Kubernetes.Discovery
...
-                        log.Warn("Unable to authorise credentials, see verbose log for details.");
-                        log.Verbose($"Unable to authorise credentials for Instance Profile: {instanceProfileException}");
-                        log.Verbose($"Unable to authorise credentials for Environment Variables: {environmentVariablesException}");
-                        return false;
-                    }
-                }
-            }
+            var credentials = Credentials.Type == "account"
+                ? new BasicAWSCredentials(Credentials.Account.AccessKey, Credentials.Account.SecretKey)
+                : FallbackCredentialsFactory.GetCredentials();
```
scottmerchant, line 21: Using FallbackCredentialsFactory from AWS library to go through all fallback credential options in default order.

PR #929, opened 2022-10-12
Updating EKS authentication to use 'aws eks get token' and use the aws-iam-authenticator as a fallback

IsaacCalligeros95:
This PR addresses an issue where the EKS authentication api was hard coded to the v1alpha1 endpoint. As a part of this change due to compatibility [issues](https://github.com/aws/aws-cli/issues/6920) with `aws-iam-authenticator` versions earlier than `0.5.3` and `kubectl` versions > 1.23.6 `aws cli` authentication using `aws eks get-token` has been added and prioritised. Newer versions of kubectl  (1.24.0 and on) are using the v1beta1 endpoint which breaks on our old hardcoded v1alpha1 implementation. 

Please note: kubectl 1.24.0 requires versions of aws-iam-authenticator 0.5.5 or later (0.5.4 is a pre-release version). For versions of Kubectl 1.23.6 and earlier aws-iam-authenticator 0.5.3 and earlier should be used. The aws cli will work as expected across both of these versions. https://github.com/aws/aws-cli/issues/6920#issuecomment-1121229515)

Kubectl: 1.23.6
aws-cli 2.8.3. Health Check ✅
aws-iam-authenticator 0.5.3. Health Check ✅
aws-iam-authenticator 0.5.9. Health Check ❌

Kubectl 1.25.3
aws-cli 2.8.3. Health Check ✅
aws-iam-authenticator 0.5.3. Health Check ❌
aws-iam-authenticator 0.5.9. Health Check ✅


mjhilton:
This looks like a fairly fundamental change to the way we authorize against an EKS cluster. Completely removing the support for IAM Authenticator and requiring a new dependency for these steps is quite a big change.

I think we could do a bit more graceful implementation here, which might look something like:

```pseudocode
if (Customer Has AWS CLI in Environment) {
    if (AWS CLI Version >= 1.16.156) {
        Use "aws eks get-token" auth method 
        return
    } else {
        Echo "AWS CLI is available but needs to be upgraded to use aws eks get-token"
    }
}

if (Customer Has IAM Authenticator in Environment) {
    Notify Of IAM Authenticator Deprecation
    Use "aws-iam-authenticator" auth method 
}

Echo "Can't find any way to authenticate, you need to install AWS CLI >=  1.16.156"
```

Combined with some kind of change to the way we determine the API to use (rather than hardcoding to an old one), that should provide customers with a reasonable migration path, while still supporting the new things.

source/Calamari.Tests/KubernetesFixtures/InstallTools.cs
```diff
@@ -58,12 +59,9 @@ public async Task Install()
                 KubectlExecutable = await DownloadCli("Kubectl",
                     async () =>
                     {
```
IsaacCalligeros95, line 61: Switched back to the latest version of kubectl, this is in the context of tests and should help us spot issues earlier. This failed previously on an kubectl and aws-cli version mismatch. https://github.com/aws/aws-cli/issues/6920

source/Calamari.Tests/KubernetesFixtures/Approved/KubernetesContextScriptWrapperFixture.ExecutionWithEKS_IAMAuthenticator.approved.txt
```diff
@@ -6,5 +6,12 @@
 [Verbose] "kubectl" config use-context octocontext --request-timeout=1m
 [Verbose] "kubectl" config set-cluster octocluster --insecure-skip-tls-verify=true --request-timeout=1m
 [Info] Creating kubectl context to <server> (namespace calamari-testing) using EKS cluster name my-eks-cluster
+[Error] The EKS cluster Url specified should contain a valid aws region name
```
IsaacCalligeros95, line 9: Should I pass through the URL here?

source/Calamari.Tests/KubernetesFixtures/KubernetesRunsScriptFixture.cs
```diff
@@ -26,7 +26,7 @@ namespace Calamari.Tests.KubernetesFixtures
     [TestFixture]
     public class KubernetesContextScriptWrapperFixture
     {
-        const string ServerUrl = "<server>";
+        const string ServerUrl = "https://ThisIsAFakeServerUrlWithTheRegion.gr7.ap-southeast-2.eks.amazonaws.com";
```
IsaacCalligeros95, line 29: As a part of this change we've switched to using the `aws eks get-token` from the aws cli. In the context of kubectl config this requires a region argument to make a valid request. I haven't been able to find anything indicating that any EKS clusters wouldn't match [this format](https://docs.aws.amazon.com/general/latest/gr/rande.html). If there are issues with this it'll fall through to `aws-iam-authenticator`

source/Calamari.Tests/KubernetesFixtures/KubernetesRunsScriptFixture.cs
```diff
@@ -365,10 +383,22 @@ CalamariResult ExecuteScriptInternal(ICommandLineRunner runner, IScriptWrapper w
...
         {
+            IVariables Variables;
+            public RecordOnly(IVariables variables)
+            {
+                Variables = variables;
+            }
+            
             public CommandResult Execute(CommandLineInvocation invocation)
             {
+                // If were running an aws command (we check the version and get the eks token endpoint) we want to see the actual commands result.
```
IsaacCalligeros95, line 394: When checking the aws cli version & get token commands I didn't want to mock a reply. This is a bit hacky in that we have a RecordOnly CommandLineRunner and then create a new CommandLineRunner that records inside it. I'm happy to take another look at this later & any guidance for a refactor would be appreciated.

source/Calamari/Kubernetes/KubernetesContextScriptWrapper.cs
```diff
@@ -421,6 +425,59 @@ void SetupContextForAmazonServiceAccount(string @namespace, string clusterUrl, s
...
+                    try
+                    {
+                        var region = Regex.Match(clusterUrl, @"(us(-gov)?|ap|ca|cn|eu|sa)-(central|(north|south)?(east|west)?)-\d", RegexOptions.IgnoreCase).Value;
+                        if (string.IsNullOrWhiteSpace(region))
+                        {
+                            log.Error("The EKS cluster Url specified should contain a valid aws region name");
+                        }
+                        
+                        var awsCliCommandRes = ExecuteCommandAndReturnOutput(aws, "--version").FirstOrDefault();
+                        var awsCliVersion = Regex.Match(awsCliCommandRes, @"aws-cli\/(\d+\.)?(\d+\.)?(\*|\d+)").Value.Replace("aws-cli/", string.Empty);
```
YihaoOct, line 443: Here we are not using different parts of SemVer, but just treat it as a whole string. Therefore it is not necessary to use regex here since it is a bit less easy to read. The `aws --version` command prints the versions of dependencies on a space separated line, eg:
```
aws-cli/2.8.4 Python/3.9.11 Darwin/21.4.0 exe/x86_64 prompt/off
```
We can simply split it:
```cs
var awsCliVersion = awsCliCommandRes.Split()
    .FirstOrDefault(versions => versions.StartsWith("aws-cli"))
    .Replace("aws-cli/", string.Empty);
```

source/Calamari/Kubernetes/KubernetesContextScriptWrapper.cs
```diff
@@ -421,6 +425,59 @@ void SetupContextForAmazonServiceAccount(string @namespace, string clusterUrl, s
...
+                        var awsCliSemVer = new SemanticVersion(awsCliVersion);
+                        var eksGetTokenVersion = new SemanticVersion("1.16.156");
+                        var validAwsVersion = awsCliSemVer.CompareTo(eksGetTokenVersion) == 1;
```
YihaoOct, line 446: The documentation only specifies the meaning when the result of `CompareTo` is greater than zero, less than zero and equal to zero. The meaning of exact number is unspecified. I suggest using `> 0` over `== 1`.

source/Calamari/Kubernetes/KubernetesContextScriptWrapper.cs
```diff
@@ -421,6 +425,59 @@ void SetupContextForAmazonServiceAccount(string @namespace, string clusterUrl, s
...
 
+                if (!TrySetAws())
+                {
+                    log.Warn("Could not find the aws cli, falling back to the aws-iam-authenticator.");
+                }
+                else
+                {
+                    try
+                    {
+                        var region = Regex.Match(clusterUrl, @"(us(-gov)?|ap|ca|cn|eu|sa)-(central|(north|south)?(east|west)?)-\d", RegexOptions.IgnoreCase).Value;
```
YihaoOct, line 436: Looks like the service url for EKS has a fixed format: https://docs.aws.amazon.com/general/latest/gr/eks.html

Should we use the format to get the region section (for example, trimming the `.amazonaws.com` from the end and so on), rather than matching with the regex? The risk for using regex is that amazon may expand to other regions in the future. In that case we need to come back add modify the regex pattern, which could be error-prone

source/Calamari/Kubernetes/KubernetesContextScriptWrapper.cs
```diff
@@ -421,6 +425,59 @@ void SetupContextForAmazonServiceAccount(string @namespace, string clusterUrl, s
...
+                            return;
+                        }
+
+                        log.Info($"aws cli version: {awsCliSemVer.ToString()} does not support the \"aws eks get-token\" command. Falling back to aws-iam-authenticator. Please update to a version later than 1.16.156");
+
+                    }
+                    catch (Exception e)
+                    {
+                        log.Verbose($"Unable to authenticate to {clusterUrl} using the aws cli. Failed with the error: {e}.");
+                        log.Info("Falling back to aws-iam-authenticator.");
```
YihaoOct, line 477: I understand that this means that if we use `aws eks get-token` to authenticate, but then the authentication fails, we will attempt to use `aws-iam-authenticator` to authenticate again. I'm not sure if this is intended (or am I understanding this correctly). Also note that there might be other stuff than authentication that may throw exceptions in the try block, for example, `new SemanticVersion()` might fail if an invalid version string is passed in; and the `JObject.Parse` method may also throw if it fails to parse. I think we should probably let the exception happen so we have a chance to fix them.

YihaoOct (changes requested):
Thanks for putting this together and set up the tests 👍 Some suggestions and questions here~

source/Calamari.Tests/KubernetesFixtures/Approved/KubernetesContextScriptWrapperFixture.ExecutionWithEKS_IAMAuthenticator.approved.txt
```diff
@@ -6,5 +6,12 @@
 [Verbose] "kubectl" config use-context octocontext --request-timeout=1m
 [Verbose] "kubectl" config set-cluster octocluster --insecure-skip-tls-verify=true --request-timeout=1m
 [Info] Creating kubectl context to <server> (namespace calamari-testing) using EKS cluster name my-eks-cluster
+[Error] The EKS cluster Url specified should contain a valid aws region name
+[Verbose] Unable to authenticate to <server> using the aws cli. Failed with the error: System.ArgumentNullException: Value cannot be null. (Parameter 's')
+   at System.IO.StringReader..ctor(String s)
+   at Newtonsoft.Json.Linq.JObject.Parse(String json, JsonLoadSettings settings)
+   at Newtonsoft.Json.Linq.JObject.Parse(String json)
+   at Calamari.Kubernetes.KubernetesContextScriptWrapper.SetupKubectlAuthentication.SetupContextForAmazonServiceAccount(String namespace, String clusterUrl, String user) in C:\Users\isaac\source\repos\Calamari\source\Calamari\Kubernetes\KubernetesContextScriptWrapper.cs:line 456.
```
IsaacCalligeros95, line 14: Note: I think throwing an error here to notify why the aws-cli auth failed is the right thing but the fact the error includes the line number etc is a real pain for these approval tests.

source/Calamari.Tests/KubernetesFixtures/Approved/KubernetesContextScriptWrapperFixture.ExecutionWithEKS_IAMAuthenticator.approved.txt
```diff
@@ -6,12 +6,7 @@
...
 [Info] Creating kubectl context to <server> (namespace calamari-testing) using EKS cluster name my-eks-cluster
-[Error] The EKS cluster Url specified should contain a valid aws region name
-[Verbose] Unable to authenticate to <server> using the aws cli. Failed with the error: System.ArgumentNullException: Value cannot be null. (Parameter 's')
-   at System.IO.StringReader..ctor(String s)
-   at Newtonsoft.Json.Linq.JObject.Parse(String json, JsonLoadSettings settings)
-   at Newtonsoft.Json.Linq.JObject.Parse(String json)
-   at Calamari.Kubernetes.KubernetesContextScriptWrapper.SetupKubectlAuthentication.SetupContextForAmazonServiceAccount(String namespace, String clusterUrl, String user) in C:\Users\isaac\source\repos\Calamari\source\Calamari\Kubernetes\KubernetesContextScriptWrapper.cs:line 456.
-[Info] Falling back to aws-iam-authenticator.
+[Warn] Unable to authenticate to <server> using the aws cli.
+[Warn] Falling back to aws-iam-authenticator.
```
IsaacCalligeros95, line 10: If we're logging these as warnings they'll show with the orange ! in logs, for now we're happy with this being an Info level log. We're not trying to push people off of aws-iam-authenticator

source/Calamari.Tests/KubernetesFixtures/Approved/KubernetesContextScriptWrapperFixture.ExecutionWithEKS_IAMAuthenticator.approved.txt
```diff
@@ -6,12 +6,7 @@
 [Verbose] "kubectl" config use-context octocontext --request-timeout=1m
 [Verbose] "kubectl" config set-cluster octocluster --insecure-skip-tls-verify=true --request-timeout=1m
 [Info] Creating kubectl context to <server> (namespace calamari-testing) using EKS cluster name my-eks-cluster
-[Error] The EKS cluster Url specified should contain a valid aws region name
-[Verbose] Unable to authenticate to <server> using the aws cli. Failed with the error: System.ArgumentNullException: Value cannot be null. (Parameter 's')
-   at System.IO.StringReader..ctor(String s)
```
IsaacCalligeros95: We can still log this error, just with the message only. Avoiding the stacktrace line number

source/Calamari.Tests/KubernetesFixtures/Approved/KubernetesContextScriptWrapperFixture.ExecutionWithEKS_IAMAuthenticator.approved.txt
```diff
@@ -6,12 +6,7 @@
 [Verbose] "kubectl" config use-context octocontext --request-timeout=1m
 [Verbose] "kubectl" config set-cluster octocluster --insecure-skip-tls-verify=true --request-timeout=1m
 [Info] Creating kubectl context to <server> (namespace calamari-testing) using EKS cluster name my-eks-cluster
```
IsaacCalligeros95, line 8: Add verbose log of attempting to authenticate with aws-cli

source/Calamari/Kubernetes/KubernetesContextScriptWrapper.cs
```diff
@@ -425,25 +425,22 @@ void SetupContextForAmazonServiceAccount(string @namespace, string clusterUrl, s
...
-                {
-                    log.Warn("Could not find the aws cli, falling back to the aws-iam-authenticator.");
-                }
-                else
+                if (TrySetAws())
                 {
                     try
                     {
-                        var region = Regex.Match(clusterUrl, @"(us(-gov)?|ap|ca|cn|eu|sa)-(central|(north|south)?(east|west)?)-\d", RegexOptions.IgnoreCase).Value;
+                        var region = clusterUrl.Replace(".eks.amazonaws.com", "").Split('.').Last();
```
IsaacCalligeros95, line 432: 1. Look at extract to methods
2. Look at extracting to encapsulated classes/logic

source/Calamari/Kubernetes/KubernetesContextScriptWrapper.cs
```diff
@@ -425,25 +425,22 @@ void SetupContextForAmazonServiceAccount(string @namespace, string clusterUrl, s
...
-                        var eksGetTokenVersion = new SemanticVersion("1.16.156");
-                        var validAwsVersion = awsCliSemVer.CompareTo(eksGetTokenVersion) == 1;
+                        var awsCliVersion = awsCliCommandRes.Split()
+                                                            .FirstOrDefault(versions => versions.StartsWith("aws-cli"))
+                                                            .Replace("aws-cli/", string.Empty);
+
+                        var validAwsVersion = new Version(awsCliVersion).CompareTo(new Version("1.16.156")) > 0;
 
                         if (validAwsVersion)
                         {
```
IsaacCalligeros95, line 446: Break up these ExecuteCommand lines etc into functions, Hide the how

source/Calamari/Kubernetes/KubernetesContextScriptWrapper.cs
```diff
@@ -421,6 +425,79 @@ void SetupContextForAmazonServiceAccount(string @namespace, string clusterUrl, s
...
+                        var awsCliVersion = GetAwsCliVersion();
+
+                        if (new Version(awsCliVersion).CompareTo(new Version("1.16.156")) > 0)
+                        {
+                            var apiVersion = GetEksClusterApiVersion(clusterName, region);
+                            SetKubeConfigAuthenticationToAwsCli(user, clusterName, region, apiVersion);
+                            return;
+                        }
+
+                        log.Verbose($"aws cli version: {awsCliVersion} does not support the \"aws eks get-token\" command. Please update to a version later than 1.16.156");
```
mjhilton: A bit of shuffling of things around here can make the code a little more cohesive. Bring the calculation of the Cluster Region down closer to where it's used to make the purpose clearer, eg:

```suggestion                       
                        var awsCliVersion = GetAwsCliVersion();
                        if (new Version(awsCliVersion).CompareTo(new Version("1.16.156")) > 0)
                        {
                            var region = GetEksClusterRegion(clusterUrl);
                            if (!string.IsNullOrWhiteSpace(region))
                            {
                                var apiVersion = GetEksClusterApiVersion(clusterName, region);
                                SetKubeConfigAuthenticationToAwsCli(user, clusterName, region, apiVersion);
                                return;
                            }

                            log.Verbose("The EKS cluster Url specified should contain a valid aws region name");
                        }

                        log.Verbose($"aws cli version: {awsCliVersion} does not support the \"aws eks get-token\" command. Please update to a version later than 1.16.156");
```

source/Calamari/Kubernetes/KubernetesContextScriptWrapper.cs
```diff
@@ -421,6 +425,79 @@ void SetupContextForAmazonServiceAccount(string @namespace, string clusterUrl, s
...
+                    {
+                        var region = GetEksClusterRegion(clusterUrl);
+                        if (string.IsNullOrWhiteSpace(region))
+                        {
+                            log.Verbose("The EKS cluster Url specified should contain a valid aws region name");
+                        }
+                        
+                        var awsCliVersion = GetAwsCliVersion();
+
+                        if (new Version(awsCliVersion).CompareTo(new Version("1.16.156")) > 0)
```
mjhilton, line 442: A few things I'm wondering about this `Version` usage:
* Could the `GetAwsCliVersion` method directly return a `Version` rather than a `string`?
* Could we extract the comparison Version to a named field/constant, so that it's got a name, something like `minmumAwsCliVersionForAuth`?
* Is it possible to just directly use comparison operators rather than the `CompareTo` method? Then it could look more like:
    ```
    if (awsCliVersion > minimumAwsCliVersionForAuth)
    ```

mjhilton (commented):
Looking much better with those small refactorings. It makes the code a lot easier to read.

A couple of further tweaks suggested to further improve the readability.

I've got a suggestion for some cleanup of the `InstallTools` class, but it's a slightly more involved one, so I'd like to work through that together rather than try to do it on this PR discussion.

source/Calamari/Kubernetes/KubernetesContextScriptWrapper.cs
```diff
@@ -737,13 +847,38 @@ bool TrySetKubectl()
...
+                    var clientVersion = JsonConvert.DeserializeAnonymousType(kubeCtlVersionJson, new { clientVersion = new { gitVersion = "1.0.0" } });
+                    kubectlVersionString = clientVersion?.clientVersion?.gitVersion?.TrimStart('v');
+                    if (kubectlVersionString != null)
+                    {
+                        kubectlVersion = Maybe<Version>.Some(new Version(kubectlVersionString));
+                        return;
+                    }
+                }
+                catch (Exception)
+                {
```
IsaacCalligeros95, line 874: This is a bit of a code smell/redundant is it worth changing?

source/Calamari/Kubernetes/KubernetesContextScriptWrapper.cs
```diff
@@ -421,11 +427,99 @@ void SetupContextForAmazonServiceAccount(string @namespace, string clusterUrl, s
...
+
+            bool TrySetKubeConfigAuthenticationToAwsCli(string clusterName, string clusterUrl, string user)
+            {
+                log.Verbose("Attempting to authenticate with aws-cli");
+                if (TrySetAws())
+                {
+                    try
+                    {
+                        var awsCliVersion = GetAwsCliVersion();
+                        var minmumAwsCliVersionForAuth = new Version("1.16.156");
```
YihaoOct, line 447: typo: `minmum` -> `minimum`

source/Calamari/Kubernetes/KubernetesContextScriptWrapper.cs
```diff
@@ -421,11 +427,99 @@ void SetupContextForAmazonServiceAccount(string @namespace, string clusterUrl, s
...
+                {
+                    log.Verbose("Could not find the aws cli, falling back to the aws-iam-authenticator.");
+                }
+
+                return false;
+            }
+            
+            string GetEksClusterRegion(string clusterUrl) => clusterUrl.Replace(".eks.amazonaws.com", "").Split('.').Last();
+
+            Version GetAwsCliVersion()
```
YihaoOct, line 478: I think the `Version` here is a `System.Runtime.Version`, which is used to represent [the version number of an assembly, operating system, or the common language runtime](https://learn.microsoft.com/en-us/dotnet/api/system.version?view=net-6.0). In short, it is a .net specific thing, and the format is in `major.minor.build.revision`, which differs from SemVer. The outcome that it also works for SemVers in our program is a coincident.

I think we should directly use the `SemanticVersion` class. IMHO correct mapping of the version parts is more important than the ability to use the overloaded `<` operator for readability.

source/Calamari/Kubernetes/KubernetesContextScriptWrapper.cs
```diff
@@ -421,11 +427,99 @@ void SetupContextForAmazonServiceAccount(string @namespace, string clusterUrl, s
...
+                    }
+                    catch (Exception e)
+                    {
+                        log.Verbose($"Unable to authenticate to {clusterUrl} using the aws cli. Failed with error message: {e.Message}");
+                    }
+                }
+                else
+                {
+                    log.Verbose("Could not find the aws cli, falling back to the aws-iam-authenticator.");
+                }
```
YihaoOct, line 471: It doesn't matter but just a recommendation of styles - feel free to ignore:
Since the `else` block is short, I would prefer using a "guard clause" style, so that we can reduce one level of indentation on the main branch:

```cs
if (!TrySetAws())
{
    log.Verbose();
    return false;
}

// the previous if logic
try
{ //...
```

YihaoOct (changes requested):
Thanks Isaac, some comments there:

source/Calamari/Kubernetes/KubernetesContextScriptWrapper.cs
YihaoOct: I am not too sure about the context here, but here I could not easily understand what kind of Exception we are trying to handle here. Reading the code inside the `try` block, I think the Exception might come from JSON serialisation. In this case I think it would be better if we preserve the original Exception, or at least log it out, rather than swallowing it. It may help future debugging.
For example, in the future we could realise that the version isn't set, which caused some bugs in the code, but I imagine it would take a lot of debugging or reasoning about the code logic to find out it is the JSON format that caused the error

IsaacCalligeros95: I was just trying to ignore any possible exceptions from the kubectl version command. I've tested this command on older versions of kubectl but just have some concerns hence the try block. This version is only used for as a part of fetching the EKS api version and I think the Maybe<> handles the case of it not being set but I'll add logging in case it's not.

YihaoOct: I've searched for the usages of `kubeCtlVersion`, and it was only read from 1 place. I think it will perhaps be clearer if we make this a return value of this method, and include this as a parameter in methods that needs to access it.

Thereby we could avoid the style of invoking a method for the side effect which changes a global state (i.e. kubeCtlVersion). This also make this part of logic more testable, and give us more confidence that the `catch` block only hides problems associated with version parsing

mjhilton (approved):
Ship iiiiiit

YihaoOct (approved):
LGTM 👍

PR #931, opened 2022-10-17
Make code signing and compression async

IsaacCalligeros95:
Minor improvements to make code signing & compression async for builds. DotnetPublish hasn't been included as Nuke publish throws exceptions when running on multiple projects in multiple threads. 

Code signing still valid.
image


mik-ky (approved):
:+1:

build/Build.cs
```diff
@@ -294,9 +301,16 @@ void PublishPackage(CalamariPackageMetadata calamariPackageMetadata)
                 .SetRuntime(calamariPackageMetadata.Architecture)
                 .SetOutput(outputDirectory)
             );
-            SignDirectory(outputDirectory);
+            
+            if (!project.Name.Contains("Tests"))
```
mik-ky, line 305: :+1:

nit:`EndsWith` would be more precise with our naming conventions, but I don't think it will matter too much.

PR #932, opened 2022-10-18
Set default regex match timeout to prevent ReDoS

YihaoOct:
## Background

[sc-6247]

This PR addresses [this issue](https://github.com/OctopusDeploy/ResearchAndDevelopment/issues/479).

There are a couple of ReDoS vulnerabilities found in [OctopusDeploy](https://github.com/OctopusDeploy/OctopusDeploy). These issues were addressed by globally setting the default Regex match timeout to 1 second in [this PR](https://github.com/OctopusDeploy/OctopusDeploy/pull/12586).

There are no known vulnerabilities found yet in this repository. However, since we also use Regex, there could be attack surfaces in the future. If they were exploited, it could result in that the worker or server running the Calamari program maxing out CPU and being unable to respond. Therefore, we are preventively adopting a similar approach to globally limit the match timeout for Regex to be 1 second for Calamari and flavours as well.

The default timeout can be overridden where needed:
```cs
new Regex( --, --, TimeSpan.FromSeconds(10));
```

### DIFF
--- a/source/Calamari.Common/CalamariFlavourProgramAsync.cs
+++ b/source/Calamari.Common/CalamariFlavourProgramAsync.cs
@@ -31,6 +31,8 @@ namespace Calamari.Common
 {
     public abstract class CalamariFlavourProgramAsync
     {
+        internal static readonly TimeSpan DefaultRegexMatchTimeout = TimeSpan.FromSeconds(1);
+        
         readonly ILog log;
 
         protected CalamariFlavourProgramAsync(ILog log)
@@ -99,6 +101,10 @@ namespace Calamari.Common
         {
             try
             {
+                // adds default max timout to avoid ReDoS attacks
+                // https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regex.matchtimeout
+                AppDomain.CurrentDomain.SetData("REGEX_DEFAULT_MATCH_TIMEOUT", DefaultRegexMatchTimeout);
+                
                 SecurityProtocols.EnableAllSecurityProtocols();
                 var options = CommonOptions.Parse(args);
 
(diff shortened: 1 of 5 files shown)


source/Calamari.Common/CalamariFlavourProgramAsync.cs
```diff
@@ -99,6 +101,10 @@ protected async Task<int> Run(string[] args)
         {
             try
             {
+                // adds default max timout to avoid ReDoS attacks
+                // https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regex.matchtimeout
+                AppDomain.CurrentDomain.SetData("REGEX_DEFAULT_MATCH_TIMEOUT", DefaultRegexMatchTimeout);
+                
                 SecurityProtocols.EnableAllSecurityProtocols();
```
mjhilton, line 108: It would be nice to follow the convention here and create a static method on a class that describes what is happening. You could then either keep the reasoning comment in that class, or perhaps remove its necessity through naming.

source/Calamari.Common/CalamariFlavourProgramAsync.cs
```diff
@@ -31,6 +31,8 @@ namespace Calamari.Common
 {
     public abstract class CalamariFlavourProgramAsync
     {
+        internal static readonly TimeSpan DefaultRegexMatchTimeout = TimeSpan.FromSeconds(1);
```
mjhilton, line 34: I think they went with 10 seconds in the Server implementation of this.

YihaoOct: In Server it is 1 second: https://github.com/OctopusDeploy/OctopusDeploy/blob/3804cabaaf0e162a9b898b00c26e157207174fa2/source/Octopus.Server/Configuration/AppDomainConfiguration.cs#L8

The 10 seconds in PR description was to show an example of manually overriding the default :)

mjhilton: Nice, thanks.

mjhilton (approved):
LGTM 👍

PR #933, opened 2022-10-20
Bumping from netcoreapp3.1 to net6.0

IsaacCalligeros95:
Core 3.1 is coming to EOL on the [13th of December 2022](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core). Internally we will be removing it from build agents. This PR is meant to remove all dependencies on core3.1 from Calamari.
[sc-9747]

### DIFF
--- a/build-local.sh
+++ b/build-local.sh
@@ -30,7 +30,7 @@ WarningMessage="${Yellow}\
 ║ WARNING:                                               ║
 ║ Building Calamari on a non-windows machine will result ║
 ║ in Calmari and Calamari.Cloud nuget packages being     ║
-║ built against net6.0. This means that some      ║
+║ built against net6.0. This means that some             ║
 ║ steps may not work as expected because they require a  ║
 ║ .Net Framework compatible Calamari Nuget Package.      ║
 ╬════════════════════════════════════════════════════════╬\
--- a/build-local.sh
+++ b/build-local.sh
@@ -30,7 +30,7 @@ WarningMessage="${Yellow}\
 ║ WARNING:                                               ║
 ║ Building Calamari on a non-windows machine will result ║
 ║ in Calmari and Calamari.Cloud nuget packages being     ║
-║ built against netcoreapp3.1. This means that some      ║
+║ built against net6.0. This means that some      ║
 ║ steps may not work as expected because they require a  ║
 ║ .Net Framework compatible Calamari Nuget Package.      ║
 ╬════════════════════════════════════════════════════════╬\
--- a/build/Build.cs
+++ b/build/Build.cs
@@ -190,7 +190,7 @@ namespace Calamari.Build
                       if (!OperatingSystem.IsWindows())
                           Log.Warning("Building Calamari on a non-windows machine will result "
                                       + "in the {DefaultNugetPackageName} and {CloudNugetPackageName} "
-                                      + "nuget packages being built as .Net Core 3.1 packages "
+                                      + "nuget packages being built as .Net Core 6.0 packages "
                                       + "instead of as .Net Framework 4.0 and 4.5.2 respectively. "
                                       + "This can cause compatibility issues when running certain "
                                       + "deployment steps in Octopus Server",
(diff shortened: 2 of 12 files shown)


build-local.sh
```diff
@@ -30,7 +30,7 @@ WarningMessage="${Yellow}\
 ║ WARNING:                                               ║
 ║ Building Calamari on a non-windows machine will result ║
 ║ in Calmari and Calamari.Cloud nuget packages being     ║
-║ built against netcoreapp3.1. This means that some      ║
+║ built against net6.0. This means that some      ║
```
mjhilton, line 33: Whitespace oh noes

source/Calamari.Tests/Fixtures/Integration/Proxies/ProxyEnvironmentVariablesGeneratorFixture.cs
```diff
@@ -232,7 +232,7 @@ void AssertProxyBypassed(IEnumerable<EnvironmentVariable> result)
         
         void AssertUnauthenticatedSystemProxyUsed(IEnumerable<EnvironmentVariable> output)
         {
-#if !NETCOREAPP3_1
+#if !NETCOREAPP
```
mjhilton, line 235: These constants are usually defined in the relevant `csproj`. It looks like at some stage, things got out of sync. 

The [constant I can see](https://github.com/OctopusDeploy/Calamari/pull/933/files#diff-68d19028bf7ced97e653357fa76afc2adf2ac2be0bd449d507a8d20e639c3375R21) which would be defined for `net6.0` is `NETCORE` rather than `NETCOREAPP`. We probably need to use that here for it to work as you expect.

source/Calamari/Calamari.csproj
```diff
@@ -20,10 +20,10 @@
...
   <PropertyGroup Condition="!$([MSBuild]::IsOSUnixLike())">
-    <TargetFrameworks>net40;net452;netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>net40;net452;net6.0</TargetFrameworks>
   </PropertyGroup>
   <PropertyGroup Condition="$([MSBuild]::IsOSUnixLike())">
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(TargetFramework)' == 'net40' OR '$(TargetFramework)' == 'net452' ">
     <DefineConstants>$(DefineConstants);IIS_SUPPORT;WINDOWS_CERTIFICATE_STORE_SUPPORT</DefineConstants>
```
mjhilton, line 29: I'm surprised that there don't seem to be any constants defined for `net6.0` (unless there's something in the base msbuild files that automatically comes with the framework?).

Are there any parts of the Calamari project where we're switching based on constants other than the above two (`IIS_SUPPORT`, `WINDOWS_CERTIFICATE_STORE_SUPPORT`)?

mjhilton (commented):
Overall it's looking good, but there's a couple of things I'd like to understand before marking ✅

build/Build.cs
```diff
@@ -397,8 +397,8 @@ void CompressCalamariProject(Project project)
                           actions.Add(() =>
                           {
                               var publishedLocation =
-                                  DoPublish("Calamari.Tests", Frameworks.NetCoreApp31, nugetVersion, rid);
-                              var zipName = $"Calamari.Tests.netcoreapp.{rid}.{nugetVersion}.zip";
+                                  DoPublish("Calamari.Tests", Frameworks.Net60, nugetVersion, rid);
+                              var zipName = $"Calamari.Tests.{rid}.{nugetVersion}.zip";
```
IsaacCalligeros95, line 401: I've removed the netcoreapp slug here and updated the artifact dependencies from `Calamari.Tests.netcoreapp.%rid%.nuget` -> `Calamari.Tests*.%rid%.nuget` for now to work in the interim for both versions. The * can be removed if this change merges through.

mjhilton (approved):
LGTM!

PR #935, opened 2022-10-21
Ensure tests actually run

YihaoOct:
[sc-7405]

In our previous migration, some migrated tests have a `[Test] public void MyTest()` signature, which does not actually run to the completion of asynchronous code if any. This caused some tests to pass without actually executing.

When having the return type `void`, the test method is treated as synchronous, and returns successfully immediately:

Screen Shot 2022-10-21 at 6 21 42 PM

Ideally they should have the signature `[Test] public async Task MyTest()` to execute correctly and make things clear. So this PR changes the wrong method signatures to this. 

That said, if the return type is `Task` but the `async` keyword is missing, NUnit still waits until the returned `Task` is resolved, therefore the tests are executed correctly:

Screen Shot 2022-10-21 at 6 22 40 PM

For this reason, the test methods which has the signatures like `[Test] public Task MyTest()` (lacking `async`) still works - no need to modify them.

### DIFF
--- a/source/Calamari.AzureResourceGroup.Tests/AzureResourceGroupActionHandlerFixture.cs
+++ b/source/Calamari.AzureResourceGroup.Tests/AzureResourceGroupActionHandlerFixture.cs
@@ -65,10 +65,10 @@ namespace Calamari.AzureResourceGroup.Tests
         }
 
         [Test]
-        public void Deploy_with_template_in_package()
+        public async Task Deploy_with_template_in_package()
         {
             var packagePath = TestEnvironment.GetTestPath("Packages", "AzureResourceGroup");
-            CommandTestBuilder.CreateAsync<DeployAzureResourceGroupCommand, Program>()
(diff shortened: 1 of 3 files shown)


IsaacCalligeros95 (approved):
Looking at other tests they tend to follow the public Task Blah and then return the task itself.
 
`        public Task EnsureAllContainerRegistrationsAreMet()
        {
            return CommandTestBuilder.CreateAsync<MyCommand, MyProgram>()
                              .Execute();
        }`
        
I'm happy with the async await method here just raising the point. Thanks for cleaning these up!

YihaoOct:
Thanks Isaac, the approach you described is the same as the second screenshot (just with return value). In the codebase I found inconsistent conventions. This approach was used alongside with explicit `async` + `await`. Although this approach works, it is implicitly relying on the fact that NUnit waits until the returned `Task` to complete, which is more of NUnit's API details. I think using explicit `await`s is clearer, and by adding `async` in the test method signature, it prevents mistaking `Task` for `void` (and the test still compiles and passs)

PR #940, opened 2022-10-28
Enable partial build and package for local testing

mjhilton:
# Background
The `.\build-local.ps1` script currently has a few minor time-saving optimisations, but still can take 10-15 minutes to prepare a whole Calamari package. That's way too long for an effective development loop.

# Changes
This PR adds a parameter to the build script which allows you to specify an array of Calamari Flavours to build, with everything else excluded. The core Calamari is always included, because this is needed for package acquisition, so it's very unlikely that you'll be able to get a functioning Deployment without it.

## Results
Full local build, ready for testing in Server (with a single Flavour) takes around 2.5 minutes. Build time would increase linearly for each extra flavour included.

# Usage
Execute the `build-local` script with a `-PartialBuildFlavours` parameter.

To build only AzureScripting (aka "Run an Azure Script"):
```pwsh
.\build-local.ps1 -PartialBuildFlavours Calmari.AzureScripting
```
To build AzureScripting and Terraform:
```pwsh
.\build-local.ps1 -PartialBuildFlavours Calmari.AzureScripting, Calamari.Terraform
```

### DIFF
--- a/.nuke/build.schema.json
+++ b/.nuke/build.schema.json
@@ -81,6 +81,12 @@
           "type": "boolean",
           "description": "Run packing step in parallel"
         },
+        "PartialBuildFlavours": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
         "Partition": {
           "type": "string",
           "description": "Partition to use on CI"
@@ -126,7 +132,6 @@
               "CheckForbiddenWords",
               "Clean",
               "Compile",
-              "CopySashimiPackagesForConsolidation",
               "CopyToLocalPackages",
               "Pack",
               "PackageConsolidatedCalamariZip",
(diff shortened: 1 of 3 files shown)


mjhilton:
I believe there's still some fundamental rework we could do to what is now a very large and sprawling build process. I'm pretty sure there's a stack of duplication and unnecessary work done in it, which could be slimmed down and sped up. But in the interest of smaller safer changes, this helps.

.nuke/build.schema.json
```diff
@@ -126,7 +132,6 @@
               "CheckForbiddenWords",
               "Clean",
               "Compile",
-              "CopySashimiPackagesForConsolidation",
```
mjhilton, line 129: I believe this is a stage that was previously removed, just the build schema wasn't committed.

PR #941, opened 2022-10-28
Warn and prevent use of Azure-CLI v2.40

mjhilton:
I started working on this PR to detect and warn customers about OctopusDeploy/Issues#7782, however when I was doing local testing, after going down a number of rabbit holes, I found that no matter what I do Az CLI 2.40 completely freezes when called by Octopus. Even `az version` doesn't work. 

The Bash version of the check represents what I previously had for Powershell, but you can see the Powershell one has been expanded to try to work around this hanging problem, to no avail.

I don't know if we'll revive this one at some time later, but it's not one that I can figure out on Friday afternoon!


mjhilton:
Abandoning this one.

PR #943, opened 2022-11-06
Fix azure resource group tests

YihaoOct:
[sc-29613]

This PR fixes failing tests in AzureResourceGroup that we discovered failing earlier. This was caused by template file not being added to context when the template is supplied in-line. The template file creation logic is in Sashimi half and now lives in Server.

Rider recommendations are also applied, and auto-indentation was applied to code.

### DIFF
--- a/source/Calamari.AzureResourceGroup.Tests/AzureResourceGroupActionHandlerFixture.cs
+++ b/source/Calamari.AzureResourceGroup.Tests/AzureResourceGroupActionHandlerFixture.cs
@@ -85,6 +85,7 @@ namespace Calamari.AzureResourceGroup.Tests
         public async Task Deploy_with_template_inline()
         {
             var packagePath = TestEnvironment.GetTestPath("Packages", "AzureResourceGroup");
+            var templateFileContent = File.ReadAllText(Path.Combine(packagePath, "azure_website_template.json"));
             var paramsFileContent = File.ReadAllText(Path.Combine(packagePath, "azure_website_params.json"));
             var parameters = JObject.Parse(paramsFileContent)["parameters"].ToString();
 
@@ -98,6 +99,7 @@ namespace Calamari.AzureResourceGroup.Tests
                                                context.Variables.Add(SpecialVariables.Action.Azure.ResourceGroupTemplateParameters, parameters);
 
                                                context.WithFilesToCopy(packagePath);
+                                               AddTemplateFiles(context, templateFileContent, paramsFileContent);
                                            })
                               .Execute();
         }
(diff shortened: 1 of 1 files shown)


source/Calamari.AzureResourceGroup.Tests/AzureResourceGroupActionHandlerFixture.cs
```diff
@@ -69,37 +66,40 @@ public async Task Deploy_with_template_in_package()
...
+                .Execute();
         }
 
         [Test]
         public async Task Deploy_with_template_inline()
         {
             var packagePath = TestEnvironment.GetTestPath("Packages", "AzureResourceGroup");
-            var paramsFileContent = File.ReadAllText(Path.Combine(packagePath, "azure_website_params.json"));
+            var templateFileContent = await File.ReadAllTextAsync(Path.Combine(packagePath, "azure_website_template.json"));
+            var paramsFileContent = await File.ReadAllTextAsync(Path.Combine(packagePath, "azure_website_params.json"));
```
YihaoOct, line 86: Using the async version and immediately await the result is the same as just calling the synchronised version. I just applied what Rider recommended here since it would be a more consistent default when two approaches are equivalent.

source/Calamari.AzureResourceGroup.Tests/AzureResourceGroupActionHandlerFixture.cs
```diff
@@ -69,37 +66,40 @@ public async Task Deploy_with_template_in_package()
...
-                                           {
-                                               AddDefaults(context);
-                                               context.Variables.Add(SpecialVariables.Action.Azure.ResourceGroupDeploymentMode, "Complete");
-                                               context.Variables.Add("Octopus.Action.Azure.TemplateSource", "Inline");
-                                               context.Variables.Add(SpecialVariables.Action.Azure.ResourceGroupTemplate, File.ReadAllText(Path.Combine(packagePath, "azure_website_template.json")));
-                                               context.Variables.Add(SpecialVariables.Action.Azure.ResourceGroupTemplateParameters, parameters);
-
-                                               context.WithFilesToCopy(packagePath);
-                                           })
-                              .Execute();
```
YihaoOct, line 102: I applied "auto-indentation" for these code blocks, which makes the indentation pattern the same as that in Server. The old indentation pattern indents the parameter list of a method invocation to start from the same column. This pattern is hard to maintain. If we refactor the name of a method and the new name has a different length, then all the following lines of chained method invocation need to change the indentation accordingly to maintain a meaningful indentation style. This makes the diff trace unnecessarily large.

mjhilton (approved):
LGTM

PR #944, opened 2022-11-07
Fix installation script for GCP on Windows

YihaoOct:
[sc-29615]

This PR fixes tests for GCP when running on Window 2012 agents.

Previously there is a `PathTooLongException` thrown when running test set up on Windows 2012. The test setup downloads the `gcloud` CLI tool and extracts it. Due to the length of the extraction path, and the limitation that the file name could not exceed 260 characters on Windows 2012, the extraction fails. The tests are running fine on Linux and NetCore on Windows. It only failed when targeting .net framework and running directly on Window 2012 hosts.

Though the `PathTooLongException` could be avoided by upgrading target framework from net461 to net462, the path length limitation is on the host's file system, so long paths will still be unaddressable on Windows 2012 (resulting in "part of the path could not be found" error).

Therefore, this PR resolves this issue by shortening the installation path name. If the content of the `gcloud` installation package changes in the future and the file names get even longer, it is possible that this test will fail again. But currently I can't find a more systematic approach to address this. In the future, we should re-evaluate whether it is necessary to keep the test suite running on Windows 2012; we could just run it from a more recent version of Windows.

This PR also applies Rider's recommendations on code style where possible.

### DIFF
--- a/source/Calamari.GoogleCloudScripting.Tests/Calamari.GoogleCloudScripting.Tests.csproj
+++ b/source/Calamari.GoogleCloudScripting.Tests/Calamari.GoogleCloudScripting.Tests.csproj
@@ -7,7 +7,7 @@
         <IsPackable>false</IsPackable>
     </PropertyGroup>
     <PropertyGroup Condition="!$([MSBuild]::IsOSUnixLike())">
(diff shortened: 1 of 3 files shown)


IsaacCalligeros95 (approved):
LGTM

PR #948, opened 2022-11-09
Copy tools to the output folder for AzureScripting tests

YihaoOct:
[sc-29612]

This is a first step towards fixing failing integration tests in AzureScripting. It copies the required tools to the output folder so the tests can utilise them.

Previously the tests are failing because the tools could not be resolved. This error shadows the original test failure we were trying to resolve.


IsaacCalligeros95 (approved):
LGTM

PR #950, opened 2022-11-10
Fix Azure Scripting Issue with TLS version on Windows 2016

YihaoOct:
[sc-29612]

This PR fixes the TLS version mismatch issue for AzureScripting tests when running with PowerShell Desktop on Windows 2016.


IsaacCalligeros95 (approved):
LGTM

PR #955, opened 2022-11-10
Kubectl auth refactoring - step 5

mjhilton:
This step pulls the concern of managing log message redactions out of the core logic and puts it in its own class, ensuring the `CommandLineTool`s don't need to know about this behaviour at all.


mjhilton:

PR #956, opened 2022-11-14
Kubectl auth refactoring: step 6

mjhilton:
This step extracts a new class to represent `AzureCli`, and fixes up some approvals which had gotten out of sync


mjhilton:

PR #957, opened 2022-11-14
Kubectl auth refactoring: step 7

mjhilton:
This step extracts a method: `AzureCli.ConfigureAksKubeCtlAuthentication`, which manages configuring the Kubectl auth for Azure authentication


mjhilton:

PR #960, opened 2022-11-17
Warn customers if their GKE auth tooling is incorrect

mjhilton:
# Background
From `kubctl` 1.26 onward, some auth code which used to be integrated in the tool (for doing Azure- and GCloud-specific OAuth flows) has been removed, making it the vendor's responsibility to publish authentication plugins that implement their required auth logic. 

This PR is designed to help customers choose the correct tooling for their situation, and guide them to the right information if they're not in a compatible state.

## Before
If customers run into an incompatible scenario (eg they use `kubectl` >= 1.26, but don't have the `gke-gcloud-auth-plugin` installed), their auth to the cluster and therefore any step that tries to deploy to the cluster will fail. 

The warnings provided by the `kubectl` tooling may or may not be clear enough for the customer to understand what's gone wrong or what they need to do to fix it - they're out of our control. 

## After
If customers are in a situation where their GKE auth would fail, we write a warning message to the logs, and direct them to a link that can help them remediate the scenario.

Fixes OctopusDeploy/Issues#7621.
Addresses [sc-6271] (internal).

## Notes/implementation details
This PR also bundles a series of refactorings that were done to make the code easier to reason about before implementing the main change. 

The Azure equivalent is fine as-is, because the way we authenticate to Azure never invoked the Azure-specific auth from `kubectl`; instead it generates generic certificate-based auth credentials directly into the `kubeconfig`. 

The warning message for the problematic GKE configuration uses an `oc.to` link which is configured via short.io. It currently points just to [the GCloud blog about this change](https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke), but we could/should probably create our own first-party blog content to put the changes in clearer context for Octopus users. This can be done at any time without a code change, we just update where the `oc.to` link points to.

### DIFF
--- a/source/Calamari/Kubernetes/Authentication/AzureKubernetesServicesAuth.cs
+++ b/source/Calamari/Kubernetes/Authentication/AzureKubernetesServicesAuth.cs
@@ -1,0 +1,51 @@
+﻿using System;
+using Calamari.Common.Plumbing.Logging;
+using Calamari.Common.Plumbing.Variables;
+using Calamari.Kubernetes.Integration;
+
+namespace Calamari.Kubernetes.Authentication
+{
+    public class AzureKubernetesServicesAuth
+    {
+        readonly AzureCli azureCli;
+        readonly Kubectl kubectlCli;
+        readonly IVariables deploymentVariables;
+
+        /// <summary>
+        /// This class is responsible for configuring the kubectl auth against an AKS cluster for an Octopus Deployment
+        /// </summary>
+        /// <param name="azureCli"></param>
+        /// <param name="kubectlCli"></param>
+        /// <param name="deploymentVariables"></param>
+        public AzureKubernetesServicesAuth(AzureCli azureCli, Kubectl kubectlCli, IVariables deploymentVariables)
+        {
+            this.azureCli = azureCli;
+            this.kubectlCli = kubectlCli;
+            this.deploymentVariables = deploymentVariables;
+        }
+
+        public bool TryConfigure(string @namespace, string kubeConfig)
+        {
+            if (!azureCli.TrySetAz())
+                return false;
+
+            var disableAzureCli = deploymentVariables.GetFlag("OctopusDisableAzureCLI");
+            if (!disableAzureCli)
+            {
+                var azEnvironment = deploymentVariables.Get("Octopus.Action.Azure.Environment") ?? "AzureCloud";
+                var subscriptionId = deploymentVariables.Get("Octopus.Action.Azure.SubscriptionId");
+                var tenantId = deploymentVariables.Get("Octopus.Action.Azure.TenantId");
+                var clientId = deploymentVariables.Get("Octopus.Action.Azure.ClientId");
+                var password = deploymentVariables.Get("Octopus.Action.Azure.Password");
+                azureCli.ConfigureAzAccount(subscriptionId, tenantId, clientId, password, azEnvironment);
+
+                var azureResourceGroup = deploymentVariables.Get("Octopus.Action.Kubernetes.AksClusterResourceGroup");
+                var azureCluster = deploymentVariables.Get(SpecialVariables.AksClusterName);
+                var azureAdmin = deploymentVariables.GetFlag("Octopus.Action.Kubernetes.AksAdminLogin");
+                azureCli.ConfigureAksKubeCtlAuthentication(kubectlCli, azureResourceGroup, azureCluster, @namespace, kubeConfig, azureAdmin);
+            }
+
+            return true;
+        }
+    }
+}
--- a/source/Calamari/Kubernetes/Integration/AzureCli.cs
+++ b/source/Calamari/Kubernetes/Integration/AzureCli.cs
@@ -1,0 +1,95 @@
+﻿using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Calamari.Common.Features.Processes;
+using Calamari.Common.Plumbing;
+using Calamari.Common.Plumbing.Logging;
+
+namespace Calamari.Kubernetes.Integration
+{
+    public class AzureCli : CommandLineTool
+    {
+        public AzureCli(ILog log, ICommandLineRunner commandLineRunner, string workingDirectory, Dictionary<string, string> environmentVars)
+            : base(log, commandLineRunner, workingDirectory, environmentVars)
+        {
+        }
+
+        public bool TrySetAz()
+        {
+            var foundExecutable = CalamariEnvironment.IsRunningOnWindows
+                ? ExecuteCommandAndReturnOutput("where", "az.cmd").FirstOrDefault()
+                : ExecuteCommandAndReturnOutput("which", "az").FirstOrDefault();
+
+            if (string.IsNullOrEmpty(foundExecutable))
+            {
+                log.Error("Could not find az. Make sure az is on the PATH.");
+                return false;
+            }
+
+            ExecutableLocation = foundExecutable.Trim();
+            return true;
+        }
+
+        public void ConfigureAzAccount(string subscriptionId,
+                                       string tenantId,
+                                       string clientId,
+                                       string password,
+                                       string azEnvironment)
+        {
+            environmentVars.Add("AZURE_CONFIG_DIR", Path.Combine(workingDirectory, "azure-cli"));
+
+            TryExecuteCommandAndLogOutput(ExecutableLocation,
+                                          "cloud",
+                                          "set",
+                                          "--name",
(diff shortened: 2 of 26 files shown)


source/Calamari/Kubernetes/KubernetesContextScriptWrapper.cs
```diff
@@ -246,26 +244,20 @@ bool TrySetupContext(string kubeConfig, string @namespace, string accountType)
...
-                        return false;
-                    }
-
-                    ConfigureAzAccount();
+                    var azureCli = new AzureCli(log, commandLineRunner, workingDirectory, environmentVars);
+                    var azureAuth = new AzureKubernetesServicesAuth(azureCli, kubectlCli, variables);
 
-                    SetupContextForAzureServicePrincipal(kubeConfig, @namespace);
+                    if (!azureAuth.TryConfigure(@namespace, kubeConfig))
+                        return false;
```
YihaoOct, line 251: Thanks for extracting this 👍 I'm thinking it would be nice if we could extract all such authentication methods into their own classes and give them a common interface (the already extracted Azure Auth and GCloud Auth do demonstrate a common pattern). Then we could probably remove this large if-else statement here. Something like:
```cs
IAuthenticationProvider auth = AuthenticationProviderFactory.Create(variables, accountType);
return auth.Configure(variables, accountType);
```

But there are too many changes to make to do this, and additional risks, which make this probably not worthwhile. I'm happy the way it currently is.

mjhilton: We're on the same wavelength :) I was just telling @IsaacCalligeros95 on Tuesday about my plans to do exactly this.

I think this would be a really positive direction for this code, but agree that the scope of refactoring is at its limit for one PR. I can add it as a TODO comment so that the next person to touch it works toward it, though.

source/Calamari/Kubernetes/Authentication/AzureKubernetesServicesAuth.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    public class AzureKubernetesServicesAuth
+    {
+        readonly AzureCli azureCli;
+        readonly Kubectl kubectlCli;
+        readonly IVariables deploymentVariables;
+
+        /// <summary>
+        /// This class is responsible for configuring the kubectl auth against an AKS cluster for an Octopus Deployment
+        /// </summary>
```
YihaoOct, line 16: I really like this one-liner summary document comment. Sometimes I feel the class/method names are not the best place to express the intention (and just like comments, names can get out of date as well). We should probably use more of these.

source/Calamari/Kubernetes/Integration/CommandLineTool.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        public string ExecutableLocation { get; protected set; }
+
+        protected bool TryExecuteCommandAndLogOutput(string exe, params string[] arguments)
+        {
+            var result = ExecuteCommandAndLogOutput(new CommandLineInvocation(exe, arguments));
+            return result.ExitCode == 0;
+        }
+
+        protected CommandResult ExecuteCommandAndLogOutput(CommandLineInvocation invocation)
```
IsaacCalligeros95, line 33: Do we want all these commands along with output logged? I think this is a change of functionality and in the case of some of the `get-credentials` style calls we'd be logging returned tokens?

IsaacCalligeros95 (commented):
LGTM, I'm just curious about the logging of CLI commands and if anything has actually changed there. I might be mistaken but just concerned the outputs from commands that fetch tokens would have a change in behaviour.

source/Calamari/Kubernetes/Integration/GCloud.cs
```diff
@@ -0,0 +1,110 @@
...
+            }
+
+            if (!string.IsNullOrEmpty(zone))
+            {
+                environmentVars.Add("CLOUDSDK_COMPUTE_ZONE", zone);
+            }
+
+            if (!useVmServiceAccount)
+            {
+                if (jsonKey == null)
```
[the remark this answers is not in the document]
mjhilton: Yep, we can see from the [class extraction](https://github.com/OctopusDeploy/Calamari/pull/960/commits/800a74e296a383cab8f3c929abc1c83dc384e0a1#diff-a201f377c4880673f195e26a67d8323bb7cac40ce3dec8745fcb63a1bb55e335L283) that the `accountVariable` was only ever used to select the right `jsonkey`, so should be no functional difference between the two.

source/Calamari/Kubernetes/Integration/CommandLineTool.cs
mjhilton: The commands [were always logged by default before, the naming was just never explicit about it](https://github.com/OctopusDeploy/Calamari/pull/960/commits/354732537d5ab491782d1b9078fe7bb63323273c#diff-a201f377c4880673f195e26a67d8323bb7cac40ce3dec8745fcb63a1bb55e335L914). 

Where the commands weren't previously logged, I've introduced the [ExecuteCommandAndReturnOutput](https://github.com/OctopusDeploy/Calamari/pull/960/commits/354732537d5ab491782d1b9078fe7bb63323273c#diff-b42431012ab7a6abafd9cd1104372a0a755df8abdd8292c6e93c775ffb9d581aR79) method, which after the CLI class refactors was only internal calls.

mjhilton: The [assent tests](https://github.com/OctopusDeploy/Calamari/pull/960/commits/354732537d5ab491782d1b9078fe7bb63323273c#diff-391bf7573d07d10295da10f256fa5e03290fbce52c977651657598cb2ddf177e) validate that no extra command logging has been introduced unless it was done intentionally :)

source/Calamari/Kubernetes/Integration/GCloud.cs
```diff
@@ -0,0 +1,110 @@
...
+using System.IO;
+using System.Linq;
+using Calamari.Common.Features.Processes;
+using Calamari.Common.Plumbing;
+using Calamari.Common.Plumbing.FileSystem;
+using Calamari.Common.Plumbing.Logging;
+
+namespace Calamari.Kubernetes.Integration
+{
+    public class GCloud : CommandLineTool
```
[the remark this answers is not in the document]
mjhilton: Yeah it would, but I might do that at a later time.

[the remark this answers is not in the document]
mjhilton: Not gonna change this one at this stage.

source/Calamari/Kubernetes/Integration/AzureCli.cs
```diff
@@ -0,0 +1,95 @@
...
+            if (adminLogin)
+            {
+                arguments.Add("--admin");
+                clusterName += "-admin";
+            }
+
+            var result = ExecuteCommandAndLogOutput(new CommandLineInvocation(ExecutableLocation, arguments.ToArray()));
+            result.VerifySuccess();
+
+            kubectlCli.ExecuteCommandAndAssertSuccess("config", "set-context", clusterName, $"--namespace={@clusterNamespace}");
```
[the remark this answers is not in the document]
mjhilton: Could do, not gonna do now.

source/Calamari/Kubernetes/Integration/CommandLineTool.cs
IsaacCalligeros95: I think your right that this isn't changing functionality. The assent tests do [swallow a number of command results](https://github.com/OctopusDeploy/Calamari/blob/17f77e2500753f73061a965b27e841fa2a49d32c/source/Calamari.Tests/KubernetesFixtures/KubernetesRunsScriptFixture.cs#L430). I'm going to pull the branch down and give it a run just to step through and understand this properly.

YihaoOct (approved):
LGTM 👍

source/Calamari/Kubernetes/Integration/CommandLineTool.cs
IsaacCalligeros95: Confirmed locally on EKS there are no additional outputs, sorry for the hold up!

IsaacCalligeros95 (approved):
LGTM

source/Calamari/Kubernetes/Integration/CommandLineTool.cs
mjhilton: Thanks for being detailed in your review, I appreciate it :)

PR #961, opened 2022-11-17
Add support for inheriting AWS web identities in containers

mcasperson:
This PR adds support for inheriting web identities from a worker running in an EKS pod.

There are two changes in this PR:

1. Cloud discovery adds a fallback that calls `AssumeRoleWithWebIdentityCredentials.FromEnvironmentVariables()`, and changes the order of login precedence fallabcks to match those of the `aws` CLI (documented [here](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-quickstart.html#cli-configure-quickstart-precedence))
2. The context established for K8s and AWS steps also calls  `AssumeRoleWithWebIdentityCredentials.FromEnvironmentVariables()`, and change the order to match the `aws` CLI.

No UI changes are required, as steps that have an AWS login but do not have an account will now attempt to use web roles (i.e. roles assigned to service accounts in EKS pods) in addition to instance roles (i.e. those assigned to EC2 instances).

This PR fixes https://github.com/OctopusDeploy/Issues/issues/7904

### DIFF
--- a/source/Calamari.Aws/Kubernetes/Discovery/AwsAuthenticationDetails.cs
+++ b/source/Calamari.Aws/Kubernetes/Discovery/AwsAuthenticationDetails.cs
@@ -34,9 +34,10 @@ namespace Calamari.Aws.Kubernetes.Discovery
             else
             {
                 // The sequence of fallbacks trying to log in with credentials exposed by the worker.
-                if (!(TryGetInstanceProfileAwsCredentials(log, out credentials)
-                      || TryGetEnvironmentVariablesAwsCredentials(log, out credentials)
-                      || TryGetAssumeRoleWithWebIdentityCredentials(log, out credentials)))
+                // This follows the precidence document at https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-quickstart.html#cli-configure-quickstart-precedence
+                if (!(TryGetEnvironmentVariablesAwsCredentials(log, out credentials)
+                      || TryGetAssumeRoleWithWebIdentityCredentials(log, out credentials)
+                      || TryGetInstanceProfileAwsCredentials(log, out credentials)))
                 {
                     log.Warn("Unable to authorise credentials, see verbose log for details.");
                     return false;
--- a/source/Calamari.Aws/Kubernetes/Discovery/AwsAuthenticationDetails.cs
+++ b/source/Calamari.Aws/Kubernetes/Discovery/AwsAuthenticationDetails.cs
@@ -33,28 +33,13 @@ namespace Calamari.Aws.Kubernetes.Discovery
             }
             else
             {
-                try
+                // The sequence of fallbacks trying to log in with credentials exposed by the worker.
+                if (!(TryGetInstanceProfileAwsCredentials(log, out credentials)
+                      || TryGetEnvironmentVariablesAwsCredentials(log, out credentials)
+                      || TryGetAssumeRoleWithWebIdentityCredentials(log, out credentials)))
                 {
-                    // If not currently running on an EC2 instance,
-                    // this will throw an exception.
-                    credentials = new InstanceProfileAWSCredentials();
-                }
-                // Catching a generic Exception because AWS SDK throws undocumented exceptions.
-                catch (Exception instanceProfileException)
-                {
-                    try
-                    {
-                        // The last attempt is trying to use Environment Variables.
-                        credentials = new EnvironmentVariablesAWSCredentials();
-                    }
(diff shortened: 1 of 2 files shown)


source/Calamari.CloudAccounts/AwsEnvironmentGeneration.cs
```diff
@@ -46,11 +46,35 @@ public static async Task<AwsEnvironmentGeneration> Create(ILog log, IVariables v
...
-            await PopulateSuppliedKeys();
-            await PopulateKeysFromInstanceRole();
+
+            if (!await PopulateSuppliedKeys())
+            {
+                if (!await LoginFallback())
+                {
+                    throw new Exception("AWS-LOGIN-ERROR-0006: "
+                                        + "Failed to login via external credentials assigned to the worker. "
+                                        + $"For more information visit {log.FormatLink("https://g.octopushq.com/AwsCloudFormationDeploy#aws-login-error-0006")}");
```
mjhilton, line 56: I think we need to be doing any these links using `oc.to` rather than `g.octopushq.com`

source/Calamari.CloudAccounts/AwsEnvironmentGeneration.cs
```diff
@@ -132,26 +154,27 @@ void PopulateCommonSettings()
...
             if (!String.IsNullOrEmpty(accessKey))
             {
                 EnvironmentVars["AWS_ACCESS_KEY_ID"] = accessKey;
                 EnvironmentVars["AWS_SECRET_ACCESS_KEY"] = secretKey;
                 if (!await verifyLogin())
                 {
-                    throw new Exception("AWS-LOGIN-ERROR-0005: Failed to verify the credentials. " +
-                                             "Please check the keys assigned to the Amazon Web Services Account associated with this step. " +
-                                             $"For more information visit {log.FormatLink("https://g.octopushq.com/AwsCloudFormationDeploy#aws-login-error-0005")}");
+                    throw new Exception("AWS-LOGIN-ERROR-0005: Failed to verify the credentials. " + "Please check the keys assigned to the Amazon Web Services Account associated with this step. " + $"For more information visit {log.FormatLink("https://g.octopushq.com/AwsCloudFormationDeploy#aws-login-error-0005")}");
```
mjhilton, line 165: Potentially here too.

mjhilton (approved):
This looks like a reasonable change to me.

source/Calamari.CloudAccounts/AwsEnvironmentGeneration.cs
mcasperson: Good catch - I'll make those changes.

source/Calamari.CloudAccounts/AwsEnvironmentGeneration.cs
mcasperson: :+1: I'll update this.

source/Calamari.Aws/Kubernetes/Discovery/AwsAuthenticationDetails.cs
```diff
@@ -33,28 +33,14 @@ public bool TryGetCredentials(ILog log, out AWSCredentials credentials)
...
-                    // this will throw an exception.
-                    credentials = new InstanceProfileAWSCredentials();
-                }
-                // Catching a generic Exception because AWS SDK throws undocumented exceptions.
-                catch (Exception instanceProfileException)
+                // The sequence of fallbacks trying to log in with credentials exposed by the worker.
+                // This follows the precedence document at https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-quickstart.html#cli-configure-quickstart-precedence
+                if (!(TryGetEnvironmentVariablesAwsCredentials(log, out credentials)
+                      || TryGetAssumeRoleWithWebIdentityCredentials(log, out credentials)
+                      || TryGetInstanceProfileAwsCredentials(log, out credentials)))
```
scottmerchant, line 40: Just a note that in my (untested) fix [I opted to use](https://github.com/OctopusDeploy/Calamari/pull/927/files#r1030943727) the `FallbackCredentialsFactory.GetCredentials()` in the AWS SDK which would give discovery all (?) AWS fallback options to the user in the default order as specified by AWS. It might be worth trying to see if it handles the case you're facing as it might save us from running into another edge case in the future.

scottmerchant: Nevermind, I just saw this is merged :) we can potentially use the above if another similar issue crops up again, but for now I think it's fine 👍

PR #968, opened 2022-12-07
Fix argument handling for Powershell 7.3

YihaoOct:
Internal ticket: [sc-34031]

This PR fixes issue [#7910](https://github.com/OctopusDeploy/Issues/issues/7910).

When a PowerShell script is running on an environment with PowerShell 7.3+ installed, the arguments passed to native commands are not resolved correctly if they are surrounded by quotes. For example, for the command `az login --username="asdf"`, before PowerShell 7.3, `asdf` is passed to `az` as the username, while in 7.3+, `"asdf"` is passed as the username (quotation marks preserved). This breaks the azure context set up code.

This PR addresses this by setting the [feature flag](https://learn.microsoft.com/en-gb/powershell/scripting/learn/experimental-features?view=powershell-7.3#psnativecommandargumentpassing) `$PSNativeCommandPassing` to `Legacy` so that the original behaviour is preserved. Since we cannot control which version of PowerShell customers are running, we couldn't directly remove the quotation marks, because doing this will break the scripts if they are run on lower version of PowerShell. 

This flag is set in `Bootstrap.ps1` since it is executed before any other PowerShell scripts, setting the flag globally in this file would prevent this issue from happening in various other PowerShell scripts in other flavours, not just AzureScripting.


.nuke/build.schema.json
```diff
@@ -126,7 +126,6 @@
               "CheckForbiddenWords",
               "Clean",
               "Compile",
-              "CopySashimiPackagesForConsolidation",
               "CopyToLocalPackages",
               "Pack",
               "PackageConsolidatedCalamariZip",
```
YihaoOct, line 131: This file changes every time we run a local build, which I think adds irrelevant traces to PRs. This file is solely used for [shell completion](https://nuke.build/faq/#build-structure), shall we just gitignore it? It will be auto-generated again locally when the build runs locally next time.

mik-ky (approved):
LGTM 👍

.nuke/build.schema.json
YihaoOct: I added this file to .gitignore and removed it.

PR #971, opened 2022-12-15
Updating GKE tests to include GKE_Gcloud_Auth_Plugin

IsaacCalligeros95:
Updating GKE tests to include the `GKE_GCLOUD_AUTH_PLUGIN`. As of kubectl 1.26 this is the default auth mechanism for account based authentication to GKE see https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke for details.


source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureGke.cs
```diff
@@ -48,7 +48,8 @@ protected override void ExtractVariablesFromTerraformOutput(JObject jsonOutput)
         {
             return new Dictionary<string, string>
             {
-                { "GOOGLE_CLOUD_KEYFILE_JSON", ExternalVariables.Get(ExternalVariable.GoogleCloudJsonKeyfile) }
+                { "GOOGLE_CLOUD_KEYFILE_JSON", ExternalVariables.Get(ExternalVariable.GoogleCloudJsonKeyfile) },
+                { "USE_GKE_GCLOUD_AUTH_PLUGIN", "TRUE"}
```
IsaacCalligeros95, line 52: I'm not 100% sure if this is required, according to the [docs ](https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke)you can run into issues. Locally the test doesn't run for me so I'm relying on the pipeline and with the slow feedback cycle decided to include it incase.

mjhilton: If we've got a version of `kubectl` that requires the `gke-gcloud-auth-plugin` then it shouldn't be needed; this flag was introduced during the beta/testing releases of these changes to enable customer to opt-in early.

It can't hurt to have it though :)

mjhilton (approved):
Seems reasonable to me

source/Calamari.Tests/KubernetesFixtures/InstallTools.cs
```diff
@@ -217,6 +218,45 @@ public async Task InstallGCloud()
...
+            var stdOut = new StringBuilder();
+            var stdError = new StringBuilder();
+            var commandExitCode = SilentProcessRunner.ExecuteCommand(executable,
+                                                                     arguments,
+                                                                     workingDirectory,
+                                                                     variables,
+                                                                     (Action<string>)(s => stdOut.AppendLine(s)),
+                                                                     (Action<string>)(s => stdError.AppendLine(s))).ExitCode;
+
+            if (commandExitCode != 0)
```
YihaoOct, line 254: If whenever the exit code is not 0 an exception is thrown, that means the exit code in return value will always be 0. So we don't really need the `ExitCode` field in the result, and we are not using it anyway.

source/Calamari.Tests/KubernetesFixtures/InstallTools.cs
```diff
@@ -217,6 +218,45 @@ public async Task InstallGCloud()
...
+            {
+                throw new Exception($"{stdOut}{stdError}");
```
YihaoOct, line 256: We could add more structure to the message, for example, put stdOut and stdError on separate lines and give a heading to each. Simply concatenating them may not give an easy-to-read message.

source/Calamari.Tests/KubernetesFixtures/KubernetesRunsScriptFixture.cs
```diff
@@ -404,6 +413,13 @@ void InstallAwsCli(InstallTools tools)
             Task.Run(() => installAwsCliTask);
             Task.WaitAll(installAwsCliTask);
         }
+        
+        void InstallGCloud(InstallTools tools)
+        {
+            var installGCloud = tools.InstallGCloud();
+            Task.Run(() => installGCloud);
+            Task.WaitAll(installGCloud);
+        }
```
YihaoOct, line 422: I know this is done in other methods as well, but it looks a bit strange to me. Maybe we could test using `async`/`await` instead? I don't think I fully understand the current pattern 🤔 Looks like it is trying to use `Task.Run` and `Task.WaitAll` to avoid using `async`, but I'm not sure what's the reason behind

YihaoOct (changes requested):
It looks good; there're some small suggestions :)

source/Calamari.Tests/KubernetesFixtures/KubernetesRunsScriptFixture.cs
IsaacCalligeros95: Using Async await here meant the calling signature becomes `InstallTools(async (tools) => await InstallGCloud(tools));` There might be a cleaner way that I'm missing but went with this approach to keep the call signature cleaner.

source/Calamari.Tests/KubernetesFixtures/KubernetesRunsScriptFixture.cs
```diff
@@ -394,26 +408,34 @@ CalamariResult ExecuteScriptInternal(ICommandLineRunner runner, IScriptWrapper w
...
+            await tools.InstallAwsCli();
+        }
+
+        static async Task InstallGCloud(InstallTools tools)
         {
-            var installAwsCliTask = tools.InstallAwsCli();
-            Task.Run(() => installAwsCliTask);
-            Task.WaitAll(installAwsCliTask);
+            await tools.InstallGCloud();
         }
```
APErebus, line 428: There was no reason I could see not to have these as fully `async`.

source/Calamari.Tests/KubernetesFixtures/KubernetesRunsScriptFixture.cs
```diff
@@ -394,26 +408,34 @@ CalamariResult ExecuteScriptInternal(ICommandLineRunner runner, IScriptWrapper w
...
 
-        class RecordOnly :  ICommandLineRunner
+        class RecordOnly : ICommandLineRunner
         {
             IVariables Variables;
-            public RecordOnly(IVariables variables)
+            readonly InstallTools installTools;
+
+            public RecordOnly(IVariables variables, InstallTools installTools)
```
APErebus, line 435: pass in the install tools rather than a static variable

source/Calamari.Tests/KubernetesFixtures/InstallTools.cs
APErebus: @YihaoOct I'm not sure exactly what you are referring to here. As you saying the `SilentProcessRunner.ExecuteCommand` will throw an exception on a non-zero exit code? I couldn't see that in the code

source/Calamari.Tests/KubernetesFixtures/InstallTools.cs
```diff
@@ -204,7 +205,7 @@ public async Task InstallGCloud()
             using (var client = new HttpClient())
             {
                 GcloudExecutable = await DownloadCli("gcloud",
-                    () => Task.FromResult<(string, string)>(("346.0.0", string.Empty)),
+                    () => Task.FromResult<(string, string)>(("436.0.0", string.Empty)),
```
APErebus, line 208: bumped to latest

source/Calamari.Tests/KubernetesFixtures/InstallTools.cs
```diff
@@ -218,6 +219,45 @@ public async Task InstallGCloud()
                         return GetGcloudExecutablePath(destinationDirectoryName);
                     });
             }
+            
+            if (CalamariEnvironment.IsRunningOnWindows)
+            {
+                InstallGkeAuthPlugin();
+            }
```
APErebus, line 226: This is currently limited to windows because of extra stuff needed on *nix systems & mac

source/Calamari.Tests/KubernetesFixtures/InstallTools.cs
YihaoOct: This PR has been a bit old 😅 In the original version I commented on: https://github.com/OctopusDeploy/Calamari/pull/971/commits/5fb98b5a02daa6cb6fd0d67561d595e99aaa8970

The ExitCode used to be part of the return value of this method. I was saying that if we can reach the return statement without an exception, then it means the ExitCode has to be 0. Therefore we don't need to return it.

It's not a concern now, but because of the place I put the comment on, it wan't picked up as `Outdated`

APErebus: ah, all good 👍

APErebus:
@YihaoOct @IsaacCalligeros95 @scme0 I've updated this PR with some minor clean-up as to try and get this all working again, even if it's just windows tests

source/Calamari.Tests/KubernetesFixtures/InstallTools.cs
```diff
@@ -218,6 +219,45 @@ public async Task InstallGCloud()
...
+                                                                     (Action<string>)(s => stdOut.AppendLine(s)),
+                                                                     (Action<string>)(s => stdError.AppendLine(s))).ExitCode;
+
+            if (commandExitCode != 0)
+            {
+                throw new InvalidOperationException($"stdOut: {stdOut}, stdError: {stdError}");
+            }
+
+            return stdOut.ToString().Trim('\r', '\n');
         }
```
YihaoOct, line 261: I was reading other parts of the same file, looks like this pattern has been done multiple times elsewhere, such as in `InstallAwsCli` and `AddExecutePermission`. Maybe we can make all of them use this helper methods so it is more consistent?

YihaoOct (approved):
Looks good 👍 I've made a comment

source/Calamari.Tests/KubernetesFixtures/InstallTools.cs
APErebus: yeah, I can fix that up 👍

APErebus: @YihaoOct I made those changes as suggested in [c4420aa](https://github.com/OctopusDeploy/Calamari/pull/971/commits/c4420aabc791995df618372379b69cc53dbb561f)

APErebus:
[sc-50665]

PR #972, opened 2022-12-20
Adding release notes for calamari

IsaacCalligeros95:
Adding the release notes from the Calamari pipeline to the Calamari.Consolidated.nuspec definition for the package. This change isn't much of a value add, and the Calamari package isn't publicly consumable, so doesn't technically come under Sensible Defaults. Feedback pushing against whether this is worthwhile adding is welcome.

### DIFF
--- a/build/Build.cs
+++ b/build/Build.cs
@@ -479,7 +479,14 @@ namespace Calamari.Build
             _ => _.DependsOn(PackageConsolidatedCalamariZip)
                   .Executes(() =>
                             {
-                                NuGetPack(s => s.SetTargetPath(BuildDirectory / "Calamari.Consolidated.nuspec")
+                                var releaseNotes = IsLocalBuild ? "Local" : File.ReadAllText(ArtifactsDirectory / "ReleaseNotes.md");
+                                
+                                var nuspec = BuildDirectory / "Calamari.Consolidated.nuspec";
+                                var text = File.ReadAllText(nuspec);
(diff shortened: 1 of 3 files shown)


build/Build.cs
```diff
@@ -479,9 +479,19 @@ void CompressCalamariProject(Project project)
...
                   .Executes(() =>
                             {
-                                NuGetPack(s => s.SetTargetPath(BuildDirectory / "Calamari.Consolidated.nuspec")
-                                                .SetVersion(NugetVersion.Value)
-                                                .SetOutputDirectory(ArtifactsDirectory));
+                                var releaseNotes = IsLocalBuild ? "Local" : File.ReadAllText(RootDirectory / "releasenotes" / "ReleaseNotes.md");
+                                
+                                var nuspec = BuildDirectory / "Calamari.Consolidated.nuspec";
+                                var text = File.ReadAllText(nuspec);
+                                text = text.Replace("$releaseNotes$", releaseNotes);
```
YihaoOct, line 486: There is a native solution to replace the `$token$` provided by [nuget](https://learn.microsoft.com/en-gb/nuget/reference/nuspec#replacement-tokens). You just need to provide that via a `-properties` argument. This is available as the `.SetProperties()` method inside the `NuGetPack(...)` builder. So you don't need to manually replace the token here.

YihaoOct (changes requested):
Looks good, but I think there is a simpler way to replace the token, maybe worth a try

YihaoOct (approved):
LGTM

PR #974, opened 2022-12-23
Fixes problems with terraform and AKS during integration test setup

mjhilton:
This PR fixes the setup for tests which are failing in the pipeline due to changes in Azure and the way it uses terraform.

* Terraform/Kubernetes no longer generates a default secret for service accounts, so `default_secret_name` is now empty. This PR updates the `tf` so that we explicitly create a token for the service account
* The syntax for RBAC in the azurerm provider has changed from an object to just a property
* Marginally improved the error message we'll see in the logs so that instead of "Expected 0, but got 1", we'll see "Expected 0 because `terraform apply` should run without error and exit cleanly during infrastructure setup, but got 1"


IsaacCalligeros95 (approved):
LGTM, nice work!

PR #977, opened 2023-01-12
Look at size on disk when calculating compression ratio

mjhilton:
The previous iteration of the Compression Ratio calculation used the compressed size reported by the archive, which wasn't giving an accurate picture and was skewing the ratio.

This PR uses the file size on disk as the compressed size, which gives a correct representation of the Ratio.

### DIFF
--- a/source/Calamari.Common/Features/Packages/Decorators/ArchiveLimits/EnforceCompressionRatioDecorator.cs
+++ b/source/Calamari.Common/Features/Packages/Decorators/ArchiveLimits/EnforceCompressionRatioDecorator.cs
@@ -36,7 +36,7 @@ namespace Calamari.Common.Features.Packages.Decorators.ArchiveLimits
                     {
                         var compressedSize = archiveInfo.Length;
                         var uncompressedSize = archive.TotalUncompressSize;
-                        var compressionRatio = uncompressedSize == 0 ? 0 : (double)uncompressedSize / compressedSize;
+                        var compressionRatio = compressedSize == 0 ? 0 : (double)uncompressedSize / compressedSize;
 
                         if (compressionRatio > maximumCompressionRatio)
                         {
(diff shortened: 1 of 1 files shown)


source/Calamari.Common/Features/Packages/Decorators/ArchiveLimits/EnforceCompressionRatioDecorator.cs
```diff
@@ -30,11 +31,12 @@ bool TryVerifyCompressionRatio(string packageFile, out ArchiveLimitException? ex
...
                 if (maximumCompressionRatio >= 1)
                 {
+                    var archiveInfo = new FileInfo(packageFile);
                     using (var archive = ArchiveFactory.Open(packageFile))
                     {
+                        var compressedSize = archiveInfo.Length;
                         var uncompressedSize = archive.TotalUncompressSize;
-                        var compressedSize = archive.TotalSize;
-                        var compressionRatio = uncompressedSize == 0 ? 0 : uncompressedSize / compressedSize;
+                        var compressionRatio = uncompressedSize == 0 ? 0 : (double)uncompressedSize / compressedSize;
```
YihaoOct, line 39: Hi Matt, it looks to me that if `uncompressedSize == 0`, then the ratio would just automatically be `0`, so I am not sure what this ternary expression is aiming for here. I do see if `compressedSize == 0`, then there would be a division by zero exception. If we wanted to guard against this exception, should it be `var compressionRatio = compressedSize == 0 ? 0 : (double)uncompressedSize / compressedSize;` ?

Maybe there is an assumption that when `compressedSize` is 0, `upcompressedSize` is 0 as well, but I think this could be more explicit.

YihaoOct (changes requested):

source/Calamari.Common/Features/Packages/Decorators/ArchiveLimits/EnforceCompressionRatioDecorator.cs
mjhilton: You're very right: the coalesce is intended to prevent divide-by-zero, but I got them the wrong way around. I'll change it :)

YihaoOct (approved):
LGTM

PR #979, opened 2023-01-16
Fixes incorrect compressed archive size reporting

mjhilton:
As with [the archive limit enforcement code](https://github.com/OctopusDeploy/Calamari/pull/977), the metric logging code has been using an incorrect value for compressed size of archives, throwing out the compression ratio metrics.

This PR fixes that for logging.


mik-ky (approved):
LGTM after looking at the referenced PR.

PR #984, opened 2023-01-26
Bumps Octostache to 3.3.0 to include new `calc` operator

mjhilton:
Introduces ability to do basic calculations in variables substitution, eg `#{calc 1+2}` evaluates to `3`.


mik-ky (approved):
Looks good - can confirm the right version bumps :+1:

PR #986, opened 2023-01-30
Reverts netfx upgrades for `master`

mjhilton:
Our recent `netfx` upgrades from 4.5.2/4.6.1 to 4.8 caused some issues with certain steps running in the Worker Tools execution containers.

We rolled back the version of Calamari in use in Server, but in doing so also nuked some other changes we want to keep.

This PR merges forward the reverts to the netfx upgrade commits so we can re-release the changes we want to keep; the netfx changes can be re-integrated later.


mik-ky (approved):
Looks good. Thanks @mjhilton

PR #987, opened 2023-01-31
Update kubernetes to set health check container feed and image on discovered targets

scottmerchant:
This is the Calamari side of this change, the Server side can be [found here](https://github.com/OctopusDeploy/OctopusDeploy/pull/16435).

The basic gist of this is that the Container Image and Feed are not passed through in the Discovery Context from the k8s steps which trigger the discovery so that any discovery k8s targets have the health check containers set to the same values.

### DIFF
--- a/source/Calamari/Kubernetes/Commands/KubernetesDiscoveryCommand.cs
+++ b/source/Calamari/Kubernetes/Commands/KubernetesDiscoveryCommand.cs
@@ -68,8 +68,8 @@ namespace Calamari.Kubernetes.Commands
                 return ExitStatus.Success;
             }
 
-            var containerFeed = additionalVariables.GetValueOrDefault(SpecialVariables.Container.Feed);
-            var containerImage = additionalVariables.GetValueOrDefault(SpecialVariables.Container.Image);
+            var containerFeed = additionalVariables.TryGetValue(SpecialVariables.Container.Feed, out var feed) ? feed : null;
+            var containerImage = additionalVariables.TryGetValue(SpecialVariables.Container.Image, out var image) ? image: null;
 
             log.Verbose($"Found {clusters.Count} candidate clusters.");
             var discoveredTargetCount = 0;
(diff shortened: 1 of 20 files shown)


source/Calamari/Kubernetes/Commands/KubernetesDiscoveryCommand.cs
```diff
@@ -142,24 +146,29 @@ bool TryGetDiscoveryContextJson(out string json)
...
                 scope = discoveryContext?.Scope;
-                
+                additionalVariables = discoveryContext?.AdditionalVariables != null
+                    ? new Dictionary<string, string>(discoveryContext.AdditionalVariables, StringComparer.OrdinalIgnoreCase)
+                    : new Dictionary<string, string>();
+
                 if (type != null && scope != null)
                     return true;
-                
+
```
gb-8, line 171: The error message logged in the `catch` block could be better.
Parsing could fail even `Type` or `Scope` can be extracted. E.g. when it includes an `AdditionalVariables` property that cannot be parsed as a dictionary.

gb-8: The above comment is redundant now that the `AdditionalVariables` property is gone, but the messages could still be improved to say `"Authentication Type"` instead of just `Type`.

source/Calamari/Kubernetes/Commands/KubernetesDiscoveryCommand.cs
```diff
@@ -153,13 +154,13 @@ bool TryGetDiscoveryContextJson(out string json)
...
                 type = discoveryContext?.Authentication?.Type;
                 scope = discoveryContext?.Scope;
-                
+
                 if (type != null && scope != null)
                     return true;
-                
+
                 log.Error($"Could not extract Type or Scope from {ContextVariableName}, the data is in the wrong format.");
```
gb-8, line 164:
```suggestion
                log.Error($"Could not extract Authentication Type or Scope from {ContextVariableName}, the data is in the wrong format.");
```

gb-8 (approved):
Looks good. Just left a nit pick on an existing error message that can be improved.

PR #988, opened 2023-02-02
Fix slow syncing due to nuke

veochen-octopus:
Looks like Calamari suffers from the same syncing issue due to `Nuke` scanning too aggressively. Same fix applied here as copied from https://github.com/OctopusDeploy/OctopusDeploy/pull/16288

EDIT: The slowness is not consistently reproducible. Happened to me when I loaded up Calamari for the first time in a long time. But seems fine for others who work on Calamari a lot. So this fix is more for future proofing and better practice.

### DIFF
--- a/build/_build.csproj
+++ b/build/_build.csproj
@@ -13,12 +13,16 @@
         <Nullable>enable</Nullable>
     </PropertyGroup>
 
+    <PropertyGroup>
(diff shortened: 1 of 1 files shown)


IsaacCalligeros95 (approved):
LGTM

PR #995, opened 2023-02-28
Retrieve k8s status

YihaoOct:
This PR is part of the [Kubernetes Object Status narrative](https://docs.google.com/document/d/1jyEHJt4s8bTTR4okm9024MHWq0JApVo18xVEcAhYytI/edit?usp=sharing) (internal). It adds a new ScriptWrapper that is enabled if the user enables the Kubernetes Object Status feature, and runs after a Kubernetes deployment is executed. It checks the statuses of Kubernetes resources deployed (including indirectly) within a step, and sends the information to Server via service message on update.

This PR supports Deployments, ReplicaSets, Pods, Services, EndpointSlices and ConfigMaps. For unsupported resource types, the status will still be sent back to Server, but the status is always deemed successful and no extra information fields will be added.

For Pods, the status is now solely decided by the "Phase" field, which isn't the same as the "Status" column of "kubectl get pod" command. We will improve this later PRs.

In the case of blue-green deployment, or deployments that have "wait until completion" option selected, the status check wrapper will be skipped. The behaviour would be the same as before.


source/Calamari/Kubernetes/ResourceStatus/ResourceStatusChecker.cs
```diff
@@ -0,0 +1,108 @@
...
+                    resourceEntry.Value.Removed = true;
+                    removed.Add(resourceEntry.Value);
+                }
+            }
+            return removed;
+        }
+
+        private void SendServiceMessage(Resource resource)
+        {
+            // TODO: update this for database
```
johnsimons, line 97: Do we still need this comment?

johnsimons: I actually don't know what it means!

YihaoOct: The current service message is in the shape that is to be handled as a "DAT" type of log entry. This is in order for the Server to store it in the logs. So it has the Service Message name of "logData", type of "k8s-status" and everything else is in the "data" field.

But once we change to the database solution, I imagine we will implement an IHandleServiceMessage and update database there. In this case, we could use a better schema, for example, the name could be "k8s-resource-status" rather than "logData" which doesn't make too much sense if we don't store it in the logs. Also things like ActionId and Cluster name could be added directly as Service Message properties, so the Server doesn't even need to inspect the content of the data

johnsimons: cool, I'm changing it now

source/Calamari/Kubernetes/ResourceStatus/ResourceUpdateReporter.cs
```diff
@@ -0,0 +1,90 @@
...
+            this.log = log;
+        }
+        
+        public void ReportUpdatedResources(IDictionary<string, Resource> originalStatuses, IDictionary<string, Resource> newStatuses)
+        {
+            var createdOrUpdatedResources = GetCreatedOrUpdatedResources(originalStatuses, newStatuses);
+            var removedResources = GetRemovedResources(originalStatuses, newStatuses);
+            foreach (var resource in createdOrUpdatedResources.Concat(removedResources))
+            {
+                log.Verbose($"Resource updated: {JsonConvert.SerializeObject(resource)}");
```
johnsimons, line 34: I don't think we need this?

YihaoOct: I just removed the verbose log

source/Calamari.Tests/Calamari.Tests.csproj
```diff
@@ -10,6 +10,7 @@
     <ApplicationIcon />
     <StartupObject />
     <RuntimeIdentifiers>win-x64;linux-x64;osx-x64;linux-arm;linux-arm64</RuntimeIdentifiers>
+    <LangVersion>7.3</LangVersion>
```
johnsimons, line 13: Why do we need this @YihaoOct ?

source/Calamari.Tests/KubernetesFixtures/ResourceStatus/KubernetesYamlTests.cs
```diff
@@ -0,0 +1,31 @@
...
+            identifier.Name.Should().Be("nginx");
+            identifier.Namespace.Should().Be("default");
+        }
+
+        [Test]
+        public void ShouldHandleMultipleResourcesDefinedInTheSameFile()
+        {
+            var input = TestFileLoader.Load("multiple-resources.yaml");
+            var got = KubernetesYaml.GetDefinedResources(input);
+            got.Should().HaveCount(3);
```
johnsimons, line 28: I think we should also validate that we parsed the correct combo of type + name + namespace for each resource

source/Calamari.Tests/KubernetesFixtures/ResourceStatus/ResourceRetrieverTests.cs
```diff
@@ -0,0 +1,121 @@
...
+
+            var kind = arguments[1];
+            var nameProvided = arguments.Length >= 3 && !arguments[2].StartsWith("-");
+            
+            string output;
+            if (!nameProvided)
+            {
+                // when resource name is not provided, return all resources of that kind in a List response
+                var items = new JArray(data.Where(item => item.SelectToken("$.kind").Value<string>() == kind));
+                output = $"{{items: {items}}}" ;
```
johnsimons, line 94: it may be nicer to user the `yield return` syntax in this function?

source/Calamari.Common/Features/Scripting/ScriptWrapperPriorities.cs
```diff
@@ -7,6 +7,16 @@ namespace Calamari.Common.Features.Scripting
     /// </summary>
     public static class ScriptWrapperPriorities
     {
+        /// <summary>
+        /// The priority for script wrappers that configure kubernetes authentication
+        /// </summary>
+        public const int KubernetesContextPriority = 1002;
```
johnsimons, line 13: Should this be `ScriptWrapperPriorities.CloudAuthenticationPriority + 2` ?

source/Calamari/Kubernetes/ResourceStatus/ResourceRetriever.cs
```diff
@@ -0,0 +1,65 @@
...
+            }
+            return resources;
+        }
+    
+        private Resource GetResource(ResourceIdentifier resourceIdentifier, IKubectl kubectl)
+        {
+            var result = kubectl.ExecuteCommandAndReturnOutput(new[]
+            {
+                "get", resourceIdentifier.Kind, resourceIdentifier.Name, "-o json", $"-n {resourceIdentifier.Namespace}"
+            }).Join("");
```
johnsimons, line 42: use `string.empty`

source/Calamari/Kubernetes/ResourceStatus/ResourceRetriever.cs
```diff
@@ -0,0 +1,65 @@
...
+            var childKind = parentResource.ChildKind;
+            if (string.IsNullOrEmpty(childKind))
+            {
+                return Enumerable.Empty<Resource>();
+            }
+            
+            var result = kubectl.ExecuteCommandAndReturnOutput(new[]
+            {
+                "get", childKind, "-o json", $"-n {parentResource.Namespace}"
+            }).Join("");
```
johnsimons, line 57: use `string.empty`

source/Calamari/Kubernetes/ResourceStatus/ResourceStatusChecker.cs
```diff
@@ -0,0 +1,132 @@
...
+        /// unitl the deployment timeout is met, or the deployment has succeeded or failed after stably.
+        /// </summary>
+        /// <returns>true if all resources have been deployed successfully, otherwise false</returns>
+        bool CheckStatusUntilCompletionOrTimeout(IEnumerable<ResourceIdentifier> resourceIdentifiers, 
+            ICountdownTimer deploymentTimer, 
+            ICountdownTimer stabilizationTimer, 
+            IKubectl kubectl);
+    }
+
+    public enum DeploymentStatus
```
johnsimons, line 25: maybe move to its own file

johnsimons (changes requested):

source/Calamari.Tests/Calamari.Tests.csproj
YihaoOct: Good catch, I added this to make Rider pick up some incompatible syntax in higher versions of C# that will not build against .net 4 such as file-scoped namespaces and so on. But it turned out there are still some syntax that wasn't captured even with this. This means 7.3 is not the correct version to use to guarantee successful compilation for older .net targets.

I will try to find a better version to set to facilitate local development

source/Calamari.Common/Features/Scripting/ScriptWrapperPriorities.cs
YihaoOct: They are the same value, right? Since all priorities are now defined in the same class next to each other, later developers should be able to pretty easily figure out the order when reading this class.

source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureBase.cs
```diff
@@ -184,17 +183,6 @@ protected void DoDiscovery(AwsAuthenticationDetails authenticationDetails)
...
-        protected class ServiceMessageCollectorLog : InMemoryLog
-        {
-            public List<ServiceMessage> ServiceMessages { get; } = new List<ServiceMessage>();
-            public override void WriteServiceMessage(ServiceMessage serviceMessage)
-            {
-                ServiceMessages.Add(serviceMessage);
-                
-                base.WriteServiceMessage(serviceMessage);
-            }
-        }
```
YihaoOct, line 197: I incorporated this into `InMemoryLog` - didn't see why we wouldn't do so since `InMemoryLog` is only used in tests.

zentron: This looks like a good change.
To keep your PR clean and focused though, I'd recommend this change, be made in its own PR.
The smaller the set of changes the easier it is for the reviewer to understand the full context

source/Calamari/Kubernetes/ResourceStatus/ResourceStatusReportWrapper.cs
```diff
@@ -0,0 +1,128 @@
...
+            // TODO this won't handle configMaps defined together with a Deploy a Container step
+            var customResourceFileName =
+                variables.Get("Octopus.Action.KubernetesContainers.CustomResourceYamlFileName");
+            var knownFileNames = new[]
+            {
+                "secret.yml", customResourceFileName, "deployment.yml", "service.yml", "ingress.yml",
+            };
+            foreach (var file in knownFileNames)
+            {
+                if (!fileSystem.FileExists(file))
```
zentron, line 115: Does this mean that if a user opts to load the config from a package that _doesnt_ match one of those names above, then it wont take effect?

source/Calamari/Kubernetes/ResourceStatus/ResourceUpdateReporter.cs
```diff
@@ -0,0 +1,96 @@
...
+        {
+            var createdOrUpdated = new List<Resource>();
+            foreach (var resource in newStatuses)
+            {
+                if (!originalStatuses.ContainsKey(resource.Key) || resource.Value.HasUpdate(originalStatuses[resource.Key]))
+                {
+                    createdOrUpdated.Add(resource.Value);
+                }
+            }
+            return createdOrUpdated;
```
zentron, line 54:
```suggestion
                   return newStatuses
                   .Where(resource => !originalStatuses.ContainsKey(resource.Key) || resource.Value.HasUpdate(originalStatuses[resource.Key]))
                   .Select(resource => resource.Value);
```

source/Calamari/Kubernetes/ResourceStatus/ResourceUpdateReporter.cs
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        private static IEnumerable<Resource> GetRemovedResources(IDictionary<string, Resource> originalStatuses, IDictionary<string, Resource> newStatuses)
+        {
+            var removed = new List<Resource>();
+            foreach (var resource in originalStatuses)
+            {
+                if (!newStatuses.ContainsKey(resource.Key))
+                {
+                    resource.Value.Removed = true;
```
zentron, line 64: This mutation of the resource list passed in was unexpected

source/Calamari/Kubernetes/ResourceStatus/ResourceUpdateReporter.cs
```diff
@@ -0,0 +1,96 @@
...
+        private void SendServiceMessage(Resource resource)
+        {
+            var parameters = new Dictionary<string, string>
+            {
+                {"type", "k8s-status"},
+                {"actionId", variables.Get("Octopus.Action.Id")},
+                {"taskId", variables.Get(KnownVariables.ServerTask.Id)},
+                {"targetId", variables.Get("Octopus.Machine.Id")},
+                {"environmentId", variables.Get(DeploymentEnvironment.Id)},
+                {"spaceId", variables.Get("Octopus.Space.Id")},
```
zentron, line 80: It feels like this operation shouldnt care about spaceId, environmentId etc. Thats just a result of the task its being logged against.
We should ideally be able to determine this from the context of the operation over in Octopus

source/Calamari/Kubernetes/ResourceStatus/Resources/Deployment.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        public Deployment(JObject json) : base(json)
+        {
+            Replicas = FieldOrDefault("$.status.replicas", 0);
+            Available = FieldOrDefault("$.status.availableReplicas", 0);
+            Ready = FieldOrDefault("$.status.readyReplicas", 0);
+            UpToDate = FieldOrDefault("$.status.updatedReplicas", 0);
+            Unavailable = FieldOrDefault("$.status.unavailableReplicas", 0);
+    
+            if (UpToDate == Replicas && Available == Replicas && Ready == Replicas)
```
zentron, line 24: minor, but this seems simple enough it could get away with a ternary expression

source/Calamari/Kubernetes/ResourceStatus/Resources/ResourceStatus.cs
```diff
@@ -0,0 +1,7 @@
+namespace Calamari.Kubernetes.ResourceStatus.Resources
+{
+    public enum ResourceStatus
+    {
+        InProgress, Successful, Failed
```
zentron, line 5: Minor formatting consistency, single line per enum

zentron (changes requested):
This looks like some great progress.

source/Calamari/Kubernetes/Integration/Kubectl.cs
```diff
@@ -5,14 +5,22 @@
...
 using Calamari.Common.Plumbing.Logging;
-using Calamari.Common.Plumbing.Variables;
 using Newtonsoft.Json;
 using Octopus.CoreUtilities;
 using Octopus.Versioning.Semver;
 
 namespace Calamari.Kubernetes.Integration
 {
-    public class Kubectl : CommandLineTool
+    public interface IKubectl
```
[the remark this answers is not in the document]
YihaoOct: This interface was added because I wanted to swap the implementation with a mock in the tests, and at the same time I wanted to rely on the existing `Kubectl` implementation which contains the methods to locate custom kubectl tool etc.

Thinking about it now, it would nicer to have a smaller interface that is only used in the status check part (maybe called `IKubectlGet`) which takes the existing `Kubectl` tool as a dependency. This will make the mock implementation much easier to write as well.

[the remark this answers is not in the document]
YihaoOct: Just updated

source/Calamari/Kubernetes/ResourceStatus/ResourceStatusReportWrapper.cs
YihaoOct: When customer selects a filename from a package, the name is set to the `Octopus.Action.KubernetesContainers.CustomResourceYamlFileName` variable. It is included in the list, so it should take effect.

source/Calamari/Kubernetes/ResourceStatus/ResourceUpdateReporter.cs
YihaoOct: @johnsimons Based on your work on the database side, shall we remove some of these IDs if they are not used?

johnsimons: I am using all of these @YihaoOct

johnsimons: Given we have the info here, I don't see the reason to add more work to the server to do these lookups.
@zentron apart from sending this info across the wire, is there a technical reason to remove these?

source/Calamari/Kubernetes/ResourceStatus/KubernetesYaml.cs
```diff
@@ -0,0 +1,46 @@
...
+                result.Add(GetDefinedResource(parser));
+            }
+    
+            return result;
+        }
+    
+        private static ResourceIdentifier GetDefinedResource(IParser parser)
+        {
+            var yamlObject = Deserializer.Deserialize<dynamic>(parser);
+            yamlObject["metadata"].TryGetValue("namespace", out object @namespace);
```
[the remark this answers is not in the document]
YihaoOct: Thanks for capturing this. I've added the logic to skip the status check wrapper if the wrapper that runs `kubectl apply` returns non-zero exit code. So, if we could reach this piece of code, it means `kubectl apply` was successful and so the `metadata` section shouldn't be missing, because it should contain the required information "name". But just to keep it robust, I've also added try-catch around this piece of code so now it omits bad yaml definitions rather than blowing up.

source/Calamari/Kubernetes/ResourceStatus/ResourceUpdateReporter.cs
zentron: More an architectural reason. Its possible that the task might not be running for a project, or at some point even an environment. It also feels like a bunch of redundant information given its already connected via the server task. 
Lets catch up this morning.

source/Calamari/Kubernetes/ResourceStatus/ResourceStatusReportWrapper.cs
```diff
@@ -0,0 +1,128 @@
...
+            var stabilizationTimeoutSeconds = variables.GetInt32(SpecialVariables.StabilizationTimeout) ?? 10;
+            var workingDirectory = Path.GetDirectoryName(script.File);
+            var definedResources = KubernetesYaml.GetDefinedResources(content).ToList();
+
+            if (definedResources.Count == 0)
+            {
+                return result;
+            }
+
+            foreach (var proxyVariable in ProxyEnvironmentVariablesGenerator.GenerateProxyEnvironmentVariables())
```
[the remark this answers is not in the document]
YihaoOct: I replicated this logic from existing wrappers. For these environment variables, they are passed to the point where Calamari creates a processes to run commands. So we need to make sure these variables exist every time before we invoke a command. I will have a proper look at this and see how we could reduce this replication in a separate PR.

source/Calamari/Kubernetes/ResourceStatus/ResourceRetriever.cs
```diff
@@ -0,0 +1,73 @@
...
+            var childKind = parentResource.ChildKind;
+            if (string.IsNullOrEmpty(childKind))
+            {
+                return Enumerable.Empty<Resource>();
+            }
+
+            var result = kubectlGet.AllResources(childKind, parentResource.Namespace, kubectl);
+            var resources = ResourceFactory.FromListJson(result);
+            return resources.Where(resource => resource.OwnerUids.Contains(parentResource.Uid)).ToList();
+        }
```
zentron, line 71: Moving the assignment into where the resource is created will allow us to cut out the `EnrichWithChildren` method
```suggestion
            return resources.Where(resource => resource.OwnerUids.Contains(parentResource.Uid))
                            .Select(child =>
                                    {
                                        child.Children = GetChildrenResources(child, kubectl);
                                        return child;
                                    })
                            .ToList();
```

line 30 above then also just becomes
```
            foreach (var resource in resources)
            {
                resource.Children = GetChildrenResources(resource, kubectl);
            }
```

Generally I like to avoid objects being modified outside the scope of their creation when possible so we can consider them immutable outside that context.

zentron (changes requested):
Couple minor refactoring requests but its looking pretty close.

zentron (approved):
Nice, looking forward to being able to actually test this through a real instance

johnsimons (approved):

PR #997, opened 2023-03-07
Fix failing Calamari build

scottmerchant:
The build is failing because one call seems to return json as separate log lines instead of a single log line containing the json blob. Unsure why it has randomly started doing this as I can't see a reason.. but it could be the AWS CLI? Unsure if we updated a version of something that would affect this.

[sc-42546]

### DIFF
--- a/source/Calamari.Tests/KubernetesFixtures/Approved/KubernetesContextScriptWrapperFixture.ExecutionWithEKS_IAMAuthenticator.approved.txt
+++ b/source/Calamari.Tests/KubernetesFixtures/Approved/KubernetesContextScriptWrapperFixture.ExecutionWithEKS_IAMAuthenticator.approved.txt
@@ -8,7 +8,7 @@
 [Verbose] "kubectl" config set-cluster octocluster --insecure-skip-tls-verify=true --request-timeout=1m
 [Info] Creating kubectl context to <server> (namespace calamari-testing) using EKS cluster name my-eks-cluster
 [Verbose] Attempting to authenticate with aws-cli
-[Verbose] Unable to authenticate to <server> using the aws cli. Failed with error message: Error reading JObject from JsonReader. Path '', line 0, position 0.
+[Verbose] Unable to authenticate to <server> using the aws cli. Failed with error message: Unexpected character encountered while parsing value: P. Path '', line 0, position 0.
 [Verbose] Attempting to authenticate with aws-iam-authenticator
 [Verbose] "kubectl" config set-credentials octouser --exec-command=aws-iam-authenticator --exec-api-version=client.authentication.k8s.io/v1beta1 --exec-arg=token --exec-arg=-i --exec-arg=my-eks-cluster --request-timeout=1m
 [Verbose] "kubectl" get namespace calamari-testing --request-timeout=1m
--- a/source/Calamari.Tests/KubernetesFixtures/InstallTools.cs
+++ b/source/Calamari.Tests/KubernetesFixtures/InstallTools.cs
@@ -209,7 +209,7 @@ namespace Calamari.Tests.KubernetesFixtures
                         var downloadUrl = GetGcloudDownloadLink(tuple.version);
                         var fileName = GetGcloudZipFileName(tuple.version);
 
-                        await DownloadGcloud(GetGcloudZipFileName(tuple.version),
+                        await DownloadGcloud(fileName,
                             client,
                             downloadUrl,
                             destinationDirectoryName);
--- a/source/Calamari.Tests/KubernetesFixtures/InstallTools.cs
+++ b/source/Calamari.Tests/KubernetesFixtures/InstallTools.cs
@@ -133,7 +133,7 @@ namespace Calamari.Tests.KubernetesFixtures
                 AwsCliExecutable = await DownloadCli("aws",
                     () =>
                     {
-                        var version = "2.11.0";
+                        var version = "latest";
                         return Task.FromResult((version, GetAwsCliDownloadLink(version)));
                     },
                     async (destinationDirectoryName, tuple) =>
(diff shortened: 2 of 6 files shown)


source/Calamari.Tests/KubernetesFixtures/KubernetesRunsScriptFixture.cs
```diff
@@ -238,7 +238,6 @@ public void ExecutionWithEKS_IAMAuthenticator()
         }
         
         [Test]
-        [WindowsTest] // This test requires the aws cli tools. Currently only configured to install on Windows
```
scottmerchant, line 241: This is not true anymore, I don't think. I've removed it to see if the test passes on linux.

scottmerchant: I was wrong - linux needs sudo to install the tool, which blows up the test on the workers. 😞

scottmerchant:
I've added some logging because it looks like the output is not simply the JSON like it is on my local machine. Once I know what we receive I can add some code to parse it correctly.

source/Calamari.Tests/KubernetesFixtures/Approved/KubernetesContextScriptWrapperFixture.ExecutionWithEKS_IAMAuthenticator.approved.txt
```diff
@@ -8,7 +8,7 @@
 [Verbose] "kubectl" config set-cluster octocluster --insecure-skip-tls-verify=true --request-timeout=1m
 [Info] Creating kubectl context to <server> (namespace calamari-testing) using EKS cluster name my-eks-cluster
 [Verbose] Attempting to authenticate with aws-cli
-[Verbose] Unable to authenticate to <server> using the aws cli. Failed with error message: Error reading JObject from JsonReader. Path '', line 0, position 0.
+[Verbose] Unable to authenticate to <server> using the aws cli. Failed with error message: Unexpected character encountered while parsing value: P. Path '', line 0, position 0.
```
scottmerchant, line 11: This test apparently was always failing to parse the authentication object because it passes in a bogus region `<server>` so I just updated the error as it's changed now that I have joined all the log lines before parsing.

source/Calamari.Tests/KubernetesFixtures/KubernetesRunsScriptFixture.cs
johnsimons: This is interesting because we are using `sudo` in https://github.com/OctopusDeploy/Calamari/blob/d203d663372dd02ae33920ab420eabf2e3a5e99a/source/Calamari.Tests/KubernetesFixtures/InstallTools.cs#L166
Couldn't we do the same here too @scottmerchant ?

johnsimons:
@scottmerchant I think the issue is that we are not using a fixed version for `awscli`.
This version [here](https://github.com/OctopusDeploy/Calamari/blob/303bf359f5c5135aab080fc4b48f57927b7c1c01/source/Calamari.Tests/KubernetesFixtures/InstallTools.cs#L136) is not being used as part of the URL.
According to the [doco](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-version.html), the URL should be something like https://awscli.amazonaws.com/awscli-exe-linux-x86_64-2.0.30.zip

source/Calamari.Tests/KubernetesFixtures/KubernetesRunsScriptFixture.cs
scottmerchant: Ah yep I guess so - I didn't look to much into it as originally it wasn't running on linux so I just thought I'd try and see how it goes. I think it's probably worth just getting this change in to get the build green so Yihao can get his PR going. I'll fix the cli version as you suggested though 👍

scottmerchant:
> @scottmerchant I think the issue is that we are not using a fixed version for `awscli`. This version [here](https://github.com/OctopusDeploy/Calamari/blob/303bf359f5c5135aab080fc4b48f57927b7c1c01/source/Calamari.Tests/KubernetesFixtures/InstallTools.cs#L136) is not being used as part of the URL. According to the [doco](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-version.html), the URL should be something like https://awscli.amazonaws.com/awscli-exe-linux-x86_64-2.0.30.zip

Ah yep good idea. I'll add a change to make it fixed in this PR.

scottmerchant:
@johnsimons I've updated the install tools area to download a specific version of the CLI and I've updated it to download the latest version `v2.11.0`.

source/Calamari.Tests/KubernetesFixtures/InstallTools.cs
```diff
@@ -133,8 +133,8 @@ public async Task InstallAwsCli()
                 AwsCliExecutable = await DownloadCli("aws",
                     () =>
                     {
-                        var version = "2.8.3";
-                        return Task.FromResult((version, GetAwsCliDownloadLink()));
+                        var version = "latest";
+                        return Task.FromResult((version, GetAwsCliDownloadLink(version)));
```
scottmerchant, line 137: Good point from Yihao was that we want this test to fail if latest aws cli doesn't work with our stuff so I updated this to download the latest but you can optionally change it on your dev machine if you want to test a specific version.

Note: on mac it doesn't install the version it downloads - you need to manually install the version you want to run. 🤷‍♂️

johnsimons: it feels like we have a mix of testing low and high ranges

scottmerchant: I suppose we could fix the version as long as it's noted in the doco.

johnsimons: Aren't the other versions fixed for the other tooling?

johnsimons: Maybe something to ask others, what is our stance on this?

scottmerchant: I have fixed the version to what it was meant to be fixed to originally. Can you please give it a final check over and let me know if there are any other issues? 🙏

johnsimons (approved):

PR #998, opened 2023-03-09
Consolidate ServiceMessageCollectorLog and InMemoryLog

YihaoOct:
In some Kubernetes fixtures we need to assert on the service message sent to the server. Since `InMemoryLog` does not provide methods to record the service messages, a `ServiceMessageCollectorLog` class was extended from it to provide this capability. This was needed because before all Calamari flavours were incorporated, the Kubernetes fixtures and `InMemoryLog` lived in separate repositories.

Now all flavours are consolidated in this single repository, the service message collection methods should be moved into the `InMemoryLog` itself.

### DIFF
--- a/source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureAks.cs
+++ b/source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureAks.cs
@@ -134,9 +134,6 @@ namespace Calamari.Tests.KubernetesFixtures
         [TestCase(true)]
         public void DiscoverKubernetesClusterWithAzureServicePrincipalAccount(bool setHealthCheckContainer)
         {
-            var serviceMessageCollectorLog = new ServiceMessageCollectorLog();
-            Log = serviceMessageCollectorLog;
-
             var scope = new TargetDiscoveryScope("TestSpace",
                 "Staging",
                 "testProject",
@@ -197,10 +194,10 @@ namespace Calamari.Tests.KubernetesFixtures
                 KubernetesDiscoveryCommand.CreateKubernetesTargetServiceMessageName,
                 serviceMessageProperties);
 
-            serviceMessageCollectorLog.ServiceMessages.Should()
-                                      .ContainSingle(s => s.Properties["name"] == targetName)
-                                      .Which.Should()
-                                      .BeEquivalentTo(expectedServiceMessage);
+            Log.ServiceMessages.Should()
+                .ContainSingle(s => s.Properties["name"] == targetName)
+                .Which.Should()
+                .BeEquivalentTo(expectedServiceMessage);
         }
     }
 }
--- a/source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureBase.cs
+++ b/source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureBase.cs
@@ -150,19 +150,16 @@ namespace Calamari.Tests.KubernetesFixtures
             AwsAuthenticationDetails authenticationDetails,
             Dictionary<string,string> properties)
         {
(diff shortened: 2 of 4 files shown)


source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureAks.cs
```diff
@@ -134,8 +134,7 @@ public void UnreachableK8Cluster_ShouldExecuteTargetScript()
         [TestCase(true)]
         public void DiscoverKubernetesClusterWithAzureServicePrincipalAccount(bool setHealthCheckContainer)
         {
-            var serviceMessageCollectorLog = new ServiceMessageCollectorLog();
-            Log = serviceMessageCollectorLog;
+            Log = new InMemoryLog();
```
johnsimons, line 137: I don't think we need this

source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureBase.cs
```diff
@@ -150,19 +150,18 @@ protected void DoDiscovery(AwsAuthenticationDetails authenticationDetails)
             AwsAuthenticationDetails authenticationDetails,
             Dictionary<string,string> properties)
         {
-            var serviceMessageCollectorLog = new ServiceMessageCollectorLog();
-            Log = serviceMessageCollectorLog;
+            Log = new InMemoryLog();
```
johnsimons, line 153: I don't think we need this

source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureEks.cs
```diff
@@ -350,23 +350,22 @@ public void DiscoverKubernetesClusterWithNoValidCredentials()
                     Role = new AwsAssumedRole { Type = "noAssumedRole" },
                     Regions = new []{region}
                 };
-                
-                var serviceMessageCollectorLog = new ServiceMessageCollectorLog();
-                Log = serviceMessageCollectorLog;
+
+                Log = new InMemoryLog();
```
johnsimons, line 354: I don't think we need this

source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureEks.cs
```diff
@@ -395,22 +394,21 @@ public void DiscoverKubernetesClusterWithInvalidAccountCredentials()
                 Regions = new []{region}
             };
             
-            var serviceMessageCollectorLog = new ServiceMessageCollectorLog();
-            Log = serviceMessageCollectorLog;
+            Log = new InMemoryLog();
```
johnsimons, line 397: I don't think we need this

johnsimons (changes requested):
LGTM but I don't think we need to reset the Log, can you look at that @YihaoOct

source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureAks.cs
YihaoOct: Yes, I think you are right. The `Log` is `InMemoryLog` by default: https://github.com/OctopusDeploy/Calamari/blob/d249e6bb18e38afc6249d15e32ea18ceb32ef797/source/Calamari.Tests/Helpers/CalamariFixture.cs#L25

zentron (approved):
Nice change. Thanks for thinking about how we can continue to use and extend the existing InMemoryLog and for pulling this out into its own PR.

Hopefully that feels like the right approach as compared with keeping in the original PR.

johnsimons (approved):

PR #999, opened 2023-03-15
Handle disabled timeouts

YihaoOct:
This PR handles the scenario where the timeouts are disabled.

When the total deployment timeout is disabled, it is treated as infinite timeout that never completes.
When the stabilisation timeout is disabled, it is treated as zero timeout which does not stabilise.

These timeouts are treated as disabled if unset.

Screenshot 2023-03-20 at 12 38 53 PM

### DIFF
--- a/source/Calamari/Kubernetes/ResourceStatus/ResourceStatusReportWrapper.cs
+++ b/source/Calamari/Kubernetes/ResourceStatus/ResourceStatusReportWrapper.cs
@@ -59,8 +59,10 @@ namespace Calamari.Kubernetes.ResourceStatus
             Dictionary<string, string> environmentVars)
         {
             var customKubectlExecutable = variables.Get(SpecialVariables.CustomKubectlExecutable);
-            var deploymentTimeoutSeconds = variables.GetInt32(SpecialVariables.DeploymentTimeout) ?? 100;
-            var stabilizationTimeoutSeconds = variables.GetInt32(SpecialVariables.StabilizationTimeout) ?? 10;
+            var deploymentTimeoutSeconds = variables.GetInt32(SpecialVariables.DeploymentTimeout) ?? 0;
+            var stabilizationTimeoutSeconds = variables.GetInt32(SpecialVariables.StabilizationTimeout) ?? 0;
+            var deploymentTimeoutEnabled = variables.GetFlag(SpecialVariables.DeploymentTimeoutEnabled);
+            var stabilizationTimeoutEnabled = variables.GetFlag(SpecialVariables.StabilizationTimeoutEnabled);
             var workingDirectory = Path.GetDirectoryName(script.File);
             
             var result = NextWrapper.ExecuteScript(script, scriptSyntax, commandLineRunner, environmentVars);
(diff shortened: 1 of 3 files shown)


source/Calamari/Kubernetes/ResourceStatus/ResourceStatusReportWrapper.cs
```diff
@@ -59,8 +59,10 @@ public bool IsEnabled(ScriptSyntax syntax)
             Dictionary<string, string> environmentVars)
         {
             var customKubectlExecutable = variables.Get(SpecialVariables.CustomKubectlExecutable);
-            var deploymentTimeoutSeconds = variables.GetInt32(SpecialVariables.DeploymentTimeout) ?? 100;
-            var stabilizationTimeoutSeconds = variables.GetInt32(SpecialVariables.StabilizationTimeout) ?? 10;
+            var deploymentTimeoutSeconds = variables.GetInt32(SpecialVariables.DeploymentTimeout) ?? 0;
+            var stabilizationTimeoutSeconds = variables.GetInt32(SpecialVariables.StabilizationTimeout) ?? 0;
```
YihaoOct, line 63: These defaults were added for the ease of debugging only. They should actually be set from the frontend. When the default values are set from the frontend, the defaults here do not take effect. So they are removed to avoid confusion.

source/Calamari/Kubernetes/ResourceStatus/ResourceStatusReportWrapper.cs
```diff
@@ -92,14 +92,15 @@ public bool IsEnabled(ScriptSyntax syntax)
                 return new CommandResult(string.Empty, 1);
             }
 
-            var stabilizingTimer = new StabilizingTimer(
-                new CountdownTimer(TimeSpan.FromSeconds(deploymentTimeoutSeconds)),
-                new CountdownTimer(TimeSpan.FromSeconds(stabilizationTimeoutSeconds)));
+            var deploymentTimer = deploymentTimeoutSeconds == 0
+                ? new InfiniteCountdownTimer() as ICountdownTimer
+                : new CountdownTimer(TimeSpan.FromSeconds(deploymentTimeoutSeconds)) as ICountdownTimer;
```
johnsimons, line 97: no need to the case ` as ICountdownTimer`

source/Calamari/Kubernetes/SpecialVariables.cs
```diff
@@ -16,6 +16,8 @@ public static class SpecialVariables
         public const string DeploymentWait = "Octopus.Action.KubernetesContainers.DeploymentWait";
         public const string DeploymentTimeout = "Octopus.Action.Kubernetes.DeploymentTimeout";
         public const string StabilizationTimeout = "Octopus.Action.Kubernetes.StabilizationTimeout";
+        public const string DeploymentTimeoutEnabled = "Octopus.Action.Kubernetes.DeploymentTimeoutEnabled";
```
johnsimons, line 19: These don't seem to be used anywhere?

johnsimons (changes requested):
a few little comments @YihaoOct

source/Calamari/Kubernetes/SpecialVariables.cs
YihaoOct: Yes, sorry I forgot to remove this

source/Calamari/Kubernetes/ResourceStatus/ResourceStatusReportWrapper.cs
YihaoOct: It fails compilation for lower versions of C# on TC, like this:

```
C:\BuildAgent\work\e0cefbed4ad11812\source\Calamari\Kubernetes\ResourceStatus\ResourceStatusReportWrapper.cs(97,47): error CS8957: Conditional expression is not valid in language version 7.3 because a common type was not found between 'InfiniteCountdownTimer' and 'CountdownTimer'. To use a target-typed conversion, upgrade to language version 9.0 or greater. [C:\BuildAgent\work\e0cefbed4ad11812\source\Calamari\Calamari.csproj]
```

johnsimons: Wondering if we could explicitly declare the type instead fo `var`?

johnsimons (approved):
I left a small comment. LGTM

source/Calamari/Kubernetes/ResourceStatus/ResourceStatusReportWrapper.cs
YihaoOct: I tried but that didn't work either, because the right hand side of the assignment is evaluated first and the error occurred there. Without the "target-typed conversion" feature,  the type we declared on the left hand side didn't take effect until the right-hand side is evaluated.

PR #1000, opened 2023-03-20
Handle unset namespace in yaml and remove unused properties sent to server

YihaoOct:
Before this change, if a user deploys a "Raw Yaml" step and the yaml file does not explicitly set the .metadata.namespace field, we will use "default" as the namespace regardless the namespace field directly. This is not the correct behaviour.

This PR pass the set namespace field from the step configuration to the yaml parser, so that the set namespace is respected when the namespace is missing from yaml.

This PR also removes some properties from the service message sent to server which are no longer in use.


source/Calamari/Kubernetes/ResourceStatus/ResourceUpdateReporter.cs
```diff
@@ -67,12 +67,9 @@ private void SendServiceMessage(Resource resource, bool removed)
                 {"actionId", variables.Get("Octopus.Action.Id")},
                 {"stepName", $"Step {variables.Get("Octopus.Action.Number")}: {variables.Get("Octopus.Action.StepName")}"},
                 {"taskId", variables.Get(KnownVariables.ServerTask.Id)},
-                {"targetId", variables.Get("Octopus.Machine.Id")},
```
johnsimons, line 70: this one should not be deleted

source/Calamari/Kubernetes/ResourceStatus/ResourceUpdateReporter.cs
```diff
@@ -67,12 +67,9 @@ private void SendServiceMessage(Resource resource, bool removed)
...
                 {"targetName", variables.Get("Octopus.Machine.Name")},
-                {"environmentId", variables.Get(DeploymentEnvironment.Id)},
                 {"spaceId", variables.Get("Octopus.Space.Id")},
                 {"tenantId", variables.Get(DeploymentVariables.Tenant.Id)},
```
johnsimons, line 72: this one can be deleted

YihaoOct: Do you mean tenantId? @johnsimons

johnsimons: yes

PR #1003, opened 2023-03-24
Fix Azure Web App Target Discovery by changing discovery code to use the Resource Graph API

scottmerchant:
We had a customer report that when they change tags on their Azure Resources (Azure Web App, App Service or Web App Slot) that the tags weren't reflected correctly in target discovery.

I have since discovered that this is an issue with Azure's API in that updates take more time to propagate to different regions. The discovery process worked perfectly on my machine (in the Australia South East region) but was producing incorrect results in cloud (Australia East region).

After raising a ticket with Microsoft, they suggested I use the Resource Graph API which shouldn't have the same issues. I've tested it and while there is a short time where it might get the old tags (about 1 minute) it does update after that so I think these changes solve the problem.

[sc-44562]


source/Calamari.AzureAppService/Behaviors/TargetDiscoveryBehaviour.cs
```diff
@@ -27,9 +23,9 @@ namespace Calamari.AzureAppService.Behaviors
 {
     public class TargetDiscoveryBehaviour : IDeployBehaviour
     {
-        private const string WebAppSlotsType = "sites/slots";
-        private const string WebAppType = "sites";
-        private const int PageSize = 500;
+        private const string WebAppSlotsType = "microsoft.web/sites/slots";
+        private const string WebAppType = "microsoft.web/sites";
```
MissedTheMark, line 27: It might be nice to have a short comment explaining what these values represent and where they came from.

source/Calamari.AzureAppService.Tests/TargetDiscoveryBehaviourIntegrationTestFixture.cs
```diff
@@ -141,6 +153,8 @@ public async Task Execute_WebAppWithNonMatchingTags_CreatesNoTargets()
 
             await CreateOrUpdateTestWebApp(tags);
 
+            await Task.Delay(waitForAzureToProcessPeriod);
```
zentron, line 156: Rather than a hard delay, can we use the Eventually semantics to keep checking for a period of time.

zentron (approved):
Only question is around an improvement to the test. Once thats been actioned\responded then feel free to merge

source/Calamari.AzureAppService.Tests/TargetDiscoveryBehaviourIntegrationTestFixture.cs
scottmerchant: Unfortunately we don't have the ShouldEventually utility in Calamari but yeah I can implement a crude version of it here.

source/Calamari.AzureAppService/Behaviors/TargetDiscoveryBehaviour.cs
```diff
@@ -182,11 +175,16 @@ public static ServiceMessage CreateWebAppTargetCreationServiceMessage(string? re
...
-        public static AsyncPageable<GenericResource> GetResources(this SubscriptionResource subscription,
-            string resourceType, int pageSize, CancellationToken cancellationToken)
+        public static async Task<AzureResource[]> GetResourcesByType(this ArmClient armClient, params string[] types)
         {
-            return subscription.GetGenericResourcesAsync($"resourceType eq 'Microsoft.web/{resourceType}'",
-                top: pageSize, cancellationToken: cancellationToken);
+            var tenant = armClient.GetTenants().First();
+            var typeCondition = types.Any()
+                ? $"| where {string.Join(" or ", types.Select(t => $"type == '{t}'"))} |"
+                : string.Empty;
```
MissedTheMark, line 183: I'd pull `string.Join(" or ", types.Select(t => $"type == '{t}'"))` into a new variable or two above this - it'll make it easier to see that the `| where XYZ |` clause is actually simple.
```suggestion
            var typesToRetrieveClause = string.Join(" or ", types.Select(t => $"type == '{t}'"));
            var typeCondition = types.Any()
                ? $"| where { typesToRetrieveClause } |"
                : string.Empty;
```

MissedTheMark (approved):
I'm reviewing this from a perspective of not knowing anything about the Graph Resource API, but looks good generally.  Couple of comments.

source/Calamari.AzureAppService/Azure/AzureResource.cs
```diff
@@ -0,0 +1,52 @@
...
+        public string ResourceGroup { get; set; }
+
+        [JsonProperty("tags")]
+        public Dictionary<string,string> Tags { get; set; }
+
+        [JsonIgnore]
+        public bool IsSlot => Type.EndsWith("/slots");
+
+        [JsonIgnore]
+        public string SlotName
```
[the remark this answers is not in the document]
scottmerchant: Done 👍

source/Calamari.AzureAppService/Behaviors/TargetDiscoveryBehaviour.cs
scottmerchant: Yep good call 👍

source/Calamari.AzureAppService/Behaviors/TargetDiscoveryBehaviour.cs
scottmerchant: Yep good idea - now to find some doco explaining these values 😅

source/Calamari.AzureAppService.Tests/TargetDiscoveryBehaviourIntegrationTestFixture.cs
scottmerchant: I have copied the ShouldEventually class and related utility classes to Calamari and massaged them to fit into the Calamari world so other tests can use them too - it might be worth using this for some of the flakey tests we have.

PR #1004, opened 2023-03-26
Rename Status to ResourceStatus

YihaoOct:
This PR changes the `Status` property on Kubernetes resource classes to `ResourceStatus`. This is to prevent the naming clashing with the `Status` column of a `kubectl get` command to a Pod and a PVC.


johnsimons (approved):
LGTM

PR #1005, opened 2023-03-27
Implement Pod status

YihaoOct:
[sc-44236]

This PR adds more comprehensive handling of Pod statuses. 

We attempt to create as close description as users could get from the `STATUS` column `kubectl get` command. The implemented scenarios are covered in [this spreadsheet](https://docs.google.com/spreadsheets/d/13zaYtw6V0OVT_8M57bq2W6vCLbk33lCqZQkT6TXgd7g/edit#gid=0) (internal). Please also see the unit tests to see the behaviour in difference scenarios.

The original `Phase` property which was a placeholder to be used in the walking skeleton is now removed.

### DIFF
--- a/source/Calamari/Kubernetes/ResourceStatus/Resources/Pod.cs
+++ b/source/Calamari/Kubernetes/ResourceStatus/Resources/Pod.cs
@@ -14,10 +14,10 @@ namespace Calamari.Kubernetes.ResourceStatus.Resources
             var phase = Field("$.status.phase");
             var initContainerStatuses = data
                 .SelectToken("$.status.initContainerStatuses")
-                ?.ToObject<ContainerStatus[]>() ?? Array.Empty<ContainerStatus>();
+                ?.ToObject<ContainerStatus[]>() ?? new ContainerStatus[] { };
             var containerStatuses = data
                 .SelectToken("$.status.containerStatuses")
-                ?.ToObject<ContainerStatus[]>() ?? Array.Empty<ContainerStatus>();
+                ?.ToObject<ContainerStatus[]>() ?? new ContainerStatus[] { };
 
             (Status, ResourceStatus) = GetStatus(phase, initContainerStatuses, containerStatuses);
         }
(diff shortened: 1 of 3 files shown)


source/Calamari/Kubernetes/ResourceStatus/Resources/Pod.cs
```diff
@@ -1,36 +1,121 @@
...
+                        ? GetStatus(containerStatuses) 
+                        : GetInitializingStatus(initContainerStatuses);
+                case "Failed":
+                    return (GetReason(containerStatuses.FirstOrDefault()), ResourceStatus.Failed);
+                case "Succeeded":
+                    return (GetReason(containerStatuses.FirstOrDefault()), ResourceStatus.Successful);
                 default:
-                    ResourceStatus = ResourceStatus.Failed;
-                    break;
+                    return ("Running", ResourceStatus.Successful);
```
johnsimons, line 50: Should this return `ResourceStatus.InProgress`

source/Calamari/Kubernetes/ResourceStatus/Resources/Pod.cs
```diff
@@ -1,36 +1,121 @@
...
-                case"Succeeded": 
-                case "Running":
-                    ResourceStatus = ResourceStatus.Successful;
-                    break;
                 case "Pending":
-                    ResourceStatus = ResourceStatus.InProgress;
-                    break;
+                    if (!initContainerStatuses.Any() && !containerStatuses.Any())
+                    {
+                        return ("Pending", ResourceStatus.InProgress);
```
johnsimons, line 40: Maybe this one should return the "queued" status?

source/Calamari/Kubernetes/ResourceStatus/Resources/Pod.cs
YihaoOct: This is perhaps something worth discussion. I think Successful makes the most sense for most scenarios. In a typical deployment, a container runs a long-running process and will never terminate unless error occurs (thinking a web server). In this scenario, the final state that the user desires is that all Pods are in a "Running" state. In other words, if all Pods are "Running", for the deployment that's a success. So `Successful` reflects the status better than `InProgress`.

source/Calamari/Kubernetes/ResourceStatus/Resources/Pod.cs
YihaoOct: We don't have the state for "queued" defined for a k8s resource and we don't know what's the definition of "queued" or how to differentiate it from "InProgress". I imagine you may think "queued" is appropriate because it is already one of the states we use in frontend as you showed me today. But in my opinion, using the frontend display status to drive the business logic would not be a proper choice. Also I think the frontend "queued" status may have other implications in other contexts, which may not be proper in the k8s status check context.

source/Calamari/Kubernetes/ResourceStatus/Resources/Pod.cs
johnsimons: makes sense

source/Calamari/Kubernetes/ResourceStatus/Resources/Pod.cs
johnsimons: no worries, sounds ok

johnsimons (approved):
nice test coverage

PR #1006, opened 2023-03-28
Fix pod status

YihaoOct:
[sc-44236]

There was a bug in the [previous implementation](https://github.com/OctopusDeploy/Calamari/pull/1005) of Pod status which I discovered while manually testing other stuff.

When the container of the Pod is in a `CrashLoopBackOff` state, the `phase` of the Pod is actually `Running` instead of `Pending`. I incorrectly thought it was `Pending`, so the implementation wasn't correct as well. 

I've modified the relevant test case and fixed the implementation. I also updated the [doc](https://docs.google.com/spreadsheets/d/13zaYtw6V0OVT_8M57bq2W6vCLbk33lCqZQkT6TXgd7g/edit#gid=0) (internal).

Screenshot 2023-03-28 at 3 10 30 PM

### DIFF
--- a/source/Calamari.Tests/KubernetesFixtures/ResourceStatus/Resources/PodTests.cs
+++ b/source/Calamari.Tests/KubernetesFixtures/ResourceStatus/Resources/PodTests.cs
@@ -113,7 +113,7 @@ namespace Calamari.Tests.KubernetesFixtures.ResourceStatus.Resources
         public void WhenContainerFailsExecutionAndRestarting_TheStatusShouldShowCrashLoopBackOff()
         {
             var podResponse = new PodResponseBuilder()
-                .WithPhase("Pending")
+                .WithPhase("Running")
                 .WithContainerStatuses(new State[]
                 {
                     new State { Waiting = new Waiting { Reason = "CrashLoopBackOff" } }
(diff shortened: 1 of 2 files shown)


source/Calamari.Tests/KubernetesFixtures/ResourceStatus/Resources/PodTests.cs
```diff
@@ -195,7 +195,7 @@ public void WhenContainerIsRunningWithoutErrors_TheStatusShouldShowRunning()
         }
 
         [Test]
-        public void WhenMoreThanOneContainersFailed_TheStatusShouldShowTheFirstError()
+        public void WhenMoreThanOneContainersHaveErrors_TheStatusShouldShowTheFirstError()
```
johnsimons, line 198:
```suggestion
        public void WhenMoreThanOneContainerHasErrors_TheStatusShouldShowTheFirstError()
```

source/Calamari/Kubernetes/ResourceStatus/Resources/Pod.cs
```diff
@@ -67,16 +67,26 @@ private static (string, ResourceStatus) GetInitializingStatus(ContainerStatus[]
...
+            return ("Running", ResourceStatus.Successful);
         }
         
         private static string GetReason(ContainerStatus status)
         {
+            // In real scenario this shouldn't happen, but we give it a default value just in case
             if (status == null)
             {
-                return "Pending";
+                return "-";
```
johnsimons, line 89: Why not an empty string?

johnsimons (approved):
LGTM, just a few comments/suggestions

PR #1007, opened 2023-03-28
Support other resources including secrets and configMaps defined in the "Deploy Container" step

YihaoOct:
[sc-44227]

The "Deploy Kubernetes Container" step allows users to add other resources including Secrets, ConfigMaps, Services and Ingresses. Like in the screenshot below:

Screenshot 2023-03-28 at 3 51 57 PM

The current implementation does not show these extra resources in the status check. This PR adds the support for these resources.

In a "Deploy Kubernetes Container" step with a Secret, a ConfigMap and a Service defined, we now can show the resource status for all of those (we haven't implemented Ingresses yet, but once it is implemented this should work as well).

Screenshot 2023-03-28 at 3 44 59 PM

*Note these logs are added ad hoc just to indicate the status checks are performed. They don't exist in the actual PR*

### DIFF
--- a/source/Calamari.Tests/KubernetesFixtures/ResourceStatus/KubernetesYamlTests.cs
+++ b/source/Calamari.Tests/KubernetesFixtures/ResourceStatus/KubernetesYamlTests.cs
@@ -65,5 +65,26 @@ namespace Calamari.Tests.KubernetesFixtures.ResourceStatus
 
             got.Should().BeEquivalentTo(expected);
         }
+
+        [Test]
+        public void ShouldHandleMultipleYamlFiles()
+        {
+            var manifest = TestFileLoader.Load("single-deployment.yaml");
+            var multipleFileInput = new string[] {manifest, manifest};
+            var got = KubernetesYaml.GetDefinedResources(multipleFileInput, string.Empty);
+            var expected = new ResourceIdentifier[]
+            {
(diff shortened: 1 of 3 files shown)


johnsimons (changes requested):
@YihaoOct the changes LGTM.
But you need test coverage for the new added types, Secrets and ConfigMaps?

YihaoOct:
@johnsimons This PR does not introduce new types. 

This PR simply reads two extra variables `Octopus.Action.KubernetesContainer.ComputedConfigMapName` and `.....ComputedSecretName` and create resource identifiers for them so we could retrieve information for them. We need to do this because the ConfigMap and Secret that are defined in "Deploy Kubernetes Container" step are stored in yaml files. They are created imperatively using `kubectl create configmap ...` rather than `kubectl apply`. And the configMaps created in this way will have names like `my-configmap-Deployments-123` which is saved in the variables. So we just need to read this variable to know what to look for. 

I have tests in this [draft PR](https://github.com/OctopusDeploy/Calamari/pull/1009/files#diff-577fc9fcaebf27f4b906875b48d68cccb8044107904125a36681248080ef25ff) in which we cover all the resources we support including Secrets and ConfigMaps. Does this look like something you are thinking of?

johnsimons (approved):
LGTM

PR #1009, opened 2023-03-29
Add all supported resources and their fields

YihaoOct:
[sc-44238]

This PR adds the support to all kubernetes resource types that we want to support:

- Pod
- ReplicaSet
- Deployment
- StatefulSet
- DaemonSet
- Job
- CronJob
- Service
- EndpointSlice
- Ingress
- ConfigMap
- Secret
- PersistentVolumeClaim

The fields we support are documented in [this spreadsheet](https://docs.google.com/spreadsheets/d/1jO3JUnOYI4fKXRsdL7VBPur56QjM5jYQkaWnV5NNF6I/edit#gid=0) (internal).

Please also see the unit tests for each resource type which contains examples of what properties are sent.

### DIFF
--- a/source/Calamari.Tests/KubernetesFixtures/ResourceStatus/Resources/PersistentVolumeClaimTests.cs
+++ b/source/Calamari.Tests/KubernetesFixtures/ResourceStatus/Resources/PersistentVolumeClaimTests.cs
@@ -1,0 +1,54 @@
+using Calamari.Kubernetes.ResourceStatus.Resources;
+using FluentAssertions;
+using NUnit.Framework;
+
+namespace Calamari.Tests.KubernetesFixtures.ResourceStatus.Resources
+{
+    [TestFixture]
+    public class PersistentVolumeClaimTests
+    {
+        [Test]
+        public void ShouldCollectCorrectProperties()
+        {
+            const string input = @"{
+    ""kind"": ""PersistentVolumeClaim"",
+    ""metadata"": {
+        ""name"": ""my-pvc"",
+        ""namespace"": ""default"",
+        ""uid"": ""01695a39-5865-4eea-b4bf-1a4783cbce62""
+    },
+    ""spec"": {
+        ""volumeName"": ""pvc-08cdb1f6-42e4-4938-b4c7-0576030a8da6"",
+        ""storageClassName"": ""standard""
+    },
+    ""status"": {
+        ""phase"": ""Bound"",
+        ""accessModes"": [
+            ""ReadWriteOnce""
+        ],
+        ""capacity"": {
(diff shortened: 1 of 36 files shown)


source/Calamari.Tests/KubernetesFixtures/ResourceStatus/Resources/PersistentVolumeClaimTests.cs
```diff
@@ -0,0 +1,55 @@
...
+                Kind = "PersistentVolumeClaim",
+                Name = "my-pvc",
+                Namespace = "default",
+                Uid = "01695a39-5865-4eea-b4bf-1a4783cbce62",
+                Status = "Bound",
+                Volume = "pvc-08cdb1f6-42e4-4938-b4c7-0576030a8da6",
+                Capacity = "1Mi",
+                AccessModes = new string []
+                {
+                    "ReadWriteOnce"
```
YihaoOct, line 47: In the actual output of `kubectl` this is `RWO` instead. But looks like the long form is more readable, and this does not require conversion, so I just kept the long form.

source/Calamari.Tests/KubernetesFixtures/ResourceStatus/Resources/PodTests.cs
```diff
@@ -28,10 +62,10 @@ public void WhenInitContainerIsBeingCreatedWithoutErrors_TheStatusShouldShowNumb
         {
             var podResponse = new PodResponseBuilder()
                 .WithPhase("Pending")
-                .WithInitContainerStatuses(new State[]
+                .WithInitContainerStates(new ContainerState[]
```
YihaoOct, line 65: Renamed some classes so they match the name of the entities in the [API reference](https://kubernetes.io/docs/reference/kubernetes-api/).

source/Calamari/Kubernetes/ResourceStatus/Resources/DaemonSet.cs
```diff
@@ -0,0 +1,57 @@
...
+                   || last.Available != Available
+                   || last.NodeSelector != NodeSelector;
+        }
+
+        private static string FormatNodeSelectors(Dictionary<string, string> nodeSelectors)
+        {
+            var selectors = nodeSelectors
+                .ToList()
+                .OrderBy(_ => _.Key)
+                .ThenBy(_ => _.Value)
```
YihaoOct, line 51: Technically the order of iterating a Dictionary is not guaranteed to be stable (though in reality it often is). We don't want the iteration order to affect the outcome because that will cause unnecessary updates sent to server.

johnsimons (approved):

PR #1011, opened 2023-04-14
Fix environment variable

YihaoOct:
This is to fix a bug in Kubernetes resources status check. We need to set the environment variable for kubeconfig to be the correct location. Without this the authentication will not be set up properly unless deploying to a local cluster.

### DIFF
--- a/source/Calamari/Kubernetes/ResourceStatus/ResourceStatusReportWrapper.cs
+++ b/source/Calamari/Kubernetes/ResourceStatus/ResourceStatusReportWrapper.cs
@@ -96,6 +96,15 @@ namespace Calamari.Kubernetes.ResourceStatus
                 return result;
             }
 
+            
+            var kubeConfig = Path.Combine(workingDirectory, "kubectl-octo.yml");
+
+            if (environmentVars == null)
+            {
+                environmentVars = new Dictionary<string, string>();
+            }
+            environmentVars.Add("KUBECONFIG", kubeConfig);
+            
             foreach (var proxyVariable in ProxyEnvironmentVariablesGenerator.GenerateProxyEnvironmentVariables())
             {
                 environmentVars[proxyVariable.Key] = proxyVariable.Value;


source/Calamari/Kubernetes/ResourceStatus/ResourceStatusReportWrapper.cs
```diff
@@ -96,6 +96,15 @@ public bool IsEnabled(ScriptSyntax syntax)
                 return result;
             }
 
+            
+            var kubeConfig = Path.Combine(workingDirectory, "kubectl-octo.yml");
+
+            if (environmentVars == null)
```
zentron, line 102: Is tjhis able to be null? Nothing else seems to do that check here. Ideally we should have null checking on for this project but it should be probably done earlier in the flow rather than mid way through.

zentron (approved):
Approved to get the build running, but comment to take a look at around the null check.

source/Calamari/Kubernetes/ResourceStatus/ResourceStatusReportWrapper.cs
YihaoOct: Yes, we found it to be null when we ran it. That's why we added this checking.
In the other wrapper the same has been done as well: https://github.com/OctopusDeploy/Calamari/blob/b7c5f152fb2a768761ef45372dcd79f554823a66/source/Calamari/Kubernetes/KubernetesContextScriptWrapper.cs#L60-L63

I think the root cause is this default value here: https://github.com/OctopusDeploy/Calamari/blob/b7c5f152fb2a768761ef45372dcd79f554823a66/source/Calamari.Common/Features/Scripting/ScriptEngine.cs#L45

A proper fix should remove this default value, and add an empty Dictionary initialisation everywhere this method is invoked.

zentron: > A proper fix should remove this default value, and add an empty Dictionary initialisation everywhere this method is invoked.

Sounds good :D

PR #1012, opened 2023-04-16
Do not return early if the manifests are empty

YihaoOct:
Fixes [sc-46510]

We use to return early when there are nothing in the manifests file. This does not work if the only resources that are deployed are not coming from manifest files. eg secrets and configMaps have their computed names saved in variables rather than in manifest files.

### Before
Screenshot 2023-04-17 at 11 51 26 AM

### After
Screenshot 2023-04-17 at 11 52 00 AM


johnsimons (approved):
LGTM

PR #1013, opened 2023-04-17
Prefer AzModules over AzRmContext on Windows Powershell

IsaacCalligeros95:
If available, prefer using AzModules over AzRM for Windows PowerShell, see https://github.com/OctopusDeploy/Issues/issues/7732 for details. 

**For context:**
In `Initialize-AzContext` (PowerShell Core path) when holding both AzureRM and Azure PowerShell modules, we check to see if AzureRM is available with `Get-Command "Disable-AzureRMContextAutosave" -errorAction SilentlyContinue`, this throws the warning we're seeing. This is to use [Enable-AzureRmAlias](https://learn.microsoft.com/en-us/powershell/azure/migrate-from-azurerm-to-az?view=azps-9.6.0#option-2-use-compatibility-mode-with-enable-azurermalias) for compatibility with older commands. These modules contain overlapping commands, so to install, you have to -AllowClobber `Install-Module -Force -Name Az -AllowClobber`. We do this in WorkerTools. 

Using AzureRM exclusively, does not print to the warning, but does log a warning
image

When both are available, we turn off context autosave, without this, all authentication occurs in memory and isolates each session from the context changes in other sessions. This is done using `Disable-AzRMContextAutosave -Scope Process`, the warning we're seeing.

There's two paths to this issue:
* Using Windows PowerShell, we use AzureRM modules
* Using PowerShell Core we check for the `Disable-AzureRMContextAutosave` command, this writes to stdwarning

This change prefers Azure PowerShell modules over AzureRM if available and suppresses the warning when checking for `Disable-AzureRMContextAutosave`

[sc-41867]
Fixes https://github.com/OctopusDeploy/Issues/issues/7732

**Before:** 
Windows Worker Tools, using PowerShell Core:
image

Using Windows PowerShell shows no warning.

**After:** 
Windows Worker Tools, using PowerShell Core:
image

Windows Worker Tools, using Windows PowerShell:
image

Linux Worker Tools, using PowerShell Core: 
image

TODO: Spinning up in a cloud environment to test on Linux workers before merging (unlikely to cause any issues).

### DIFF
--- a/source/Calamari.AzureScripting/Scripts/AzureContext.ps1
+++ b/source/Calamari.AzureScripting/Scripts/AzureContext.ps1
@@ -116,7 +116,9 @@ function Initialize-AzContext {
 
     # Turn off context autosave, as this will make all authentication occur in memory, and isolate each session from the context changes in other sessions
     Write-Host "##octopus[stdout-verbose]"
+    Set-Item -Path Env:\SuppressAzureRmModulesRetiringWarning -Value $true
     Disable-AzContextAutosave -Scope Process
+    Set-Item -Path Env:\SuppressAzureRmModulesRetiringWarning -Value $false
     Write-Host "##octopus[stdout-default]"
 
     $AzureEnvironment = Get-AzEnvironment -Name $OctopusAzureEnvironment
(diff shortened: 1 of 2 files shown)


hnrkndrssn (approved):
👍

PR #1014, opened 2023-04-17
Yihao/k8s status integration tests

YihaoOct:
[sc-44307]
This PR adds some integration tests to cover the logic in the Kubernetes resource status check wrapper.

To test everything in a live cluster provides some extra assurance, however this is hard to set up and inefficient to execute (the current live k8s cluster tests takes a long time and are very flaky). So I made the tradeoff not to run the tests in a live cluster, but just attempting to cover as much logic as possible for the script wrapper with mocks. Despite not ideal, the added tests should give us more confidence and prevent regression.

### DIFF
--- a/source/Calamari.Common/CalamariFlavourProgram.cs
+++ b/source/Calamari.Common/CalamariFlavourProgram.cs
@@ -123,7 +123,6 @@ namespace Calamari.Common
 
             builder.RegisterAssemblyTypes(assemblies)
                 .AssignableTo<IScriptWrapper>()
-                .Where(wrapper => wrapper.IsPublic)
                 .Except<TerminalScriptWrapper>()
                 .As<IScriptWrapper>()
                 .SingleInstance();
--- a/source/Calamari.Common/CalamariFlavourProgram.cs
+++ b/source/Calamari.Common/CalamariFlavourProgram.cs
@@ -123,6 +123,7 @@ namespace Calamari.Common
 
             builder.RegisterAssemblyTypes(assemblies)
                 .AssignableTo<IScriptWrapper>()
+                .Where(wrapper => wrapper.IsPublic)
                 .Except<TerminalScriptWrapper>()
                 .As<IScriptWrapper>()
                 .SingleInstance();
--- a/source/Calamari.Tests/KubernetesFixtures/ResourceStatus/ResourceStatusReportWrapperTests.cs
+++ b/source/Calamari.Tests/KubernetesFixtures/ResourceStatus/ResourceStatusReportWrapperTests.cs
@@ -90,7 +90,7 @@ namespace Calamari.Tests.KubernetesFixtures.ResourceStatus
             wrapper.IsEnabled(Syntax).Should().BeFalse();
         }
 
-        [Test]
+        //[Test]
         public void FindsCorrectManifestFiles()
         {
             var variables = new CalamariVariables();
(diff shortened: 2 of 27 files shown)


source/Calamari.Tests/KubernetesFixtures/ResourceStatus/ResourceStatusReportWrapperTests.cs
```diff
@@ -0,0 +1,240 @@
...
+            variables.Set(SpecialVariables.ClusterUrl, "https://localhost");
+            variables.Set(SpecialVariables.ResourceStatusCheck, "True");
+            variables.Set(SpecialVariables.DeploymentWait, "wait");
+            
+            var wrapper = new ResourceStatusReportWrapper(variables, log, fileSystem, statusChecker);
+
+            wrapper.IsEnabled(Syntax).Should().BeFalse();
+        }
+
+        //[Test]
```
johnsimons, line 93: just delete the code

source/Calamari.Tests/KubernetesFixtures/ResourceStatus/ResourceStatusReportWrapperTests.cs
```diff
@@ -0,0 +1,240 @@
...
+
+        [Test]
+        public void NotEnabled_WhenWaitForDeploymentIsSelected()
+        {
+            var variables = new CalamariVariables();
+            var log = new SilentLog();
+            var fileSystem = new TestCalamariPhysicalFileSystem();
+            var statusChecker = new MockResourceStatusChecker();
+            
+            variables.Set(Deployment.SpecialVariables.EnabledFeatureToggles, "KubernetesDeploymentStatusFeatureToggle");
```
johnsimons, line 83: i noticed these 3 lines are repeated pretty much on every test, is this something we can do differently, eg `CreateCalamariVariables`

johnsimons (changes requested):
Really nice to see test coverage on this code path.
A lot of similar code is shared between `FindsConfigMapsDeployedInADeployContainerStep` and `FindsSecretsDeployedInADeployContainerStep`. I think we could create a method that takes the assertion as a function, thoughts?

source/Calamari.Common/CalamariFlavourProgram.cs
```diff
@@ -123,6 +123,7 @@ protected virtual void ConfigureContainer(ContainerBuilder builder, CommonOption
 
             builder.RegisterAssemblyTypes(assemblies)
                 .AssignableTo<IScriptWrapper>()
+                .Where(wrapper => wrapper.IsPublic)
```
johnsimons, line 126: Instead of modifying this for all the tests let's enable/disable the wrapper on the test itself, see comment below

source/Calamari.Tests/KubernetesFixtures/ResourceStatus/ResourceStatusReportWrapperTests.cs
```diff
@@ -0,0 +1,235 @@
...
+            CheckedResources = resourceIdentifiers.ToList();
+            return true;
+        }
+    }
+
+    internal class StubScriptWrapper : IScriptWrapper
+    {
+        public int Priority { get; } = 1;
+        public IScriptWrapper NextWrapper { get; set; }
+        public bool IsEnabled(ScriptSyntax syntax) => true;
```
johnsimons, line 224: Set this to be `false` by default, and then override it via a `SetEnable` method ?

johnsimons (approved):
LGTM, but it seems the builds are failing 😞

PR #1015, opened 2023-04-17
Wait until all Pods are removed

YihaoOct:
[sc-46519]

This PR makes sure that all exceeding Pods associated with a Deployments are completely removed before we mark the deployment as successful. Previously we do not wait for these Pods to go away, as a result, at the time the deployment has been successful, there are still Pods displayed on the screen.


johnsimons (approved):
LGTM

PR #1018, opened 2023-04-21
Add logging to test output when running discovery tests

scottmerchant:
I realised when looking at a flakey test that the output doesn't include the discovery code logging because of the way logging is setup for these tests.

The only time the InMemoryLogs are logged to the test output are when a script is executed but discovery doesn't use a script so it's never logged.

I've added in some code which will log the discovery code logs to test output after discovery has been executed.

### DIFF
--- a/source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureBase.cs
+++ b/source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureBase.cs
@@ -63,10 +63,7 @@ namespace Calamari.Tests.KubernetesFixtures
         {
             var calamariResult = ExecuteScriptInternal(new CommandLineRunner(Log, variables), wrapper, scriptName);
 
-            foreach (var message in Log.Messages)
-            {
-                Console.WriteLine($"[{message.Level}] {message.FormattedMessage}");
-            }
+            WriteLogMessagesToTestOutput();
 
             return calamariResult;
         }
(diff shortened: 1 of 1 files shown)


source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureBase.cs
```diff
@@ -176,10 +176,17 @@ protected void DoDiscovery(AwsAuthenticationDetails authenticationDetails)
...
-                                       .Argument("extensions", string.Join(',', extensions)));
+                var result = InvokeInProcess(Calamari()
+                       .Action(KubernetesDiscoveryCommand.Name)
+                       .Argument("variables", variablesFile.FilePath)
+                       .Argument("extensions", string.Join(',', extensions)));
+
+                foreach (var message in Log.Messages)
+                {
+                    Console.WriteLine($"[{message.Level}] {message.FormattedMessage}");
+                }
```
MissedTheMark, line 187: Given that this uses the base class's Log property (and that this is a dupe of code above), is it worth creating a method to do this in that base class (ie in CalamariFixture)?  I assume this functionality would come in handy in other tests.

MissedTheMark (approved):
Looks reasonable to me.  Just one comment, I'll let you decide whether it's worth doing.

source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureBase.cs
scottmerchant: I see what you're saying but in `CalamariFixture` the logger is an `InMemoryLogger` which actually does log to StdOut and StdErr as well. In this base class, the Logger is changed to a `DoNotDoubleLog` logger which overrides the writing to StdOut and StdErr so that the logging can be done manually. Basically `CalamariFixture` by default doesn't have this issue, only `KubernetesContextScriptWrapperLiveFixtureBase`. I can, however move the dupe code to a method in this class which I will do for neatness 👍

MissedTheMark: Oh sorry, I missed that this used `KubernetesContextScriptWrapperLiveFixtureBase` - my bad 🙇

scottmerchant: No problems! the naming of the loggers and general structure of these tests doesn't make it super easy to understand 😅

PR #1020, opened 2023-04-28
Fixing some logging bugs in the new kubernetes raw yaml command

scottmerchant:
I noticed that when multiple yaml files are applied that the resources are returned in a custom List object in JSON which we aren't handling properly so it looks like this:
image

This fix means we parse the object correctly and pull out the objects which looks like this:
image

I have also updated the Batch numbering to be 1 based instead of 0 (zero) based:
image


MissedTheMark (approved):
Nice!

PR #1022, opened 2023-05-09
Fix countdown timer with 0 duration

YihaoOct:
[sc-48027]

There is a bug in the timer so if it is configured with 0 timeout, the `HasCompleted` method will return true due to the way we do the comparison.

This caused steps configured with 0 stabilisation timeouts to fail immediately while some resources are still in-progress.

### DIFF
--- a/source/Calamari/Kubernetes/ResourceStatus/CountdownTimer.cs
+++ b/source/Calamari/Kubernetes/ResourceStatus/CountdownTimer.cs
@@ -31,7 +31,7 @@ namespace Calamari.Kubernetes.ResourceStatus
         public void Start() => stopwatch.Start();
         public void Reset() => stopwatch.Reset();
         public bool HasStarted() => stopwatch.IsRunning;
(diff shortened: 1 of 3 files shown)


source/Calamari/Kubernetes/ResourceStatus/CountdownTimer.cs
```diff
@@ -31,7 +31,7 @@ public CountdownTimer(TimeSpan duration)
         public void Start() => stopwatch.Start();
         public void Reset() => stopwatch.Reset();
         public bool HasStarted() => stopwatch.IsRunning;
-        public bool HasCompleted() => stopwatch.Elapsed >= duration;
+        public bool HasCompleted() => HasStarted() && stopwatch.Elapsed >= duration;
```
YihaoOct, line 34: This could be fixed with just changing `>=` to `>`, but I think this way communicate the intention better. A timer must be started before it completes.

zentron (approved):

PR #1023, opened 2023-05-10
Add countOfChecks to the service message

YihaoOct:
This PR adds a `countOfChecks` field to the service message for kubernetes object status check. This field reflects which round of check this service message belongs to. It will be used by server to prevent updating a newer kubernetes resource record with a stale one.

### DIFF
--- a/source/Calamari.Tests/KubernetesFixtures/ResourceStatus/ResourceUpdateReporterTests.cs
+++ b/source/Calamari.Tests/KubernetesFixtures/ResourceStatus/ResourceUpdateReporterTests.cs
@@ -46,7 +46,7 @@ namespace Calamari.Tests.KubernetesFixtures.ResourceStatus
                     bool.FalseString
                 });
             
-            serviceMessages.Select(message => message.Properties["countOfChecks"])
+            serviceMessages.Select(message => message.Properties["checkCount"])
                 .Should().BeEquivalentTo(new string[]
                 {
                     "1",
@@ -80,7 +80,7 @@ namespace Calamari.Tests.KubernetesFixtures.ResourceStatus
                 {
                     new KeyValuePair<string, string>("name", "nginx"),
                     new KeyValuePair<string, string>("removed", bool.FalseString),
-                    new KeyValuePair<string, string>("countOfChecks", "1")
+                    new KeyValuePair<string, string>("checkCount", "1")
                 });
         }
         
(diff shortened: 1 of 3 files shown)


source/Calamari.Tests/KubernetesFixtures/ResourceStatus/ResourceUpdateReporterTests.cs
```diff
@@ -102,6 +110,7 @@ public void ReportsRemovedResourcesCorrectly()
                 {
                     new KeyValuePair<string, string>("name", "redis"),
                     new KeyValuePair<string, string>("removed", bool.TrueString),
+                    new KeyValuePair<string, string>("countOfChecks", "1"),
```
scottmerchant, line 113: Is there a way we can get a test to produce a second update to make sure we're getting a count of 2?

scottmerchant: (or more than 2 if you want)

scottmerchant (commented):
Just a couple of comments from me.

As an aside, I think this will also help me write my tests as I'm receiving status update messages and with this I can check the last status update to make sure that it's correct 👍

source/Calamari/Kubernetes/ResourceStatus/ResourceUpdateReporter.cs
```diff
@@ -79,7 +79,8 @@ private void SendServiceMessage(Resource resource, bool removed)
                 {"namespace", resource.Namespace},
                 {"status", resource.ResourceStatus.ToString()},
                 {"data", JsonConvert.SerializeObject(resource)},
-                {"removed", removed.ToString()}
+                {"removed", removed.ToString()},
+                {"countOfChecks", countOfChecks.ToString()}
```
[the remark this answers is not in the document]
YihaoOct: `checkCount` sounds good to me! Not a native speaker so I don't have a good feeling of how the name sounds

source/Calamari.Tests/KubernetesFixtures/ResourceStatus/ResourceUpdateReporterTests.cs
YihaoOct: Because the method that gets tested simply put whatever the number is passed in into the service message, we can't test the incrementation of the number here. The current test structure captures what is happening during one check but not the whole cycle.
It would be nice if we add some checks to the incrementation of the numbers etc. Because we are considering changing the logic of stabilisation timeout and execution timeout, I think it is better to add the test in after the logic is re-implemented, which will make writing such tests much easier.

scottmerchant (approved):
LGTM

zentron:
LGTM. I'm still have some reservations about an in-memory counter, but this is very much a two-way door that we can reconsider if we need to. 👍

zentron (approved):

PR #1024, opened 2023-05-11
Port K8s scripts to Native Calamari Commands part 1

scme0:


scme0:
Closed because this is now redundant.

PR #1025, opened 2023-05-12
Move RedactedValuesLogger to Calamari.Common and make it the default and registered logger

scottmerchant:
A small PR as part of [[sc-47324](https://app.shortcut.com/octopusdeploy/story/47324/move-the-k8s-execution-logic-into-calamari-from-server)]

This `RedactedValuesLogger` is a small wrapper for loggers to allow secret values to be shielded from logging output.

This is currently only used in the Kubernetes script steps but I thought it would be better to make it the default so other places can leverage it but also to ensure that when we do use it, that specific logger instance is also everywhere else. (If you create a `RedactedValuesLogger` and add some values to it but then when another class logs those values, it logs to a standard `ConsoleLogger`, the values won't be redacted!


source/Calamari.AzureAppService/Behaviors/TargetDiscoveryBehaviour.cs
```diff
@@ -155,6 +155,8 @@ public async Task Execute(RunningDeployment runningDeployment)
             catch (JsonException ex)
             {
                 Log.Warn($"Target discovery context from variable {contextVariableName} is in wrong format: {ex.Message}");
+                Log.Verbose("Raw JSON:");
+                Log.Verbose(json);
```
MissedTheMark, line 159: Are there any security implications of logging the entire json string?

source/Calamari.Common/Plumbing/Logging/RedactedValuesLogger.cs
```diff
@@ -17,8 +15,9 @@ public RedactedValuesLogger(ILog innerLogger)
             log = innerLogger;
         }
 
-        public void AddValueToRedact(string value, string replacement)
+        public void AddValueToRedact(string? value, string replacement)
```
MissedTheMark, line 18: I don't like having a nullable value here - if there's a location which may pass a null value, can we do the check there instead?

source/Calamari.Common/Plumbing/Extensions/LogExtensions.cs
```diff
@@ -23,5 +23,16 @@ public static void LogMetric(this ILog log, string metricName, object metricValu
 
             log.WriteServiceMessage(new ServiceMessage(ServiceMessageNames.CalamariDeploymentMetric.Name, serviceMessageParameters));
         }
+
+        public static void AddValueToRedact(this ILog log, string? value, string replacement)
```
MissedTheMark, line 27: See my comment on `RedactedValuesLogger.cs` regarding the nullable value.

source/Calamari.Common/Plumbing/Extensions/LogExtensions.cs
```diff
@@ -23,5 +23,16 @@ public static void LogMetric(this ILog log, string metricName, object metricValu
...
+        {
+            if (!(log is RedactedValuesLogger redactedValuesLogger))
```
MissedTheMark, line 29: How will the consumer of this method know if their logger is able to redact?

source/Calamari.Common/CalamariFlavourProgram.cs
```diff
@@ -25,21 +25,21 @@
...
 {
-    public abstract class CalamariFlavourProgram
+    public abstract class  CalamariFlavourProgram
     {
         protected readonly ILog Log;
 
         protected CalamariFlavourProgram(ILog log)
         {
-            Log = log;
+            Log = new RedactedValuesLogger(log);
```
MissedTheMark, line 34: Will using this everywhere have any performance/memory implications? What about if it's used to redact large log strings? If so, should there be XML comments warning of that?  
(not saying there definitely are, but just a reminder to consider these things)

MissedTheMark (commented):
Couple of things I'd like you to consider before we merge this.

source/Calamari.AzureAppService/Behaviors/TargetDiscoveryBehaviour.cs
scottmerchant: Shoot this is not meant to be there! Good catch.

source/Calamari.Common/CalamariFlavourProgram.cs
scottmerchant: I think there will be a minor hit but because the vast majority of times we don't have any values to redact, it will just be returning the value without attempting any sort of replacements anyway. If there are values to redact then we definitely want to pay the price for replacing those strings.

The main problem I'm trying to solve with this change is that if we move things around (that's what I'm planning to do in subsequent PRs) we can't rely on the fact that the logger is a redacted logger, and if it's not, we can't rely on things getting redacted correctly, unless we say that the default logger is a redacted logger.

I ran into this issue during my spike work and I wanted to make sure we covered it before any secrets were leaked into logs.

source/Calamari.Common/Plumbing/Extensions/LogExtensions.cs
scottmerchant: They wont... I was thinking this exception is more for developers as if they're hitting this issue, something hasn't been setup correctly. I don't expect anyone to hit this exception if they have just injected `ILog` like normal.

MissedTheMark: Sorry, I was unclear - by consumer, I meant the developers.

source/Calamari.AzureAppService/Behaviors/TargetDiscoveryBehaviour.cs
scottmerchant: Removed.

source/Calamari.Common/Plumbing/Logging/RedactedValuesLogger.cs
scottmerchant: Fixed.

source/Calamari.Common/Plumbing/Extensions/LogExtensions.cs
scottmerchant: Fixed

source/Calamari.Common/Plumbing/Extensions/LogExtensions.cs
scottmerchant: Yeah I understand, it's not ideal... the problem is that even if I register `RedactedValuesLogger` as itself so that when redacted values need to be added, the class can just inject the redacted logger directly (rather than `ILog`, that doesn't guarantee that the same logger will be used when we log stuff. This would mean that data we want redacted could be logged as-is without us knowing (unless we manually check). I went this way because I thought that an exception when trying to add redacted values is the better result.

Unsure if there is a better solution... 

CC: @zentron

scottmerchant: Solved this issue after discussion with @MissedTheMark by moving the RedactedValuesLogger functionality into the AbstractLog and I even wrote a few unit tests around the Redacting of values.

I updated the ILog interface to have the ability to add redacted values as well.

This feels so much cleaner as now we're guaranteed to redact values if they are give to the logger no matter where the logger is injected/passed etc.

MissedTheMark (approved):
LGTM 👍

PR #1026, opened 2023-05-12
Adding the initial test for the Raw Yaml Deployment Script and associated infrastructure changes

scottmerchant:
This is a part of [[sc-47324](https://app.shortcut.com/octopusdeploy/story/47324/move-the-k8s-execution-logic-into-calamari-from-server)]

This is the first test which uses the scripts taken from server (DeploymentScript.sh/ps1) and runs a test against them.

It also updates the test infrastructure to be able to run the test from the RunScriptCommand which will make it much easier to run against the RawYamlCommand and compare results, as opposed to the way the other tests are running scripts.


source/Calamari.Tests/KubernetesFixtures/Scripts/KubernetesDeployment.ps1
scottmerchant, line 1: These ps1 and sh files are copied directly from server - no changes made.

source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureBase.cs
```diff
@@ -42,42 +52,98 @@ public void Setup()
...
+        CalamariResult ExecuteWithRunScriptCommand(ICalamariFileSystem fileSystem, IEnumerable<IScriptWrapper> scriptWrappers)
         {
-            var wrappers = new List<IScriptWrapper>(new[] { wrapper });
-            if (variables.Get(Deployment.SpecialVariables.Account.AccountType) == "AmazonWebServicesAccount")
+            var oldEnvVars = new Dictionary<string, string>();
+            foreach (var envVar in GetEnvironments())
             {
-                wrappers.Add(new AwsScriptWrapper(Log, variables));
+                oldEnvVars[envVar.Key] = Environment.GetEnvironmentVariable(envVar.Key);
+                Environment.SetEnvironmentVariable(envVar.Key, envVar.Value);
```
MissedTheMark, line 98: Can we do this without needing to modify the environment variables?  My concern that this could interfere with other tests, either when running concurrently, or if this fails before the variables are reset to their old values.

MissedTheMark (commented):
LGTM generally, just one concern with modifying environment variables in tests.

source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureBase.cs
scottmerchant: This is fixed now! I had to adjust a few bits and pieces to be able to pass in the environment variables but now it works so I don't have to set them on the process 👍

scottmerchant: I'm just waiting for the build to come good...

source/Calamari/Commands/RunScriptCommand.cs
```diff
@@ -87,7 +89,7 @@ public override int Execute(string[] commandLineArguments)
                 new ExecuteScriptConvention(scriptEngine, commandLineRunner)
             };
 
-            var deployment = new RunningDeployment(packageFile, variables);
+            var deployment = runningDeploymentFactory(packageFile);
```
MissedTheMark, line 92: Using a factory here isn't really justified, IMO, and overcomplicates this.  The constructor only has two parameters (three in other places), and the `variables` field is needed elsewhere in this class anyway.
It also means you have to create the factory in the tests manually.

source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureBase.cs
```diff
@@ -104,6 +165,52 @@ protected void TestScript(IScriptWrapper wrapper, string scriptName)
...
+        }
+
+        private void SetupTempDirectoryAndVerifyResult(Action<TemporaryDirectory> addFilesAction, Func<CalamariResult> func)
+        {
+            using (var dir = TemporaryDirectory.Create())
+            {
+                var folderPath = Path.Combine(dir.DirectoryPath, "TestFolder");
+                Directory.CreateDirectory(folderPath);
+                var currentDirectory = Environment.CurrentDirectory;
+                Environment.CurrentDirectory = folderPath;
```
MissedTheMark, line 190: Is there any other way we can write to the temporary path? Or do the scripts use Environment.CurrentDirectory?

MissedTheMark (approved):
Looks good to me, I'd just like you to remove the factory as per my comment below 👍

source/Calamari/Commands/RunScriptCommand.cs
scottmerchant: That's fair - I have another PR to make this use dependency injection properly which I'll do as a separate PR.

source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureBase.cs
scottmerchant: I have figured out why it wasn't working. There are placed in the code where we use `Path.GetFilePath()` on a relative path so it uses the `Environment.CurrentDirectory` which is obviously not right if we're working in a different directory. I've updated the areas that affect this test.

source/Calamari/Commands/RunScriptCommand.cs
scottmerchant: Oh I just realised why I did it - I've done it this way so I can slip in Environment Variables in the test - if it's `new`ed up in RunScriptCommand I have no way of passing environment variables to it. This way I can setup the factory in my tests to pass in the environment variables when it creates the instance.

PR #1027, opened 2023-05-12
Add missing AWS Auth Convention for Raw Yaml Command

scottmerchant:
This is a part of [[sc-47324](https://app.shortcut.com/octopusdeploy/story/47324/move-the-k8s-execution-logic-into-calamari-from-server)]

During test writing I noticed straight away that this command did not support AWS auth because it was missing a key part of that. This adds the AwsAuthConvention class which wraps `AwsEnvironmentGeneration.Create()`.

Dependent on https://github.com/OctopusDeploy/Calamari/pull/1035 merging first


zentron (dismissed):
Perhaps silly question, but why is this seperate to the auth setup done inside KubernetesAuthContextConvention?

scottmerchant:
No, good question. It's currently used in a ScriptWrapper which executes on every script (not just K8S or AWS related scripts...). Script wrappers are not nice...

The `KubernetesAuthContextConvention` just wraps `SetupKubectlAuthentication` which is used in a separate script wrapper so I didn't want to add extra functionality to the Convention to make it more confusing...

source/Calamari/Kubernetes/AwsAuthConventionFactoryFactory.cs
```diff
@@ -0,0 +1,22 @@
+using System;
+using Calamari.Common.Plumbing.Logging;
+using Calamari.Common.Plumbing.Variables;
+using Calamari.Deployment.Conventions;
+
+namespace Calamari.Kubernetes
+{
+    public class AwsAuthConventionFactoryFactory
```
scottmerchant, line 8: You read that right AwsAuthConventionFactory... Factory!

scottmerchant: If you inject a Lazy, it does check if the type in the Lazy is registered, but it doesn't check that type's dependencies so if you create another Lazy with in that, autofac doesn't complain if it's not actually a registered interface.

source/Calamari.Aws/Integration/AwsAuthConvention.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.Threading.Tasks;
+using Calamari.CloudAccounts;
+using Calamari.Common.Commands;
+using Calamari.Common.Plumbing.Logging;
+using Calamari.Common.Plumbing.Variables;
+using Calamari.Deployment.Conventions;
+
+namespace Calamari.Aws.Integration
+{
+    public class AwsAuthConvention : IInstallConvention
```
scottmerchant, line 11: I went back to use the `AwsAuthConvention` because I was getting issues in tests etc when I tried to add it to `SetupKubectlAuthentication`. We can combine it later I guess but for now it's easier to just keep it as a separate convention.

source/Calamari/Kubernetes/AwsAuthConventionFactoryFactory.cs
MissedTheMark: This feels so dirty 😁 But if it works, it works!
A comment explaining what this does, and why you've done this would be great - I.e. something like explain that we're stopping Autofac from checking the dependency, and that we're doing it because of dotnet 4 causing issues for xyz etc.

MissedTheMark: I wonder if there's another name we could give this which might be less cryptic than FactoryFactory, like AwsAuthConventionFactoryWrapper or something?

source/Calamari/Kubernetes/Commands/KubernetesApplyRawYamlCommand.cs
```diff
@@ -93,10 +96,19 @@ public override int Execute(string[] commandLineArguments)
                 new ConfigurationTransformsConvention(new ConfigurationTransformsBehaviour(fileSystem, variables, configurationTransformer, transformFileLocator, log)),
                 new ConfigurationVariablesConvention(new ConfigurationVariablesBehaviour(fileSystem, variables, replacer, log)),
                 new StructuredConfigurationVariablesConvention(new StructuredConfigurationVariablesBehaviour(structuredConfigVariablesService)),
+            };
+
+            if (variables.Get(Deployment.SpecialVariables.Account.AccountType) == "AmazonWebServicesAccount")
```
MissedTheMark, line 101: Is there somewhere where the "AmazonWebServicesAccount" constant can be defined and referenced here?

MissedTheMark (approved):
Looks reasonable to me.  My only concern would be in making absolutely sure that this still works in 4.0 - will any of the existing tests give us confidence with that?
Still approving this though, I'll let you make that call 👍

source/Calamari/Kubernetes/Commands/KubernetesApplyRawYamlCommand.cs
scottmerchant: I have just gone through and updated all magic strings for Account types to consts but it touches a few files so I might do it as a separate PR if that's ok with you. Good call out though 👍

source/Calamari/Kubernetes/AwsAuthConventionFactoryFactory.cs
scottmerchant: I have changed the class name and added a long winded comment. I also slightly updated the factory so that it doesn't need to be passed the `ILog` and `IVariables` as I realised they can just be injected.

source/Calamari.Aws/Integration/AwsAuthConvention.cs
scottmerchant: CC @zentron

MissedTheMark (approved):
LGTM.  I'm happy with this, and after our chat today I think we should be good to go 👍

zentron (approved):
Looks good. So long as it passes all tests then this seems cleaner 👍

PR #1029, opened 2023-05-15
Fix flakey tests which use AppServiceIntegrationTest

scottmerchant:
I had to fix these tests because the OneTimeSetup was intermittently failing which means you can't just mute them.

Changing the way the name is generated seems to have fixed the problem.

Flakey tests story: [[sc-49124](https://app.shortcut.com/octopusdeploy/story/49124/calamari-flakey-tests)]


MissedTheMark (approved):
Approving this so you can keep moving further on your other PRs, with the caveat of not seeing how this would fix the issue, but trusting you that it does 😅

PR #1030, opened 2023-05-15
Simplify k8s status

YihaoOct:
This PR simplifies the logic of Kubernetes object status checks.

- Only top level resources are included for computing the overall status of a deployment.
- Pods status now solely decided by the `phase` field, instead of failing deployment on further criteria on the container statuses. The displayed status field is still kept, such as `ImagePullBackOff`, but it does not fail the Pod.
- Stabilisation timeout is removed. Only the execution timeout is kept. Within the timeout, we fail or succeed the step immediately once we see the success/failure status from the top level resources. Once the timeout is reached, the deployment fails.
- We added a variable `Octopus.Action.Kubernetes.WaitForJobs`. When this variable is not set, Jobs are treated as always successful. When this variable is set to true, we wait for the Job to be successful within the timeout.


source/Calamari.Tests/KubernetesFixtures/ResourceStatus/Resources/PodTests.cs
```diff
@@ -154,10 +155,10 @@ public void WhenContainerFailsExecutionAndRestarting_TheStatusShouldShowCrashLoo
                 })
                 .Build();
     
-            var pod = (Pod)ResourceFactory.FromJson(podResponse);
+            var pod = (Pod)ResourceFactory.FromJson(podResponse, new Options());
 
             pod.Status.Should().Be("CrashLoopBackOff");
-            pod.ResourceStatus.Should().Be(KubernetesResources.ResourceStatus.Failed);
+            pod.ResourceStatus.Should().Be(KubernetesResources.ResourceStatus.Successful);
```
YihaoOct, line 161: Because `CrashLoopBackOff` state happens after the Pod has reached the `Running` phase, in the simplified logic we'll report the status as successful. This is not ideal, but I'm refraining from making the logic too complicated. The `CrashLoopBackOff` state will still be displayed, and the controlling object such as ReplicaSet will not have this Pod as one of the ready Pods, so the overall deployment will not succeed anyway. The only exception is when the user is trying to deploy a Pod directly from YAML, which is a discouraged use case. In this case the deployment will be successful while the Pod is in a `CrashLoopBackOff`.

source/Calamari/Kubernetes/ResourceStatus/Options.cs
```diff
@@ -0,0 +1,7 @@
+namespace Calamari.Kubernetes.ResourceStatus
+{
+    public class Options
```
zentron, line 3: A more descriptive name like `ResourceStatusOptions` will make it easier to follow what this is used for.

source/Calamari.Tests/KubernetesFixtures/ResourceStatus/ResourceStatusCheckerTests.cs
```diff
@@ -0,0 +1,214 @@
...
+        }
+
+        [Test]
+        public void FailureBeforeTimeout_ShouldReturnAsFailed()
+        {
+            var retriever = new TestRetriever();
+            var reporter = new TestReporter();
+            var resourceStatusChecker = new ResourceStatusChecker(retriever, reporter, new InMemoryLog());
+
+            var timer = new TestTimer(5);
```
zentron, line 71: Its not clear to me why some of these tests use `TestTimer(5)` and others `TestTimer(2)`
Perhaps more explicit timers would make this clearer.

source/Calamari.Tests/KubernetesFixtures/ResourceStatus/Resources/PodTests.cs
zentron: To to be clear, if we a pod goes into `CrashLoopBackOff` then the K8S cluster considers it to be `Running`? Thats surprising. I would have thought that it would still be failed. So from the users perspective on the status page we will report this pod as successful (even if the replica set isnt yet successful)

source/Calamari/Kubernetes/ResourceStatus/ResourceStatusChecker.cs
```diff
@@ -42,78 +40,64 @@ public ResourceStatusChecker(IResourceRetriever resourceRetriever, IResourceUpda
...
-                        break;
-                    case DeploymentStatus.Failed:
-                        log.Verbose("Resource status check terminated during stabilization period with some failed resources");
-                        break;
-                    default:
-                        break;
-                }
-
-                return false;
+                timer.Tick();
```
zentron, line 70: It wasn't until reading the timer class that I realised `Tick()` is actually a `Sleep`. Consider using the name `Sleep` or `Wait` or something that indicates what its doing. The clock semantics is less important than being clear about what line of code is blocking.

zentron (changes requested):
Looking good, looks like we're not far off.

source/Calamari.Tests/KubernetesFixtures/ResourceStatus/Resources/PodTests.cs
YihaoOct: Yes it is. From this diagram, you can see the `CrashLoopBackOff` state happens when the Pod enters a `Running` phase.

YihaoOct: Forgot to put in the diagram link: https://miro.com/app/board/uXjVPkFe6vA=/

source/Calamari/Kubernetes/ResourceStatus/Options.cs
YihaoOct: I didn't use a long name because it is already in the fully qualified name of this class `Calamari.Kubernetes.ResourceStatus.Options`. If the short name will result in a clash with another `Option` class, people can use it as `using ResourceStatusOptions = Calamari.Kubernetes.ResourceStatus.Options`.

Maybe adding some XMLDoc would be helpful here?

YihaoOct: Added docs for this one.

zentron (approved):

PR #1032, opened 2023-05-17
Make KubernetesCommandBase and add tests for each auth type

scottmerchant:
Moving the common auth to a base command class so that we can test the auth with a testable command. This means we don't have to run actual deployments when testing the auth types. We do this already with the script steps by using a simple script eg: `kubectl cluster-info`

[sc-47324]


source/Calamari.Tests/KubernetesFixtures/TestableKubernetesDeploymentCommand.cs
```diff
@@ -0,0 +1,68 @@
...
+            Func<ConfigurationVariablesConvention> configurationVariablesFactory,
+            Func<StructuredConfigurationVariablesConvention> structuredConfigurationVariablesFactory,
+            IAwsAuthConventionFactory awsAuthConventionFactoryFactory,
+            Func<KubernetesAuthContextConvention> kubernetesAuthContextFactory,
+            ConventionProcessor.Factory conventionProcessorFactory, RunningDeployment.Factory runningDeploymentFactory,
+            ICalamariFileSystem fileSystem, IExtractPackage extractPackage) : base(deploymentJournalWriter, variables,
+            kubectl, delegateInstallFactory, substituteInFilesFactory, configurationTransformationFactory,
+            configurationVariablesFactory, structuredConfigurationVariablesFactory, awsAuthConventionFactoryFactory,
+            kubernetesAuthContextFactory, conventionProcessorFactory, runningDeploymentFactory, fileSystem,
+            extractPackage)
```
MissedTheMark, line 37: See my comment here: https://github.com/OctopusDeploy/Calamari/pull/1042#pullrequestreview-1456727969

MissedTheMark (commented):
The concept of this is good, we just need to have another look at the factories.

source/Calamari/Kubernetes/Commands/KubernetesDeploymentCommandBase.cs
```diff
@@ -0,0 +1,131 @@
...
+            this.structuredConfigVariablesService = structuredConfigVariablesService;
+            this.kubectl = kubectl;
+
+            Options.Add("package=", "Path to the NuGet package to install.",
+                v => pathToPackage = new PathToPackage(Path.GetFullPath(v)));
+        }
+
+        protected abstract IEnumerable<IInstallConvention> CommandSpecificConventions();
+
+        public override int Execute(string[] commandLineArguments)
```
scottmerchant, line 65: most of the content of this method is exactly what is removed from `KubernetesApplyRawYamlCommand`.

source/Calamari/Kubernetes/Commands/KubernetesApplyRawYamlCommand.cs
```diff
@@ -75,70 +54,13 @@ public override int Execute(string[] commandLineArguments)
                 throw new InvalidOperationException(
                     "Unable to execute the Kubernetes Apply Raw YAML Command because the appropriate feature has not been enabled.");
```
scottmerchant, line 56: A lot of what is removed below has been moved to here: https://github.com/OctopusDeploy/Calamari/pull/1032/files#r1218793262

source/Calamari.Tests/KubernetesFixtures/TestableKubernetesDeploymentCommand.cs
```diff
@@ -0,0 +1,64 @@
...
+using Calamari.Common.Plumbing.Logging;
+using Calamari.Common.Plumbing.Variables;
+using Calamari.Deployment.Conventions;
+using Calamari.Kubernetes.Commands;
+using Calamari.Kubernetes.Integration;
+
+namespace Calamari.Tests.KubernetesFixtures
+{
+    [Command(Name, Description = "used for tests")]
+    public class TestableKubernetesDeploymentCommand : KubernetesDeploymentCommandBase
```
MissedTheMark, line 18: I don't love that we have a different command for testing, even if the bulk of the logic in contained in the base class.
Is this because GatherAndApplyRawYamlConvention and ResourceStatusReportConvention can't be used in a test?

scottmerchant: This is only used to test the Auth components of the command base. There are still tests which do full deployments as well using the real `KubernetesApplyRawYamlCommand`. This is the equivalent of running the kubernetes script but not using the real deployment script, instead only using a dummy script. You can see that in tests like `AuthorisingWithAmazonAccount`, `AuthorisingWithPodServiceAccountToken`, `AuthorisingWithAzureServicePrincipal` etc. In those cases when using the script it will use a dummy script of `kubectl cluster-info` instead of the long complicated KubernetesDeployment.sh/ps1 scripts.

MissedTheMark: Cool.  As long as we have test coverage of the real command, I'm good 👍

scottmerchant: Yep definitely, I just didn't want to do full deployments for every auth type as it's not really required so I tried to make an equivalent to a dummy script.

MissedTheMark (approved):
LGTM.  I like that we're running the same tests against the old and new approaches.

PR #1033, opened 2023-05-18
Enforce C# compiler version to 7.3

scottmerchant:
Calamari is build with .Net Framework 4.0 and 4.5.2 as well as .Net 6.0.

On Mac and linux this translates to only building with .Net 6.0 which is fine except that Rider just then uses the latest compatible C# compiler version which is higher than what is compatible with .Net Framework.

This change enforces the compiler version to 7.3 so that it doesn't trip people up who are developing for Calamari on a Mac.

You can see [here](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version#defaults) that all .Net framework versions max out at v7.3 but .Net 6 uses c# 10.

### DIFF
--- a/source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureAks.cs
+++ b/source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureAks.cs
@@ -184,7 +184,7 @@ namespace Calamari.Tests.KubernetesFixtures
                 { "awsAssumeRole", bool.FalseString }
             };
 
-            if (scope.HealthCheckContainer is not null)
+            if (scope.HealthCheckContainer != null)
             {
                 serviceMessageProperties.Add("healthCheckContainerImageFeedIdOrName", scope.HealthCheckContainer.FeedIdOrName);
                 serviceMessageProperties.Add("healthCheckContainerImage", scope.HealthCheckContainer.ImageNameAndTag);
--- a/source/Directory.Build.props
+++ b/source/Directory.Build.props
@@ -1,0 +1,5 @@
+<Project>
+ <PropertyGroup>
+   <LangVersion>7.3</LangVersion>
+ </PropertyGroup>
+</Project>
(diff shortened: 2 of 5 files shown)


source/Directory.Build.props
```diff
@@ -0,0 +1,5 @@
+<Project>
+ <PropertyGroup>
+   <LangVersion>7.3</LangVersion>
```
scottmerchant, line 3: This sets the version for all projects which don't set their own version.

YihaoOct: I do remember having syntax errors building against .net 4.0 despite I set the language version to 7.3. 
I guess in the documentation, by saying All .Net Framework uses default version of 7.3, they only included the supported versions of .net framework, which are 4.8 and 4.6(?). .net 4.0 is simply ignored because it was deprecated 8 year ago: https://devblogs.microsoft.com/dotnet/support-ending-for-the-net-framework-4-4-5-and-4-5-1/. It is understandable they don't include this version in "All".

Still, this is a good step forward, at least we have fewer syntax errors! Thank you

scottmerchant: Ah damn, that's a good point. Yeah It might be "good enough" until we update to latest .net framework version anyway. One day we'll get this solution to .net 6 😅

source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureAks.cs
```diff
@@ -184,7 +184,7 @@ public void DiscoverKubernetesClusterWithAzureServicePrincipalAccount(bool setHe
                 { "awsAssumeRole", bool.FalseString }
             };
 
-            if (scope.HealthCheckContainer is not null)
+            if (scope.HealthCheckContainer != null)
```
scottmerchant, line 187: Turns out this was the only issue keeping Calamari.Tests from being set to C# version 7.3 so I've just updated it to use the older style of null check.

scottmerchant: Rider doesn't like when you change compiler versions for projects as it will show errors where there are none until you delete the bin and obj folders for that project and rebuild.

YihaoOct (approved):
LGTM

PR #1034, opened 2023-05-18
Reverse the script wrapper order

YihaoOct:
## Issue

When deploying to EKS, a customer saw an exception thrown, complaining `KUBECONFIG` variable added twice to a Dictionary.

Octopus K8S Object Status Error

## Reason

This issue was related to https://github.com/OctopusDeploy/Calamari/pull/1011 which we put in earlier in order to fix the bug where the `KUBECONFIG` variable isn't set for kubernetes resource status check. It turns out that fix wasn't a proper one, but rather shadows the actual underlying problem.

The actual cause is that the resource status check wrapper runs BEFORE the kubernetes context wrapper. The order did not matter because the first thing the resource status check wrapper does is to invoke the next wrapper.

However, the resource status check wrapper actually need access the environment variables set by the kubernetes context wrapper, which includes `KUBECONFIG`. When the resource status check wrapper runs first, these variables are not populated yet, causing errors connecting to remote kubernetes clusters.

In #1011 we duplicated the variable setup code, so we were able to use the correct kubeconfig file to authenticate. For local and Azure clusters this worked fine without errors. This is because the  resource status check wrapper is the first in the chain, and here the `environmentVars` variable is `null`. This is passed to the next wrapper, which is the kubernetes context wrapper. Because in both wrappers this parameter was `null`, they created separate Dictionary instances and added their variables. There's no clashes.

However when the status check runs on AWS, the AWS script wrapper was added before these. In the AWS script wrapper, the `environmentVars` have been set to a non-null value. Therefore, when this value is passed to the resource status check wrapper and then the kubernetes context wrapper, they are using the same Dictionary instance, and both will attempt to add the same key `KUBECONFIG`, causing an exception.

## Fix

The proper fix for this is to swap the order of status check wrapper and kubernetes context wrapper. Then we don't need to set up those variables for a second time.

### Before

Screenshot 2023-05-19 at 3 31 01 PM
Screenshot 2023-05-19 at 3 36 04 PM

### After

Screenshot 2023-05-19 at 3 21 19 PM

### DIFF
--- a/source/Calamari/Kubernetes/KubernetesContextScriptWrapper.cs
+++ b/source/Calamari/Kubernetes/KubernetesContextScriptWrapper.cs
@@ -48,6 +48,11 @@ namespace Calamari.Kubernetes
         {
             var workingDirectory = Path.GetDirectoryName(script.File);
 
+            if (environmentVars == null)
+            {
+                environmentVars = new Dictionary<string, string>();
+            }
+            
             var setupKubectlAuthentication = new SetupKubectlAuthentication(variables,
                                                                             log,
                                                                             scriptSyntax,
--- a/source/Calamari/Kubernetes/KubernetesContextScriptWrapper.cs
+++ b/source/Calamari/Kubernetes/KubernetesContextScriptWrapper.cs
@@ -48,11 +48,6 @@ namespace Calamari.Kubernetes
         {
             var workingDirectory = Path.GetDirectoryName(script.File);
 
-            if (environmentVars == null)
-            {
-                environmentVars = new Dictionary<string, string>();
-            }
-            
             var setupKubectlAuthentication = new SetupKubectlAuthentication(variables,
                                                                             log,
                                                                             scriptSyntax,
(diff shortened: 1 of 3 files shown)


scottmerchant (approved):
LGTM.

The script wrapper setup is pretty fragile in general but some of this will end up becoming defunct when we move to using Kubernetes Commands instead. Once that transition is complete we can remove the Kubernetes Context and Status Check script wrappers completely, which will remove this issue all together. For now I think this is good 👍

source/Calamari/Kubernetes/KubernetesContextScriptWrapper.cs
```diff
@@ -48,11 +48,6 @@ public bool IsEnabled(ScriptSyntax syntax)
         {
             var workingDirectory = Path.GetDirectoryName(script.File);
 
-            if (environmentVars == null)
```
johnsimons, line 51: Why remove this?

johnsimons (approved):

source/Calamari/Kubernetes/KubernetesContextScriptWrapper.cs
YihaoOct: Damn, I removed it from the wrong wrapper. Thanks for spotting this!

PR #1035, opened 2023-05-19
Remove the idea of extensions from Calamari

scottmerchant:
While porting kubernetes script steps to commands, we noticed that the extensions idea doesn't really make sense anymore as we always ship those assemblies anyway. We just need to ensure they are only loaded when calamari is not .net4.


source/Calamari.Aws/Integration/AwsScriptWrapper.cs
```diff
@@ -16,7 +19,11 @@ public class AwsScriptWrapper : IScriptWrapper
         readonly ILog log;
         readonly IVariables variables;
         public int Priority => ScriptWrapperPriorities.CloudAuthenticationPriority;
-        bool IScriptWrapper.IsEnabled(ScriptSyntax syntax) => true;
+
+        bool IScriptWrapper.IsEnabled(ScriptSyntax syntax) =>
+            variables.Get(SpecialVariables.Account.AccountType) == "AmazonWebServicesAccount" ||
+            string.Equals(variables.Get(AwsSpecialVariables.Authentication.UseInstanceRole), bool.TrueString,
+                StringComparison.InvariantCultureIgnoreCase);
```
scottmerchant, line 26: This script wrapper would blow up if you didn't have the correct variables set with your AWS Creds (and previously it wouldn't be loaded if you didn't specify the AWS extension) so I had to put some checks in (which mirror those in Server in `WithAwsCalamariExtensionsIfRequired()` method.

source/Calamari.Tests/Fixtures/Manifest/ExecuteManifestCommandFixture.cs
```diff
@@ -94,8 +92,7 @@ CalamariResult ExecuteCommand(VariableDictionary variables, string extensions =
                 return Invoke(Calamari()
                               .Action("execute-manifest")
                               .Argument("variables", variablesFile.FilePath)
-                              .Argument("sensitiveVariablesPassword", "GB8KdBqYRlgAON9ISUPdnQ==")
-                              .Argument("extensions", extensions));
```
scottmerchant, line 98: Turns out tests were using extensions to load the `Calamari.Tests` assembly so they could have test commands and stuff.. so I had to change some things around to get everything to work again.

source/Calamari.Aws/Integration/AwsScriptWrapper.cs
MissedTheMark, line 26: Felt a bit clunky to read this, might be worth modifying it to something like this:
```suggestion
        bool IScriptWrapper.IsEnabled(ScriptSyntax syntax)
        {
            var accountType = variables.Get(SpecialVariables.Account.AccountType);
            var useAwsInstanceRole = variables.Get(AwsSpecialVariables.Authentication.UseInstanceRole);
            return accountType == "AmazonWebServicesAccount" ||
                string.Equals(useAwsInstanceRole, bool.TrueString, StringComparison.InvariantCultureIgnoreCase);
         }
```

MissedTheMark (commented):
Looks good to me, though I do feel that you should revert the program -> runner changes as we discussed.
I think I'd like to get @zentron to approve this though - lets see if we can get him to have a quick look at it this afternoon.

MissedTheMark (approved):
Approving, given that @zentron has approved the overall concept on the server side.

zentron:
Looks good Thanks for breaking this up into its own piece.  🙇

PR #1038, opened 2023-05-23
Fix resource status on raw yaml command

scottmerchant:
After [this change](https://github.com/OctopusDeploy/Calamari/pull/1034), we need to ensure that the existing EnvironmentVariables are passed in as the `KUBECONFIG` variable is no long set again in the Resource Status Executor.

### DIFF
--- a/source/Calamari/Kubernetes/Conventions/ResourceStatusReportConvention.cs
+++ b/source/Calamari/Kubernetes/Conventions/ResourceStatusReportConvention.cs
@@ -18,7 +18,7 @@ namespace Calamari.Kubernetes.Conventions
 
         public void Install(RunningDeployment deployment)
         {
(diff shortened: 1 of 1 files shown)


YihaoOct (changes requested):
Maybe we should just make the environmentVariables a non-optional parameter, so to make clear that the responsibility of handling null stuff would be on the caller. This will prevent this kind of errors I think.

scottmerchant:
> Maybe we should just make the environmentVariables a non-optional parameter, so to make clear that the responsibility of handling null stuff would be on the caller. This will prevent this kind of errors I think.

Yep good point, nice pick up. I've made it non-optional 👍

YihaoOct (approved):
LGTM 👍

PR #1036, opened 2023-05-22
Add kubelogin to AKS cluster auth

IsaacCalligeros95:
Adds support for [Kubelogin](https://github.com/Azure/kubelogin) to AKS cluster authentication. This is done using the [convert-kubeconfig](https://azure.github.io/kubelogin/cli/convert-kubeconfig.html) command and will require kubelogin to be on path for all AKS deployments. 

Kubelogin can be installed simply by running `az aks install-cli`, for further details see https://azure.github.io/kubelogin/install.html. 

## Results
Executing locally with kubelogin for aad-rbac aks cluster
image

In Execution containers (Windows 5.1.0) for AAD rbac and rbac 
image

Execution containers (Ubuntu 5.1.0) 
image

PR #1041, opened 2023-05-25
Kubelogin e2e

veochen-octopus:
E2E tests to verify the `kubelogin` changes in https://github.com/OctopusDeploy/Calamari/pull/1036. To make this PR easy to read and review, I've kept the old fixture `KubernetesContextScriptWrapperLiveFixtureAks` unchanged. This uses a cluster with local account enabled, like before. `kubelogin` is invoked wherever appropriate.

An additional fixture is added based on the existing fixture
- It's called `KubernetesContextScriptWrapperLiveFixtureAksLocalAccessDisabled`
- The terraform files are in the `AKS-local-access-disabled` folder
- This creates a cluster with local access disabled, and therefore with the service account, role, role binding and secret all removed
- The test cases for authorizing with a service account token and a client certificate are also removed as they're associated with a service account. This leaves only one authorization test which uses the service principal.


source/Calamari.Tests/KubernetesFixtures/Terraform/Clusters/AKS-local-access-disabled/aks.tf
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  role_based_access_control_enabled = true
+  local_account_disabled            = true
+
+  azure_active_directory_role_based_access_control {
+    managed            = true
+    azure_rbac_enabled = true
+  }
+}
```
veochen-octopus, line 35: These are the key configs for creating a cluster with local access disabled.

benPearce1 (approved):

PR #1042, opened 2023-05-25
Add test for KubernetesApplyRawYamlCommand and tidy up implementation

scottmerchant:
I've updated the kubernetes deployment test to test both the raw yaml command and the raw yaml deployment script. 

I've also updated the test to run calamari as a "black box" instead of `new`ing everything up, we just run the calamari out of process with the correct env vars and variables. This means we are also testing that the autofac registrations (including registering things from different assemblies) are done correctly. It also reduces the boilerplate in the test infrastructure.

As a follow on from the above - I'm able to revert a few changes I made previously to help pass in the correct working directory and environment variables when calling calamari in-process as they are no longer required. I've also updated the `KubernetesApplyRawYamlCommand` to only compile in .NET 4.5.1 and .NET 6 with precompile flags so that we can directly reference`AwsAuthConvention` directly without factories etc.

[sc-47324]


source/Calamari/Kubernetes/Commands/KubernetesApplyRawYamlCommand.cs
```diff
@@ -4,66 +4,77 @@
...
+            Func<SubstituteInFilesConvention> substituteInFilesFactory,
+            Func<ConfigurationTransformsConvention> configurationTransformationFactory,
+            Func<ConfigurationVariablesConvention> configurationVariablesFactory,
+            Func<StructuredConfigurationVariablesConvention> structuredConfigurationVariablesFactory,
+            IAwsAuthConventionFactory awsAuthConventionFactory,
+            Func<KubernetesAuthContextConvention> kubernetesAuthContextFactory,
+            Func<GatherAndApplyRawYamlConvention> gatherAndApplyRawYamlFactory,
+            Func<ResourceStatusReportConvention> resourceStatusReportFactory,
+            ConventionProcessor.Factory conventionProcessorFactory,
+            RunningDeployment.Factory runningDeploymentFactory,
```
MissedTheMark, line 58: I'm not sure this is an improvement - what problem are we solving by converting these all to factories?  I can't spot anywhere in your other PRs that requires this (correct me if I'm wrong).
Seeing these makes me concerned that we're either using DI for DI's sake, or we have a design issue with our command.

MissedTheMark (changes requested):
See comment below.  The number of factories here concerns me.

source/Calamari.Common/Features/Packages/ExtractPackage.cs
```diff
@@ -23,16 +23,16 @@ public ExtractPackage(ICombinedPackageExtractor combinedPackageExtractor, ICalam
             this.log = log;
         }
 
-        public void ExtractToStagingDirectory(PathToPackage? pathToPackage, IPackageExtractor? customPackageExtractor = null)
+        public void ExtractToStagingDirectory(PathToPackage? pathToPackage, IPackageExtractor? customPackageExtractor = null, string? workingDirectory = null)
         {
-            var targetPath = Path.Combine(Environment.CurrentDirectory, "staging");
+            var targetPath = Path.Combine(workingDirectory ?? Environment.CurrentDirectory, "staging");
```
zentron, line 28: Whats the reason that this step needs to extract the package to a working directory rather than the Current Directory?  Based on the implimentation of the value being passed in up the chain it looks like it will typically resolve to Environment.CurrentDirectory anyway?

source/Calamari/Kubernetes/Commands/KubernetesApplyRawYamlCommand.cs
```diff
@@ -73,52 +62,58 @@ public override int Execute(string[] commandLineArguments)
...
+                    }
+
+                    kubectl.WorkingDirectory = d.CurrentDirectory;
+                    kubectl.EnvironmentVariables = d.EnvironmentVariables;
                 }),
-                new SubstituteInFilesConvention(new SubstituteInFilesBehaviour(substituteInFiles)),
-                new ConfigurationTransformsConvention(new ConfigurationTransformsBehaviour(fileSystem, variables, configurationTransformer, transformFileLocator, log)),
-                new ConfigurationVariablesConvention(new ConfigurationVariablesBehaviour(fileSystem, variables, replacer, log)),
-                new StructuredConfigurationVariablesConvention(new StructuredConfigurationVariablesBehaviour(structuredConfigVariablesService)),
+                conventionFactory.Resolve<SubstituteInFilesConvention>(),
```
zentron, line 90: why is this command now using a convention factory rather than just new-ing it up like the others?

It might be an idea to consider how we can update how the relationship for _all_ commands and conventions work but its not clear why this one in particular needs to be different. Presumably for testing?

zentron (dismissed):

source/Calamari.Common/Features/Packages/ExtractPackage.cs
scottmerchant: Ah yeah this was for tests, but is now not required - I might back out these workingDirectory changes.

source/Calamari/Kubernetes/SetupKubectlAuthentication.cs
```diff
@@ -45,7 +45,7 @@ public class SetupKubectlAuthentication
             this.workingDirectory = workingDirectory;
         }
 
-        public CommandResult Execute(string accountType)
+        public CommandResult Execute(string accountType, Kubectl kubectl = null)
```
zentron, line 48: Whats the reason that sometimes we can pass through the kubectl (based on where its called in the  `KubernetesAuthContextConvention` presumably via the container) and in the other case (via `KubernetesContextScriptWrapper`) we dont?

scottmerchant: Yep - basically allowing the wrapper to continue to function until we can remove it (once all kubectl script steps are moved to calamari commands).

source/Calamari/Kubernetes/Commands/KubernetesApplyRawYamlCommand.cs
scottmerchant: After our chat this morning I've reverted the injection of the conventions as well as made a few other tidy ups. Looking pretty clean now I think.

source/Calamari.Common/Features/Packages/ExtractPackage.cs
scottmerchant: This has been reverted now.

source/Calamari/Kubernetes/Commands/KubernetesApplyRawYamlCommand.cs
scottmerchant: I've reverted this after discussing with Mark.

source/Calamari/Kubernetes/SetupKubectlAuthentication.cs
scottmerchant: I've managed to remove this and put it as a constructor dependency. It's not the wrappers responsibility to create the kubectl instance which makes more sense and will go away once we can delete the wrapper.

source/Calamari.Common/Commands/RunningDeployment.cs
```diff
@@ -7,9 +7,6 @@ namespace Calamari.Common.Commands
 {
     public class RunningDeployment
     {
-        public delegate RunningDeployment Factory(string? packageFilePath,
-            Dictionary<string, string>? environmentVariables = null);
```
scottmerchant, line 11: Because the tests are now executing calamari out-of-process, we can specify the environment variables for the executable before it runs so we don't need to do any fance in-process passing in specific environment variables anymore. This means that this previous change to RunningDeployment can be removed and we can new it up like we were originally.

scottmerchant: This also means we no longer need to register `RunningDeployment` in the Calamari Flavour Program classes.

source/Calamari/Kubernetes/Commands/KubernetesApplyRawYamlCommand.cs
```diff
@@ -20,109 +22,124 @@
...
+                new ConfigurationVariablesConvention(new ConfigurationVariablesBehaviour(fileSystem, variables,
+                    new ConfigurationVariablesReplacer(variables, log), log)),
+                new StructuredConfigurationVariablesConvention(
+                    new StructuredConfigurationVariablesBehaviour(structuredConfigVariablesService))
             };
 
             if (variables.Get(Deployment.SpecialVariables.Account.AccountType) == "AmazonWebServicesAccount")
             {
-                conventions.Add(awsAuthConventionFactory.Create());
+                conventions.Add(new AwsAuthConvention(log, variables));
```
scottmerchant, line 117: I've added precompiler flags to this whole class so it won't be compiled in .NET 4.0 this means we can new up the `AwsAuthConvention` here without factories etc etc.

source/Calamari/Kubernetes/Conventions/GatherAndApplyRawYamlConvention.cs
```diff
@@ -19,16 +19,16 @@ public class GatherAndApplyRawYamlConvention: IInstallConvention
...
         private readonly ILog log;
         private readonly ICalamariFileSystem fileSystem;
-        private readonly ICommandLineRunner commandLineRunner;
+        private readonly Kubectl kubectl;
 
         public GatherAndApplyRawYamlConvention(
             ILog log,
             ICalamariFileSystem fileSystem,
-            ICommandLineRunner commandLineRunner)
+            Kubectl kubectl)
```
scottmerchant, line 27: Kubectl is now passed in as a constructor level parameter so that we only have a single instance. If we have multiple instances then each one will print out logs saying it has successfully being setup which doesn't look too good in the task log.

source/Calamari/Kubernetes/Conventions/GatherAndApplyRawYamlConvention.cs
```diff
@@ -135,19 +129,15 @@ private IEnumerable<Resource> ApplyBatchAndReturnResources(int index, string glo
                 var token = JToken.Parse(outputJson);
 
                 List<Resource> lastResources;
-                if (token.Type == JTokenType.Array)
-                {
-                    lastResources = token.ToObject<List<Resource>>();
-                }
```
scottmerchant, line 141: This check was removed because I thought it was required at first when I added it but after more testing it seems like I was just mistaken when porting the logic from bash/powershell to c# so I've removed it to reduce confusion for future engineers.

source/Calamari/Kubernetes/Conventions/GatherAndApplyRawYamlConvention.cs
```diff
@@ -135,19 +129,15 @@ private IEnumerable<Resource> ApplyBatchAndReturnResources(int index, string glo
...
-                else if (token["kind"]?.ToString() != "List" ||
+                if (token["kind"]?.ToString() != "List" ||
                     (lastResources = token["items"]?.ToObject<List<Resource>>()) == null)
                 {
                     lastResources = new List<Resource> { token.ToObject<Resource>() };
                 }
 
                 foreach (var resource in lastResources)
                 {
-                    log.Info($"'{resource.Kind}/{resource.Metadata.Name}' created.");
```
scottmerchant, line 150: Removing this punctuation means the output is the same as the old deployment scripts

source/Calamari/Kubernetes/KubernetesContextScriptWrapper.cs
```diff
@@ -52,11 +53,13 @@ public bool IsEnabled(ScriptSyntax syntax)
             {
                 environmentVars = new Dictionary<string, string>();
             }
-            
+
+            var kubectl = new Kubectl(variables, log, commandLineRunner, workingDirectory, environmentVars);
```
scottmerchant, line 57: For wrappers I'm newing up kubectl here instead of inside the components (eg: inside `SetupKubectlAuthentication`). The affective behaviour is the same for the deployment scripts where kubectl instances are created numerous times, each time printing messages to the task log.

source/Calamari/Kubernetes/ResourceStatus/ResourceStatusReportWrapper.cs
```diff
@@ -50,6 +61,11 @@ public bool IsEnabled(ScriptSyntax syntax)
             Dictionary<string, string> environmentVars)
         {
             var workingDirectory = Path.GetDirectoryName(script.File);
+            var kubectl = new Kubectl(variables, log, commandLineRunner, workingDirectory, environmentVars);
```
scottmerchant, line 64: Same as https://github.com/OctopusDeploy/Calamari/pull/1042/files#r1218780595

MissedTheMark (approved):
Looks good to me 👍   Thanks for addressing those issues that were raised.  I like that the Net4 incompatible stuff doesn't get compiled for it at all now, makes a lot more sense than trying to hide it from like we were before.

PR #1044, opened 2023-05-26
Add additional tests for raw yaml command

scottmerchant:
In this PR I've added more tests for the raw yaml command (and deployment scripts) to cover failure cases and to ensure the logging is the same (or similar in some cases).

[sc-47324]


source/Calamari.Tests/KubernetesFixtures/Scripts/KubernetesDeployment.sh
```diff
@@ -227,6 +227,9 @@ function deploy_customresources {
                 done
             else
                 write_plainerror "\"kubectl apply -o json\" returned invalid JSON."
+                write_plainerror "---------------------------"
+                write_plainerror "$newCustomResources"
+                write_plainerror "---------------------------"
```
scottmerchant, line 232: This was a difference I found between `KubernetesDeployment.ps1` and `KubernetesDeployment.sh`. I've made it match so I didn't have adjust the test to handle different platform/script types but in the Command it behaves the same way as the Original `KubernetesDeployment.ps1` script.

source/Calamari/Kubernetes/Commands/KubernetesDeploymentCommandBase.cs
```diff
@@ -122,12 +122,18 @@ public override int Execute(string[] commandLineArguments)
...
-                conventionRunner.RunConventions();
+                conventionRunner.RunConventions(logExceptions: false);
                 deploymentJournalWriter.AddJournalEntry(conventionRunner.Deployment, true, pathToPackage);
             }
-            catch (Exception)
+            catch (Exception e)
             {
                 deploymentJournalWriter.AddJournalEntry(conventionRunner.Deployment, false, pathToPackage);
+
+                if (e is KubernetesDeploymentFailedException || e is TimeoutException)
```
scottmerchant, line 132: These changes are to make the logs look neater and more user friendly. Basically, the conventions we are running handle known exceptions by logging the appropriate information and then throwing either a `KubernetesDeploymentFailedException` or a `TimeoutException`. We don't need to relog them again with a stack trace so instead we just just return -1.

source/Calamari.Common/Plumbing/FileSystem/CalamariPhysicalFileSystem.cs
```diff
@@ -206,17 +206,17 @@ public virtual IEnumerable<string> EnumerateFiles(string parentDirectoryPath, pa
             var parentDirectoryInfo = new DirectoryInfo(parentDirectoryPath);
 
             return searchPatterns.Length == 0
-                ? parentDirectoryInfo.GetFiles("*", SearchOption.TopDirectoryOnly).Select(fi => fi.FullName)
-                : searchPatterns.SelectMany(pattern => parentDirectoryInfo.GetFiles(pattern, SearchOption.TopDirectoryOnly).Select(fi => fi.FullName));
+                ? parentDirectoryInfo.EnumerateFiles("*", SearchOption.TopDirectoryOnly).Select(fi => fi.FullName)
+                : searchPatterns.SelectMany(pattern => parentDirectoryInfo.EnumerateFiles(pattern, SearchOption.TopDirectoryOnly).Select(fi => fi.FullName));
```
MissedTheMark, line 210: Given that we're only after the full name of the files, we should be able to get a significant performance increase (~2x) if we use the static version of this call, i.e. `Directory.EnumerateFiles(parentDirectoryPath, "*", SearchOption.TopDirectoryOnly)`.
Same applies below too.  At the moment, we're loading all the info about the files, not just their name.

MissedTheMark (commented):
Couple of comments to begin with, will continue on Monday.

source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureAks.cs
```diff
@@ -124,9 +149,7 @@ public void UnreachableK8Cluster_ShouldExecuteTargetScript()
             variables.Set($"{clientCert}.CertificatePem", aksClusterClientCertificate);
             variables.Set($"{clientCert}.PrivateKeyPem", aksClusterClientKey);
 
-            var wrapper = CreateWrapper();
-
-            TestScript(wrapper, "Test-Script");
+            DeployWithNonKubectlTestScriptAndVerifyResult();
```
scottmerchant, line 152: Now all tests for kubernetes are done out-of-process, instead of in process.

source/Calamari.Tests/KubernetesFixtures/Scripts/KubernetesDeployment.ps1
```diff
@@ -263,6 +262,22 @@ function Deploy-CustomResources() {
...
+
+	Write-PlainError "`"kubectl apply -o json`" returned invalid JSON:"
+	Write-PlainError "---------------------------"
+	Write-PlainError "$json"
+	Write-PlainError "---------------------------"
+	Write-PlainError "This can happen with older versions of kubectl. Please update to a recent version of kubectl."
+	Write-PlainError "See https://github.com/kubernetes/kubernetes/issues/58834 for more details."
+	Write-PlainError "Custom resources will not be saved as output variables, and will not be automatically cleaned up."
+}
+
```
scottmerchant, line 280: As annoying as it is, I had to update this script so that it would produce the same output as the KubernetesDeployment.sh script. Unfortunately it was not producing the same output which is really annoying to test against.

source/Calamari.Common/Features/Scripting/WindowsPowerShell/Bootstrap.ps1
```diff
@@ -292,6 +292,40 @@ function Write-Verbose
...
+as a PS script that can be invoked from a batch file, for instance.
+
+Note that PS by default sends ALL its streams to *stdout* when invoked from
+cmd.exe.
+
+This function acts similarly to Write-Host in that it simply calls
+.ToString() on its input; to get the default output format, invoke
+it via a pipeline and precede with Out-String.
+
+taken from https://stackoverflow.com/a/15669365/19177210
```
scottmerchant, line 311: I lifted this from a stack overflow post - is that ok?

source/Calamari.Common/Plumbing/FileSystem/CalamariPhysicalFileSystem.cs
```diff
@@ -397,11 +397,16 @@ string GetTempBasePath()
...
             var path = Path.Combine(GetTempBasePath(), Guid.NewGuid().ToString());
-            if (!Directory.Exists(path))
-                Directory.CreateDirectory(path);
+            CreateDirectory(path);
             return path;
         }
 
+        public void CreateDirectory(string directory)
+        {
+            if (!Directory.Exists(directory))
```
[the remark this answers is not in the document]
scottmerchant: Fixed 👍

source/Calamari.Common/Plumbing/FileSystem/CalamariPhysicalFileSystem.cs
scottmerchant: Fixed! Turns out this wasn't setup in the correct way (it wasn't using AlphaFS for .net framework builds). I've fixed it so now it's done properly. Once we're off .net framework this `CalamariPhysicalFileSystem` class should be a lot simpler.

source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureBase.cs
```diff
@@ -214,34 +212,22 @@ protected void DoDiscovery(AwsAuthenticationDetails authenticationDetails)
...
-                       .Argument("extensions", string.Join(',', extensions)));
+            variables.Add(KubernetesDiscoveryCommand.ContextVariableName,
+                JsonConvert.SerializeObject(discoveryContext));
 
-                WriteLogMessagesToTestOutput();
-
-                return result;
-            }
+            SetupTempDirectoryAndVerifyResult(null,
+                (dir, _) => ExecuteCommand(KubernetesDiscoveryCommand.Name, dir, null), shouldSucceed: true);
```
scottmerchant, line 230: I've updated target discovery tests to execute out-of-process too so that all kubernetes tests are consistent.

source/Calamari.Tests/KubernetesFixtures/Scripts/KubernetesDeployment.ps1
scottmerchant: These updates could be ported back to server but I'm not sure its worth it as we'll be phasing out these script steps soon anyway. 🤷‍♂️

source/Calamari.Tests/KubernetesFixtures/Scripts/KubernetesDeployment.sh
scottmerchant: Same as for the powershell script, I don't think it's worth porting it back to server as we're phasing out the use of the scripts anyway.

source/Calamari.Common/Plumbing/FileSystem/CalamariPhysicalFileSystem.cs
```diff
@@ -201,22 +201,18 @@ IEnumerable<FileSystemInfo> EnumerateWithGlob(string parentDirectoryPath, params
...
+        public virtual IEnumerable<string> EnumerateFiles(
+            string parentDirectoryPath,
+            params string[] searchPatterns)
         {
-            var parentDirectoryInfo = new DirectoryInfo(parentDirectoryPath);
-
-            return searchPatterns.Length == 0
-                ? parentDirectoryInfo.GetFiles("*", SearchOption.TopDirectoryOnly).Select(fi => fi.FullName)
-                : searchPatterns.SelectMany(pattern => parentDirectoryInfo.GetFiles(pattern, SearchOption.TopDirectoryOnly).Select(fi => fi.FullName));
+            return Directory.EnumerateFiles(parentDirectoryPath, searchPatterns);
```
MissedTheMark, line 208: Does `Directory.EnumerateFiles(..)` actually accept an array of search patterns here?  I can't find it in the docs.

MissedTheMark: Looking below, is the intent here to use `StandardDirectory.EnumerateFiles?`

source/Calamari.Common/Plumbing/FileSystem/CalamariPhysicalFileSystem.cs
```diff
@@ -201,22 +201,18 @@ IEnumerable<FileSystemInfo> EnumerateWithGlob(string parentDirectoryPath, params
...
+        public virtual IEnumerable<string> EnumerateFilesRecursively(
+            string parentDirectoryPath,
+            params string[] searchPatterns)
         {
-            var parentDirectoryInfo = new DirectoryInfo(parentDirectoryPath);
-
-            return searchPatterns.Length == 0
-                ? parentDirectoryInfo.GetFiles("*", SearchOption.AllDirectories).Select(fi => fi.FullName)
-                : searchPatterns.SelectMany(pattern => parentDirectoryInfo.GetFiles(pattern, SearchOption.AllDirectories).Select(fi => fi.FullName));
+            return Directory.EnumerateFilesRecursively(parentDirectoryPath, searchPatterns);
```
MissedTheMark, line 215: As above ☝️

source/Calamari.Common/Plumbing/FileSystem/FileOperations.cs
```diff
@@ -110,6 +116,31 @@ public IEnumerable<string> EnumerateDirectoriesRecursively(string path)
             return Directory.EnumerateDirectories(path, "*", SearchOption.AllDirectories);
         }
 
+        public virtual IEnumerable<string> EnumerateFiles(
```
MissedTheMark, line 119: My apologies if I've missed them, but this and the other new methods in this class are prime candidates for unit tests.

MissedTheMark: (I like these, btw 😁)

source/Calamari.Shared/Calamari.Shared.csproj
```diff
@@ -84,7 +84,7 @@
     <PackageReference Include="Microsoft.Web.Deployment" Version="3.6.0" />
     <PackageReference Include="NuGet.CommandLine" Version="2.8.6" />
     <PackageReference Include="NuGet.Core" Version="2.14.0" />
-    <PackageReference Include="AlphaFS" Version="2.1.3-octopus0006" />
+    <PackageReference Include="AlphaFS" Version="2.2.6" />
```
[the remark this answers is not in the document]
MissedTheMark: I'm a big fan of updating external dependencies where we can, given proper consideration to the risks, so usually I'd be all thumbs up here.
The old version number `2.1.3-octopus0006` made me think it might be a custom Octopus branch though, and indeed it does look like it: https://github.com/OctopusDeploy/AlphaFS

source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureBase.cs
```diff
@@ -89,44 +84,81 @@ CalamariResult ExecuteDirectlyWithScriptEngine(IReadOnlyList<IScriptWrapper> wra
...
+            DeployWithScriptAndVerifyResult("echo running target script...");
+        }
+
+        private void DeployWithScriptAndVerifyResult(string script,
+            Func<string, string> addFilesOrPackageFunc = null, bool shouldSucceed = true)
+        {
+            DeployWithScriptAndVerifyResult(() => (script, script), addFilesOrPackageFunc, shouldSucceed);
+        }
+
+        private void DeployWithScriptAndVerifyResult(Func<(string bashScript, string powershellScript)> scriptFactory,
```
MissedTheMark, line 114: I think we should just pass in the two scripts as strings here, instead of the function - we're not gaining anything by it that I can see, and it makes it more complex to understand.

source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureBase.cs
```diff
@@ -89,44 +84,81 @@ CalamariResult ExecuteDirectlyWithScriptEngine(IReadOnlyList<IScriptWrapper> wra
...
+        {
+            SetupTempDirectoryAndVerifyResult(addFilesOrPackageFunc, (d, p) =>
+            {
+                var (bashScript, powershellScript) = scriptFactory();
                 variables.Set(SpecialVariables.Action.Script.ScriptBodyBySyntax(ScriptSyntax.Bash),
-                    File.ReadAllText(Path.Combine(scriptPath, "KubernetesDeployment.sh")));
+                    bashScript);
                 variables.Set(SpecialVariables.Action.Script.ScriptBodyBySyntax(ScriptSyntax.PowerShell),
-                    File.ReadAllText(Path.Combine(scriptPath, "KubernetesDeployment.ps1")));
+                    powershellScript);
```
MissedTheMark, line 123: Can we just set the variables within the `DeployWithScriptAndVerifyResult` method instead of doing it within the  `SetupTempDirectoryAndVerifyResult` method?

source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureBase.cs
```diff
@@ -89,44 +84,81 @@ CalamariResult ExecuteDirectlyWithScriptEngine(IReadOnlyList<IScriptWrapper> wra
...
 
-                return ExecuteCommand(RunScriptCommand.Name, d);
-            });
+                return ExecuteCommand(RunScriptCommand.Name, d, p);
```
MissedTheMark, line 125: If we changed the above, would anything other than the `ExecuteCommand(RunScriptCommand.Name, d, p)` method need to be called in the `SetupTempDirectoryAndVerifyResult` method? In that case, we could just move the `ExecuteCommand` call into that method.

At the moment, passing multiple func calls into it makes it pretty hard to follow what's going on.

source/Calamari/Kubernetes/Conventions/GatherAndApplyRawYamlConvention.cs
```diff
@@ -41,7 +45,7 @@ public void Install(RunningDeployment deployment)
             var directories = GroupFilesIntoDirectories(deployment, globs, variables);
 
             var resources = new List<Resource>();
-            foreach (var (directory, index) in directories.Select((d,i) => (d,i)))
+            foreach (var (directory, index) in directories.Select((d, i) => (d, i)))
```
MissedTheMark, line 48: "Back in my day, we didn't have any of them fancy foreach statements.  I wore an onion on my belt, which was the style at the time..." 😁 
But TBH, it's what you really want here:
```suggestion
            for(int i = 0; i < directories.Count(); i++){
```

MissedTheMark: Otherwise, if you want to keep it set-based, something like this should work fine:
```suggestion
            var resources = directories.SelectMany((directory, index) => ApplyBatchAndReturnResources(index, globs[index], directory))
```

source/Calamari/Kubernetes/Conventions/GatherAndApplyRawYamlConvention.cs
```diff
@@ -77,12 +81,23 @@ private List<string> GroupFilesIntoDirectories(RunningDeployment deployment, str
             var directories = new List<string>();
             foreach (var (glob, idx) in globs.Select((g, i) => (g, i)))
```
MissedTheMark, line 82: As above, here.

source/Calamari/Kubernetes/Conventions/GatherAndApplyRawYamlConvention.cs
```diff
@@ -77,12 +81,23 @@ private List<string> GroupFilesIntoDirectories(RunningDeployment deployment, str
...
                 {
-                    fileSystem.CopyFile(file, Path.Combine(directory, Path.GetFileName(file)));
+                    var relativeFilePath = file.Path ?? file.Stem;
+                    var fullPath = Path.Combine(deployment.CurrentDirectory, relativeFilePath);
+                    var targetPath = Path.Combine(directory, relativeFilePath);
+                    var targetDirectory = Path.GetDirectoryName(targetPath);
+                    if (targetDirectory != null)
+                    {
+                        fileSystem.CreateDirectory(targetDirectory);
+                    }
```
MissedTheMark, line 99: Can we just do this here?  It saves us a retrieval of the path info.   I'm not massively stressed either way, but depending on the number of files returned it could be a bit faster.
```suggestion
                        fileSystem.CreateDirectory(targetPath); // Only creates the directory if it doesn't exist
```

MissedTheMark (changes requested):
Few things here that aren't critical, but the AlphaFS change could well be, hence the 🛑 change request here.

source/Calamari/Kubernetes/Conventions/GatherAndApplyRawYamlConvention.cs
scottmerchant: Does this work if the path includes the file name?

MissedTheMark: Oh right! No, I imagine it wouldn't.

source/Calamari.Common/Features/Scripting/WindowsPowerShell/Bootstrap.ps1
```diff
@@ -292,6 +292,39 @@ function Write-Verbose
...
+	}
+	if ($InputObject) {
+		[void] $outFunc.Invoke($InputObject.ToString())
+	} else {
+		[string[]] $lines = @()
+		$Input | % { $lines += $_.ToString() }
+		[void] $outFunc.Invoke($lines -join "`r`n")
+	}
+}
+
```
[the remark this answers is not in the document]
scottmerchant: I think that's basically saying if you run `Write-StdErr` with no arg, it'll print a new line. My main concern with removing it would be that it might not meet other devs' expectations if nothing happens when it's used.

Also technically this function would now be available to customers to use in their own scripts - but it's not advertised anywhere so the only way they would know is if they inspected this code.

source/Calamari.Common/Features/Scripting/WindowsPowerShell/Bootstrap.ps1
scottmerchant: Yeah I'm not sure - don't want to get done for copywrite but there isn't many ways you can actually write this...

I'm tempted to just write it as a warning instead in powershell and not check the error level in the test.

source/Calamari.Common/Plumbing/FileSystem/CalamariPhysicalFileSystem.cs
scottmerchant: It's super confusing in a code review because `Dictionary` here is referring to a property on `CalamariPhysicalFileSystem` which is of type `IDictionary` and is an instance of `StandardDirectory` or `LongPathsDirectory` depending on the runtime of the Calamari.

So it's not just the `Dictionary` static method...

source/Calamari.Common/Plumbing/FileSystem/CalamariPhysicalFileSystem.cs
```diff
@@ -389,19 +385,22 @@ string GetTempBasePath()
...
-            if (!Directory.Exists(path))
-                Directory.CreateDirectory(path);
+            CreateDirectory(path);
             return path;
         }
 
+        public void CreateDirectory(string directory)
+        {
+            Directory.CreateDirectory(directory);
+        }
```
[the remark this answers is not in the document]
scottmerchant: It's unfortunately required because Dictionary is a property not just the static class. It can contain different implementations depending on the calamari build - for .net Framework it uses `LongPathsDirectory` for everything else it uses `StandardDictionary`.

source/Calamari.Shared/Calamari.Shared.csproj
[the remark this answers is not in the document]
scottmerchant: Reverting this to the original version. We'll phase out AlphaFS once we're off the .NET Framework.

source/Calamari.Common/Plumbing/FileSystem/CalamariPhysicalFileSystem.cs
MissedTheMark: Ah gotcha 👍

source/Calamari.Common/Plumbing/FileSystem/CalamariPhysicalFileSystem.cs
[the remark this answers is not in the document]
MissedTheMark: Oh right, I missed that this is a public method.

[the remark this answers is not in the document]
MissedTheMark: Also, Dictionary => Directory for a bunch of your comments here 😁

source/Calamari.Common/Plumbing/FileSystem/CalamariPhysicalFileSystem.cs
scottmerchant: *Directory

They both start with D so they are interchangeable right? 🤦‍♂️

source/Calamari.Common/Plumbing/FileSystem/CalamariPhysicalFileSystem.cs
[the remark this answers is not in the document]
scottmerchant: woops 🤦‍♂️

source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureBase.cs
scottmerchant: This is a good point - I've fixed this whole lot of methods and removed the guff. It's a lot neater now.

source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureBase.cs
scottmerchant: Fixed (see other comment)

source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureBase.cs
scottmerchant: Fixed (see other comment)

source/Calamari.Common/Plumbing/FileSystem/FileOperations.cs
MissedTheMark: As per our Slack conversation, I'm ok with you adding unit tests for these in a separate PR, given that this is only currently used by commands being a feature toggle.

source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureBase.cs
```diff
@@ -50,135 +44,55 @@ public void Setup()
             SetTestClusterVariables();
         }
 
-        protected KubernetesContextScriptWrapper CreateWrapper(ICalamariFileSystem fileSystem = null)
+        protected void DeployWithDeploymentScriptAndVerifyResult(Func<string, string> addFilesOrPackageFunc = null, bool shouldSucceed = true)
```
MissedTheMark, line 47: :chef-kiss: much cleaner!

source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureEks.cs
```diff
@@ -490,17 +540,117 @@ public void DiscoverKubernetesClusterWithInvalidAccountCredentials()
...
+                {
+                    using (var readStream = File.OpenRead(pathToCustomResource))
+                    {
+                        archive.AddEntry(pathInPackage, readStream);
+                        using (var writeStream = File.OpenWrite(pathToPackage))
+                        {
+                            archive.SaveTo(writeStream, CompressionType.Deflate);
+                        }
+                    }
+                }
```
MissedTheMark, line 648: Don't worry about actually changing this, but just FYI:
```suggestion
                using (var archive = ZipArchive.Create())
                using (var readStream = File.OpenRead(pathToCustomResource))
                {
                    archive.AddEntry(pathInPackage, readStream);
                    using var writeStream = File.OpenWrite(pathToPackage); // This is fine as a using declaration because it'll be disposed within this block.
                    archive.SaveTo(writeStream, CompressionType.Deflate);
                }
```

source/Calamari.Common/Plumbing/FileSystem/FileOperations.cs
scottmerchant: Any chance we can do those tests in a subsequent PR? I think this should be ok as it's just passing the buck to the built-in methods anyways but yeah happy to add some tests afterwards.

MissedTheMark (approved):
Looks good to me 👍 
Thanks for implementing those `Func` changes, it's nice and easy to follow now 🎉

source/Calamari.Common/Features/Scripting/WindowsPowerShell/Bootstrap.ps1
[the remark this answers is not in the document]
scottmerchant: I've removed it - just used `Write-Warning` instead.

source/Calamari.Common/Features/Scripting/WindowsPowerShell/Bootstrap.ps1
scottmerchant: I've removed this.

PR #1046, opened 2023-05-29
Add extra check to ensure that the AWSScriptWrapper runs for aws script steps

scottmerchant:
A Windows E2E test started failing after I merged my change to remove Extensions into master because it was deploying an AWS Script. 

The AWS Script utilises the `AwsScriptWrapper` where before it was always enabled if it was side loaded via an extension, now it's always loaded but won't execute unless the right variables are set. The Aws Script Step doesn't set the AccountType variable so I also have to check the AwsAccount.Variable variable, to ensure that there is account creds there so that the script wrapper can run.

Removal of Extensions PRs:
Server: https://github.com/OctopusDeploy/OctopusDeploy/pull/18227
Calamari: https://github.com/OctopusDeploy/Calamari/pull/1035

### DIFF
--- a/source/Calamari.Aws/Deployment/AwsSpecialVariables.cs
+++ b/source/Calamari.Aws/Deployment/AwsSpecialVariables.cs
@@ -1,10 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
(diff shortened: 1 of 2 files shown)


MissedTheMark (approved):
LGTM

PR #1047, opened 2023-05-30
Remove --short flag on kubectl version call

IsaacCalligeros95:
We output the kubctl version using `kubectl version --client --short` this prints to stdwarn as --short is being deprecated in future versions of kubectl. The output from this command isn't being used directly, it's purely for logging. This change removes the short flag due to deprecation.

Removing --short also throws a warning, I've settled on --output=yaml, prefering the human readability. Json is available as well.
image

Fixes https://github.com/OctopusDeploy/Issues/issues/8082

### DIFF
--- a/source/Calamari/Kubernetes/Integration/Kubectl.cs
+++ b/source/Calamari/Kubernetes/Integration/Kubectl.cs
@@ -50,7 +50,7 @@ namespace Calamari.Kubernetes.Integration
                 ExecutableLocation = customKubectlExecutable;
             }
 
-            if (TryExecuteKubectlCommand("version", "--client", "--short"))
+            if (TryExecuteKubectlCommand("version", "--client", "--output=yaml"))
             {
                 log.Verbose($"Found kubectl and successfully verified it can be executed.");
                 return true;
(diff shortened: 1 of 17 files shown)


veochen-octopus (approved):
Nice one 👍

source/Calamari/Kubernetes/Integration/Kubectl.cs
```diff
@@ -50,7 +50,7 @@ public bool TrySetKubectl()
                 ExecutableLocation = customKubectlExecutable;
             }
 
-            if (TryExecuteKubectlCommand("version", "--client", "--short"))
+            if (TryExecuteKubectlCommand("version", "--client", "--output=yaml"))
```
benPearce1, line 53: this is the only combination that doesn't print a deprecation notice

benPearce1 (approved):

YihaoOct:
Nice, I fixed the warning message in the health check itself in this PR: https://github.com/OctopusDeploy/OctopusDeploy/pull/18129

But I didn't notice these were used in the context wrapper as well. It's good to remove these and keep things consistent. 👍

IsaacCalligeros95:
> Nice, I fixed the warning message in the health check itself in this PR: [OctopusDeploy/OctopusDeploy#18129](https://github.com/OctopusDeploy/OctopusDeploy/pull/18129)
> 
> But I didn't notice these were used in the context wrapper as well. It's good to remove these and keep things consistent. 👍

Oh sweet, I missed that I was seeing it during deployments. Should we update the output type on one of these to json/yaml to keep them in sync? I'm not fussed which option we go with and happy to put up the PR for it

YihaoOct:
Nah, I think both are the same. They serve different purposes, the health check one checks the connection to the cluster, and the one in the context wrapper checks whether kubectl has been correctly set up at the execution location. They just happen to use the `version` command to do the checks. The logs are just side effect, as long as they don't go to stderr any format should be fine. So I'm happy to keep it as is now 😃

PR #1049, opened 2023-05-30
Add test which deploys multiple files with different glob paths

scottmerchant:
The new functionality of the Deploy Raw Yaml Command is that it can deploy multiple yaml files from a package at once using specified glob paths. This test uses a mixed bag of different types of glob paths with several different resource types to ensure they all deploy correctly and in the correct order according to the glob paths specified.

[sc-47324]


MissedTheMark (approved):
Quick review, but LGTM 👍

PR #1052, opened 2023-05-31
Fix structured variable replacement

YihaoOct:
[sc-46778]

When a YAML comment is inserted between a mapping key and its value, the YamlDotNet library does not handle the emission of YAML document correctly. Examples are:

```yaml
hello: # comment
  - world
```

becomes 

```yaml
hello # this is a comment:
- world
```

and

```yaml
hello:
  # comment
  - world
```

becomes

```yaml
hello
# this is a comment
:
- world
```

This is an issue with YamlDotNet itself, see: https://github.com/aaubry/YamlDotNet/issues/812.

We were aware of this issue when this piece of code was first written, we used a workaround of reversing the order of the comment node event and the mapping value node event, so that the YamlDotNet emitter emits the correct code.

However, this workaround was only applied to the inline comment case, not the block comment case. So this PR extends the fix to block comments as well.

This PR also bumps the YamlDotNet library to the latest version.

### DIFF
--- a/source/Calamari.Common/Calamari.Common.csproj
+++ b/source/Calamari.Common/Calamari.Common.csproj
@@ -65,7 +65,7 @@
         <PackageReference Include="Octostache" Version="3.3.0" />
         <PackageReference Include="SharpCompress" Version="0.24.0" />
         <PackageReference Include="XPath2" Version="1.0.12" />
-        <PackageReference Include="YamlDotNet" Version="8.1.2" />
+        <PackageReference Include="YamlDotNet" Version="13.1.0" />
         <PackageReference Include="System.ValueTuple" Version="4.4.0" />
     </ItemGroup>
 
--- a/source/Calamari.Common/Calamari.Common.csproj
+++ b/source/Calamari.Common/Calamari.Common.csproj
@@ -66,7 +66,6 @@
         <PackageReference Include="SharpCompress" Version="0.24.0" />
         <PackageReference Include="XPath2" Version="1.0.12" />
         <PackageReference Include="YamlDotNet" Version="13.1.0" />
-        <PackageReference Include="System.ValueTuple" Version="4.4.0" />
     </ItemGroup>
 
     <ItemGroup>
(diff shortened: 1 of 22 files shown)


source/Calamari.Common/Calamari.Common.csproj
```diff
@@ -65,7 +65,7 @@
         <PackageReference Include="Octostache" Version="3.3.0" />
         <PackageReference Include="SharpCompress" Version="0.24.0" />
         <PackageReference Include="XPath2" Version="1.0.12" />
-        <PackageReference Include="YamlDotNet" Version="8.1.2" />
+        <PackageReference Include="YamlDotNet" Version="13.1.0" />
```
YihaoOct, line 68: Bumping this to the latest version. 
To be more specific, we want to include this fix: https://github.com/aaubry/YamlDotNet/commit/c8a70f0eebf2889238dc35047b70705a1bf61ce3

Otherwise the output block comment will lose its indentation.

source/Calamari.Common/Features/StructuredVariables/YamlFormatVariableReplacer.cs
```diff
@@ -106,22 +107,24 @@ void LogReplacement(string key)
                                 && structureWeAreReplacing.Value.startEvent.Path == node.Path)
                             {
                                 outputEvents.AddRange(ParseFragment(structureWeAreReplacing.Value.replacementValue,
-                                                                    mappingStart.Event.Anchor,
-                                                                    mappingStart.Event.Tag));
+                                                                    mappingStart.Event.Anchor.Value,
```
YihaoOct, line 110: Typing adjustment due to version bump of the library

source/Calamari.Shared/Calamari.Shared.csproj
```diff
@@ -61,7 +61,6 @@
     <PackageReference Include="SharpCompress" Version="0.24.0" />
     <PackageReference Include="Sprache" Version="2.3.0" />
     <PackageReference Include="System.ValueTuple" Version="4.4.0" />
-    <PackageReference Include="YamlDotNet" Version="8.1.2" />
```
YihaoOct, line 64: Removed this reference here because it is already included ass a transitive dependency. Removing it here avoids ambiguity.

PR #1056, opened 2023-06-01
Enhance logging

YihaoOct:
This PR enhance the logging for Kubernetes Object Status to help customers understand deployment failures.

- We fail the step by producing an Error level message, so that when the step fails the reason shows directly on the Task Logs without having to expand the raw logs
- We add a list of failed or in-progress resources to the Verbose level log, in case we need to understand which resources went wrong

Screenshot 2023-06-06 at 1 40 46 PM

Screenshot 2023-06-02 at 4 06 46 PM


source/Calamari/Kubernetes/ResourceStatus/ResourceStatusReportExecutor.cs
```diff
@@ -78,12 +78,7 @@ public void ReportStatus(string workingDirectory, ICommandLineRunner commandLine
                 ? new InfiniteTimer(TimeSpan.FromSeconds(PollingIntervalSeconds)) as ITimer
                 : new Timer(TimeSpan.FromSeconds(timeoutSeconds), TimeSpan.FromSeconds(PollingIntervalSeconds));
 
-            var completedSuccessfully = statusChecker.CheckStatusUntilCompletionOrTimeout(definedResources, timer, kubectl, new Options() {  WaitForJobs = waitForJobs});
-
-            if (!completedSuccessfully)
-            {
-                throw new TimeoutException("Not all resources have deployed successfully within timeout");
```
johnsimons, line 85: We still need to throw the exception, otherwise the status code of Calamari wqill be zero right?

source/Calamari/Kubernetes/ResourceStatus/ResourceStatusReportWrapper.cs
```diff
@@ -57,14 +57,7 @@ public bool IsEnabled(ScriptSyntax syntax)
                 return result;
             }
 
-            try
-            {
-                statusReportExecutor.ReportStatus(workingDirectory, commandLineRunner, environmentVars);
-            }
-            catch (Exception e)
-            {
-                return new CommandResult("K8s Resource Status Reporting", 1, e.Message, workingDirectory);
```
johnsimons, line 66: We need this to set the status code, otheriwse Calamari will report 0 as status code and that is assumed to have succceded in Server

YihaoOct: Good catch. Thanks! Looks like without a proper exit code the status is "succeed with warnings". I will fix this soon

source/Calamari/Kubernetes/ResourceStatus/ResourceStatusReportExecutor.cs
YihaoOct: Fixed this. But I think it is better to communicate whether it has been successful using return value rather than exceptions. So I made this method to return a `bool`.

source/Calamari/Kubernetes/Conventions/ResourceStatusReportConvention.cs
```diff
@@ -18,7 +19,11 @@ public ResourceStatusReportConvention(ResourceStatusReportExecutor statusReportE
 
         public void Install(RunningDeployment deployment)
         {
-            statusReportExecutor.ReportStatus(deployment.CurrentDirectory, commandLineRunner, deployment.EnvironmentVariables);
+            var successful = statusReportExecutor.ReportStatus(deployment.CurrentDirectory, commandLineRunner, deployment.EnvironmentVariables);
+            if (!successful)
+            {
+                throw new TimeoutException("Not all resources have deployed successfully within timeout");
+            }
```
YihaoOct, line 26: I think it is better to communicate the result of the status check execution via the return value rather than an exception. But I noticed that the exception is necessary here for it to work via conventions, so I moved the previous exception here. @scottmerchant please call out if there are something I missed here, thanks

scottmerchant: Just FYI some of this is going to conflict with my changes. Can we hold off on merging this? I'll help resolve the conflicts after I've merged my things but it might take a few days.

scottmerchant: Main reason is I'm going on leave soon and I really want to get my stuff in before I go... 🙏

YihaoOct: Thanks @scottmerchant for the info. No problems, but I wonder if the conflict is from changing from exception to return value? If so, will it prevent the conflict if I change this back to the existing mechanism, just add more logs?

scottmerchant: Yeah if you kept it the same with just some extra logs it would be ok for now. We can do the change to use return values later if you want. In the new future I'm making changes to remove the `ResourceStatusReportConvention` because it won't be executed via an Install convention then...

scottmerchant: But I would still hold off merging for the moment please.

YihaoOct: Thanks, I will hold off merging. In the meantime, since refactoring the return value isn't the goal for this PR, I will just revert that part and keep the changes only to logging.

YihaoOct: I've reverted that part of refactoring. I will add you as a reviewer to this PR and let me know when you think this PR is good to be merged @scottmerchant

source/Calamari/Kubernetes/ResourceStatus/ResourceStatusReportExecutor.cs
```diff
@@ -59,7 +59,6 @@ public void ReportStatus(string workingDirectory, ICommandLineRunner commandLine
             if (!definedResources.Any())
             {
                 log.Verbose("No defined resources are found, skipping resource status check");
-                return;
```
johnsimons, line 62: why did we remove this ?

YihaoOct: Sorry I forgot to remove update this in the back-and-forth changes regarding the return value. Fixed

johnsimons (approved):
LGTM

PR #1057, opened 2023-06-01
Fix structured variable replacement

YihaoOct:
[sc-46778]

When a YAML comment is inserted between a mapping key and its value, the YamlDotNet library does not handle the emission of YAML document correctly. Examples are:

```yaml
hello: # comment
  - world
 ```

becomes

```yaml
hello # this is a comment:
- world
```

and

```yaml
hello:
  # comment
  - world
```

becomes

```yaml
hello
# this is a comment
:
- world
```

This is an issue with YamlDotNet itself, see: https://github.com/aaubry/YamlDotNet/issues/812.

We were aware of this issue when this piece of code was first written, we used a workaround of reversing the order of the comment node event and the mapping value node event, so that the YamlDotNet emitter emits the correct code.

However, this workaround was only applied to the inline comment case, not the block comment case. So this PR extends the fix to block comments as well.


johnsimons (approved):
Nice work @YihaoOct

PR #1058, opened 2023-06-02
Remove ReleaseNotes from PackCalamariConsolidatedNugetPackage

scottmerchant:
This should fix the build after the ReleaseNotes snapshot dependency and artifact dependency were removed from the team city build.


YihaoOct (approved):
LGTM

PR #1059, opened 2023-06-04
Update object status to execute in parallel

scme0:
Updating the Kubernetes Raw Yaml Command to send resources to the `ResourceStatusReportExecutor` after each batch of yaml has been applied.

This means that status updates will appear more quickly than if we wait to apply all batches before starting to report Resource Statuses.


source/Calamari/Kubernetes/ResourceStatus/ResourceStatusChecker.cs
```diff
@@ -30,75 +41,144 @@ public class ResourceStatusChecker : IResourceStatusChecker
...
+                return result;
+            });
+        }
+
+        public void AddResources(IEnumerable<ResourceIdentifier> newResources)
+        {
+            var newResourcesList = newResources.ToList();
+            log.Info($"Resource Status Check: {newResourcesList.Count} new resources have been added.");
+            AddResourcesInternal(newResourcesList);
+            timer?.Restart();
```
scottmerchant, line 146: Main thing to note out is that the way this works is that the timer will start when you start the `ResourceStatusReportExecutor` but it will be restarted everytime you add additional resources. So if you start the executor and take too long to add more it will time out. Unsure on exactly what that would look like but not sure if that's really a problem?

I could be worth trying to setup something that makes applying some yaml take a long time.. but I'm not sure if that's possible.

source/Calamari/Kubernetes/ResourceStatus/ResourceStatusChecker.cs
```diff
@@ -30,75 +41,144 @@ public class ResourceStatusChecker : IResourceStatusChecker
         internal const string MessageDeploymentSucceeded = "Resource status check completed successfully because all resources are deployed successfully";
         internal const string MessageDeploymentFailed = "Resource status check terminated with errors because some resources have failed";
         internal const string MessageInProgressAtTheEndOfTimeout = "Resource status check terminated because the timeout has been reached but some resources are still in progress";
-        
+
         private readonly IResourceRetriever resourceRetriever;
         private readonly IResourceUpdateReporter reporter;
         private readonly ILog log;
+        private ITimer timer;
+        private readonly HashSet<ResourceIdentifier> resources = new HashSet<ResourceIdentifier>();
```
scottmerchant, line 49: `HashSet<>` is used to enforce uniqueness. I've also added equality operators to ResourceIdentifier because we can't use `record`s here. 😞

source/Calamari/Kubernetes/ResourceStatus/ResourceStatusChecker.cs
```diff
@@ -30,75 +41,144 @@ public class ResourceStatusChecker : IResourceStatusChecker
...
-            }
-            while (!timer.HasCompleted() && deploymentStatus == DeploymentStatus.InProgress);
+                var result = new ResourceStatusTaskResult();
+                do
+                {
+                    ResourceIdentifier[] definedResources;
+                    bool stillWaitingForMoreResources;
+                    lock (resourceLock)
+                    {
+                        stillWaitingForMoreResources = waitingForMoreResources;
```
scottmerchant, line 108: Set the field value of `waitingForMoreResources` to a local value here so that we ensure that we get at least 1 pass of checks before it completes as otherwise it may miss some resources if the timing is just right.

source/Calamari/Kubernetes/ResourceStatus/ResourceStatusChecker.cs
```diff
@@ -112,7 +192,18 @@ private static DeploymentStatus GetDeploymentStatus(List<Resource> resources, Li
...
         }
 
-        private void LogInProgressResources(List<Resource> definedResourceStatuses, Dictionary<string, Resource> resourceStatuses, List<ResourceIdentifier> definedResources)
+        private void AddResourcesInternal(IEnumerable<ResourceIdentifier> newResources)
+        {
+            foreach (var resource in newResources)
+            {
+                lock (resourceLock)
+                {
+                    resources.Add(resource);
```
scottmerchant, line 201: There is no `AddRange` extension method for `HashSet`.. at least that I could see. 😢

source/Calamari/Kubernetes/ResourceStatus/ResourceStatusReportExecutor.cs
```diff
@@ -34,16 +36,25 @@ public class ResourceStatusReportExecutor
             this.kubectl = kubectl;
         }
 
-        public void ReportStatus(string workingDirectory)
+        public bool ReportStatus(string workingDirectory)
```
scottmerchant, line 39: Non-async method is for ScriptWrapper usage.

source/Calamari/Program.cs
```diff
@@ -67,10 +67,12 @@ protected override void ConfigureContainer(ContainerBuilder builder, CommonOptio
             builder.RegisterType<CalamariCertificateStore>().As<ICertificateStore>().SingleInstance();
             builder.RegisterType<DeploymentJournalWriter>().As<IDeploymentJournalWriter>().SingleInstance();
             builder.RegisterType<PackageStore>().As<IPackageStore>().SingleInstance();
+#if !NET40
```
scottmerchant, line 70: Just not compiling stuff that can never be used by .net 4.0 versions of Calamari as it means we can reference Tasks/async without it failing on the .net 4.0 build.

YihaoOct (changes requested):
Nice work on making the status checker able to run in parallel 👍 This will be very helpful when we begin to implement the checks for Helm.

Now the status checker keeps a separate thread, and has various statuses to maintain. It requires the caller to invoke the methods in a particular sequence, otherwise a runtime error is thrown (e.g. wait for completion on a checker that hasn't been started). That requires some additional knowledge for the caller. I think we can perhaps make it clearer and safer by _"making invalid states unrepresentable"_. Here is some pseudo code for what I am thinking:

```
class StatusChecker:
  Start(ResourceIdentifiers[]) -> TimedChecker
  StartAndWaitForResources() -> WaitForResourcesChecker

class WaitForResourcesChecker:
  AddResources(ResourceIdentifier[]) -> WaitForResourcesChecker
  CompleteAddingResources() -> TimedChecker

class TimedChecker:
  WaitForCompletionOrTimeout() -> Result
```

Here each class only contains valid methods, and returns a different type. So it won't get into an invalid state. In this implementation, the `AddResources` method of `WaitForResourcesChecker` cancels the current checking task, and creates a new started `WaitForResourcesChecker` that checks for a larger resource list, making the resources field immutable.

Usage when no more resources are needed:
```
var result = new StatusChecker(resources).Start().WaitForCompletionOrTimeout();
```

Usage when new resources are required:
```
var checker = new StatusChecker(initialResources).StartAndWaitForResources();
foreach (var rs in additionalResourceLists)
{
  checker = checker.AddResources(rs);
}
var result = checker.CompleteAddingResources().WaitForCompletionAndTimeout();
```

source/Calamari/Kubernetes/ResourceStatus/Resources/ResourceIdentifier.cs
```diff
@@ -1,14 +1,15 @@
+using System;
+
 namespace Calamari.Kubernetes.ResourceStatus.Resources
 {
    /// <summary>
    /// Identifies a unique resource in a kubernetes cluster
    /// </summary>
-   public class ResourceIdentifier
+   public class ResourceIdentifier : IEquatable<ResourceIdentifier>
```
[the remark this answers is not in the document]
scottmerchant: Great point! I forgot about structs 🤦‍♂️

source/Calamari/Kubernetes/ResourceStatus/ResourceStatusChecker.cs
```diff
@@ -30,75 +41,144 @@ public class ResourceStatusChecker : IResourceStatusChecker
...
+
         private readonly IResourceRetriever resourceRetriever;
         private readonly IResourceUpdateReporter reporter;
         private readonly ILog log;
+        private ITimer timer;
+        private readonly HashSet<ResourceIdentifier> resources = new HashSet<ResourceIdentifier>();
+        private readonly object resourceLock = new object();
+        private bool isCheckingStatus;
+        private bool waitingForMoreResources;
+        private Task<ResourceStatusTaskResult> statusCheckTask;
```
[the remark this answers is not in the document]
scottmerchant: I was going to say "boolean operations are atomic" but now that I search for a source it looks like that doesn't mean they are thread safe: https://stackoverflow.com/questions/29411961/c-sharp-and-thread-safety-of-a-bool

I'll follow up on this (and also the `Task<ResourceStatusTaskResult>`). Good pickup!

source/Calamari/Kubernetes/ResourceStatus/ResourceStatusChecker.cs
scottmerchant: Yeah fair enough - I wasn't sure if the timeout meant "after everything is applied" or "for the whole application process". Do you think it's worth discussing this in the channel with the team?

YihaoOct: Yeah, I think this is a problem that needs input from product and design

scottmerchant:
@YihaoOct 
Thanks for your suggestion on restructure. I like it so I'll give it a go and let you know when you can have another look 👍

source/Calamari/Kubernetes/ResourceStatus/ResourceStatusChecker.cs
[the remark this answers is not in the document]
YihaoOct: Yeah, I think perhaps the best way is just not to share them at all. Every time we change state or add more resources, we simply cancel the current checker and create a new checker. See the review comments for the idea. We can still guarantee the total time by reusing the started Timer instance.

source/Calamari/Kubernetes/ResourceStatus/ResourceFinder.cs
```diff
@@ -0,0 +1,116 @@
...
+        private readonly IVariables variables;
+        private readonly ICalamariFileSystem fileSystem;
+
+        public ResourceFinder(IVariables variables, ICalamariFileSystem fileSystem)
+        {
+            this.variables = variables;
+            this.fileSystem = fileSystem;
+        }
+
+        public IEnumerable<ResourceIdentifier> FindResources(string workingDirectory)
```
scme0, line 21: This code is used only by the script wrapper to get resources from files so I've moved it to a separate class so we don't have to worry about it for the Raw Yaml Command.

source/Calamari/Kubernetes/ResourceStatus/ResourceStatusReportExecutor.cs
```diff
@@ -1,160 +1,48 @@
...
-            var defaultNamespace = variables.Get(SpecialVariables.Namespace, "default");
-            // When the namespace on a target was set and then cleared, it's going to be "" instead of null
-            if (string.IsNullOrEmpty(defaultNamespace))
-            {
-                defaultNamespace = "default";
-            }
-            var timeoutSeconds = variables.GetInt32(SpecialVariables.Timeout) ?? 0;
-            var waitForJobs = variables.GetFlag(SpecialVariables.WaitForJobs);
-
-            var manifests = ReadManifestFiles(workingDirectory).ToList();
```
scme0, line 48: All this code that gets resources from files has been moved to the `ResourceRetriever` class.

scme0:
@YihaoOct Apologies in advance for making the review bigger with my most recent changes. I think it's looking better now, do you want to have another look?

source/Calamari/Kubernetes/ResourceStatus/RunningResourceStatusCheck.cs
```diff
@@ -0,0 +1,158 @@
...
+            else
+            {
+                log.Verbose("Resource Status Check: Waiting for resources to be applied.");
+            }
+            statusCheckTask = RunNewStatusCheck(initialResources);
+        }
+
+        public async Task<bool> WaitForCompletionOrTimeout()
+        {
+            await taskLock.WaitAsync();
```
YihaoOct, line 69: 👍

source/Calamari/Kubernetes/ResourceStatus/Timer.cs
```diff
@@ -9,43 +10,23 @@ namespace Calamari.Kubernetes.ResourceStatus
...
-        public Timer(TimeSpan duration, TimeSpan interval)
-        {
-            this.duration = duration;
-            this.interval = interval;
-            stopwatch = new Stopwatch();
-        }
-
+        private readonly Stopwatch stopwatch = new Stopwatch();
+        public TimeSpan Interval { get; set; } = TimeSpan.FromSeconds(1);
+        public TimeSpan Duration { get; set; } = Timeout.InfiniteTimeSpan;
```
YihaoOct, line 25: If possible, I still think it is better to have these in the constructor to enforce consistency. The default values are a little bit unobvious, eg. we'll never actually use the default interval I think. Good work on `Timeout.InfiniteTimeSpan` though, I didn't know about it.

YihaoOct (approved):
Looks pretty neat 👍 It's good from my perspective as long as the team has agreed on the interpretation of execution timeout value.

source/Calamari/Kubernetes/ResourceStatus/Timer.cs
scme0: Good point - I've updated this to use Delegate Factories and I think it looks pretty neat.

scme0: I also just realised that the `WaitForInterval` method is using `Thread.Sleep`. I've updated it to use `Task.Delay` (as an async method) because it is generally advised to avoid `Thread.Sleep` and we're in an async context when we call the method anyway.

scme0:
@YihaoOct I think from the [thread with the team](https://octopusdeploy.slack.com/archives/C04V68JCM1B/p1687239482634189) it looks like we're going with the timer restarting for each update with new resources. This is currently how it's implemented as a new timer is created each time a new KOS task is created so as long as the build goes green I think we are good to merge.

PR #1064, opened 2023-06-14
Make sure we don't check for config maps and secrets that do not exist

YihaoOct:
When a Kubernetes user uses the Deploy a Kubernetes Container step, and have the ConfigMap and/or Secret features enabled (which is the default), if the user does not actually include the config map or secret (leaving the configuration section empty), the object status check will pass successfully, because it attempts to find the config map or secret that was never created.

Screenshot 2023-06-14 at 3 47 59 PM

This PR adds some checks that are equivalent to the counterpart in the script which actually does the deployment: https://github.com/OctopusDeploy/OctopusDeploy/blob/d30286031e3970f6f468cdc12a2d2526b5aa0e98/source/Octopus.Kubernetes/ActionHandler/Scripts/KubernetesDeployment.sh#L124-L128

After the change we no longer attempt to do checks on configmaps or secrets that have not been created.

Screenshot 2023-06-14 at 3 46 07 PM

### DIFF
--- a/build/Build.cs
+++ b/build/Build.cs
@@ -478,15 +478,12 @@ namespace Calamari.Build
         Target PackCalamariConsolidatedNugetPackage =>
             _ => _.DependsOn(PackageConsolidatedCalamariZip)
                   .Executes(() =>
-                            {
-                                var releaseNotes = IsLocalBuild ? "Local" : File.ReadAllText(RootDirectory / "releasenotes" / "ReleaseNotes.md");
-                                
-                                NuGetPack(s => s.SetTargetPath(BuildDirectory / "Calamari.Consolidated.nuspec")
-                                            .SetProperty("releaseNotes", releaseNotes)
-                                            .SetBasePath(BuildDirectory)
-                                            .SetVersion(NugetVersion.Value)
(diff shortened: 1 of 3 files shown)


build/Build.cs
```diff
@@ -478,15 +478,12 @@ void CompressCalamariProject(Project project)
...
-                                            .SetBasePath(BuildDirectory)
-                                            .SetVersion(NugetVersion.Value)
-                                            .SetOutputDirectory(ArtifactsDirectory));
-                            });
+                  {
+                      NuGetPack(s => s.SetTargetPath(BuildDirectory / "Calamari.Consolidated.nuspec")
+                          .SetBasePath(BuildDirectory)
+                          .SetVersion(NugetVersion.Value)
+                          .SetOutputDirectory(ArtifactsDirectory));
+                  });
```
YihaoOct, line 486: Need to back-port this so it builds properly on TC. See: https://github.com/OctopusDeploy/Calamari/pull/1058

johnsimons (approved):
LGTM, but it would be nice to have some kind of test @YihaoOct ?

PR #1066, opened 2023-06-22
Remove unused env var: SSH_OctopusAPITester_PrivateKey

akirayamamoto:
I noticed a reference to an environment variable that is not used: `SSH_OctopusAPITester_PrivateKey`.

This PR removes such variable.


scme0 (approved):
Happy with this as long as the build goes green 👍

PR #1068, opened 2023-06-26
Support authentication to private GKE clusters

APErebus:
Adds support for new special variable `Octopus.Action.Kubernetes.GkeUseClusterInternalIp` which is used to set the `--internal-ip` flag during the GCloud authentication call.

This allows for private clusters to be authenticated against as they do not have a public IP address

Fixes https://github.com/OctopusDeploy/Issues/issues/8225

Shortcut story: [sc-51633]


scme0 (approved):
Looks good 🚀

PR #1069, opened 2023-06-26
Filter service messages when checking for command result errors

mik-ky:
There was a recent change related to send masking messages to stderr to fix sensitive variables leaking - https://github.com/OctopusDeploy/Calamari/pull/1008.

This, however, introduced a bug which caused any bash scripts to fail when `Octopus.Action.FailScriptOnErrorOutput` was set to `true`, since as part of the bootstrapping, errors were being sent to `stderr`. This change filters out service messages to determine whether a script has any errors.

[sc-52118]

### DIFF
--- a/source/Calamari.Common/Features/Processes/CommandResult.cs
+++ b/source/Calamari.Common/Features/Processes/CommandResult.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Linq;
 
 namespace Calamari.Common.Features.Processes
 {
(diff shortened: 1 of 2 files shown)


benPearce1 (approved):

source/Calamari.Common/Features/Processes/CommandResult.cs
```diff
@@ -19,7 +17,7 @@ public CommandResult(string command, int exitCode, string? additionalErrors = nu
 
         public string? Errors { get; }
 
-        public bool HasErrors => !string.IsNullOrWhiteSpace(Errors);
+        public bool HasErrors => !string.IsNullOrWhiteSpace(Errors) && !IsServerMessage(Errors);
```
benPearce1, line 20: Nit: this could do with a test

PR #1071, opened 2023-06-28
Handle child steps naming differently

johnsimons:
In this PR we are handling child steps a bit differently.
We are checking whether the current executing step is a child step or not by looking at the `Octopus.Action.Number` variable, this will give us a whole number eg 1, 3, 4 if it is a normal step but if it is a child step the value will be a fraction number eg 1.2, 1.3, 1.5.
So we use this distinction to select a different code path to build the step name.
For a normal step, we use `Octopus.Step.Number` concatenated with `Octopus.Step.Name` but for a child step we use `Octopus.Action.Number` concatenated with `Octopus.Action.Name`

Fixes https://github.com/OctopusDeploy/Issues/issues/8219

### DIFF
--- a/source/Calamari/Kubernetes/ResourceStatus/ResourceUpdateReporter.cs
+++ b/source/Calamari/Kubernetes/ResourceStatus/ResourceUpdateReporter.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using Calamari.Common.Plumbing.Logging;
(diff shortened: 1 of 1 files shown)


source/Calamari/Kubernetes/ResourceStatus/ResourceUpdateReporter.cs
```diff
@@ -64,11 +65,21 @@ private static IEnumerable<Resource> GetRemovedResources(IDictionary<string, Res
...
         {
+            var actionNumber = variables.Get("Octopus.Action.Number", string.Empty);
+            var stepNumber = variables.Get("Octopus.Step.Number");
+            var stepName = variables.Get("Octopus.Step.Name");
+            
+            if (actionNumber.IndexOf(".", StringComparison.Ordinal) > 0)
+            {
+                stepNumber = actionNumber;
+                stepName = variables.Get("Octopus.Action.Name");
+            }
```
YihaoOct, line 76: Looks good, but it seems if there is only one action under a step, then the action number and action name will just be the step number and step name, so perhaps we can directly use those. See in the screenshot that the action name and step name, action number and step number are the same

Screenshot 2023-07-03 at 5 21 13 PM

YihaoOct (approved):
LGTM, a comment

source/Calamari/Kubernetes/ResourceStatus/ResourceUpdateReporter.cs
johnsimons: So we are only changing the naming behaviour for child steps, so I am not sure I understand what you saying @YihaoOct ?

YihaoOct: I mean, why not directly use `Octopus.Action.Number` and `Octopus.Action.Name` instead of those for steps for all?

johnsimons: Ah, because `Octopus.Action.Name` may not be the same as `Octopus.Step.Name`.
Via the API, they can be different 😉

johnsimons: We want those names to be the same as when the users are looking at the process itself.

PR #1072, opened 2023-06-29
Add support for WaitForDebugger to CalamariFlavourProgramAsync

APErebus:
When discussing how to debug calamari locally with @scme0, we realised that the async calamari don't support the `Octopus.Calamari.WaitForDebugger` variable.

This PR just adds the same loop as is in the non-async program.


scme0 (approved):
Nice! thanks for doing that 👍

PR #1073, opened 2023-06-29
Check if KOS is enabled before we do status check from the command

YihaoOct:
This PR addresses https://github.com/OctopusDeploy/Issues/issues/8236.

We did not check if we should run status checks if the status check is done via a Raw YAML command. This PR adds this check.

See discussions in: https://octopusdeploy.slack.com/archives/C03SG0LFJHX/p1687989153045139 (internal link).

### DIFF
--- a/source/Calamari.Tests/KubernetesFixtures/Commands/KubernetesApplyRawYamlCommandFixture.cs
+++ b/source/Calamari.Tests/KubernetesFixtures/Commands/KubernetesApplyRawYamlCommandFixture.cs
@@ -37,7 +37,7 @@ namespace Calamari.Tests.KubernetesFixtures.Commands
         }
         
         [Test]
-        public void WhenResourceStatusIsEnabled_ShouldNotRunStatusChecks()
+        public void WhenResourceStatusIsEnabled_ShouldRunStatusChecks()
         {
             var variables = new CalamariVariables()
             {
--- a/source/Calamari.Tests/KubernetesFixtures/Commands/KubernetesApplyRawYamlCommandFixture.cs
+++ b/source/Calamari.Tests/KubernetesFixtures/Commands/KubernetesApplyRawYamlCommandFixture.cs
@@ -1,0 +1,74 @@
+using Calamari.Common.Features.Packages;
+using Calamari.Common.Features.Processes;
+using Calamari.Common.Features.StructuredVariables;
+using Calamari.Common.Features.Substitutions;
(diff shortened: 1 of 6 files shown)


source/Calamari/Kubernetes/Commands/KubernetesApplyRawYamlCommand.cs
```diff
@@ -58,9 +61,11 @@ public override int Execute(string[] commandLineArguments)
 
         protected override async Task<bool> ExecuteCommand(RunningDeployment runningDeployment)
         {
-            var gatherAndApplyRawYamlExecutor =
-                new GatherAndApplyRawYamlExecutor(log, fileSystem, kubectl);
-
+            if (!variables.GetFlag(SpecialVariables.ResourceStatusCheck))
```
YihaoOct, line 64: We do not check `DeploymentStyle` and `DeploymentWait` here, although we do this in the [script wrapper](https://github.com/OctopusDeploy/Calamari/blob/1f6e2a6b523eb0a7259d7f47d231d54cbda74636/source/Calamari/Kubernetes/ResourceStatus/ResourceStatusReportWrapper.cs#L38-L44). This is because these settings are only for Deploy-a-container step. We don't have these settings for RAW YAML.

source/Calamari.Tests/KubernetesFixtures/Commands/KubernetesApplyRawYamlCommandFixture.cs
```diff
@@ -0,0 +1,74 @@
...
+            var resourceStatusCheck = Substitute.For<IResourceStatusReportExecutor>();
+            var command = CreateCommand(variables, resourceStatusCheck);
+            
+            command.Execute(new string[]{ });
+
+            resourceStatusCheck.ReceivedCalls().Should().BeEmpty();
+        }
+        
+        [Test]
+        public void WhenResourceStatusIsEnabled_ShouldNotRunStatusChecks()
```
scme0, line 40:
```suggestion
        public void WhenResourceStatusIsEnabled_ShouldRunStatusChecks()
```

source/Calamari/Kubernetes/Commands/KubernetesApplyRawYamlCommand.cs
scme0: Nice and simple. Good work!

scme0 (approved):
Looks good! 🚀 Just a typo in your test method.

PR #1076, opened 2023-07-03
Handle child steps naming differently

johnsimons:
In this PR we are handling child steps a bit differently.
We are checking whether the current executing step is a child step or not by looking at the `Octopus.Action.Number` variable, this will give us a whole number eg 1, 3, 4 if it is a normal step but if it is a child step the value will be a fraction number eg 1.2, 1.3, 1.5.
So we use this distinction to select a different code path to build the step name.
For a normal step, we use `Octopus.Step.Number` concatenated with `Octopus.Step.Name` but for a child step we use `Octopus.Action.Number` concatenated with `Octopus.Action.Name`

Fixes https://github.com/OctopusDeploy/Issues/issues/8219

### DIFF
--- a/source/Calamari/Kubernetes/ResourceStatus/ResourceUpdateReporter.cs
+++ b/source/Calamari/Kubernetes/ResourceStatus/ResourceUpdateReporter.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


YihaoOct (approved):

johnsimons:
@YihaoOct can you also tick this one, it is just the same PR to apply to `master`

PR #1079, opened 2023-07-05
Fix for RequiresDotNetCore attribute

hnrkndrssn:
The `NETCORE` constant used by this attribute haven't been defined in the `csproj` file, this PR adds this definition.

[sc-53301]


veochen-octopus (approved):
LG 👍

PR #1080, opened 2023-07-06
Update Raw Yaml Step to use existing Glob Implementation

scme0:
Simply updating the Raw Yaml Step to use the existing Glob implementation so that it aligns with the implementation used by Variable Substitution etc.

There are some limitations with the Glob implementation where it basically just supports wildcards (eg: ** and *).

At this stage it's the only option without a lot of rework and testing.


MissedTheMark (approved):
Looks good to me.  Might be nice to add some tests around the glob stuff here later.

PR #1081, opened 2023-07-07
Add groups support to Glob implementation

scme0:
This PR adds back in support for groups (`{...,...}`, `[...-...]`, and `[...]`) in glob patterns.

See this issue: https://github.com/OctopusDeploy/Issues/issues/3320 for more info about the backwards compatibility issues we were resolving by removing the grouping functionality.

The way this implementation maintains backwards compatibility is that:
1. for `{}`, if it does not contain a comma, it searches only for the literal. eg: `My/Folder/Has/{Brackets}/config.yaml` will **only** resolve to that exact path.
2. for `[]` if it does not contain a `-`, it searches for all group options AND the literal. eg: `My/Folder/[Has]/Brackets/config.yaml` will search for:
`My/Folder/[Has]/Brackets/config.yaml`
`My/Folder/H/Brackets/config.yaml`
`My/Folder/a/Brackets/config.yaml`
`My/Folder/s/Brackets/config.yaml`
3. for any `[]{}` that span over a directory separator or intersect with other brackets (eg: `[My/Cool]/Path.config` and `My/C[o{o]l}/Path.config`), the brackets and their contents will only be taken as literals.
4. for any `[]{}` that don't have an equivalent opening or closing bracket (eg: `My/[Cool/Path.config`) will only be taken as literals.

I feel like this is a good balance between having new glob functionality and maintaining compatibility for those with brackets in their paths. If we want, we can phase out the bracket support and introduce escaping support to allow users to escape brackets to make them literals. I would suggest we do this later though and this is a good first step in that direction.

With this change, I don't think there is much chance of a user having a folder with brackets and also having folders with a single character so it should be fine. **I'm interested in getting opinions on this**.

We could also role this out with a feature toggle so that it's easy to rollback if a customer has issues.

[sc-52786]


APErebus:
> We could also role this out with a feature toggle so that it's easy to rollback if a customer has issues.

This seems eminently sensible.

source/Calamari.Common/Plumbing/FileSystem/GlobExpressions/CharGroupRetriever.cs
```diff
@@ -0,0 +1,95 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace Calamari.Common.Plumbing.FileSystem.GlobExpressions
+{
+    public static class CharGroupRetriever
```
APErebus, line 8: Is it worth having a set of tests _just_ for this specific logic? (rather than the overall result in `CalamariPhysicalFileSystemFixture`).

source/Calamari.Common/Plumbing/FileSystem/GlobExpressions/StringGroupRetriever.cs
```diff
@@ -0,0 +1,73 @@
+using System.Collections.Generic;
+using System.IO;
+
+namespace Calamari.Common.Plumbing.FileSystem.GlobExpressions
+{
+    public static class StringGroupRetriever
```
APErebus, line 6: same re. unit tests

source/Calamari.Tests/Fixtures/Integration/FileSystem/CalamariPhysicalFileSytemFixture.cs
APErebus, line 1: Nit on this filename `CalamariPhysicalFileSytemFixture.cs` -> `CalamariPhysicalFileSystemFixture.cs`

source/Calamari.Common/Plumbing/FileSystem/GlobExpressions/Ungrouper.cs
```diff
@@ -0,0 +1,49 @@
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Calamari.Common.Plumbing.FileSystem.GlobExpressions
+{
+    public static class Ungrouper
```
APErebus, line 6: The name of this class is bothering me. I don't have a great suggestion, but it feels a bit ambiguous/generic

APErebus (commented):
Logic looks good to me. 👍 

I do think a feature flag might make the most sense, just in case we have reports from customers with weird path structures

source/Calamari.Common/Plumbing/FileSystem/GlobExpressions/Ungrouper.cs
scme0: I agree. What about `GlobExpressionGroupResolver`?

source/Calamari.Tests/Fixtures/Integration/FileSystem/CalamariPhysicalFileSytemFixture.cs
scme0: Good catch, thanks.

source/Calamari.Common/Plumbing/FileSystem/GlobExpressions/CharGroupRetriever.cs
scme0: Yep that's a fair call. I'll put some in.

source/Calamari.Common/Plumbing/FileSystem/GlobExpressions/Ungrouper.cs
APErebus: Yeah, that's a better name 👍

scme0:
@APErebus Unfortunately adding a FeatureToggle for this was a bit of a pain in the ass in that I had to move the FeatureToggles enum from Calamari.Shared to Calamari.Common which meant I had to update all the references to a different namespace. This has blown out the number of files changes, so sorry about that!

APErebus:
> @APErebus Unfortunately adding a FeatureToggle for this was a bit of a pain in the ass in that I had to move the FeatureToggles enum from Calamari.Shared to Calamari.Common which meant I had to update all the references to a different namespace. This has blown out the number of files changes, so sorry about that!

All good, I'll take another look :)

source/Calamari.Tests/Fixtures/Integration/FileSystem/GlobExpression/CharGroupRetrieverFixture.cs
```diff
@@ -0,0 +1,54 @@
...
+        [TestCase("Her[e/h]ello.txt", "")] // [] separated by directory separator
+        [TestCase("Here/[te{s]t,mp}.txt", "")] // [] intersects with {}
+
+        // Returns groups
+        [TestCase("Here/[hi].txt", "{startIndex:5,length:4,options:['h','i','[hi]']}")] // Basic char group, will also return literal expression.
+        [TestCase("Here/[a-c].txt", "{startIndex:5,length:5,options:['a','b','c']}")] // Range char Group
+        [TestCase("This/I[st]/Fi[rn]e.png",
+            "{startIndex:6,length:4,options:['s','t','[st]']};" +
+            "{startIndex:13,length:4,options:['r','n','[rn]']}")] // Two char groups
+        [TestCase("Here/[?*.].txt", "{startIndex:5,length:5,options:['?','*','.','[?*.]']}")] // Other characters
```
APErebus, line 27: It's a minor nit but by putting a JSON string here, you lose the connection to the code model definition.. Did you considering something like `[TestCaseSource]` and having that return pre-defined `Group` objects.

The tests are really good, i'm just concerned having the JSON strings are your test cases

scme0: My aim with using JSON like this was mainly brevity. I also find it a bit more readable compared to using a TestCaseSource implementation.

I've updated the code to use a TestCaseSource but to my eye it's harder to parse quickly because of the required boilerplate in C# compared to writing the class in json.

Example:
image

I agree it disconnects a bit from the `Group` model and won't show compile time errors if someone updates the model which is a pain. I'm happy to concede here but I wish there were a better, cleaner option!

APErebus: Yeah, I can understand where you are coming from. I think it's ok to lose a bit of brevity to gain the compiler support.
There could be some extra stuff we could do to make the test cases themselves easier to read if you want to play with it? I'm not fussed at this point

APErebus (approved):
Looks really good mate. I appreciate you working through my comments and thoughts.

Let's get it tested and get it in

source/Calamari.Tests/Fixtures/Integration/FileSystem/GlobExpression/CharGroupRetrieverFixture.cs
scme0: In the end I agree and have come around because of the case where I would prefer the compiler complain than to find out tests have broken after the fact.

scme0:
As discussed, during testing I found a weird edge case with the old functionality where if a `{}` group was added *after* a `*`, it would be evaluated correctly. We decided that instead of trying to incorporate this bizarre edge case into the new version of `Glob`, that we would bring back the old version and rename it to `LegacyGlob`. This means the original functionality is untouched and the new implementation can not be concerned with replicated the old behaviour.

I've update Calamari File System to call the appropriate implementation of `Glob` based on the `GlobMode` enum passed in. It defaults to `LegacyMode` if an invalid value is given.

You've heard of `GodMode`... well we have `GlobMode`!

scme0:
I've tested the above code and it now works as I expect with the edge case intact when the feature toggle is switched off 👍

source/Calamari.Common/Plumbing/FileSystem/LegacyGlob.cs
```diff
@@ -514,16 +515,4 @@ public bool IsMatch(string input)
...
-    {
-        internal static IEnumerable<TSource> DistinctBy<TSource, TKey>(this IEnumerable<TSource> source,
-            Func<TSource, TKey> keySelector)
-        {
-            var knownKeys = new HashSet<TKey>();
-            foreach (var element in source)
-                if (knownKeys.Add(keySelector(element)))
-                    yield return element;
-        }
-    }
```
scme0, line 528: This bad boy has been moved to it's own file.

source/Calamari.Tests/Fixtures/Integration/FileSystem/CalamariPhysicalFileSystemFixture.cs
```diff
@@ -132,14 +130,29 @@ string CreateFile(params string[] relativePath)
...
         [TestCase(@"Config/*.config", "c.config")]
-        [TestCase(@"Config/Feature1/*.config", "f1-a.config", 2)]
-        [TestCase(@"Config/Feature1/*.config", "f1-b.config", 2)]
+        [TestCase(@"Config/Feature1/*.config", "f1-a.config", 3)]
+        [TestCase(@"Config/Feature1/*.config", "f1-b.config", 3)]
+        [TestCase(@"Config/Feature1/*.config", "f1-c.config", 3)]
         [TestCase(@"Config/Feature2/*.config", "f2.config")]
-        public void EnumerateFilesWithGlob(string pattern, string expectedFileMatchName, int expectedQty = 1)
+        [TestCase(@"Config/Feature1/*-{a,b}.config", "f1-a.config", 2, 2)]
+        [TestCase(@"Config/Feature1/*-{a,b}.config", "f1-b.config", 2, 2)]
```
scme0, line 141: I wrote these tests after I started changing the `Glob` class. If I had put them in first, I would have seen that the current implementation handles this specific case (having the `*` before the `{}` group).

source/Calamari.Common/Plumbing/FileSystem/CalamariPhysicalFileSystem.cs
```diff
@@ -186,16 +186,25 @@ void EnsureDirectoryDeleted(string path, FailureOptions failureOptions)
...
-        IEnumerable<FileSystemInfo> EnumerateWithGlob(string parentDirectoryPath, bool enableGroupSupport, params string[] globPattern)
+        IEnumerable<FileSystemInfo> EnumerateWithGlob(string parentDirectoryPath, GlobMode globMode, params string[] globPattern)
         {
+            IEnumerable<FileSystemInfo> Expand(string path)
+            {
+                return globMode switch
+                {
+                    GlobMode.GroupExpansionMode => Glob.ExpandPattern(path),
+                    _ => LegacyGlob.Expand(path)
+                };
```
APErebus, line 202: man I love switch expressions

APErebus (approved):
That's a really nice solution. Great work 👍

PR #1082, opened 2023-07-07
Fix issue where global glob will load kubectl-octo.yml context file

scme0:
Originally we were extracting the package directly to the staging area which is fine if you're directly referencing a single file but now that we are using glob patterns, we need to extract the package into it's own folder so that a global glob won't load any other files that happen to be in the staging directory.

In this PR I've updated the step to extract the to a package folder inside the staging directory. I needed to update some other dependencies so that they could accept a subfolder to check, rather than just using the current directory on the `DeploymentProcess`.

[sc-52814]

### DIFF
--- a/source/Calamari.Common/Features/Behaviours/ConfigurationVariablesBehaviour.cs
+++ b/source/Calamari.Common/Features/Behaviours/ConfigurationVariablesBehaviour.cs
@@ -25,7 +25,7 @@ namespace Calamari.Common.Features.Behaviours
             IVariables variables,
             IConfigurationVariablesReplacer replacer,
             ILog log,
-            string? subdirectory = "")
+            string subdirectory = "")
         {
             this.fileSystem = fileSystem;
             this.variables = variables;
--- a/source/Calamari.Common/Features/Behaviours/ConfigurationVariablesBehaviour.cs
+++ b/source/Calamari.Common/Features/Behaviours/ConfigurationVariablesBehaviour.cs
@@ -1,4 +1,4 @@
-﻿using System;
+﻿using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
 using Calamari.Common.Commands;
@@ -8,6 +8,7 @@ using Calamari.Common.Plumbing.FileSystem;
 using Calamari.Common.Plumbing.Logging;
 using Calamari.Common.Plumbing.Pipeline;
 using Calamari.Common.Plumbing.Variables;
+using Octopus.CoreUtilities.Extensions;
 
 namespace Calamari.Common.Features.Behaviours
 {
@@ -17,13 +18,20 @@ namespace Calamari.Common.Features.Behaviours
         readonly IVariables variables;
         readonly IConfigurationVariablesReplacer replacer;
         readonly ILog log;
+        private readonly string subdirectory;
 
-        public ConfigurationVariablesBehaviour(ICalamariFileSystem fileSystem, IVariables variables, IConfigurationVariablesReplacer replacer, ILog log)
+        public ConfigurationVariablesBehaviour(
+            ICalamariFileSystem fileSystem,
+            IVariables variables,
+            IConfigurationVariablesReplacer replacer,
+            ILog log,
+            string? subdirectory = "")
         {
             this.fileSystem = fileSystem;
             this.variables = variables;
             this.replacer = replacer;
             this.log = log;
+            this.subdirectory = subdirectory;
         }
 
         public bool IsEnabled(RunningDeployment context)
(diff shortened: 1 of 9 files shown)


source/Calamari/Kubernetes/Commands/KubernetesDeploymentCommandBase.cs
```diff
@@ -80,36 +84,36 @@ public override int Execute(string[] commandLineArguments)
             {
                 new DelegateInstallConvention(d =>
                 {
+                    var workingDirectory = d.CurrentDirectory;
+                    var stagingDirectory = Path.Combine(workingDirectory, StagingDirectoryPath);
+                    var packageDirectory = Path.Combine(stagingDirectory, PackageDirectoryPath);
```
APErebus, line 89: How comfortable are we that this won't clash with any other folder called `package` in `staging`? Could the user create this folder or could extracting an actual package go here?

I'm just concerned with the lack of uniqueness (like we have for the working directory)...

scme0: I don't think users have any control over the contents of the staging directory, it's just a place where Calamari stores files and things for processing the specific command. The unique working directory is required as if a Calamari process fails to clean-up correctly, we want to make sure we don't have clashes.

APErebus: ok, sounds good to me

source/Calamari.Common/Features/Behaviours/ConfigurationVariablesBehaviour.cs
```diff
@@ -17,13 +18,20 @@ public class ConfigurationVariablesBehaviour : IBehaviour
...
         readonly ILog log;
+        private readonly string subdirectory;
 
-        public ConfigurationVariablesBehaviour(ICalamariFileSystem fileSystem, IVariables variables, IConfigurationVariablesReplacer replacer, ILog log)
+        public ConfigurationVariablesBehaviour(
+            ICalamariFileSystem fileSystem,
+            IVariables variables,
+            IConfigurationVariablesReplacer replacer,
+            ILog log,
+            string? subdirectory = "")
```
YihaoOct, line 28: I'm noticing the `?` after `string`. Since this arg is default to `""`, can it really be `null`? If it is `null`, then `Path.Combine` will throw an Exception I think.

source/Calamari.Common/Features/Packages/ExtractPackage.cs
```diff
@@ -23,18 +23,25 @@ public ExtractPackage(ICombinedPackageExtractor combinedPackageExtractor, ICalam
...
         }
 
         public void ExtractToStagingDirectory(PathToPackage? pathToPackage, string extractedToPathOutputVariableName)
         {
-            var targetPath = Path.Combine(Environment.CurrentDirectory, "staging");
-            fileSystem.EnsureDirectoryExists(targetPath);
-            Extract(pathToPackage, targetPath, extractedToPathOutputVariableName, null);
+            ExtractToCustomSubDirectory(
+                pathToPackage,
+                Path.Combine(Environment.CurrentDirectory, "staging"),
```
YihaoOct, line 42: Maybe we can promote `"staging"` and `"grouped"` to SpecialVariables and use that instead of string literals? I'm imagining it would be easier to understand later because we can do "find usages" on these variables. Also you can reuse these from tests.

YihaoOct (changes requested):
Looks good, just some stylish recommendations. Ideally we could have some tests if it is not too complicated to write?

source/Calamari.Common/Features/Packages/ExtractPackage.cs
scme0: Yep good idea 👍

source/Calamari.Common/Features/Behaviours/ConfigurationVariablesBehaviour.cs
scme0: Ah good catch, yep I think originally I had it as `null` default but changed it precisely because Path.Combine will take an empty string.

source/Calamari.Common/Features/Packages/ExtractPackage.cs
scme0: I'm not sure if SpecialVariables is the right place because that's referring to Variable names for `IVariables`. I've moved the literals to constants but just at the top of the relevant class for now. I think if 'grouped' is used elsewhere later we can promote it further but at the moment it's only used in this class. Staging is used in `ExtractPackage` and `KubernetesDeploymentCommandBase` (and maybe some other places but I'm not going to update them).

YihaoOct (approved):
Looks good. It would be even better if we can have some tests around this ;)

scme0:
> It would be even better if we can have some tests around this ;)

This got me thinking, why weren't the tests I wrote failing! Well it turns out I was setting something slightly wrong in the tests. I've updated the tests (tested on old code and it fails) and it succeeds with this code, so I'm adding it to this PR as a (hopefully) final commit.

PR #1083, opened 2023-07-10
Update AKS discoverer to better handle scenario's where resource group doesn't exist

APErebus:
The AKS discoverer is using the old Fluent Management API SDK, which seems to have some issues where other test suites are adding/remove resource groups.

This code does similar logic to the SDK internally where it lists all the resource groups and then gets all the clusters for each resource group. We do the same logic, but we just better handle the errors when the resource group doesn't exist

Shortcut story: [sc-44622]


scme0 (approved):
🚀

PR #1085, opened 2023-07-12
Handle 429's from Azure Management API in AppService tests

APErebus:
We were seeing flaky tests due to 429 (too many requests/rate limiting) errors being returned from the azure management API.

My suspicion is this is flaky due to other test suites making api management calls.

This PR uses a polly retry policy to retry all management API calls in the test fixture. The policy looks at the `Retry-After` header of the response to get the correct number of seconds to retry after.

### DIFF
--- a/source/Calamari.AzureAppService.Tests/AppServiceBehaviorFixture.cs
+++ b/source/Calamari.AzureAppService.Tests/AppServiceBehaviorFixture.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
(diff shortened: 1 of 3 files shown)


scme0 (approved):
Nice, looks good functionally. 👌 Some of the formatting seems less readable before to my eye though. (Some also looks a bit better so 🤷‍♂️ )

APErebus:
> Nice, looks good functionally. 👌 Some of the formatting seems less readable before to my eye though. (Some also looks a bit better so 🤷‍♂️ )

Yeah, I think that's just Rider trying to be helpful :|

PR #1086, opened 2023-07-12
Properly use await async in PublishCalamariFlavourProjects

APErebus:
We were not using async/await properly in the `PublishCalamariFlavourProjects`

It's [highly recommended](https://devblogs.microsoft.com/pfxteam/task-factory-startnew-vs-new-task-start/) to not use `new Task()` and then `task.Start()`.

We also weren't awaiting the `Task.WhenAll()`'s either, which seems _bad_:tm:

### DIFF
--- a/build/Build.cs
+++ b/build/Build.cs
@@ -213,7 +213,7 @@ namespace Calamari.Build
         Target PublishCalamariFlavourProjects =>
             _ => _
                  .DependsOn(Compile)
-                 .Executes(() =>
+                 .Executes(async () =>
                  {
                      var migratedCalamariFlavoursTests = MigratedCalamariFlavours.Flavours.Select(f => $"{f}.Tests");
                      var calamariFlavourProjects = Solution.Projects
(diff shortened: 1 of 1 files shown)


build/Build.cs
```diff
@@ -273,10 +273,10 @@ void PublishCalamariProjects(List<Project> projects)
             var packagesToPublish = crossPlatformPackages.Concat(netFxPackages);
 
             packagesToPublish.ForEach(PublishPackage);
-            Task.WhenAll(SignDirectoriesTasks);
+            await Task.WhenAll(SignDirectoriesTasks);
             
             projects.ForEach(CompressCalamariProject);
-            Task.WhenAll(ProjectCompressionTasks);
+            await Task.WhenAll(ProjectCompressionTasks);
```
APErebus, line 279: It _seems_ to me that we should be only compressing the projects _after_ we've signed the directories 🤔 

as we weren't awaiting the `Task.WhenAll`'s, the compress would always run immediately without waiting for the signing to complete

scme0 (approved):
Damn, good find! Looks good 🚀

PR #1088, opened 2023-07-13
Expose Kubernetes authentication and Object status as commands

johnsimons:
I have created a new `substitute-in-files` so that the base path can be defaulted to the current directory instead of being read from a variable.

These commands are needed by project Kustomize so we can re-use existing functionality when using the new steps framework.
This is related to this change on the [Server](https://github.com/OctopusDeploy/OctopusDeploy/pull/19123).

### DIFF
--- a/source/Calamari/Commands/KubernetesAuthenticationCommand.cs
+++ b/source/Calamari/Commands/KubernetesAuthenticationCommand.cs
@@ -61,7 +61,7 @@ namespace Calamari.Commands
                 throw new CommandException(ex.Message);
             }
             
-            variables.Set("Octopus.KubeConfig.Path", fileSystem.GetFullPath(environmentVars["KUBECONFIG"]));
+            variables.Set(SpecialVariables.KubeConfig, fileSystem.GetFullPath(environmentVars["KUBECONFIG"]));
         }
     }
     
--- a/source/Calamari/Commands/KubernetesAuthenticationCommand.cs
+++ b/source/Calamari/Commands/KubernetesAuthenticationCommand.cs
@@ -1,0 +1,70 @@
+#if !NET40
+using System.Collections.Generic;
+using Calamari.Common.Commands;
+using Calamari.Common.Features.Processes;
+using Calamari.Common.Plumbing.FileSystem;
+using Calamari.Common.Plumbing.Logging;
+using Calamari.Common.Plumbing.Variables;
+using Calamari.Kubernetes;
+using Calamari.Kubernetes.Integration;
+
+namespace Calamari.Commands
+{
+    [Command("authenticate-to-kubernetes")]
+    public class KubernetesAuthenticationCommand: Command<KubernetesAuthenticationCommandInput>
+    {
+        private readonly ILog log;
+        private readonly IVariables variables;
+        private readonly ICalamariFileSystem fileSystem;
+        private readonly ICommandLineRunner commandLineRunner;
+        private readonly Kubectl kubectl;
+
+        public KubernetesAuthenticationCommand(
+            ILog log,
+            IVariables variables,
+            ICalamariFileSystem fileSystem,
+            ICommandLineRunner commandLineRunner,
+            Kubectl kubectl)
+        {
+            this.log = log;
+            this.variables = variables;
+            this.fileSystem = fileSystem;
+            this.commandLineRunner = commandLineRunner;
+            this.kubectl = kubectl;
+        }
+
+        protected override void Execute(KubernetesAuthenticationCommandInput inputs)
+        {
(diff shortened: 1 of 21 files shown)


source/Calamari/Commands/KubernetesObjectStatusReporterCommand.cs
```diff
@@ -0,0 +1,95 @@
...
+        {
+            this.log = log;
+            this.variables = variables;
+            this.statusReportExecutor = statusReportExecutor;
+            this.kubectl = kubectl;
+        }
+
+        private bool IsEnabled()
+        {
+            var resourceStatusEnabled = variables.GetFlag(SpecialVariables.ResourceStatusCheck);
```
johnsimons, line 36: @YihaoOct wondering if some of these props can be typed inputs?
Example:
-  "Octopus.Action.Kubernetes.ResourceStatusCheck"
-  "Octopus.Action.Kubernetes.DeploymentTimeout"
-   "Octopus.Action.KubernetesContainers.DeploymentStyle" <- Maybe this one does not make sense?
-  "Octopus.Action.KubernetesContainers.DeploymentWait" <- This one is for backward compatibility right, so not needed?
-  "Octopus.Action.Kubernetes.WaitForJobs"

So the idea is to have:
```c#
public class KubernetesObjectStatusReporterCommandInput
{
    public bool WaitForJobs { get; set; }
    public int DeploymentTimeout { get; set; }
    public bool Enabled { get; set; }
}
```

What do you think?

source/Calamari/Kubernetes/SetupKubectlAuthentication.cs
```diff
@@ -443,7 +439,7 @@ string CreateKubectlConfig()
 
             environmentVars.Add("KUBECONFIG", kubeConfig);
 
-            if (scriptSyntax == ScriptSyntax.Bash)
+            if (CalamariEnvironment.IsRunningOnNix || CalamariEnvironment.IsRunningOnMac)
```
johnsimons, line 442: This was just so I don't need to use the `ScriptSyntax` because I don't have it!

source/Calamari/Commands/KubernetesObjectStatusReporterCommand.cs
johnsimons: I ended up going down this path anyway

zentron:
I dont have the background behind this change, but is it creating calamari commands that can be invoked on their own? Why would a stand-alone k8s authenticate command exist outside a command that actually does something with it?

source/Calamari/Kubernetes/ResourceStatus/ResourceStatusReportExecutor.cs
```diff
@@ -26,12 +26,10 @@ public class ResourceStatusReportExecutor : IResourceStatusReportExecutor
...
         }
 
-        public IRunningResourceStatusCheck Start(IEnumerable<ResourceIdentifier> initialResources)
+        public IRunningResourceStatusCheck Start(int timeoutSeconds, bool waitForJobs, IEnumerable<ResourceIdentifier> initialResources = null)
         {
             initialResources = initialResources ?? Enumerable.Empty<ResourceIdentifier>();
-            var timeoutSeconds = variables.GetInt32(SpecialVariables.Timeout) ?? 0;
-            var waitForJobs = variables.GetFlag(SpecialVariables.WaitForJobs);
-
+
```
YihaoOct, line 32: It looks cleaner to have these variables read from a central location 👍

source/Calamari/Commands/KubernetesAuthenticationCommand.cs
```diff
@@ -0,0 +1,70 @@
...
+                var result = setupKubectlAuthentication.Execute(accountType);
+                result.VerifySuccess();
+            }
+            catch (CommandLineException ex)
+            {
+                log.Error("KubeConfig authentication failed.");
+                throw new CommandException(ex.Message);
+            }
+            
+            variables.Set("Octopus.KubeConfig.Path", fileSystem.GetFullPath(environmentVars["KUBECONFIG"]));
```
YihaoOct, line 64: Basically the commands communicate among each other using special variables, like this one `Octopus.KubeConfig.Path`. One command sets this and a following command reads this. I think using a string literal here will be harder to maintain after we finish our work.

In this PR it is still easy to figure out which command sets this variable and which reads this variable. But I imagine once we merge this PR, we'll lose track. Let's extract these special variable name strings to the `SpecialVariables` class. In this way, if another developers want to figure out where a certain variables was set, he can do so by "Find usages" on that special variable, which is easier.

YihaoOct (approved):
LGTM, left a comment, but I don't think it is blocking

PR #1090, opened 2023-07-21
only convert kube config if kubelogin available

benPearce1:
[sc-54002]


source/Calamari/Kubernetes/Integration/Kubelogin.cs
```diff
@@ -24,13 +24,17 @@ public bool TrySetKubeLogin()
             if (string.IsNullOrEmpty(foundExecutable))
             {
                 log.Error("Could not find kubelogin. Make sure kubelogin is on the PATH.");
```
IsaacCalligeros95, line 26: I think this should be updated to something along the lines of `log.Verbose("Could not find  kubelogin on PATH. The kubeconfig will not be converted to use the Exec Plugin.");`

IsaacCalligeros95 (approved):
LGTM

PR #1091, opened 2023-07-21
Change Calamari.AzureAppService to remove deprecated Azure SDK

APErebus:
The `Microsoft.Azure.Management.*` SDK's are to be deprecated at 30th of September 2023.

This PR removes this SDK and replaces it's usages with the newer `Azure.ResourceManagement.*` packages in the `Calamari.AzureAppService` and `Calamari.AzureAppService.Tests` projects.

I spent a lot of time trying to make sure that the logic didn't change too much, just how we access the API.

The new SDK has the concept of `Resources`, which are just a reference object to an azure resource, and only make an API call when `.GetAsync` (or some variant) is performed.

I consolidated the logic of how an `AzureTargetSite` is parsed and also gave it the `SubscriptionId`, so it would have enough information to create the correct `ResourceIdentifier` (via `CreateResourceIdentifier`).

We do a lot of `site.HasSlot ? {load slot thing} : {load root website thing}`, so I've tried to consolidate all this logic in the new `ArmClientExtensions` class + extension methods.

Shortcut story: [sc-44553]


source/Calamari.AzureAppService/Behaviors/AzureAppServiceBehaviour.cs
```diff
@@ -177,22 +182,45 @@ private async Task<IDeploymentSlot> FindOrCreateSlot(IWebApp client, TargetSite
...
-                new StreamContent(new FileStream(uploadZipPath, FileMode.Open, FileAccess.Read, FileShare.Read))
-                {
-                    Headers = { ContentType = new MediaTypeHeaderValue("application/octet-stream") }
-                }
-            );
 
-            if (!response.IsSuccessStatusCode) 
-                throw new Exception($"Zip upload to {zipUploadUrl} failed with HTTP Status '{response.StatusCode} {response.ReasonPhrase}'.");
+            //we add some retry just in case the web app's Kudu/SCM is not running just yet
+            var response = await RetryPolicies.TransientHttpErrorsPolicy.ExecuteAsync(async () =>
```
APErebus, line 200: I don't _love_ adding this retry here but I was getting failures in the tests as the SCM/Kudu site wasn't ready before the test ran. I guess we could adjust the test to do some availability checks before executing the test?

Happy to take feedback on this one

source/Calamari.AzureAppService.Tests/AppServiceIntegrationTest.cs
```diff
@@ -1,96 +1,113 @@
...
-            };
-            
-            //Create a retry policy that retries on 429 errors. This is because we've been getting a number of flaky test failures
-            RetryPolicy = RetryPolicyFactory.CreateForHttp429();
-            
-            await ConfigureTestResources(resourceGroup);
+                Environment.GetEnvironmentVariable(AccountVariables.ActiveDirectoryEndPoint) ?? DefaultVariables.ActiveDirectoryEndpoint;
+
+            //ask lawrence for the sandbox auto-cleanup tags
+            ResourceGroupName = $"{DateTime.UtcNow:yyyyMMdd}-{Guid.NewGuid():N}";
```
APErebus, line 46: Suggestion from @flin-8 to name the resource group with the date, so you can easily know if the resources are orphaned and can be deleted

source/Calamari.AzureAppService.Tests/AppServiceIntegrationTest.cs
```diff
@@ -115,12 +132,47 @@ protected static async Task DoWithRetries(int retries, Func<Task> action, int se
...
+        }
+
+        protected void AddAzureVariables(VariableDictionary variables)
+        {
+            variables.Add(AccountVariables.ClientId, ClientId);
+            variables.Add(AccountVariables.Password, ClientSecret);
+            variables.Add(AccountVariables.TenantId, TenantId);
+            variables.Add(AccountVariables.SubscriptionId, SubscriptionId);
+            variables.Add("Octopus.Action.Azure.ResourceGroupName", ResourceGroupName);
+            variables.Add("Octopus.Action.Azure.WebAppName", WebSiteResource.Data.Name);
```
scme0, line 145: These two magic strings look like they come from variables in the old `AzureWebAppHealthCheckActionHandlerFixtures.cs`. I'm wondering if we can use the variables here?

I think they are:
- `SpecialVariables.Action.Azure.ResourceGroupName`
- `SpecialVariables.Action.Azure.WebAppName`

APErebus: Yeah definitely. Let me fix that up

APErebus: Fixed in [9da81a6](https://github.com/OctopusDeploy/Calamari/pull/1091/commits/9da81a66f314946d5747ffbe45e601987a07237a)

source/Calamari.AzureAppService/Behaviors/AzureAppServiceBehaviour.cs
scme0: Hrm I would probably just leave it as you've done it with the retires. As long as the tests are stable like that.

scme0 (approved):
This all looks great, thanks for making the change. 

I'm a bit unsure how we can manage the rollout of this in a safe way as obviously it's not too practical to put a feature toggle in for this. Maybe we just go for it and if needed we hotfix a Calamari version revert or something?

Maybe we should catch up with Rob to discuss?

APErebus:
> This all looks great, thanks for making the change.
> 
> I'm a bit unsure how we can manage the rollout of this in a safe way as obviously it's not too practical to put a feature toggle in for this. Maybe we just go for it and if needed we hotfix a Calamari version revert or something?
> 
> Maybe we should catch up with Rob to discuss?


I chatted to @zentron and he's feeling was that if we _could_ control this changeover via a feature toggle then that would be ideal.

I think that shouldn't be the biggest issue, but will require the PR to be reworked/redone. I'm going to close this one off and redo the work in a new branch with a feature toggle.

It will result in some duplicate code, but I don't think that's the biggest concern considering we have a fixed EOL for the old SDK

PR #1093, opened 2023-07-25
Remove split by ';' as it is no longer required

scme0:
To align with the way other steps specify multiple paths, we are going to save paths as new line separated, rather than with a semicolon `;`. In this case, I can roll back extra handling I put into Calamari before which would split paths by `;` when getting the value from the `VariableDictionary`. The `GetPaths` method automatically splits by new line separators `\r` and `\n` and it trims the paths and only returns non-empty entries.

Related to [this change](https://github.com/OctopusDeploy/OctopusDeploy/pull/19325) in Server.

[sc-54749]
[sc-52829]


MissedTheMark (approved):
Looks reasonable to me!

PR #1094, opened 2023-07-28
Early exit on disabled bindings for IIS

IsaacCalligeros95:
Variable to set the binding as enabled is evaluated last, meaning we parse values unnecessarily. A customer experienced an issue where a binding was disabled and no valid values were provided in a specific environment/target scoping. These values were still being evaluated leading to failures.

Fixes https://github.com/OctopusDeploy/Issues/issues/8271

image
image
image
image

### DIFF
--- a/source/Calamari/Scripts/Octopus.Features.IISWebSite_BeforePostDeploy.ps1
+++ b/source/Calamari/Scripts/Octopus.Features.IISWebSite_BeforePostDeploy.ps1
@@ -436,6 +436,11 @@ if ($deployAsWebSite)
 	}
 
 	ForEach($binding in $bindingArray){
+		if(![Bool]::Parse($binding.enabled)) {
+    		Write-IISBinding "Ignore binding: " $binding
+    		return
+    	}
+
 		$sslFlagPart = @{$true=1;$false=0}[[Bool]::Parse($binding.requireSni)]  
 		$bindingIpAddress =  @{$true="*";$false=$binding.ipAddress}[[string]::IsNullOrEmpty($binding.ipAddress)]
 		$bindingInformation = $bindingIpAddress+":"+$binding.port+":"+$binding.host
@@ -457,12 +462,8 @@ if ($deployAsWebSite)
 		if ([Bool]::Parse($supportsSNI)) {
 			$bindingObj.sslFlags=$sslFlagPart;
 		}
-			
-		if([Bool]::Parse($binding.enabled)) {
-			$wsbindings.Add($bindingObj) | Out-Null
-		} else {
-			Write-IISBinding "Ignore binding: " $binding
-		}
+
+		$wsbindings.Add($bindingObj) | Out-Null
 	}
 
 	# For any HTTPS bindings, ensure the certificate is configured for the IP/port combination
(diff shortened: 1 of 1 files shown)


source/Calamari/Scripts/Octopus.Features.IISWebSite_BeforePostDeploy.ps1
```diff
@@ -436,6 +436,11 @@ if ($deployAsWebSite)
 	}
 
 	ForEach($binding in $bindingArray){
+		if(![Bool]::Parse($binding.enabled)) {
+    		Write-IISBinding "Ignore binding: " $binding
+    		return
```
hnrkndrssn, line 441:
```suggestion
    		continue
```

hnrkndrssn (changes requested):
Should use `continue` rather than `return` to move on to the next item in the bindings array.

source/Calamari/Scripts/Octopus.Features.IISWebSite_BeforePostDeploy.ps1
IsaacCalligeros95: I've just pushed this up and am running a build locally to test but I'm pretty confused here. I tested this in a basic script with 

```
1..3 | ForEach-Object  {
    $binding = New-Object PSObject
    $binding | Add-Member -MemberType NoteProperty -Name "enabled" -Value $false
    Write-Host "With Continue"

    if(![Bool]::Parse($binding.enabled)) {
        Write-Host "disabled"
    	continue
    }
}
```

and get 
image

Switching to return I get 
image

Running the script itself with alternating disabled toggles it looks like your right though...
image

Do you know why that above example is incorrect?

IsaacCalligeros95: Ahh https://stackoverflow.com/questions/7760013/why-does-continue-behave-like-break-in-a-foreach-object 😥 now I'm just confused why Powershell made this different on a ForEach vs a ForEach-Object. Thanks for the catch

hnrkndrssn: sigh...`return` it is.

IsaacCalligeros95: I think it's return in `ForEach-Object` but continue in `ForEach` which is this case. Because Cmdlet vs Loop... Stoopid Powershell

hnrkndrssn: ah, sorry. I'm blind, didn't realise your test script was using `ForEach-Object` instead of `ForEach` as in the pre-postdeploy script in Calamari.

hnrkndrssn (approved):

source/Calamari/Scripts/Octopus.Features.IISWebSite_BeforePostDeploy.ps1
IsaacCalligeros95: image
image

PR #1096, opened 2023-08-03
Add support for new Azure SDK's for App Services

APErebus:
Following PR to #1091.

> The `Microsoft.Azure.Management.*` SDK's are to be deprecated at 30th of September 2023.
> 
> This PR removes this SDK and replaces it's usages with the newer `Azure.ResourceManagement.*` packages in the `Calamari.AzureAppService` and `Calamari.AzureAppService.Tests` projects.
> 
> I spent a lot of time trying to make sure that the logic didn't change too much, just how we access the API.
> 
> The new SDK has the concept of `Resources`, which are just a reference object to an azure resource, and only make an API call when `.GetAsync` (or some variant) is performed.
> 
> I consolidated the logic of how an `AzureTargetSite` is parsed and also gave it the `SubscriptionId`, so it would have enough information to create the correct `ResourceIdentifier` (via `CreateResourceIdentifier`).
> 
> We do a lot of `site.HasSlot ? {load slot thing} : {load root website thing}`, so I've tried to consolidate all this logic in the new `ArmClientExtensions` class + extension methods.
>

The main difference in this PR compared to the last one is that there is now a feature toggle `ModernAzureAppServiceSdkFeatureToggle` that now toggles between the new SDK and the old SDK. I have renamed all the old behaviours, tests and models with the `Legacy` prefix and did not change anything in those classes.

The tricky bit was working where to "swap" between the legacy & new behaviours. To do this, I created some "Factory" behaviours that have access to the variables to swap which executed behaviour is run

Shortcut story: [sc-44553]


source/Calamari.AzureAppService/HealthCheckCommand.cs
```diff
@@ -39,18 +39,12 @@ public Task Execute(RunningDeployment context)
...
-            catch (RequestFailedException rfe) when (rfe.Status == 404)
-            {
-                throw new Exception($"Could not find site {siteAndSlotName} in resource group {resourceGroupName}, using Service Principal with subscription {servicePrincipal.SubscriptionNumber}", rfe);
-            }
+
+            var resourceGroupResource = client.GetResourceGroupResource(ResourceGroupResource.CreateResourceIdentifier(servicePrincipal.SubscriptionNumber, resourceGroupName));
+
+            //if the website doesn't exist, throw
+            if (!await resourceGroupResource.GetWebSites().ExistsAsync(siteAndSlotName))
+                throw new Exception($"Could not find site {siteAndSlotName} in resource group {resourceGroupName}, using Service Principal with subscription {servicePrincipal.SubscriptionNumber}");
```
APErebus, line 47: I took a moment to improve this logic. We don't need to load the resource group (and it's data), then load the website (and it's data) to see if the website exists or not, that's what `ExistsAsync` is for :)

source/Calamari.Common/FeatureToggles/FeatureToggle.cs
```diff
@@ -9,6 +9,7 @@ public enum FeatureToggle {
         SkunkworksFeatureToggle,
         MultiGlobPathsForRawYamlFeatureToggle,
         KubernetesAksKubeloginFeatureToggle,
-        GlobPathsGroupSupportFeatureToggle
+        GlobPathsGroupSupportFeatureToggle,
+        UseModernAzureAppServiceSdkFeatureToggle
```
APErebus, line 13: This is our sexy new feature toggle 🎉

source/Calamari.AzureAppService/Behaviors/Factories/AzureAppServiceDeployBehaviourFactory.cs
```diff
@@ -0,0 +1,29 @@
...
+        
+        public bool IsEnabled(RunningDeployment context) 
+            => true;
+
+        public Task Execute(RunningDeployment context)
+        {
+            return FeatureToggle.UseModernAzureAppServiceSdkFeatureToggle.IsEnabled(context.Variables) 
+                ? new AzureAppServiceBehaviour(log).Execute(context)
+                : new LegacyAzureAppServiceBehaviour(log).Execute(context);
+        }
```
APErebus, line 27: This is an example of a factory behaviour that uses the feature toggle to swap between the new behaviour and the legacy behaviour

source/Calamari.AzureAppService/Azure/AzureTargetSite.cs
```diff
@@ -0,0 +1,75 @@
+﻿using System;
+using Azure.Core;
+using Azure.ResourceManager.AppService;
+
+#nullable enable
+namespace Calamari.AzureAppService.Azure
+{
+    public class AzureTargetSite
```
APErebus, line 8: Ported the same AzureTargetSite improvements from the last PR

source/Calamari.AzureAppService/Json/LegacyConnectionStringsSetting.cs
```diff
@@ -1,14 +1,10 @@
...
-using System.Text;
-using System.Threading.Tasks;
 using Microsoft.Azure.Management.WebSites.Models;
-using Newtonsoft.Json;
+using ArmConnectionStringType = Azure.ResourceManager.AppService.Models.ConnectionStringType;
 
 namespace Calamari.AzureAppService.Json
 {
-    public class ConnectionStringSetting
+    public class LegacyConnectionStringSetting
```
APErebus, line 6: I needed a legacy version of this because the `ConnectionStringType` enum is a different value... One wonders if we should have our own version of the enum as these values are being parsed out of the variables 🤔

source/Calamari.AzureAppService/DeployAzureAppServiceCommand.cs
```diff
@@ -23,7 +22,7 @@ public class DeployAzureAppSettingsCommand : PipelineCommand
     {
         protected override IEnumerable<IDeployBehaviour> Deploy(DeployResolver resolver)
         {
-            yield return resolver.Create<AzureAppServiceBehaviour>();
+            yield return resolver.Create<AzureAppServiceDeployBehaviourFactory>();
```
APErebus, line 25: This seems _wrong_ for the `DeployAzureAppSettingsCommand` 🤔 
But I didn't want to stuff around with it, so I just kept it calling the same behaviours

source/Calamari.AzureAppService/Behaviors/Legacy/LegacyAzureAppServiceBehaviour.cs
```diff
@@ -0,0 +1,216 @@
...
+                                                                                              Content = new StreamContent(new FileStream(uploadZipPath, FileMode.Open, FileAccess.Read, FileShare.Read))
+                                                                                              {
+                                                                                                  Headers = { ContentType = new MediaTypeHeaderValue("application/octet-stream") }
+                                                                                              }
+                                                                                          };
+
+                                                                                          var r = await client.SendAsync(request);
+                                                                                          r.EnsureSuccessStatusCode();
+                                                                                          return r;
+                                                                                      });
```
YihaoOct, line 208: I understand this is not the main change of this PR, but do we want to keep the legacy class exactly the same as before? I have no strong opinions here.

source/Calamari.AzureAppService/ArmClientExtensions.cs
```diff
@@ -0,0 +1,200 @@
...
+    /// These methods are suffixed with <i>Async</i> for consistency with the <see cref="ArmClient"/>.
+    /// In the <b>Azure.ResourceManager</b> SDKs, <i>Async</i>-suffixed methods indicate an API call is made to Azure.
+    /// </remarks>
+    public static class ArmClientExtensions
+    {
+        public static async Task<SiteConfigData> GetSiteConfigDataAsync(this ArmClient armClient, AzureTargetSite targetSite)
+        {
+            return targetSite.HasSlot switch
+                   {
+                       true => (await armClient.GetWebSiteSlotConfigResource(WebSiteSlotConfigResource.CreateResourceIdentifier(
```
YihaoOct, line 28: nit: I have no strong opinion on this one, but it feels weird to switch on boolean value. I think if statements are a more natural way. But this isn't blocking, just a style choice.

YihaoOct (changes requested):
Looks good and I've checked it is consistent with https://github.com/OctopusDeploy/Calamari/pull/1091 except for the feature toggle. I put some comments around the use of factory etc

source/Calamari.AzureAppService/ArmClientExtensions.cs
APErebus: Yeah, I can understand that. I like switch statements like this because I find them a bit easier to read.

source/Calamari.AzureAppService/Behaviors/Legacy/LegacyAzureAppServiceBehaviour.cs
APErebus: I did originally, but then I started getting flaky tests, so I introduced this into the legacy code to resolve that

source/Calamari.AzureAppService/Behaviors/AzureAppServiceBehaviour.cs
```diff
@@ -177,22 +182,45 @@ private async Task<IDeploymentSlot> FindOrCreateSlot(IWebApp client, TargetSite
...
+            {
+#pragma warning disable DE0003
+                Proxy = WebRequest.DefaultWebProxy
+#pragma warning restore DE0003
+            })
+            {
+                // The HttpClient default timeout is 100 seconds: https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.timeout?view=net-5.0#remarks
+                // This timeouts with even relatively small packages: https://octopus.zendesk.com/agent/tickets/69928
+                // We'll set this to an hour for now, but we should probably implement some more advanced retry logic, similar to https://github.com/OctopusDeploy/Sashimi.AzureWebApp/blob/bbea36152b2fb531c2893efedf0330a06ae0cef0/source/Calamari/AzureWebAppBehaviour.cs#L70
+                Timeout = TimeSpan.FromHours(1)
```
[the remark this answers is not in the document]
APErebus: yeah, probably not actually 🤔

source/Calamari.AzureAppService/Behaviors/Factories/AzureAppServiceDeployContainerBehaviourFactory.cs
```diff
@@ -0,0 +1,27 @@
...
+        }
+        
+        public bool IsEnabled(RunningDeployment context) => true;
+        
+        public Task Execute(RunningDeployment context)
+        {
+            return FeatureToggle.ModernAzureAppServiceSdkFeatureToggle.IsEnabled(context.Variables) 
+                ? new AzureAppServiceDeployContainerBehaviour(log).Execute(context)
+                : new LegacyAzureAppServiceDeployContainerBehavior(log).Execute(context);        
+        }
```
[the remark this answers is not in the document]
APErebus: Yeah, I didn't love this solution. The issue is that I don't have the variables to check the feature toggle when until execute is called. I could certainly have optimised a bit by not createing a new behaviour on each execution, but I figured they are so lightweight it didn't matter that much.

I didn't feel strongly enough about designing a new `IDeployBehaviourFactory` class, but I'm willing to take another look if you want.

source/Calamari.AzureAppService/Behaviors/AzureAppServiceBehaviour.cs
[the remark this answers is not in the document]
APErebus: This is actually worth keep for now as unzipping can be a long running operation and could exceed the `100s` timeout. The retry logic won't help this.

source/Calamari.AzureAppService/Behaviors/Factories/AzureAppServiceDeployContainerBehaviourFactory.cs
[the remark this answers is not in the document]
APErebus: @YihaoOct I made a bunch of changes to remove the factories and use the `IsEnabled` flag

APErebus:
@YihaoOct I made a bunch of changes and remove the factories based on your feedback in ([f74236e](https://github.com/OctopusDeploy/Calamari/pull/1096/commits/f74236e1cd2a806da26edd63b73c1f3e0ec91e60) and [a55a403](https://github.com/OctopusDeploy/Calamari/pull/1096/commits/a55a40330108b5aff35b55b912ecae867cbedca6)

YihaoOct (approved):
LGTM. Using `Enabled` is a cleaner way 👍

source/Calamari.AzureAppService/RetryPolicies.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Calamari.AzureAppService
+{
+    public static class RetryPolicies
+    {
+        static readonly Random Jitterer = new Random();
+
+        // Based on the logic in the Polly.Extensions.Http package, but without having to include the package
+        // We add a small amount of random jitter to just offset the retries
+        public static RetryPolicy<HttpResponseMessage> TransientHttpErrorsPolicy { get; } = Policy.Handle<HttpRequestException>()
```
johnsimons, line 16: When the Cloud team migrated to this new API we pretty much removed the auto retries after reading https://learn.microsoft.com/en-us/azure/architecture/best-practices/transient-faults.
Specifically, the last point in https://learn.microsoft.com/en-us/azure/architecture/best-practices/transient-faults#manage-retry-policy-configurations

johnsimons: We actually looked at the MS source code and I am pretty sure we came to the conclusion that by default MS does a really good job in this area.

johnsimons: @x-cubed do you remember us looking at this when we converted to this new lib?

x-cubed: Yeah, it's been a while but I believe they had code to handle common cases like `429 Too Many Requests` automatically so that we didn't have to.

johnsimons: It seems you are only using this when we upload the zip, so ignore my comment!

johnsimons: That said it feels weird that we are checking for `HttpStatusCode.RequestTimeout` errors but we don't increase the timeout on subsequent retries

johnsimons:
@APErebus I am all forward to using feature flags, but I am unclear on how we are going to monitor the success or failure of this change.
Do we have enough telemetry to tell us if people are using this step and whether it was successful?

johnsimons:
So is the plan to enable the flag for all cloud instances and then monitor it for a while?

APErebus:
> So is the plan to enable the flag for all cloud instances and then monitor it for a while?

Yeah, that was the plan. I'm pretty confident that it's a safe change, but the feature toggle gives us that ability to quickly revert if there is something not right.

Good question regarding telemetry, I'm not sure if we have telemetry regarding successful step executions 🤔

source/Calamari.AzureAppService/RetryPolicies.cs
APErebus: The timeout is currently 1 hour but I think the server terminates the request well before that. This is the thing that could be fixed with that [async zip deployment improvement](https://octopusdeploy.slack.com/archives/C03SG0LFJHX/p1692832782107319)

APErebus:
@johnsimons I looked and there is good telemetry in Amplitude where we are can monitor the number of successful deployments with the feature toggle enabled and disabled

PR #1097, opened 2023-08-04
Add Vanta Tags to test AWS S3 buckets

APErebus:
@cvandal has notified us that the S3 buckets being created by the Calamari AWS tests need to have Vanta tags applied.

I also updated the bucket names to be more descriptive than the pure guids they were previously

Shortcut story: [sc-55753]

### DIFF
--- a/source/Calamari.Tests/AWS/CloudFormation/CloudFormationFixture.cs
+++ b/source/Calamari.Tests/AWS/CloudFormation/CloudFormationFixture.cs
@@ -12,7 +12,7 @@ namespace Calamari.Tests.AWS.CloudFormation
     [Category(TestCategory.RunOnceOnWindowsAndLinux)]
     public class CloudFormationFixture
     {
-        string GenerateStackName() => $"{DateTime.UtcNow:yyyy-MM-dd}-calamari-cloudform-{Guid.NewGuid():N}";
+        string GenerateStackName() => $"calamari-cloudformationfixture-{Guid.NewGuid():N}";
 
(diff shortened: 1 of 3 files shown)


source/Calamari.Tests/AWS/CloudFormation/CloudFormationFixture.cs
```diff
@@ -12,7 +12,7 @@ namespace Calamari.Tests.AWS.CloudFormation
     [Category(TestCategory.RunOnceOnWindowsAndLinux)]
     public class CloudFormationFixture
     {
-        string GenerateStackName() => $"calamariteststack{Guid.NewGuid().ToString("N").ToLowerInvariant()}";
+        string GenerateStackName() => $"calamaricloudformation{Guid.NewGuid():N}";
```
johnsimons, line 15: Does it need to be lowercase? just wondering if this is one of those things in was that has some weird requirement?

johnsimons (approved):
Lots of formatting changes unrelated but LGTM

source/Calamari.Tests/AWS/CloudFormation/CloudFormationFixture.cs
APErebus: I was having all kinds of weird issues with this not validating correctly. Apparently it does support `-`, but didn't work 🤷

APErebus:
> Lots of formatting changes unrelated but LGTM

I think the formatting was just Rider being helpful

PR #1098, opened 2023-08-09
Properly name S3 buckets in integration tests

APErebus:
Makes it easier to identify the buckets in AWS for automatic cleanup (if they don't get cleaned up by the test)


APErebus:
> I assume there are no max length issues with bucket names?

The max length is 64 chars, this is less than that :)

PR #1101, opened 2023-08-14
Handle WebException inner exceptions in TF tests

APErebus:
We see occasional scenario's where instead of `SocketException`, the inner exception is a `WebException`.

https://build.octopushq.com/test/3713854659209742989?currentProjectId=OctopusDeploy_Calamari_CalamariTerraformTests_NetFxWindowsTesting


scme0 (approved):
🚀

PR #1102, opened 2023-08-14
Reduce deployment journal size

tothegills:
The deployment journal records deployments to clean them up at a later date.
In the beginning, the deployment journal recorded everything because everything was a package being deployed.
These days, work is done that does not involve deploying a package, especially in the age of workers. With workers being involved in every deployment (instead of just the ones to the environment a deployment target is in), the deployment journal is exploding. 

This PR makes the deployment journal more selective about what it records. It no longer records journal entries not used for retention: entries where a package is not deployed to the disk. This greatly reduces the size of the journal on frequently used workers, speeding up deployments.

### DIFF
--- a/source/Calamari.Common/Plumbing/Deployment/Journal/DeploymentJournalWriter.cs
+++ b/source/Calamari.Common/Plumbing/Deployment/Journal/DeploymentJournalWriter.cs
@@ -10,7 +10,7 @@ namespace Calamari.Common.Plumbing.Deployment.Journal
 {
     public class DeploymentJournalWriter : IDeploymentJournalWriter
     {
-        private readonly ICalamariFileSystem fileSystem;
+        readonly ICalamariFileSystem fileSystem;
 
         public DeploymentJournalWriter(ICalamariFileSystem fileSystem)
         {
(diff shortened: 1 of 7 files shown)


source/Calamari.Common/Plumbing/Deployment/Journal/DeploymentJournalWriter.cs
```diff
@@ -26,16 +26,26 @@ public DeploymentJournalWriter(ICalamariFileSystem fileSystem)
...
 
             var canWrite = deployment.Variables.Get(TentacleVariables.Agent.JournalPath) != null;
 
-            if(canWrite && hasPackages && !deployment.SkipJournal)
-                journal.AddJournalEntry(new JournalEntry(deployment, wasSuccessful));
+            if (!canWrite)
+                return;
+
+            var journalEntry = new JournalEntry(deployment, wasSuccessful);
+            if (string.IsNullOrEmpty(journalEntry.ExtractedTo))
```
mik-ky, line 45: 👍  for a detailed PR description, makes it clear why this check is put.

mik-ky (approved):

PR #1109, opened 2023-08-23
WIP Fix spaces in directory name bug raw yaml step

scme0:


APErebus (approved):
Great job. Thanks for the quick turnaround on this one.

This'll will need to go into `release/2023.3` and then `master` in Server right?

scme0:
Yep I'll do that tomorrow once we have a new version of Calamari 👍

johnsimons:
@scme0 can we add a test for this?

scme0:
Good point. I thought I setup the test to reprod but I think in my haste I forgot to commit the file. Here is a new PR to test fix: https://github.com/OctopusDeploy/Calamari/pull/1110

PR #1111, opened 2023-08-24
Add asynchronous Azure zip deployment support using new feature toggle

APErebus:
Microsoft has added the ability to specify a zip deployment as asynchronous, meaning the original request just performs the upload, not the unpacking of the zip archive.

[Documentation](https://learn.microsoft.com/en-us/azure/azure-functions/deployment-zip-push#asynchronous-zip-deployment)
> You will receive a response as soon as the zip file is uploaded with a Location header pointing to the pollable deployment status URL. When polling the URL provided in the Location header, you will receive a HTTP 202 (Accepted) response while the process is ongoing and a HTTP 200 (OK) response once the archive has been expanded and the deployment has completed successfully.

This PR adds a new feature toggle `AsynchronousAzureZipDeployFeatureToggle`, that changes _all_ zip deployments to asynchronous deployments.

We first perform the upload as per normal, but then run a Polly retry policy to monitor the operation for 3 minutes to confirm it's finished.

I chose to implement this using a `RetryAndWaitForever` policy + `Timeout` policy (which is cancellation token driven) rather than write the loop myself. Happy to take feedback on this one if people don't like it, more a personal preference, but happy to change if desired.

I picked 3min because that felt reasonable,

This is the log messages

```
Zip upload succeeded. Monitoring for deployment completion
Zip deployment not completed. Received HTTP 202. Next check in 00:00:02.
Zip deployment not completed. Received HTTP 202. Next check in 00:00:02.
Zip deployment not completed. Received HTTP 202. Next check in 00:00:02.
Finished zip deployment
```

Shortcut story: [sc-57940]

### DIFF
--- a/source/Calamari.AzureAppService/Behaviors/AzureAppServiceZipDeployBehaviour.cs
+++ b/source/Calamari.AzureAppService/Behaviors/AzureAppServiceZipDeployBehaviour.cs
@@ -33,12 +33,9 @@ namespace Calamari.AzureAppService.Behaviors
         public AzureAppServiceZipDeployBehaviour(ILog log)
         {
             Log = log;
-            Archive = new ZipPackageProvider();
         }
 
         private ILog Log { get; }
-
-        private IPackageProvider Archive { get; set; }
 
         public bool IsEnabled(RunningDeployment context) => FeatureToggle.ModernAzureAppServiceSdkFeatureToggle.IsEnabled(context.Variables);
 
@@ -100,20 +97,13 @@ namespace Calamari.AzureAppService.Behaviors
 
             var packageFileInfo = new FileInfo(variables.Get(TentacleVariables.CurrentDeployment.PackageFilePath)!);
 
-            switch (packageFileInfo.Extension)
-            {
-                case ".zip":
-                    Archive = new ZipPackageProvider();
-                    break;
-                case ".nupkg":
-                    Archive = new NugetPackageProvider();
-                    break;
-                case ".war":
-                    Archive = new WarPackageProvider(Log, variables, context);
-                    break;
-                default:
-                    throw new Exception("Unsupported archive type");
-            }
+            IPackageProvider packageProvider = packageFileInfo.Extension switch
+                                  {
+                                      ".zip" => new ZipPackageProvider(),
+                                      ".nupkg" => new NugetPackageProvider(),
+                                      ".war" => new WarPackageProvider(Log, variables, context),
+                                      _ => throw new Exception("Unsupported archive type")
+                                  };
 
             // Let's process our archive while the slot is spun up. We will await it later before we try to upload to it.
             Task<WebSiteSlotResource>? slotCreateTask = null;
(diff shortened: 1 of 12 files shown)


source/Calamari.AzureAppService/RetryPolicies.cs
```diff
@@ -18,5 +25,15 @@ public static class RetryPolicies
                                                                                                   .OrResult<HttpResponseMessage>(r => (int)r.StatusCode >= 500 || r.StatusCode == HttpStatusCode.RequestTimeout)
                                                                                                   .WaitAndRetryAsync(5,
                                                                                                                      retryAttempt => TimeSpan.FromSeconds(Math.Pow(2, retryAttempt)) + TimeSpan.FromMilliseconds(Jitterer.Next(0, 1000)));
+
+        public static RetryPolicy<HttpResponseMessage> AsynchronousZipDeploymentOperationPolicy { get; } = Policy.HandleResult<HttpResponseMessage>(r => r.StatusCode == HttpStatusCode.Accepted)
+                                                                                                                 .WaitAndRetryForeverAsync((_1,ctx) => TimeSpan.FromSeconds(2),
```
johnsimons, line 30: Should we retry forever?

source/Calamari.AzureAppService/RetryPolicies.cs
```diff
@@ -18,5 +25,15 @@ public static class RetryPolicies
...
+                                                                                                                                           (response, timeout, ctx) =>
+                                                                                                                                           {
+                                                                                                                                               if (ctx.TryGetValue(ContextKeys.Log, out var logObj) && logObj is ILog log)
+                                                                                                                                               {
+                                                                                                                                                   log.Verbose($"Zip deployment not completed. Received HTTP {(int)response.Result.StatusCode}. Next check in {timeout}.");
```
johnsimons, line 35:
```suggestion
                                                                                                                                                   log.Verbose($"Zip deployment not completed. Received HTTP {(int)response.Result.StatusCode}. Next attempt in {timeout}.");
```

source/Calamari.AzureAppService/Behaviors/AzureAppServiceZipDeployBehaviour.cs
```diff
@@ -144,7 +151,15 @@ public async Task Execute(RunningDeployment context)
             Log.Verbose($"Using deployment endpoint '{publishingProfile.PublishUrl}' from publishing profile.");
 
             Log.Info($"Uploading package to {targetSite.SiteAndSlot}");
-            await UploadZipAsync(publishingProfile, uploadPath, targetSite.ScmSiteAndSlot);
+
+            if (FeatureToggle.AsynchronousAzureZipDeployFeatureToggle.IsEnabled(context.Variables))
```
johnsimons, line 155: I may be wrong where @APErebus, but from reading the [doco](https://learn.microsoft.com/en-us/azure/azure-functions/deployment-zip-push#rest), this method seems to only be supported for `/api/zipdeploy` which is the [ZipPackageProvider](https://github.com/OctopusDeploy/Calamari/blob/6981c41c7ffad05dfc8ccf95c8f9b20043b04618/source/Calamari.AzureAppService/ArchivePackagProvider/ZipPackageProvider.cs#L11), so I think we need to restrict this path only if `Archive is ZipPackageProvider`?

johnsimons (changes requested):

source/Calamari.AzureAppService/RetryPolicies.cs
APErebus: In practice it doesn't. There just isn't a built-in `RetryAndWait` with a fixed expiration. As mentioned in the top, we wrap this in a `Timeout` policy which marks a cancellation token as cancelled, which then in turn aborts this policy.

source/Calamari.AzureAppService/Behaviors/AzureAppServiceZipDeployBehaviour.cs
APErebus: Hrm ok. This isn't something I was aware of. Thanks for the heads up. Let me check further 👍

APErebus: @johnsimons Zip & Nuget both use the `/api/zipdeploy` endpoint, so I've set them as both supporting async deployment. I added a new test to cover the nuget async deployment and it works 👍

johnsimons (approved):

PR #1114, opened 2023-08-29
Make KOS able to handle resources without namespaces

ghost:
Sometimes customers need to deploy resources that are cluster-scoped, which do not have namespaces. Currently these resources result in a `kubectl get <kind> <name> -n` command being sent, which is invalid because the namespace argument `-n` must have a value.

This PR fixes this issue by filtering out the resources that are cluster-scoped. When we have cluster-scoped resources, we exclude them from the KOS and do not show them to the customer. Instead we note them down in the task log, and tell the customer that those are not checked against.

Assuming we are using this YAML which deploys a cluster-scoped resource:
Screenshot 2023-08-29 at 3 00 33 PM

## Before
Screenshot 2023-08-29 at 3 32 27 PM

## After
Screenshot 2023-08-29 at 4 29 13 PM

### DIFF
--- a/source/Calamari/Kubernetes/ResourceStatus/ResourceStatusCheckTask.cs
+++ b/source/Calamari/Kubernetes/ResourceStatus/ResourceStatusCheckTask.cs
@@ -64,8 +64,8 @@ namespace Calamari.Kubernetes.ResourceStatus
                                                   .GetAllOwnedResources(definedResources, kubectl, options)
                                                   .ToArray();
 
-                    var namespacedDefinedResources = definedResourceStatuses
-                        .Where(resourceStatus => resourceStatus.Namespaced)
+                    var nonNamespacedDefinedResources = definedResourceStatuses
+                        .Where(resourceStatus => !resourceStatus.Namespaced)
                         .Select(resourceStatus => new ResourceIdentifier(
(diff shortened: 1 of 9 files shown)


source/Calamari/Kubernetes/ResourceStatus/ResourceStatusCheckTask.cs
```diff
@@ -62,6 +62,20 @@ public class ResourceStatusCheckTask
...
+                        .Where(resourceStatus => !resourceStatus.Namespaced)
+                        .Select(resourceStatus => new ResourceIdentifier(
+                            resourceStatus.Kind, 
+                            resourceStatus.Name,
+                            resourceStatus.Namespace))
+                        .ToHashSet();
+
+                    // Filter out cluster-wide resources
+                    definedResources = definedResources
+                        .Where(resource => !string.IsNullOrEmpty(resource.Namespace) && !nonNamespacedDefinedResources.Contains(new ResourceIdentifier(resource.Kind, resource.Name, string.Empty)))
```
YihaoOct, line 76: If either the namespace on the identifier itself is empty, or the namespace from the resource response is empty, we consider it cluster-scoped.

PR #1116, opened 2023-08-31
Fix kubectl get command when namespace is empty

YihaoOct:
When the resource to be deployed is cluster-scoped (i.e. it doesn't have a namespace), in KOS we current produce a command similar to `kubectl get <kind> <name> -n` with the argument to `-n` being empty. Because this command is invalid, the kubectl command is unable to retrieve the resource status and this causes the resource status check unable to terminate. This PR fixed the command so that it handles both resource with and without a namespace.

### DIFF
--- a/source/Calamari/Kubernetes/ResourceStatus/ResourceStatusCheckTask.cs
+++ b/source/Calamari/Kubernetes/ResourceStatus/ResourceStatusCheckTask.cs
@@ -43,10 +43,10 @@ namespace Calamari.Kubernetes.ResourceStatus
 
             var definedResources = resources.ToArray();
             var checkCount = 0;
+            var result = new Result();
             return await Task.Run(async () =>
             {
                 timer.Start();
-                var result = new Result();
                 do
                 {
                     if (cancellationToken.IsCancellationRequested)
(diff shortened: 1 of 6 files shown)


source/Calamari/Kubernetes/ResourceStatus/ResourceStatusCheckTask.cs
```diff
@@ -43,10 +43,10 @@ public class ResourceStatusCheckTask
 
             var definedResources = resources.ToArray();
             var checkCount = 0;
+            var result = new Result();
```
YihaoOct, line 46: Moved this outside the loop so it is consistent with the implementation before the glob path changes:
https://github.com/OctopusDeploy/Calamari/blob/b90549801bfc8d6f9e14ef6ee9f00a7fa80f2a1c/source/Calamari/Kubernetes/ResourceStatus/ResourceStatusChecker.cs#L47

This makes sure we send back the updated resources comparing with the last run, instead of an empty list of resources.

@scme0 Can you double check on this?

johnsimons (changes requested):
The fix LGTM, but can we add some tests for this case?

johnsimons (approved):

source/Calamari/Kubernetes/ResourceStatus/ResourceStatusCheckTask.cs
YihaoOct: Actually, moving this outside will cause it to be shared by threads and is therefore not safe. Sorry I misunderstood this part at the beginning, I'm reverting this.

PR #1117, opened 2023-08-31
Fix kubectl get command when namespace is empty

YihaoOct:
Back port of https://github.com/OctopusDeploy/Calamari/pull/1116


YihaoOct:
Closed because of wrong target branch

PR #1122, opened 2023-09-05
Set AZURE_CONFIG_DIR env var for KOS

APErebus:
We need to set the `AZURE_CONFIG_DIR` working directory for KOS so AKS clusters using Azure RBAC (`kubelogin`) can be authenticated against

### DIFF
--- a/source/Calamari/Commands/KubernetesObjectStatusReporterCommand.cs
+++ b/source/Calamari/Commands/KubernetesObjectStatusReporterCommand.cs
@@ -13,7 +13,7 @@ using Calamari.Kubernetes.ResourceStatus;
 namespace Calamari.Commands
 {
     [Command("kubernetes-object-status")]
-    public class KubernetesObjectStatusReporterCommand: Command<KubernetesObjectStatusReporterCommandInput>
+    public class KubernetesObjectStatusReporterCommand : Command<KubernetesObjectStatusReporterCommandInput>
     {
         private readonly ILog log;
         private readonly IVariables variables;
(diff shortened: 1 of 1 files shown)


johnsimons (approved):

source/Calamari/Commands/KubernetesObjectStatusReporterCommand.cs
```diff
@@ -70,10 +74,14 @@ protected override void Execute(KubernetesObjectStatusReporterCommandInput input
...
 
-        private void ConfigureKubectl()
+        private void ConfigureKubectl(string workingDirectory)
         {
             var kubeConfig = variables.Get(SpecialVariables.KubeConfig);
-            var environmentVars = new Dictionary<string, string> {{"KUBECONFIG", kubeConfig}};
+            var environmentVars = new Dictionary<string, string>
+            {
+                ["KUBECONFIG"] = kubeConfig,
+                ["AZURE_CONFIG_DIR"] = Path.Combine(workingDirectory, "azure-cli")
```
johnsimons, line 83: I wonder if we should share this part of the code so they never get out of sync?

johnsimons (approved):
@APErebus PR LGTM, but I left a small comment for consideration.
Also, we need to backport this to 2023.3, so I think you need to create a hotfix branch for that

APErebus:
> @APErebus PR LGTM, but I left a small comment for consideration. Also, we need to backport this to 2023.3, so I think you need to create a hotfix branch for that

We have a `release/2023.3` branch. I'll make the change in that branch and forward merge into this one. Closing this PR

PR #1123, opened 2023-09-05
Set AZURE_CONFIG_DIR env var for KOS

APErebus:
We need to set the `AZURE_CONFIG_DIR` working directory for KOS so AKS clusters using Azure RBAC (`kubelogin`) can be authenticated against.

Supersedes #1122 and will be forward merged to `master`

### DIFF
--- a/source/Calamari/Kubernetes/Integration/AzureCli.cs
+++ b/source/Calamari/Kubernetes/Integration/AzureCli.cs
@@ -3,6 +3,7 @@ using System.IO;
 using System.Linq;
 using Calamari.Common.Features.Processes;
 using Calamari.Common.Plumbing;
+using Calamari.Common.Plumbing.FileSystem;
 using Calamari.Common.Plumbing.Logging;
 
 namespace Calamari.Kubernetes.Integration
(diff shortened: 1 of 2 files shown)


source/Calamari/Kubernetes/Integration/AzureCli.cs
```diff
@@ -92,5 +93,10 @@ public void ConfigureAksKubeCtlAuthentication(Kubectl kubectlCli, string cluster
 
             kubectlCli.ExecuteCommandAndAssertSuccess("config", "set-context", clusterName, $"--namespace={@clusterNamespace}");
         }
+
+        public static void SetConfigDirectoryEnvironmentVariable(IDictionary<string, string> environmentVars, string workingDirectory)
+        {
+            environmentVars.Add("AZURE_CONFIG_DIR", Path.Combine(workingDirectory, "azure-cli"));
+        }
```
APErebus, line 100: @johnsimons Made this a static functionality so it's at least the directory name won't get out of sync

johnsimons (approved):

PR #1128, opened 2023-09-13
Add OIDC support for AWS

hnrkndrssn:


source/Calamari.CloudAccounts/AwsEnvironmentGeneration.cs
```diff
@@ -171,6 +182,40 @@ async Task<bool> PopulateSuppliedKeys()
                 return true;
             }
 
+            if (!string.IsNullOrEmpty(roleArn))
```
IsaacCalligeros95, line 185: AWS Auth always breaks my brain. Is it safer to just check the oidcJwt here? It's a little hard to distinguish this variable RoleArn from the AssumedRoleArn on the surface. No need to change it if you disagree/have a green build.

IsaacCalligeros95 (approved):
LGTM, one minor comment but I may be off on that point and no need to hold it up.

PR #1130, opened 2023-09-19
Support Azure OIDC accounts for cloud target discovery

IsaacCalligeros95:
Adds support for Azure OIDC accounts on cloud target discovery for K8s and Az App Service
k8soidccts
image (16)

Re-tested with latest changes
OIDC:
image

Without valid subject in AD:
image

SP:
image

### DIFF
--- a/source/Calamari.Azure/Kubernetes/Discovery/AzureKubernetesDiscoverer.cs
+++ b/source/Calamari.Azure/Kubernetes/Discovery/AzureKubernetesDiscoverer.cs
@@ -2,12 +2,10 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Net;
-using System.Threading;
 using Calamari.CloudAccounts;
 using Calamari.Common.Features.Discovery;
 using Calamari.Common.Plumbing.Logging;
 using Microsoft.Rest.Azure;
-using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
 namespace Calamari.Azure.Kubernetes.Discovery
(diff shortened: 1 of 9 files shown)


source/Calamari.Azure/Kubernetes/Discovery/AzureKubernetesDiscoverer.cs
```diff
@@ -30,10 +28,34 @@ public AzureKubernetesDiscoverer(ILog log) : base(log)
...
+            }
+            catch (Exception e)
+            {
+                Log.Warn("Could not read authentication method of target discovery context.");
+            }
+
+            IAzureAccount account;
+            string accountId;
+
+            if (accountType == "AzureOidc")
```
benPearce1, line 44: Do we have a constant that we can use here?

source/Calamari.AzureAppService/Behaviors/TargetDiscoveryBehaviour.cs
```diff
@@ -42,7 +42,47 @@ public TargetDiscoveryBehaviour(ILog log)
...
+                Log.Warn("Aborting target discovery.");
+                return;
+            }
+
+            TargetDiscoveryContext<AccountAuthenticationDetails<IAzureAccount>>? targetDiscoveryContext = null;
+            try
+            {
+                var jsonObj = JsonConvert.DeserializeObject<dynamic>(json!);
+                string authenticationMethod = jsonObj!.authentication.authenticationMethod;
+                if (authenticationMethod == "AzureOidc")
```
benPearce1, line 59: Same question as above re: constant

source/Calamari.Azure/Kubernetes/Discovery/AzureKubernetesDiscoverer.cs
```diff
@@ -30,10 +28,34 @@ public AzureKubernetesDiscoverer(ILog log) : base(log)
...
-            
-            var account = authenticationDetails.AccountDetails;
+            string accountType = string.Empty;
+            try
+            {
+                accountType = JToken.Parse(contextJson).SelectToken("authentication").SelectToken("authenticationMethod").ToString();
+            }
+            catch (Exception e)
+            {
+                Log.Warn("Could not read authentication method of target discovery context.");
```
benPearce1, line 38: Looks like this is missing an exit 
```suggestion
                Log.Warn("Could not read authentication method of target discovery context.");
                return Enumberable.Empty<KubernetesCluster>();
```

source/Calamari.Azure/Kubernetes/Discovery/AzureKubernetesDiscoverer.cs
```diff
@@ -30,10 +28,34 @@ public AzureKubernetesDiscoverer(ILog log) : base(log)
...
         public override IEnumerable<KubernetesCluster> DiscoverClusters(string contextJson)
         {
-            if (!TryGetDiscoveryContext<AccountAuthenticationDetails<AzureServicePrincipalAccount>>(contextJson, out var authenticationDetails, out _))
-                return Enumerable.Empty<KubernetesCluster>();
-            
-            var account = authenticationDetails.AccountDetails;
+            string accountType = string.Empty;
+            try
+            {
+                accountType = JToken.Parse(contextJson).SelectToken("authentication").SelectToken("authenticationMethod").ToString();
```
benPearce1, line 34: This feels a bit flaky. Is the account type available in a different variable rather than digging it out of a json blob?

source/Calamari.Common/Features/Discovery/AccountAuthenticationDetails.cs
```diff
@@ -9,15 +9,18 @@ public class AccountAuthenticationDetails<TAccountDetails> : ITargetDiscoveryAut
     {
         [JsonConstructor]
         public AccountAuthenticationDetails
-            (string type, string accountId, TAccountDetails accountDetails)
+            (string type, string accountId, string authenticationMethod, TAccountDetails accountDetails)
```
benPearce1, line 12: was the `AuthenticationMethod` and existing property on the JSON?

benPearce1 (approved):
This looks good. One comment on a JSON constructor

source/Calamari.Common/Features/Discovery/AccountAuthenticationDetails.cs
IsaacCalligeros95: Yep, it's set in server just not in Calamari.

IsaacCalligeros95: The tests didn't include this, which I updated and it does cause failures if unspecified now. I spoke to Scott yesterday and he was confident that it's there and always has been but just wasn't set in Calamari. I can play it safe and just have Service principal as a fall back if the authentication method is not set though

PR #1133, opened 2023-09-21
Use default proxy for MSAL authentication

mik-ky:
Uses a default proxy when authenticating to Azure.

Steps to check whether proxy is being used:
* Create an ARM template
* Configure Octopus server to use a custom proxy with your local proxy
* Verify access logs

**ARM deploy**
image

**Azure webapp (although deployment has failed for different reason)**
image

Also cherry-picks a commit which hasn't been applied to `release/2023.2` (but has been released as 25.5.9-hotfix3) in order to make sure it's not dropped.

Relates to https://github.com/OctopusDeploy/Issues/issues/8353


tothegills (approved):
Tested with Octopus custom proxy and IE proxy.

PR #1134, opened 2023-09-22
Include ScriptCS to run along side dotnet-script

IsaacCalligeros95:
We've seen a larger number than expected of customers having issues with net6.0 not being available and not being practical to update. As a result we've decided to add ScriptCS support back in, as a default and put dotnet-script behind a project variable, and feature toggle as a fall back. This PR re-introduces ScriptCS and adds the variable `Octopus.Action.Script.CSharp.UseDotnetScript`. 

When reviewing, most of the ScriptCS-specific changes are just re-adding previous code. The tests and `source/Calamari.Common/Features/Scripting/ScriptEngine.cs` are the core of this change.


source/Calamari.Common/Features/Scripting/ScriptEngine.cs
```diff
@@ -66,6 +67,7 @@ public ScriptSyntax[] GetSupportedTypes()
         /// </returns>
         IScriptWrapper BuildWrapperChain(ScriptSyntax scriptSyntax, IVariables variables)
         {
+            bool.TryParse(variables.Get(ScriptVariables.UseDotnetScript, "false"), out bool useDotnetScript);
```
IsaacCalligeros95, line 70: This is the core of the change. We're defaulting to ScriptCS for backward compatibility when on prem upgrades and the variable will be set based on an environment feature toggle, which can be over ridden by a project-specific variable.

source/Calamari.Tests/Fixtures/DotnetScript/DotnetScriptProxyFixture.cs
```diff
@@ -13,7 +15,7 @@ public class DotnetScriptProxyFixture : WindowsScriptProxyFixtureBase
     {
         protected override CalamariResult RunScript()
         {
-            return RunScript("Proxy.csx").result;
+            return RunScript("Proxy.csx", new Dictionary<string, string>() {{ ScriptVariables.UseDotnetScript, "true" }}).result;
```
benPearce1, line 18: nit: bool.TrueString

source/Calamari.Common/Features/Scripting/ScriptEngine.cs
benPearce1, line 70: nit: bool.FalseString

benPearce1 (approved):

PR #1119, opened 2023-08-31
Oidc Auth for Azure steps

IsaacCalligeros95:
This PR supports AzureOidc Accounts to Calamari for all Azure steps **excluding CloudService & ServiceFabric**. The associated Server changes can be found [here](https://github.com/OctopusDeploy/OctopusDeploy/pull/20042). 

As a part of this change, ServicePrincipal was implemented mostly per CalamariFlavour basis. This has been consolidated into the _Calamari.CloudAccounts_ repository, which is now shared across most Azure Calamari repositories. This adds some unnecessary overhead to the Calamari Bundles for some of these flavours, but the included Nuget Packages are relatively small ~1MB for the largest, and given the additional duplication introduced for AzureOidc Accounts looks to be worth the cost of consolidating.

[See Server PR for test cases](https://github.com/OctopusDeploy/OctopusDeploy/pull/20042).)
[sc-57210}

PR #1144, opened 2023-09-30
Fix web app ResourceManagerPublishProfileProvider GetPublishProperties

IsaacCalligeros95:
During OIDC work, this creation of `WebSiteManagementClient` mistakenly had the ResourceManaementEndpointBaseUri removed. This is causing failures to find WebApps in cases where the ResourceManagementEndpoint is not the default.

Rolling back change from https://github.com/OctopusDeploy/Calamari/pull/1119/files#diff-4633112ec25a33cf99c65bf5927db785b4db60c20354548b13a35cacc6f772f9R58

[sc-60756]

### DIFF
--- a/source/Calamari.GoogleCloudScripting.Tests/GoogleCloudActionHandlerFixture.cs
+++ b/source/Calamari.GoogleCloudScripting.Tests/GoogleCloudActionHandlerFixture.cs
@@ -55,7 +55,11 @@ namespace Calamari.GoogleCloudScripting.Tests
 
                 foreach (var result in results)
                 {
-                    if (result.TimeCreated.HasValue)
+                    // Checking date time less than to fetch gcloud versions earlier than 448.
+                    // 448 requires python 3.8 and up, currently 3.5 is available on Teamcity agents
+                    // This is intended as a temporary workaround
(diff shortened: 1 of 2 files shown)


source/Calamari.GoogleCloudScripting.Tests/GoogleCloudActionHandlerFixture.cs
```diff
@@ -55,13 +55,18 @@ private GoogleStorageObject RetrieveFileFromGoogleCloudStorage(StorageClient cli
 
                 foreach (var result in results)
                 {
-                    if (result.TimeCreated.HasValue)
+                    if (result.TimeCreated.HasValue);
```
veochen-octopus, line 58:
```suggestion
                    if (result.TimeCreated.HasValue)
```

This semicolon seems unintentional. I think it would terminate the `if` so the next code block executes unconditionally.

IsaacCalligeros95: 🤦‍♂️ Cheers for the catch, I've moved the date check inline with this and removed it

veochen-octopus (approved):

PR #1146, opened 2023-10-03
Bump Globfish

scme0:
Bumping Globfish to fix issue with directories ending in `.`. 

It also improves the performance by short-circuiting the algorithm when possible.

See https://github.com/OctopusDeploy/Globfish/pull/10 for details.


MissedTheMark (approved):
👍

PR #1150, opened 2023-10-06
Isaac/arm timeout 2023.2

IsaacCalligeros95:
Adds the default proxy to calls using the AzureResourceManagementClient.
Includes fix for release/2023.2 that is included in the latest tag, but not the release branch for S3 feeds downloading on targets. 

## Before
image (1)

## After
image

### DIFF
--- a/source/Calamari.AzureResourceGroup/DeployAzureResourceGroupBehaviour.cs
+++ b/source/Calamari.AzureResourceGroup/DeployAzureResourceGroupBehaviour.cs
@@ -75,7 +75,7 @@ namespace Calamari.AzureResourceGroup
             Func<Task<IResourceManagementClient>> createArmClient = async () =>
                                                               {
                                                                   var token = new TokenCredentials(await ServicePrincipal.GetAuthorizationToken(tenantId, clientId, password, resourceManagementEndpoint, activeDirectoryEndPoint));
-                                                                  var resourcesClient = new ResourceManagementClient(token)
+                                                                  var resourcesClient = new ResourceManagementClient(token, AuthHttpClientFactory.ProxyClientHandler())
                                                                   {
                                                                       SubscriptionId = subscriptionId,
                                                                       BaseUri = new Uri(resourceManagementEndpoint),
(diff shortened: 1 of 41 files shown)


source/Calamari.AzureResourceGroup/DeployAzureResourceGroupBehaviour.cs
```diff
@@ -75,7 +75,7 @@ public async Task Execute(RunningDeployment deployment)
             Func<Task<IResourceManagementClient>> createArmClient = async () =>
                                                               {
                                                                   var token = new TokenCredentials(await ServicePrincipal.GetAuthorizationToken(tenantId, clientId, password, resourceManagementEndpoint, activeDirectoryEndPoint));
-                                                                  var resourcesClient = new ResourceManagementClient(token)
+                                                                  var resourcesClient = new ResourceManagementClient(token, AuthHttpClientFactory.ProxyClientHandler())
```
benPearce1, line 78: What happens here if there is no proxy configured?
Does `WebRequest.DefaultWebProxy` being set to no proxy mean that it just works directly?

source/Calamari.Shared/Integration/Packages/Download/S3PackageDownloader.cs
```diff
@@ -64,6 +64,7 @@ string BuildFileName(string prefix, string version, string extension)
             }
 
             var cacheDirectory = PackageDownloaderUtils.GetPackageRoot(feedId);
+            fileSystem.EnsureDirectoryExists(cacheDirectory);
```
benPearce1, line 67: do the changes in this file belong somewhere else?

source/Calamari.AzureResourceGroup/DeployAzureResourceGroupBehaviour.cs
IsaacCalligeros95: Without a proxy configured, the ResourceManagementClient just makes calls regardless of the default proxy

IsaacCalligeros95: The images in the PR description show this

hnrkndrssn (approved):
lgtm 👍

PR #1151, opened 2023-10-06
Upissue octostache to 3.7.0

rain-on:
Octostache 3.7.0 supports using braces to enclose variable names inside of a `calc` block, thus allowing variables to appear on the left-hand-side of a divide and subtract operation


IsaacCalligeros95 (approved):
LGTM

PR #1155, opened 2023-10-10
Attempt to clean up the upload file if it was duplicated/renamed during deployment

benPearce1:
[sc-61082]

This is a workaround which probably requires a better solution.

Related to https://github.com/OctopusDeploy/Issues/issues/7382

PR #1159, opened 2023-10-11
Attempt to clean up the upload file if it was duplicated/renamed during deployment

benPearce1:
[sc-61082]

This is a workaround which probably requires a better solution.

Related to https://github.com/OctopusDeploy/Issues/issues/7382

other prs:
#1155


benPearce1:
seems I have gotten this wrong

PR #1160, opened 2023-10-11
Attempt to clean up the upload file if it was duplicated/renamed during deployment

benPearce1:
[sc-61082]

This is a workaround which probably requires a better solution.

Related to https://github.com/OctopusDeploy/Issues/issues/7382

Related to https://github.com/OctopusDeploy/Calamari/pull/1155

### DIFF
--- a/source/Calamari.AzureAppService/Behaviors/AzureAppServiceBehaviour.cs
+++ b/source/Calamari.AzureAppService/Behaviors/AzureAppServiceBehaviour.cs
@@ -5,6 +5,7 @@ using System.IO;
 using System.Linq;
 using System.Net.Http;
 using System.Net.Http.Headers;
+using System.Runtime.CompilerServices;
 using System.Threading.Tasks;
 using Calamari.AzureAppService.Azure;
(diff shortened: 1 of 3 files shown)


source/Calamari.AzureAppService/Behaviors/AzureAppServiceBehaviour.cs
```diff
@@ -5,6 +5,7 @@
 using System.Linq;
 using System.Net.Http;
 using System.Net.Http.Headers;
+using System.Runtime.CompilerServices;
```
IsaacCalligeros95, line 8: Is this required?

IsaacCalligeros95 (approved):
LGTM, with a minor comment.

source/Calamari.AzureAppService/Behaviors/AzureAppServiceBehaviour.cs
benPearce1: probably not, not sure where that came from

PR #1164, opened 2023-10-13
Isaac/enumerate files

IsaacCalligeros95:
A couple of customers have run into issues for IIS steps which specifically target netfx when running things like configuration transformation. Customers are deploying packages with ~1650 files and configuration transformation completion has gone from ~2 seconds to ~1 minute. This is causing issues as they deploy across 200+ tenants to a single deployment target and tasks are now blocking on each other taking deployment times from ~2 minutes to ~20. This change isn't ideal in the sense that it removes support for long file paths. As the original change was made to tidy up and bring consistency to this functionality I think this is safe to revert, we likely don't have customer relying on long file paths here. This change is up for discussion and whether it's worth it, so far we have this one confirmed case with a second similar issue raised that may or may not be related. I'm open to feedback on whether reverting this change is the correct thing to do, but this is a significant issue for this customer.

Fixes https://github.com/OctopusDeploy/Issues/issues/8406

Original change - https://github.com/OctopusDeploy/Calamari/pull/1044

### DIFF
--- a/source/Calamari.Common/Plumbing/FileSystem/FileOperations.cs
+++ b/source/Calamari.Common/Plumbing/FileSystem/FileOperations.cs
@@ -253,10 +253,12 @@ namespace Calamari.Common.Plumbing.FileSystem
             SearchOption searchOption,
             string[] searchPatterns)
         {
+            // Note we aren't using Alphaleonis.Win32.Filesystem.Directory.EnumerateFiles which handles long file paths due to performance issues.
+            var parentDirectoryInfo = new DirectoryInfo(parentDirectoryPath);
+
             return searchPatterns.Length == 0
-                ? Alphaleonis.Win32.Filesystem.Directory.EnumerateFiles(parentDirectoryPath, "*", searchOption)
-                : searchPatterns.SelectMany(pattern =>
-                    Alphaleonis.Win32.Filesystem.Directory.EnumerateFiles(parentDirectoryPath, pattern, searchOption));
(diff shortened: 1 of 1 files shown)


scme0 (approved):
I'm ok with this as it's how things were previously. 👍 

I wonder if we could fix this with an opt-out of long filepath support for users using the .net framework version of Calamari? Maybe via a feature toggle or something? This might solve the problem and improve performance overall for users who don't need long file path support.

At the same time, the old versions of .NET Framework are being removed in the future so maybe it's ok to have the inconsistency for now until those old versions are removed and then we can clean this up at the same time.

IsaacCalligeros95:
> I'm ok with this as it's how things were previously. 👍
> 
> I wonder if we could fix this with an opt-out of long filepath support for users using the .net framework version of Calamari? Maybe via a feature toggle or something? This might solve the problem and improve performance overall for users who don't need long file path support.
> 
> At the same time, the old versions of .NET Framework are being removed in the future so maybe it's ok to have the inconsistency for now until those old versions are removed and then we can clean this up at the same time.

An opt-out would be ideal, but I'd rather leave that for the future when we have a request around this issue. Is there an RFC or anything about the older .Net Framework deprecation changes? I haven't seen much about this and thought we'd be supporting some of these for quite a while.

scme0:
@IsaacCalligeros95 We have a deprecation plan document. I'll send it to you in Slack. I believe it shows we are going to deprecate net4 and net452 imminently.

PR #1167, opened 2023-10-19
Bypass Kubectl authentication when specified

APErebus:
# Background

For the Kubernetes Agent project, we need a way to disable Kubernetes authentication when we are running on a `KubernetesTentacle`.

This works because `kubectl` uses the container service account if no other authentication is provided. I have tested this locally and it works.

# Results

We are simply checking the Deployment Target Type Id to see what value it has. If it's set to the new target `KubernetesTentacle`, we skip auth as it will fail. All other target types will attempt to authenticate.

We've added a few unit test to ensure this check is working correctly.

Shortcut story: [sc-59583]

### DIFF
--- a/source/Calamari/Kubernetes/SetupKubectlAuthentication.cs
+++ b/source/Calamari/Kubernetes/SetupKubectlAuthentication.cs
@@ -18,12 +18,13 @@ namespace Calamari.Kubernetes
 {
     public class SetupKubectlAuthentication
     {
-        private const string KubernetesTentacleTargetTypeId = "KubernetesTentacle";
+        public const string KubernetesTentacleTargetTypeId = "KubernetesTentacle";
+        public const string SkippingAuthenticationMessage = "Running inside Kubernetes cluster, using pod service account for authentication";
 
         readonly IVariables variables;
         readonly ILog log;
         readonly ICommandLineRunner commandLineRunner;
-        private readonly Kubectl kubectl;
+        readonly IKubectl kubectl;
         readonly Dictionary<string, string> environmentVars;
         readonly string workingDirectory;
         string aws;
@@ -31,7 +32,7 @@ namespace Calamari.Kubernetes
         public SetupKubectlAuthentication(IVariables variables,
             ILog log,
             ICommandLineRunner commandLineRunner,
-            Kubectl kubectl,
+            IKubectl kubectl,
             Dictionary<string, string> environmentVars,
             string workingDirectory)
         {
@@ -47,7 +48,7 @@ namespace Calamari.Kubernetes
         {
             var errorResult = new CommandResult(string.Empty, 1);
 
-            foreach (var proxyVariable in ProxyEnvironmentVariablesGenerator.GenerateProxyEnvironmentVariables())
+            foreach (var proxyVariable in GenerateProxyEnvironmentVariables())
             {
                 environmentVars[proxyVariable.Key] = proxyVariable.Value;
             }
@@ -60,9 +61,9 @@ namespace Calamari.Kubernetes
 
             // If we are running on a Kubernetes Tentacle Target,
             // then we can rely on using the service account inside the pod for authentication.
-            if (variables.Get("Octopus.Machine.DeploymentTargetType") == KubernetesTentacleTargetTypeId)
+            if (variables.Get(MachineVariables.DeploymentTargetType) == KubernetesTentacleTargetTypeId)
             {
-                log.Info("Running inside Kubernetes cluster, using pod service account for authentication");
+                log.Info(SkippingAuthenticationMessage);
                 return new CommandResult(string.Empty, 0);
             }
 
@@ -561,6 +562,14 @@ namespace Calamari.Kubernetes
             return result.ExitCode == 0
                 ? captureCommandOutput.InfoLogs.ToArray()
                 : Enumerable.Empty<string>();
+        }
+
+        /// <remarks>
+        /// Seam for testing.
+        /// </remarks>
+        protected virtual IEnumerable<EnvironmentVariable> GenerateProxyEnvironmentVariables()
+        {
+            return ProxyEnvironmentVariablesGenerator.GenerateProxyEnvironmentVariables();
         }
     }
 }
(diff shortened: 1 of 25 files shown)


source/Calamari/Kubernetes/SetupKubectlAuthentication.cs
```diff
@@ -50,12 +52,22 @@ public CommandResult Execute(string accountType)
...
-            var kubeConfig = CreateKubectlConfig();
+            // Make sure Kubectl exists and can be run
             if (!kubectl.TrySetKubectl())
             {
                 return errorResult;
             }
 
+            // If we are running on a Kubernetes Tentacle Target,
+            // then we can rely on using the service account inside the pod for authentication.
+            if (variables.Get("Octopus.Machine.DeploymentTargetType") == KubernetesTentacleTargetTypeId)
```
scme0, line 63: @APErebus I've opted to just check if the current Deployment Target Type is the KubernetesTentacle. Does that sound reasonable to you?

scme0:
@APErebus I've been thinking about testing this and there isn't a really good way of doing that. Could this addition be tested by an E2E test of a Kubernetes Tentacle running against Server and executing a kubernetes deployment?

source/Calamari/Kubernetes/SetupKubectlAuthentication.cs
```diff
@@ -45,17 +48,27 @@ public CommandResult Execute(string accountType)
...
-            var kubeConfig = CreateKubectlConfig();
+            // Make sure Kubectl exists and can be run
             if (!kubectl.TrySetKubectl())
             {
                 return errorResult;
             }
 
+            // If we are running on a Kubernetes Tentacle Target,
+            // then we can rely on using the service account inside the pod for authentication.
+            if (variables.Get(MachineVariables.DeploymentTargetType) == KubernetesTentacleTargetTypeId)
```
zentron, line 64: Im assuming that this variable `MachineVariables.DeploymentTargetType` is different if the target is being used as a worker? 
Wondering what this would mean in future if we were running workloads on a KubernetesTarget worker but trying to connect to an external cluster? 
At the moment we will just ignore any of those other credentials used before. Is that what we want?

source/Calamari/Kubernetes/SetupKubectlAuthentication.cs
```diff
@@ -18,18 +18,21 @@ namespace Calamari.Kubernetes
...
         readonly IVariables variables;
         readonly ILog log;
         readonly ICommandLineRunner commandLineRunner;
-        private readonly Kubectl kubectl;
+        readonly IKubectl kubectl;
         readonly Dictionary<string, string> environmentVars;
         readonly string workingDirectory;
         string aws;
 
         public SetupKubectlAuthentication(IVariables variables,
```
zentron, line 32: This is becoming a mammoth file. Not a blocker but we may want to break it up a little if possible (or not, just boy-scout rule )

zentron (changes requested):

source/Calamari/Kubernetes/SetupKubectlAuthentication.cs
APErebus: In the original design of the PR we had a variable that was going to be contributed to by the deployment target.

@scme0 Do you know the answer to robs question?

scme0: If you are executing Kubernetes steps on a Kubernetes Tentacle as a worker you would be using a Kubernetes API target so the Deployment Target Type would be `"Kubernetes"` I believe.

To be honest we could make it stricter, we could say "Target Type" == "Kubernetes" do auth stuff, else if "KubernetesTentacle" skip and log, else throw exception "Unsupported Target for Kubernetes Authentication".

scme0: I guess people could be just adding their own variables in to get this auth to work but they really should be using the build in Kubernetes API target in that case.

source/Calamari/Kubernetes/SetupKubectlAuthentication.cs
scme0: That's fair enough, I'll take a look 👍

scme0: I'm going to do this in a separate few PRs. It's a bit of a scary part of the code because it does a lot of crazy stuff and while there are higher level auth integration tests, there are no lower level more detailed unit tests.

I've got a PR to add a bunch of Unit Tests and a subsequent PR to split out all the different auth context providers to make this class a lot neater.

source/Calamari/Kubernetes/SetupKubectlAuthentication.cs
scme0: @APErebus @zentron So after thinking about this more I'm still of the opinion that this is ok. The only way we allow users to authenticate against a kubernetes cluster is by using a kubernetes target (old or new). For new targets, they will not be executing on a worker because it's a tentacle and for the old one, it doesn't matter where it executes as it will be executing on behalf of the old target and this property will be set to "Kubernetes" and so it will run through the authentication as expected.

zentron: You could be right in terms of "it should just work", but we don't typically do machine specific behavioral logic within calamari and this could be unexpected, particularly when this approach still allows users to send credentials etc but it just gets ignored. 

My first thought was why not just work through the config flow and basically not create the file if no credentials are provided (which they shouldn't be in this case) so that the data itself drives the logic rather than a case of it just happening to work due to the nuances of how workers and targets are used.

In any case since this is a two way door decision, once you have had a quick investigation of a more data-driven approach suggested, I'm also fine for this comment to not be a blocker in moving forward for the time being.

My only additional comment was going to be lets make sure we have a log entry there but I can see you have that base covered already :)

source/Calamari/Kubernetes/SetupKubectlAuthentication.cs
zentron: >I've got a PR to add a bunch of Unit Tests

Nice, this was what I was hoping that breaking it up would lead to. The more we can make use of unit tests over integration tests the more stable coverage we will have.

zentron (approved):
Approved pending a quick revisit of the kubeconfig approach.

APErebus:
Closing as this PR is no longer relevant based on changes made in #1190

PR #1171, opened 2023-10-27
Isaac/iis multiplecerts

IsaacCalligeros95:
Adding support for multiple certificate stores in IIS. This change assumes the certificate store output variable is not depended on by customers

Fixes https://github.com/OctopusDeploy/Issues/issues/5096

## After
image
image

**Note: I have not been able to get certificates that are not in the `MY` store to show in the IIS web binding UI. The test output does receive the details **
image
image

image

**Test using Octopus provided certs, note still has the issue with the second Cert not showing in the SSL Certificate field**
image
image
image

### DIFF
--- a/source/Calamari/Deployment/Features/IisWebSiteBeforeDeployFeature.cs
+++ b/source/Calamari/Deployment/Features/IisWebSiteBeforeDeployFeature.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Linq;
 using System.Security.Cryptography.X509Certificates;
 using Calamari.Common.Commands;
 using Calamari.Common.Features.Deployment;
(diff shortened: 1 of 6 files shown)


source/Calamari/Deployment/Features/IisWebSiteBeforeDeployFeature.cs
```diff
@@ -57,7 +57,10 @@ static void EnsureCertificateInStore(IVariables variables, string certificateVar
                 storeName = AddCertificateToLocalMachineStore(variables, certificateVariable);
             }
 
-            Log.SetOutputVariable(SpecialVariables.Action.IisWebSite.Output.CertificateStoreName, storeName, variables);
+            var storeNamesVariable = variables.Get(SpecialVariables.Action.IisWebSite.Output.CertificateStoreName, "");
+            storeNamesVariable = string.Join(",", storeNamesVariable, storeName);
+
+            Log.SetOutputVariable(SpecialVariables.Action.IisWebSite.Output.CertificateStoreName, storeNamesVariable, variables);
```
IsaacCalligeros95, line 63: I haven't renamed this from CertificateStoreName -> CertificateStoreNames. This variable looks to be internal only  and changing from a single store name to a comma-delimited list is probably a larger change.  But I've left the variable name as is for consistency, happy to rename but figured it's a small thing either way

PR #1174, opened 2023-11-09
Handle all file extensions for S3 feeds in Calamari

IsaacCalligeros95:
Porting changes from https://github.com/OctopusDeploy/OctopusDeploy/pull/19534/files over to Calamari. This adds support for non standard file types outside the KnownExtensions of `".zip", ".tar.gz", ".tar.bz2", ".tar.gz", ".tgz", ".tar.bz" `. 

Fixes https://github.com/OctopusDeploy/Issues/issues/7668

Note I've ported the change from Server where we treat the file as a zip file in the cache. I think this is valid and still extracts to the target path as expected, at least in this case for a nupkg. Looking for feedback on this.

Not found in cache
image

When found in Cache
image

Extracted
image

.tar.gz on a WSL target
image


source/Calamari.Shared/Integration/Packages/Download/S3PackageDownloader.cs
```diff
@@ -87,35 +91,35 @@ string BuildFileName(string prefix, string version, string extension)
...
+                        if (foundFilePath != null)
+                        {
+                            fullFileName = foundFilePath;
+                        }
+
+                        if (fullFileName.IsNullOrEmpty())
                             throw new Exception($"Unable to download package {packageId} {version}: file not found");
 
-                        var localDownloadName = Path.Combine(cacheDirectory, PackageName.ToCachedFileName(packageId, version, knownExtension));
+                        var localDownloadName = Path.Combine(cacheDirectory, PackageName.ToCachedFileName(packageId, version, Extension));
```
IsaacCalligeros95, line 116: Note this extension type is hardcoded to a zip file, we do this in Server, but this may not be valid to do in Calamari. 

Is this reasonable for Linux workers?
Does this work for all package types?

egorpavlikhin: Good question and I wish I knew the answer 😆 From my understanding feeds don't always need to extract the package file but just pass it through, where the extension could be anything. Looking at the code though this seems wrong. What is supposed to happen is we extract the archive as it's stored in S3 and rezip it. Which wouldn't work with the case where we are just passing the file through. This needs some manual testing.

source/Calamari.Shared/Integration/Packages/Download/S3PackageDownloader.cs
```diff
@@ -87,35 +91,35 @@ string BuildFileName(string prefix, string version, string extension)
...
 
-                        if (!fileExists)
+                        var fullFileName = "";
+                        if (foundFilePath != null)
+                        {
+                            fullFileName = foundFilePath;
+                        }
+
+                        if (fullFileName.IsNullOrEmpty())
                             throw new Exception($"Unable to download package {packageId} {version}: file not found");
```
benPearce1, line 114:
```suggestion
                        var fullFileName = !foundFilePath.IsNullOrEmpty() ? foundFilePath : throw new Exception($"Unable to download package {packageId} {version}: file not found");
```

source/Calamari.Shared/Integration/Packages/Download/S3PackageDownloader.cs
IsaacCalligeros95: Looks like there were issues at least in the case of the double extension .tar.gz file types. I've updated this to iterate over the known file extensions, fall back to the file extension from the file itself (this doesn't handle double extensions hence it's usage as a fallback) and finally fall back to zip. I'm not sure if this is a potential issue in the Server implementation.

benPearce1 (approved):

PR #1178, opened 2023-11-17
Update Calamari to strip out the v from versions when parsing semvers

scme0:
# Background

Part of [sc-64777]

I've fixed the issue with parsing semver versions starting with a 'v' eg: `"v1.2.3"` in `Octopus.Versioning`. Unfortunately Octopus can't be updated to the newer version of Octopus.Versioning as it doesn't support net framework versions lower than net461. Once we update Calamari to net462, we can update Calamari to use the new version of `Versioning`.

# Result

This is a pretty crude workaround where we just remove the 'v' from the front of any version before passing them into any create methods in the Octopus.Versioning package. We only do this for semvers.

### DIFF
--- a/source/Calamari/Commands/DownloadPackageCommand.cs
+++ b/source/Calamari/Commands/DownloadPackageCommand.cs
@@ -6,6 +6,7 @@ using Calamari.Common.Features.Packages;
 using Calamari.Common.Features.Processes;
 using Calamari.Common.Features.Scripting;
 using Calamari.Common.Plumbing;
+using Calamari.Common.Plumbing.Extensions;
 using Calamari.Common.Plumbing.FileSystem;
 using Calamari.Common.Plumbing.Logging;
 using Calamari.Common.Plumbing.Variables;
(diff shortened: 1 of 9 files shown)


source/Calamari/Commands/DownloadPackageCommand.cs
```diff
@@ -152,7 +153,7 @@ public override int Execute(string[] commandLineArguments)
                 Guard.NotNullOrWhiteSpace(feedUri, "No feed URI was specified. Please pass --feedUri https://url/to/nuget/feed");
             }
 
-            version = VersionFactory.TryCreateVersion(packageVersion, versionFormat);
+            version = VersionFactory.TryCreateVersion(versionFormat == VersionFormat.Semver ? packageVersion.SanitiseSemVerString() : this.packageVersion, versionFormat);
```
APErebus, line 156: Is the additional `this` required here?

APErebus (commented):
Just 1 query on a `this`. If that's expected, then 👍

source/Calamari/Commands/DownloadPackageCommand.cs
scme0: Ah good call, not sure what happened there 🤔

scme0: Removed 👍

APErebus (approved):
LGTM

PR #1179, opened 2023-11-18
Remove Calamari.NonWindows.sln (all platforms to use Calamari.sln)

scme0:
# Background

# Results

Following the pattern from https://github.com/OctopusDeploy/OctopusTentacle, these updates allow users from MacOS and Linux to open the Calamari.sln and build without errors, rather than having to use a separate "NonWindows" solution which is missing projects.

[sc-64969]


tleed5 (approved):
Looks good to me and builds correctly on my M1 Macbook. Is there any changes in Teamcity that are required to support this change?

scme0:
@tleed5 Fortunately the solution is built by Nuke so all I had to do was update which Solution file is used in the Nuke build and then everything worked on TeamCity after that. It looks like the build is green so I'm pretty confident it's working as expected. Thanks for sanity checking on your machine! I also asked @EddyMoulton to sanity check on linux and it's all working as expected for him too 👌

PR #1180, opened 2023-11-18
Remove calamari tiny type implementation

scme0:
# Background

# Results

Remove Calamari TinyType implementation and add reference to the common implementation.

[sc-64965]


APErebus (approved):
LGTM!

PR #1182, opened 2023-11-18
Remove AlphaFS from Calamari

scme0:
# Background

# Results

Removing AlphaFS as all supported .NET versions support long paths natively.

[sc-64964]


source/Calamari.Tests/Fixtures/Integration/FileSystem/CalamariPhysicalFileSystemFixture.cs
```diff
@@ -366,30 +366,6 @@ public void EnumerateFilesWithGlobShouldIgnoreGroups(string directory, string gl
...
-            fileSystem.OverwriteFile("Some sample text", path + @"\test.txt");
-            fileSystem.DeleteFile(path + @"\test.txt");
-
-            while (paths.Any())
-            {
-                var pathToRemove = paths.Pop();
-                fileSystem.DeleteDirectory(pathToRemove);
-            }
-        }
-
```
scme0, line 392: It's not possible to test this with XUnit because we need to be able to set the Application Manifest (which is not possible for a unit test).

scme0: This will work "by default" now that we have a minimum support version of net462 and we already have the "LongFileSupport" in the Calamari manifest file.

APErebus (approved):
Less garbage!

PR #1183, opened 2023-11-22
Adding unit tests for SetupKubectlAuthentication

scme0:
# Background

Making changes in this file is a bit scary because there aren't many easy to run tests to ensure functionality has not changed.

# Results

I've added a bunch of tests so that we can refactor this test with a bit more assurance that we're not changing how things work.

I've made minimal changes to the production code to give us more certainty that I haven't broken anything. Unfortunately some changes were required (like making a bunch of classes take `IKubectl` instead of `Kubectl`) and some changes made testing cleaner (like updating the code to use `SpecialVariables` instead of magic strings).

# Coverage

Getting full coverage wasn't specifically the aim but the report looks good with `SetupKubectlAuthentication` getting ~94% coverage and other related classes also being well covered:
Screenshot 2023-11-23 at 10 23 11

[sc-65355]


APErebus (approved):
Looks good to me! Thanks :)

PR #1184, opened 2023-11-22
Refactor kubernetes auth by splitting out context providers to make it easier to reason about

scme0:
# Background

Building off the addition of unit tests, we can now split this code out and use the tests to ensure we aren't breaking anything.

# Results

I've opted to split out the different context providers so that it's easier to follow what's going on. Before there were different overlapping requirements which made it really difficult to follow the flow for a single context provider.

[sc-65355]


source/Calamari.Tests/KubernetesFixtures/Approved/KubernetesContextScriptWrapperFixture.ExecutionShouldFailWhenAccountTypeNotValid.approved.txt
```diff
@@ -2,8 +2,4 @@
 [Verbose] Temporary kubectl config set to <path>kubectl-octo.yml
 [Verbose] "kubectl" version --client --output=yaml --request-timeout=1m
 [Verbose] Found kubectl and successfully verified it can be executed.
-[Verbose] "kubectl" config set-cluster octocluster --server=<server> --request-timeout=1m
-[Verbose] "kubectl" config set-context octocontext --user=octouser --cluster=octocluster --namespace=calamari-testing --request-timeout=1m
-[Verbose] "kubectl" config use-context octocontext --request-timeout=1m
-[Verbose] "kubectl" config set-cluster octocluster --insecure-skip-tls-verify=true --request-timeout=1m
```
scme0, line 8: It no longer makes any kubectl config calls if the account type is not valid - which is better.

source/Calamari.Tests/KubernetesFixtures/Authentication/SetupKubectlAuthenticationFixture.cs
```diff
@@ -181,6 +181,36 @@ public void Execute_WithUsernamePasswordAndCertificates_ConfiguresAuthentication
             }, opts => opts.WithStrictOrdering());
         }
 
+        [Test]
+        public void Execute_WithOnlyCertificates_ConfiguresAuthenticationCorrectly()
```
scme0, line 185: I missed this case in the previous PR. This case is when you provide now account type but you do provide a set of certificates.

source/Calamari/Kubernetes/ContextProviders/AwsContextProvider.cs
```diff
@@ -0,0 +1,53 @@
...
+            this.workingDirectory = workingDirectory;
+        }
+
+        public bool TrySetContext(string @namespace, string accountType, string clusterUrl, string clientCert, string skipTlsVerification)
+        {
+            var eksUseInstanceRole = variables.GetFlag(AwsSpecialVariables.Authentication.UseInstanceRole);
+            if (accountType != AccountTypes.AmazonWebServicesAccount && !eksUseInstanceRole)
+                return false;
+
+            if (!new CertificateAuthenticationContextProvider(variables, log, kubectl)
```
zentron, line 43: Why is this one not out on its own?

source/Calamari/Kubernetes/ContextProviders/CertificateAuthenticationContextProvider.cs
```diff
@@ -0,0 +1,106 @@
...
+            this.kubectl = kubectl;
+        }
+
+        public bool TrySetContext(
+            string @namespace,
+            string clusterUrl,
+            string clientCert,
+            string skipTlsVerification)
+        {
+            kubectl.ExecuteCommandAndAssertSuccess("config",
```
zentron, line 30: All the others providers have a  `try-set` pattern where they first perform some test, and then perform a command to modify the context.
Why is this one different?

source/Calamari/Kubernetes/ContextProviders/AwsContextProvider.cs
scme0: Certificates Auth is weird, it's used in addition to Aws, UserPass and Token Auth, or by itself.

source/Calamari/Kubernetes/ContextProviders/CertificateAuthenticationContextProvider.cs
scme0: I think the others have an obvious thing to check but this one doesn't. I guess we could just check all the required certs are there at the beginning. Probably worth adjusting 👍

APErebus:
This is supersceeded by #1184

PR #1185, opened 2023-11-23
Bypass auth if no auth variables are set

scme0:
# Background

Part of #project-k8s-agent

When a user is doing a kubernetes deployment using the Kubernetes Tentacle, they will have no authentication variables set. We need to allow this as a valid case so that the kubernetes tentacle can use the ambient authentication it has in the cluster.

[sc-59583]

# Results

This PR adjusts the `SetupKubectlAuthentication` to succeed if no properties are set. It will also remove the "KUBECONFIG" Environment Variable that was created at the start of the process.

# Known Issues

There are currently two known issues with ambient auth which haven't been fixed yet:
1. https://app.shortcut.com/octopusdeploy/story/65557/when-deploying-to-kubernetes-with-ambient-auth-context-if-no-namespace-is-specified-the-default-will-be-the-namespace-where
2. https://app.shortcut.com/octopusdeploy/story/65597/kubectl-get-create-don-t-seem-to-work-with-ambient-auth


APErebus:
Closing as this PR is no longer relevant based on changes made in #1190

PR #1186, opened 2023-11-23
Fix Azure Cloud Service tests which are using the Azure_OctopusAPITester_Certificate env var

scme0:
# Background

Slack thread: https://octopusdeploy.slack.com/archives/C03SG0LFJHX/p1700708276730729

# Results

Turns out team city is adding a trailing space on the env var which makes it fail to parse as a base64 string.

Adding a trim should solve this issue to remove the trailing space.

[sc-65471]


APErebus (approved):
Thanks mate

PR #1187, opened 2023-11-24
SPIKE: Fix ambient auth bugs

scme0:
# Background

This PR adds the ability for Calamari to do ambient auth (eg: if no variables for auth are set, don't fail, just assume that auth is given ambiently)

There are two bugs that have come out of that:
1. https://app.shortcut.com/octopusdeploy/story/65557/when-deploying-to-kubernetes-with-ambient-auth-context-if-no-namespace-is-specified-the-default-will-be-the-namespace-where
2. https://app.shortcut.com/octopusdeploy/story/65597/kubectl-get-create-don-t-seem-to-work-with-ambient-auth

# Results

This PR has random hacks that I used to try and figure out what was going on and why it wasn't working properly. I'll leave comments on bits that might be of interest.

### DIFF
--- a/source/Calamari/Kubernetes/Commands/Executors/GatherAndApplyRawYamlExecutor.cs
+++ b/source/Calamari/Kubernetes/Commands/Executors/GatherAndApplyRawYamlExecutor.cs
@@ -13,6 +13,7 @@ using Calamari.Common.Plumbing.ServiceMessages;
 using Calamari.Common.Plumbing.Variables;
 using Calamari.Kubernetes.Integration;
 using Calamari.Kubernetes.ResourceStatus.Resources;
+using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using Octopus.CoreUtilities.Extensions;
 
@@ -46,6 +47,11 @@ namespace Calamari.Kubernetes.Commands.Executors
             try
             {
                 var variables = deployment.Variables;
+                var @namespace = variables.Get(SpecialVariables.Namespace);
+                if (@namespace == null)
+                {
+                    @namespace = "default";
+                }
                 var globs = variables.GetPaths(SpecialVariables.CustomResourceYamlFileName);
                 if (globs.IsNullOrEmpty())
                     return true;
@@ -53,7 +59,7 @@ namespace Calamari.Kubernetes.Commands.Executors
                 var resources = new HashSet<Resource>();
                 foreach (var directory in directories)
                 {
-                    var res = ApplyBatchAndReturnResources(directory).ToList();
+                    var res = ApplyBatchAndReturnResources(directory, @namespace).ToList();
                     if (appliedResourcesCallback != null)
                     {
                         await appliedResourcesCallback(res.Select(r => r.ToResourceIdentifier()).ToArray());
@@ -135,7 +141,7 @@ namespace Calamari.Kubernetes.Commands.Executors
             return directories;
         }
 
-        private IEnumerable<Resource> ApplyBatchAndReturnResources(GlobDirectory globDirectory)
+        private IEnumerable<Resource> ApplyBatchAndReturnResources(GlobDirectory globDirectory, string @namespace)
         {
             var index = globDirectory.Index;
             var directoryWithTrailingSlash = globDirectory.Directory + Path.DirectorySeparatorChar;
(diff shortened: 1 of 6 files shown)


source/Calamari/Kubernetes/Commands/Executors/GatherAndApplyRawYamlExecutor.cs
```diff
@@ -151,9 +157,16 @@ private IEnumerable<Resource> ApplyBatchAndReturnResources(GlobDirectory globDir
...
             }
 
-            var result = kubectl.ExecuteCommandAndReturnOutput("apply", "-f", $@"""{globDirectory.Directory}""", "--recursive", "-o", "json");
+            log.Info($"SCME ApplyBatchAndReturnResources: {kubectl.GetHashCode()}");
+            log.Info($"SCME ApplyBatchAndReturnResources EnvVars: '{JsonConvert.SerializeObject(kubectl.environmentVars)}'");
+            log.Info($"SCME ApplyBatchAndReturnResources workingDirectory: '{kubectl.workingDirectory}'");
+            if (!TryExecuteCommandWithVerboseLoggingOnly("get", "namespace", @namespace))
+                TryExecuteCommandWithVerboseLoggingOnly("create", "namespace", @namespace);
+
+            var result = kubectl.ExecuteCommandAndReturnOutput("apply", "-f", $@"""{globDirectory.Directory}""", "--recursive", "-o", "json", "-n", @namespace);
```
scme0, line 169: I tried adding `-n <namespace>` but it won't work if the namespace is defined in the users yaml already (and they don't match).

source/Calamari/Kubernetes/Commands/KubernetesDeploymentCommandBase.cs
```diff
@@ -124,7 +124,7 @@ public override int Execute(string[] commandLineArguments)
 
             conventions.AddRange(CommandSpecificInstallConventions());
 
-            var runningDeployment = new RunningDeployment(pathToPackage, variables);
+            var runningDeployment = new RunningDeployment(pathToPackage, variables, GetEnvironmentVariables());
```
scme0, line 127: I tried setting external EnvironmentVariables into the environment variables used by Kubectl but it didn't fix the `kubectl get/create namespace ...` commands.

scme0: To be honest, I have no idea how `kubectl apply ...` does work if it has absolutely no environment variables set. Maybe it looks in a specific folder for config?

scme0: We should probably look at setting ambient environment variables like this... maybe 🤔

source/Calamari/Kubernetes/SetupKubectlAuthentication.cs
```diff
@@ -66,7 +70,7 @@ public CommandResult Execute(string accountType)
                 return CommandResult.Failure($"Unable to setup auth context for accountType {accountType}");
             }
 
-            if (!CreateNamespace(@namespace))
+            if (@namespace != "default" && !CreateNamespace(@namespace))
```
scme0, line 73: Do we need to check the namespace if it's `default`?

source/Calamari/Kubernetes/SetupKubectlAuthentication.cs
```diff
@@ -152,10 +156,38 @@ bool TrySetupContext(string kubeConfig, string @namespace, string accountType)
 
         bool CreateNamespace(string @namespace)
         {
-            if (TryExecuteCommandWithVerboseLoggingOnly("get", "namespace", @namespace))
+            log.Info($"SCME CreateNamespace: {kubectl.GetHashCode()}");
+            log.Info($"SCME CreateNamespace EnvVars: '{JsonConvert.SerializeObject(kubectl.environmentVars)}'");
+            log.Info($"SCME CreateNamespace workingDirectory: '{kubectl.workingDirectory}'");
+
+            // Load from in-cluster configuration:
+            var config = KubernetesClientConfiguration.InClusterConfig();
```
scme0, line 164: I hacked in the use of [KubernetesClient](https://github.com/kubernetes-client/csharp) to see if it would work, and it did! I had a look in the `InClusterConfig()` method and found that it was setting a bunch of things from environment variables so we might be able to do something similar to solve this problem, or we just update calamari to use KubernetesClient?

source/Calamari/Kubernetes/SetupKubectlAuthentication.cs
```diff
@@ -200,7 +232,14 @@ void ExecuteCommand(string executable, params string[] arguments)
...
         {
-            return ExecuteCommandWithVerboseLoggingOnly(new CommandLineInvocation(kubectl.ExecutableLocation, arguments.Concat(new[] { "--request-timeout=1m" }).ToArray())).ExitCode == 0;
+            var result = kubectl.ExecuteCommandAndReturnOutput(
+                arguments.Concat(new[] { "--request-timeout=1m" }).ToArray());
+            foreach (var message in result.Output.Messages)
+            {
+                log.Verbose(message.Text);
+            }
+
+            return result.Result.ExitCode == 0;
```
scme0, line 242: This is a genuinely nice clean up that we should do anyway It gets the same result as the whole method below as far as I can tell.

PR #1188, opened 2023-11-27
Robe/k8s cert refactor

zentron:
## Background
Ultimately the different auth mechanisms setup in `source/Calamari/Kubernetes/SetupKubectlAuthentication.cs` are mutually exclusive and depend on the authentication type set on the target in server
image

Unfortunately, for whatever reason we dont send back the authentication type as a variable on its own and so some of this code does checks on other properties that _would_ have been set. 

The code as it was written was a disorganised mess of multiple conditional checks which is unecessary when the different code flows are mutually exclusive.

## How to review this
I deliberately went through and made each change in its own self-contained commit to help the reviewer better undersand each refactor change that took place to illustrate the (relatively) non-changing refactor steps that took place. 

The only things really that would change is that in some cases the order of the `kubectl config set-context` commands have changed order, and some simpler assumptions can be made that, for example we wont be setting both Google Auth and Client Cert etc.

### DIFF
--- a/source/Calamari.Tests/KubernetesFixtures/Approved/KubernetesContextScriptWrapperFixture.ExecutionUsingPodServiceAccount_WithServerCert.approved.txt
+++ b/source/Calamari.Tests/KubernetesFixtures/Approved/KubernetesContextScriptWrapperFixture.ExecutionUsingPodServiceAccount_WithServerCert.approved.txt
@@ -3,9 +3,9 @@
 [Verbose] "kubectl" version --client --output=yaml --request-timeout=1m
 [Verbose] Found kubectl and successfully verified it can be executed.
 [Verbose] "kubectl" config set-cluster octocluster --server=<server> --request-timeout=1m
+[Verbose] "kubectl" config set-cluster octocluster --certificate-authority=<certificateAuthorityPath> --request-timeout=1m
 [Verbose] "kubectl" config set-context octocontext --user=octouser --cluster=octocluster --namespace=calamari-testing --request-timeout=1m
 [Verbose] "kubectl" config use-context octocontext --request-timeout=1m
-[Verbose] "kubectl" config set-cluster octocluster --certificate-authority=<certificateAuthorityPath> --request-timeout=1m
 [Info] Creating kubectl context to <server> (namespace calamari-testing) using a Pod Service Account Token
 [Verbose] "kubectl" config set-credentials octouser --token=<token> --request-timeout=1m
 [Verbose] "kubectl" get namespace calamari-testing --request-timeout=1m
--- a/source/Calamari.Tests/KubernetesFixtures/Approved/KubernetesContextScriptWrapperFixture.ExecutionUsingPodServiceAccount_WithServerCert.approved.txt
+++ b/source/Calamari.Tests/KubernetesFixtures/Approved/KubernetesContextScriptWrapperFixture.ExecutionUsingPodServiceAccount_WithServerCert.approved.txt
@@ -3,9 +3,9 @@
 [Verbose] "kubectl" version --client --output=yaml --request-timeout=1m
 [Verbose] Found kubectl and successfully verified it can be executed.
 [Verbose] "kubectl" config set-cluster octocluster --server=<server> --request-timeout=1m
-[Verbose] "kubectl" config set-cluster octocluster --certificate-authority=<certificateAuthorityPath> --request-timeout=1m
 [Verbose] "kubectl" config set-context octocontext --user=octouser --cluster=octocluster --namespace=calamari-testing --request-timeout=1m
 [Verbose] "kubectl" config use-context octocontext --request-timeout=1m
+[Verbose] "kubectl" config set-cluster octocluster --certificate-authority=<certificateAuthorityPath> --request-timeout=1m
 [Info] Creating kubectl context to <server> (namespace calamari-testing) using a Pod Service Account Token
 [Verbose] "kubectl" config set-credentials octouser --token=<token> --request-timeout=1m
 [Verbose] "kubectl" get namespace calamari-testing --request-timeout=1m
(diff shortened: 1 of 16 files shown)


source/Calamari/Kubernetes/SetupKubectlAuthentication.cs
```diff
@@ -163,25 +163,31 @@ bool TrySetupContext(string kubeConfig, string @namespace, string accountType)
...
                 const string cluster = "octocluster";
                 const string context = "octocontext";
+                
+                kubectl.ExecuteCommandAndAssertSuccess("config", "set-cluster", cluster, $"--server={clusterUrl}");
+                kubectl.ExecuteCommandAndAssertSuccess("config", "set-context", context, $"--user={user}", $"--cluster={cluster}", $"--namespace={@namespace}");
+                kubectl.ExecuteCommandAndAssertSuccess("config", "use-context", context);
+                
                 if (isUsingPodServiceAccount)
                 {
-                    SetupContextUsingPodServiceAccount(@namespace, cluster, clusterUrl, serverCert,
```
zentron, line 175: This method is removed in a later commit

source/Calamari.Tests/KubernetesFixtures/Approved/KubernetesContextScriptWrapperFixture.ExecutionUsingPodServiceAccount_WithServerCert.approved.txt
```diff
@@ -3,9 +3,9 @@
 [Verbose] "kubectl" version --client --output=yaml --request-timeout=1m
 [Verbose] Found kubectl and successfully verified it can be executed.
 [Verbose] "kubectl" config set-cluster octocluster --server=<server> --request-timeout=1m
-[Verbose] "kubectl" config set-cluster octocluster --certificate-authority=<certificateAuthorityPath> --request-timeout=1m
 [Verbose] "kubectl" config set-context octocontext --user=octouser --cluster=octocluster --namespace=calamari-testing --request-timeout=1m
 [Verbose] "kubectl" config use-context octocontext --request-timeout=1m
+[Verbose] "kubectl" config set-cluster octocluster --certificate-authority=<certificateAuthorityPath> --request-timeout=1m
```
zentron, line 8: Shuffling around the order of some of the config results in a sightly different order of kubectl commands. This _shouldnt_ really matter

source/Calamari/Kubernetes/Authentication/GoogleKubernetesEngineAuth.cs
```diff
@@ -31,7 +31,7 @@ public GoogleKubernetesEngineAuth(GCloud gcloudCli,GkeGcloudAuthPlugin authPlugi
             this.log = log;
         }
 
-        public bool TryConfigure(bool useVmServiceAccount, string @namespace)
+        public bool TryConfigure(string @namespace)
```
zentron, line 34: I _really_ want to get rid of the `bool Try...` semantics in a later commit. returning true or false everywhere rather than just throwing an exception feels unecessary when it typically is the result of some exceptional circumstances that we want to just stop continued execution on

source/Calamari.Tests/KubernetesFixtures/Authentication/SetupKubectlAuthenticationFixture.cs
```diff
@@ -133,54 +133,7 @@ public void Execute_WithNoClusterUrl_Fails()
 
             log.Received().Error("Kubernetes cluster URL is missing");
         }
-
-        [TestCase(true)]
-        [TestCase(false)]
-        public void Execute_WithUsernamePasswordAndCertificates_ConfiguresAuthenticationCorrectly(bool skipServerCertificate)
```
zentron, line 139: This test is actually not particularly meaningful since you can either authenticate with UsernamePassword, _or_ ClientCertificate. They are both different auth mechanisms and although im willing to assume that kubectl might not have a problem with setting multiple auth settings, as far as our data model is concerned, its one or the other.

Its this kind of mixing of mutually exclusive authentication mechanisms that have made the auth code so hard to follow and complex. 

Both support setting the _Server cert_, since that is used for TLS verification, but client cert is the authentication mechanism that the server uses to validate the client is who it says it is (in contrast to Username and passwrd)
image
image

source/Calamari.Tests/KubernetesFixtures/Authentication/SetupKubectlAuthenticationFixture.cs
```diff
@@ -195,7 +148,7 @@ public void Execute_WithMissingCertificateData_Fails(bool missingClientPem, bool
 
             var sut = CreateSut();
 
-            var result = sut.Execute(AccountTypes.UsernamePassword);
+            var result = sut.Execute(null);
```
zentron, line 151: Similar to above, you cant set UsernamePassword then test ClientCert logic

source/Calamari/Kubernetes/SetupKubectlAuthentication.cs
```diff
@@ -77,7 +75,7 @@ public CommandResult Execute(string accountType)
             if (!CreateNamespace(@namespace))
             {
                 log.Verbose("Could not create namespace. Continuing on, as it may not be working directly with the target.");
-            }
+            };
```
APErebus, line 78: Snuck in?

source/Calamari/Kubernetes/SetupKubectlAuthentication.cs
```diff
@@ -162,118 +97,143 @@ bool TrySetupContext(string kubeConfig, string @namespace, string accountType)
...
+                if (variables.GetFlag(SpecialVariables.SkipTlsVerification))
                 {
-                    SetupContextUsingPodServiceAccount(@namespace, cluster, clusterUrl, serverCert,
-                        skipTlsVerification,
-                        serverCertPath,
-                        context,
-                        user,
-                        podServiceAccountToken);
+                    kubectl.ExecuteCommandAndAssertSuccess("config", "set-cluster", cluster, $"--insecure-skip-tls-verify=true");
+                } else if (variables.IsSet(SpecialVariables.CertificateAuthorityPath))
```
APErebus, line 126: Minor nit on formatting

APErebus (approved):
Looks good to me! Thanks @zentron

PR #1189, opened 2023-12-01
Add support for specifying the chart directory for helm

geofflamrock:
Currently when running the Upgrade Helm Chart step, there are a set of locations that are searched for the chart:
- The directory where the package is installed
- A directory inside this named the same as the package id. This is to match the convention that the helm executable uses when packaging the chart.
- Each sub-directory

As part of adding support for sourcing files from Git repositories for the Upgrade helm chart step, we want to be allow users to specify the directory within the repository where their chart is. A Git repository might have multiple charts, and the "package id" is set to the details of the repository, so we won't match it. We don't want to use the first chart we find at either the top level or within any sub-directory.

This PR adds support for this, if the property `Octopus.Action.Helm.ChartDirectory` is specified, we'll attempt to find the chart.yaml within that. If we don't find it we'll throw an error. If the property isn't specified then we'll continue on to the rest of the search above.


source/Calamari/Kubernetes/Conventions/HelmUpgradeConvention.cs
```diff
@@ -291,37 +291,63 @@ string GetChartLocation(RunningDeployment deployment)
...
+            // Try the specific chart directory if the variable has been set
+            if (!string.IsNullOrEmpty(chartDirectoryVariable))
+            {
+                log.Verbose($"Attempting to find chart using configured directory '{chartDirectoryVariable}'");
+
+                var chartDirectory = Path.Combine(installDir, chartDirectoryVariable);
+                if (fileSystem.DirectoryExists(chartDirectory) && fileSystem.FileExists(Path.Combine(chartDirectory, "Chart.yaml")))
+                {
+                    log.Verbose($"Using chart found in configured directory '{chartDirectory}'");
+                    return chartDirectory;
```
dylanlerch, line 305: It's called out in the description that if someone has specified a directory and we don't find one, we keep checking other places. What's the reason for this decision? Would it be more predictable behaviour to just error rather than just falling back to other behaviour?

Perhaps I just don't know how this is used. But we have a file at the root and in some folder, so someone specifies a file in the specific folder. Everything is working well for many months. Someone then moves that file and forgets to update the chart directory. This would still work, but we've just unexpectedly (I think) picked up a different file.

source/Calamari.Tests/KubernetesFixtures/HelmUpgradeFixture.cs
```diff
@@ -325,6 +325,18 @@ public void AdditionalArgumentsPassed()
             result.AssertOutputMatches("[helm|\\\\helm\"] upgrade (.*) --dry-run");
         }
 
+        [Test]
+        [RequiresNonFreeBSDPlatform]
+        [RequiresNon32BitWindows]
+        [RequiresNonMac]
+        [Category(TestCategory.PlatformAgnostic)]
+        public void WhenTheChartDirectoryVariableIsSet_TheChartAtThatLocationIsUsed()
```
dylanlerch, line 333: Would it be good to have some tests for the fallback behaviour. Variable is set, but didn't find it.

I've got a separate comment with my opinions on what we should do in that case, but either way would be good to have a test for it. Or are these really expensive?

source/Calamari.Tests/KubernetesFixtures/HelmUpgradeFixture.cs
```diff
@@ -325,6 +325,18 @@ public void AdditionalArgumentsPassed()
...
+        {
+            Variables.Set(Kubernetes.SpecialVariables.Helm.ChartDirectory, "specific/location/for/my/chart");
+            var result = DeployPackage("mychart-with-specific-location-0.3.8.tar.gz");
```
dylanlerch, line 336: For extra points, would it be possible to have a way to create these archives in code so we don't have to check in a binary? Or like 'nah get stuffed mate, this is good enough'?

source/Calamari/Kubernetes/Conventions/HelmUpgradeConvention.cs
geofflamrock: Ah right, yeah this is a good point. The user is being specific about the chart location, and if it isn't there would be it be surprising to go look elsewhere? The answer is probably yes, if they want the fallback behaviour of searching for chart then don't opt into the specific location. Thanks for raising this, I'll make this change and adjust elsewhere as needed with other tests.

source/Calamari.Tests/KubernetesFixtures/HelmUpgradeFixture.cs
geofflamrock: It would be nice, I spent a bunch of time mucking around getting the right archive shape. I'll take a quick look to see if we have something already, else prob won't worry too much.

source/Calamari/Kubernetes/Conventions/HelmUpgradeConvention.cs
geofflamrock: Have made this change and added tests

dylanlerch (approved):

APErebus (approved):
Looks good to me! good stuff @geofflamrock

PR #1190, opened 2023-12-04
Support using ambient K8S auth when target is running inside a cluster

zentron:
Replaces https://github.com/OctopusDeploy/Calamari/pull/1187 and https://github.com/OctopusDeploy/Calamari/pull/1185

This change adds support for using ambient Kube credentials if it is being executed from inside a kubernetes cluster.

## Background
When a kuberentes pod is created in a cluster, the controller typically mounts some service principal certificates and sets some target environment variables that allows kubectl to fallback to these details to connect to the cluster of the standard config file is also missing. 

Shortcut story: [sc-59583], [sc-65355]

## Changes
* Added[ the change](https://github.com/OctopusDeploy/Calamari/pull/1190/files#diff-8c1ea2409377bb4363a462b8a9f7c484ea6305fd3504cd411b897d42decb382bR201) to support for ambient auth
* Remove a bunch of `bool TryX` methods to instead just throw to simplify the code flows. 
* Tidy up and rearrange the auth process

## WIP
There is an integration test in progress however due to the complexity of setting the right environment to invoke this behavior (need to run our tests in a K8S cluster) it is still in progress.


source/Calamari.Tests/KubernetesFixtures/Authentication/AmbientKubeubernetesAuthFixture.cs
```diff
@@ -0,0 +1,94 @@
+﻿using Calamari.Common.Features.Processes;
+using Calamari.Common.Plumbing.Variables;
+using Calamari.Testing.Helpers;
+using NUnit.Framework;
+
+namespace Calamari.Tests.KubernetesFixtures.Authentication
+{
+    [TestFixture]
+    public class AmbientKubeubernetesAuthFixture
```
zentron, line 9: This test is still WIP. 
Testing this functionality will be non-trivial, in that we need to provision a cluster, and then mount the _actual_ tests into and then invoke and read the results.

The side benifit is that we can use this local K8S cluster for testing some of the other components.

source/Calamari.Tests/KubernetesFixtures/Authentication/SetupKubectlAuthenticationFixture.cs
```diff
@@ -96,7 +97,7 @@ public void Setup()
             });
             kubectl = Substitute.For<IKubectl>();
             kubectl.ExecutableLocation.Returns("kubectl");
-            kubectl.TrySetKubectl().Returns(true);
```
zentron, line 99: Removed this `TryX` semantics and simplified to `X` and exceptions are thrown if things aren't configured as we would expect

source/Calamari.Tests/KubernetesFixtures/Authentication/SetupKubectlAuthenticationFixture.cs
```diff
@@ -160,12 +161,12 @@ public void Execute_WhereNamespaceDoesNotExist_NamespaceCreated()
         {
             variables.Set(SpecialVariables.ClusterUrl, ClusterUrl);
             variables.Set(SpecialVariables.Namespace, Namespace);
-
+            variables.Set(Deployment.SpecialVariables.Account.AccountType, AccountTypes.UsernamePassword);
             invocations.FailFor("kubectl", $"get namespace {Namespace} --request-timeout=1m");
 
             var sut = CreateSut();
 
-            var result = sut.Execute(AccountTypes.UsernamePassword);
```
zentron, line 168: Since we dont decide the auth mechanism on the account type alone (it also uses other variables) it seems wrong to pass the account type itself through the execute function.

source/Calamari/Kubernetes/SetupKubectlAuthentication.cs
```diff
@@ -46,148 +46,76 @@ public class SetupKubectlAuthentication
...
-            }
+            return new CommandResult(string.Empty, 0);
+        }
 
-            var outputKubeConfig = variables.GetFlag(SpecialVariables.OutputKubeConfig);
-            if (outputKubeConfig)
-            {
-                kubectl.ExecuteCommandAndAssertSuccess("config", "view");
-            }
+        string DetermineNamespace()
```
EddyMoulton, line 78: nit: 
```suggestion
        string GetNamespaceOrDefault()
```

Might be more descriptive

EddyMoulton (commented):
The refactoring looks good to me. Much, much cleaner.
Seems like it wouldn't be very much effort to carve out the ambient auth part if you wanted to get the refactoring merged in sooner.

source/Calamari.Tests/KubernetesFixtures/Authentication/SetupKubectlAuthenticationFixture.cs
EddyMoulton: My personal preference would be to go the other way and front load more variable data, then pass some kind of `IKubectlAuthenticationParams` with an implementation for each account type. That way we'll do less plucking of variables mid method.

I had a brief look at how that might end up though and feels like a bit of a can of worms, so I'm leaving this comment as food for thought but I'm not convinced it would be a worthwhile change.

source/Calamari.Tests/KubernetesFixtures/Authentication/AmbientKubeubernetesAuthFixture.cs
EddyMoulton, line 1: Also the filename for this is borked :grin:

source/Calamari/Kubernetes/SetupKubectlAuthentication.cs
zentron: Makes sense :)

source/Calamari.Tests/KubernetesFixtures/Authentication/SetupKubectlAuthenticationFixture.cs
zentron: This is the nature of the variable bag structure. What variables are needed depend entirely on the different auth mechanism. This makes it very clear that _everything_ is up for grabs :)

source/Calamari.Tests/KubernetesFixtures/Authentication/AmbientKubeubernetesAuthFixture.cs
EddyMoulton, line 1: Is this worth commiting if it's not going to be completed here?
If the work that you've done is going to bootstrap whoever looks at this test next then we can probably leave it in with a note but otherwise I'd get rid of it from this PR.

EddyMoulton (approved):

source/Calamari.Tests/KubernetesFixtures/Authentication/AmbientKubeubernetesAuthFixture.cs
APErebus: I removed this and will open a new PR for k8s changes

PR #1192, opened 2023-12-04
Fix positional Parameter test

IsaacCalligeros95:
PSCore7.4.0 introduces ambiguity between -p with -ProgressAction -PipelineVariable. This change specifically specifies the PipelineVariable.

[Credit - internal thread](https://octopusdeploy.slack.com/archives/C03SG0LFJHX/p1700690293166349?thread_ts=1700689063.728339&cid=C03SG0LFJHX)

PR #1193, opened 2023-12-04
Include tags in AWS CloudFormation changeset template

kevjt:
# Background
[sc-64624]

When using the 'Deploy an AWS CloudFormation template' step template with Change Sets, any tags specified in the template are not applied to the created resources/stack.

The cause was due to the tags not being included in the mapper that creates the AWS CreateChangeSetRequest object.

Credit to [kshatalov](https://github.com/kshatalov) for reporting and finding the fix.

# Result
Part of https://github.com/OctopusDeploy/Issues/issues/8157


source/Calamari.Tests/Fixtures/PowerShell/Scripts/Invalid.ps1
kevjt, line 1: Needed to fix failing tests
See [PR](https://github.com/OctopusDeploy/Calamari/pull/1192) for context

APErebus (approved):
Looks good to me!

PR #1196, opened 2023-12-05
Keep journal entries when retention fails to delete a deployment

tothegills:
Deployment retention can fail to clean up the deployment files. When this happens, it currently swallows exceptions and removes the journal entry from the journal. As a result, deployment files can persist forever.

Instead, this PR proposes to leave the journal entry intact. The subsequent time retention runs, another attempt to clean up the files will occur. Deployment files _should_ be cleaned up more consistently with this approach.

There is the potential that the files will never be cleaned up, and the deployment journal will expand, leading to the journal becoming a performance concern. The reports we have had where files are not cleaned up generally involve files that are being locked because the previous deployment is still running. I assume it is generally the last deployment that fails and will eventually be cleaned up by the next deployment, so journal explosion is a low risk.

### DIFF
--- a/source/Calamari/Deployment/Retention/RetentionPolicy.cs
+++ b/source/Calamari/Deployment/Retention/RetentionPolicy.cs
@@ -67,7 +67,7 @@ namespace Calamari.Deployment.Retention
             RemovedFailedPackageDownloads();
         }
 
-        ReadOnlyCollection<string> RemoveDeployments(List<JournalEntry> deploymentsToDelete, List<JournalEntry> preservedEntries)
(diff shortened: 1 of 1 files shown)


benPearce1 (approved):
👍 
Would it be worth adding a retry policy? 
If the next attempt is more likely to work, then it would be sufficient to leave it for the next run, and keep the retention step the end of the deployment fast.

tothegills:
> Would it be worth adding a retry policy?
Looking into the history, the exception was originally swallowed because there are cases where files are legitimately in use. As you've pointed out, a retry policy will add time to the retention step. I don't think it's worth it: my guess is that customers are more sensitive to the time their deployments take than whether retention completes successfully this deployment or the next.

PR #1198, opened 2023-12-08
Authenticate the AWS CLI when setting up k8s authn

EddyMoulton:
## Issue

When running the kustomize step (or any step package step) against an AWS target we did not set up the authentication for the AWS CLI.

Other AWS step package steps got around this by authenticating within the steps execution, but that's not an option for this step that targets one of many k8s cloud providers.
Other k8s steps that do work against AWS don't use the step package framework, so they can share the same env var context.

[sc-64525]

## Solution

I've added a check to see if the env is configured and set aws up correctly.

As part of this I did a bit of basic refactoring, essentially pulling the AWS specific bits out of `SetupKubectlAuthentication.cs` into specific classes like the Azure and Google cloud auth is.

## Concerns

The only bit I'm concerned about here is how setting up the AWS authentication this way (using `aws configure`) just leaves those credentials on the worker. The two issues with that are:
1. Security of leaving creds lying around.
2. Crossed wires (eg. if another step isn't setting AWS credentials correctly, it will default to use these previously configured credentials)

The reason for using `aws configure` and saving those creds to a profile file was that environment variables are not shared across different sections of a step package. In this case we would set them in a `preInstruction` script and then they aren't accessible from the main script.

## Alternatives

I explored changing the location of the AWS credential file for the duration of this step, but that of course relies on environment variables which I can't persist over the life of the step.

I also explored using profiles to minimise the risk but it seems that running `aws eks get-token` doesn't support the `--profile` option when running within kubectl

### DIFF
--- a/source/Calamari.Tests/KubernetesFixtures/Authentication/SetupKubectlAuthenticationAwsFixture.cs
+++ b/source/Calamari.Tests/KubernetesFixtures/Authentication/SetupKubectlAuthenticationAwsFixture.cs
@@ -1,0 +1,336 @@
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Calamari.Aws.Deployment;
+using Calamari.Common.Features.Processes;
+using Calamari.Common.Plumbing.FileSystem;
+using Calamari.Common.Plumbing.Logging;
+using Calamari.Common.Plumbing.Variables;
+using Calamari.Kubernetes;
+using Calamari.Kubernetes.Authentication;
+using Calamari.Kubernetes.Integration;
+using FluentAssertions;
+using NSubstitute;
+using NUnit.Framework;
+
+namespace Calamari.Tests.KubernetesFixtures.Authentication
+{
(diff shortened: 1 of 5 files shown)


source/Calamari.Tests/KubernetesFixtures/Authentication/SetupKubectlAuthenticationAwsFixture.cs
```diff
@@ -0,0 +1,336 @@
...
+            variables.Set(SpecialVariables.EksClusterName, EksClusterName);
+            variables.Set(SpecialVariables.Namespace, Namespace);
+
+            environmentVars = new Dictionary<string, string>
+            {
+                { "AWS_ACCESS_KEY_ID", "access_key" },
+                { "AWS_SECRET_ACCESS_KEY", "secret_key" },
+                { "AWS_REGION", "region" }
+            };
+        }
```
EddyMoulton, line 95: I'll commonise some of this with the main `SetupKubectlAuthentication` tests once I've re-applied my changes over Rob's refactoring

PR #1201, opened 2023-12-12
Replace all net40 builds with net462

zentron:
As we move forward with the dropping support for Win 2003, we are now at a point where we can update our Full Framework 4.0 Calamari builds to 4.6.2
This PR largely just changes the target frameworks so that we can validate that everything works all the way through from the Server as expected.

## Impacts to Server
_In theory_, since we are still building all the existing Calamari executables, there should be little to no changes we need to make as part of this change apart from tweaking some of the warnings that might show if the user is on incompatible versions of Windows. Once this merges into `master` however and Octopus Server takes a dependency on it, we will officially be not expected to work on Windows 2003. Goodbye 20 year old OS.

## What doesn't this cover
The next step in a separate PR will be to work through removing all the `#if NET40` compiler directives, dependency updates, project structure changes, and much much more that we can then introduce as a result of this update.

### DIFF
--- a/source/Calamari.Common/Calamari.Common.csproj
+++ b/source/Calamari.Common/Calamari.Common.csproj
@@ -1,15 +1,10 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
     <PropertyGroup Condition="!$([MSBuild]::IsOSUnixLike())">
-        <TargetFrameworks>net452;netstandard2.1;net40</TargetFrameworks>
+        <TargetFrameworks>net452;netstandard2.1</TargetFrameworks>
     </PropertyGroup>
     <PropertyGroup Condition="$([MSBuild]::IsOSUnixLike())">
         <TargetFramework>netstandard2.1</TargetFramework>
-    </PropertyGroup>
-
-    <PropertyGroup Condition=" '$(TargetFramework)' == 'net40' ">
-        <DefineConstants>$(DefineConstants);USE_ALPHAFS_FOR_LONG_FILE_PATH_SUPPORT;HAS_SSL3</DefineConstants>
-        <PlatformTarget>anycpu</PlatformTarget>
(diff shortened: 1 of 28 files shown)


source/Calamari.Common/Calamari.Common.csproj
```diff
@@ -1,16 +1,11 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
     <PropertyGroup Condition="!$([MSBuild]::IsOSUnixLike())">
-        <TargetFrameworks>net452;netstandard2.1;net40</TargetFrameworks>
+        <TargetFrameworks>net452;netstandard2.1</TargetFrameworks>
```
APErebus, line 4: Should this be bumped to `net462`?

source/Calamari.Shared/Calamari.Shared.csproj
```diff
@@ -25,23 +25,19 @@
     <Nullable>enable</Nullable>
   </PropertyGroup>
   <PropertyGroup Condition="!$([MSBuild]::IsOSUnixLike())">
-    <TargetFrameworks>net452;netstandard2.1;net40</TargetFrameworks>
+    <TargetFrameworks>net452;netstandard2.1;</TargetFrameworks>
```
APErebus, line 28: Ditto re. `net462`

APErebus (changes requested):
Looks like there is still a couple of left over `net452` build targets

source/Calamari.Common/Calamari.Common.csproj
zentron: nice good find. Not sure how i didnt pick that up in the find all usages :)

zentron:
Related To: https://github.com/OctopusDeploy/Issues/issues/8548

APErebus (approved):
Awesome!

PR #1202, opened 2023-12-13
Attempted fixes for flaky tests in Calamari

IsaacCalligeros95:
This PR introduces a number of changes to tests with the aim of reducing flakiness in the builds. While I've tested these and they appear to reduce the flakiness, a lot of the failures in the pipeline seems to come from concurrent runs and external dependencies. This is hard to get a proper feel for on a single PR, this change set is intended as a temporary attempt to improve the pipeline, if we don't see any benefits these changes can be reverted.

* Increase base of TransientHttpErrorsPolicy from 2 -> 2.15. This takes the maximum retry time from 32 seconds to 45 **(Note this change is in the actual step behaviour).** 
* Add a TestTransientHttpErrorsPolicy with 4 retries, and a base of 3, with a Jitterer offset of up to 10 seconds. **(This Increases back off and reduces number of requests trying to minimize 429s)**
* Move from BeginCreate to Create, this means the resource should be available by the time subsequent processes are run. **This should mean resources are ready by the time deployments are running against them, reducing the number of failed calls and retries (reduce 429s)**
* Change app service plans from P1 -> V3. We usually have 50 + of these hanging around, with 14 day deletion times, If this removes the flakiness and we can remove the 14 day retention this should save $ here. It also appears that failed runs aren't deleting their resource groups, this is surprising as clean-ups are in the base class `OneTimeTearDown` method which should always run. Successful runs seem to tear down these resources, so there may be savings by nature of reducing flakiness.
* Set `WEBSITES_CONTAINER_START_TIME_LIMIT` variable on app settings. [This is based on the java app service tests spin up time](https://stackoverflow.com/questions/51103897/increase-startup-timeout-for-azure-webapp-for-containers).

> <html>
<body>

WEBSITES_CONTAINER_START_TIME_LIMIT | Amount of time in seconds to wait for the container to complete start-up before restarting the container. Default is 230. You can increase it up to the maximum of 1800.
-- | --

</body>
</html>
https://learn.microsoft.com/en-us/azure/app-service/reference-app-settings?tabs=kudu%2Cdotnet

### DIFF
--- a/source/Calamari.AzureAppService.Tests/AppServiceBehaviourFixture.cs
+++ b/source/Calamari.AzureAppService.Tests/AppServiceBehaviourFixture.cs
@@ -5,6 +5,7 @@ using System.IO.Compression;
 using System.Linq;
 using System.Reflection;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 using System.Xml;
 using System.Xml.Linq;
@@ -394,6 +395,8 @@ namespace Calamari.AzureAppService.Tests
                                                                                       IsReserved = true
                                                                                   });
 
+                await linuxAppServicePlan.WaitForCompletionAsync(CancellationToken.None);
+
                 var linuxWebSiteResponse = await resourceGroup.GetWebSites()
                                                               .CreateOrUpdateAsync(WaitUntil.Completed,
                                                                                    $"{resourceGroup.Data.Name}-linux",
@@ -416,6 +419,9 @@ namespace Calamari.AzureAppService.Tests
                                                                                            }
                                                                                        }
                                                                                    });
+
+                await linuxWebSiteResponse.WaitForCompletionAsync(CancellationToken.None);
+                
                 WebSiteResource = linuxWebSiteResponse.Value;
             }
 
--- a/source/Calamari.AzureAppService.Tests/AppServiceBehaviourFixture.cs
+++ b/source/Calamari.AzureAppService.Tests/AppServiceBehaviourFixture.cs
@@ -435,7 +435,7 @@ namespace Calamari.AzureAppService.Tests
                                         .Execute();
 
                 // Assert
-                await DoWithRetries(10,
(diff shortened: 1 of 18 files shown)


benPearce1 (approved):
These changes seem reasonable.
The cost increase is dependant on the automated cleanup/teardown being < 14 days

source/Calamari.AzureAppService.Tests/AppServiceBehaviourFixture.cs
```diff
@@ -423,7 +435,7 @@ public async Task CanDeployZip_ToLinuxFunctionApp()
                                         .Execute();
 
                 // Assert
-                await DoWithRetries(10,
```
IsaacCalligeros95, line 426: AssertContent already retries with an exponential backoff. This time has been increased, if the container isn't available in 2 runs ~8 minutes, it generally is a different failure

source/Calamari.AzureAppService.Tests/Legacy/LegacyAppServiceIntegrationTest.cs
```diff
@@ -103,12 +101,19 @@ public async Task Cleanup()
...
-                                                                                      {
-                                                                                          var r = await client.GetAsync($"https://{hostName}/{rootPath}");
-                                                                                          r.EnsureSuccessStatusCode();
-                                                                                          return r;
-                                                                                      });
+            var response = await RetryPolicies.TestsTransientHttpErrorsPolicy.ExecuteAsync(async context =>
+                                                                                           {
+                                                                                               var r = await client.GetAsync($"https://{hostName}/{rootPath}");
+                                                                                               if (!r.IsSuccessStatusCode)
+                                                                                               {
```
IsaacCalligeros95, line 108: The deploy a war package still appears to be flakey, although I think less so. On test failures this returns a 500, theres a SO thread on setting the AppService to AlwaysOn but that already seems to be the case for these resources.

PR #1203, opened 2023-12-17
Re-add placeholder artifact for build

EddyMoulton:
I was hoping to get away without this workaround, but it's still required.

Not blocking the build, we're still on the old teamcity build step that doens't use nuke until this merges.

### DIFF
--- a/build/Build.DetermineAffectedTests.cs
+++ b/build/Build.DetermineAffectedTests.cs
@@ -13,7 +13,7 @@ namespace Calamari.Build;
 partial class Build
 {
     [Parameter(Name = "DefaultGitBranch")] readonly string MainBranchName;
-    
+
     [PublicAPI]
     Target DetermineAffectedTests =>
         target => target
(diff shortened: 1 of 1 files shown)


build/Build.DetermineAffectedTests.cs
```diff
@@ -13,15 +13,19 @@ namespace Calamari.Build;
...
     Target DetermineAffectedTests =>
         target => target
             .Executes(() =>
             {
                 if (GitVersionInfo is null)
                     throw new ArgumentNullException(nameof(GitVersionInfo));
-                
+
+                File.WriteAllText("placeholder.txt",
+                    "TeamCity can't stand to have no artifacts in a consuming configuration, so this is a placeholder");
```
MissedTheMark, line 26: I can feel the pain in this 😅

MissedTheMark (approved):
LGTM 👍

build/Build.DetermineAffectedTests.cs
EddyMoulton: :joy: The words are taken from someone elses previous work, but the feeling is all mine now

PR #1204, opened 2023-12-19
Make sure all kubectl commands use the command pipeline

APErebus:
The `--request-timeout` flag does not work when running ambient k8s auth. This was being suppressed by `DisableRequestTimeoutArgument`, however the `get namespace` and `create namespace` didn't use the `kubectl` tool, meaning this wasn't respected.

I refactored the code so that these cli calls now go through the kubectl tool, but to make sure feature parity is maintained, I had to add new code to allow for all command line outputs to be logged as verbose (which is why there was the bifurcation in the code).

Shortcut story: [sc-65597]

### DIFF
--- a/source/Calamari.Tests/KubernetesFixtures/Authentication/SetupKubectlAuthenticationFixture.cs
+++ b/source/Calamari.Tests/KubernetesFixtures/Authentication/SetupKubectlAuthenticationFixture.cs
@@ -15,6 +15,7 @@ using Calamari.Kubernetes.Authentication;
 using Calamari.Kubernetes.Integration;
 using Calamari.Tests.Helpers;
 using FluentAssertions;
+using Microsoft.Azure.Management.ContainerRegistry.Fluent.Models;
 using NSubstitute;
 using NSubstitute.ExceptionExtensions;
 using NUnit.Framework;
@@ -95,6 +96,7 @@ namespace Calamari.Tests.KubernetesFixtures.Authentication
                 if (logMessage != null) invocation.AdditionalInvocationOutputSink?.WriteInfo(logMessage);
                 return new CommandResult(invocation.Executable, isSuccess ? 0 : 1, workingDirectory: workingDirectory);
             });
+
             kubectl = Substitute.For<IKubectl>();
             kubectl.ExecutableLocation.Returns("kubectl");
             kubectl.When(x => x.ExecuteCommandAndAssertSuccess(Arg.Any<string[]>()))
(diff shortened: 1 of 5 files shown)


source/Calamari.Tests/KubernetesFixtures/Authentication/SetupKubectlAuthenticationFixture.cs
```diff
@@ -172,8 +186,8 @@ public void Execute_WhereNamespaceDoesNotExist_NamespaceCreated()
 
             invocations.TakeLast(2).Should().BeEquivalentTo(new[]
             {
-                ("kubectl", $"get namespace {Namespace} --request-timeout=1m"),
-                ("kubectl", $"create namespace {Namespace} --request-timeout=1m"),
+                ("kubectl", $"get namespace {Namespace}"),
+                ("kubectl", $"create namespace {Namespace}"),
```
APErebus, line 190: Because the test uses a substitute `IKubectl`, but the real `Kubectl` contains the request timeout flag, this arg doesn't show up in any of the test commands anymore

source/Calamari/Kubernetes/SetupKubectlAuthentication.cs
```diff
@@ -430,10 +448,10 @@ bool TrySetAws()
 
         void CreateNamespace(string @namespace)
         {
-            if (TryExecuteCommandWithVerboseLoggingOnly("get", "namespace", @namespace))
+            if (kubectl.ExecuteCommandWithVerboseLoggingOnly("get", "namespace", @namespace).ExitCode == 0)
                 return;
 
-            if (!TryExecuteCommandWithVerboseLoggingOnly("create", "namespace", @namespace))
+            if (kubectl.ExecuteCommandWithVerboseLoggingOnly("create", "namespace", @namespace).ExitCode != 0)
```
APErebus, line 454: This is where we now call the kubectl code, rather than the custom version in this class

source/Calamari/Kubernetes/Integration/CommandLineTool.cs
```diff
@@ -27,13 +27,18 @@ public class CommandLineTool
...
+
+        /// <summary>
+        /// This is a special case for when the invocation results in an error
+        /// 1) but is to be expected as a valid scenario; and
+        /// 2) we don't want to inform this at an error level when this happens.
+        /// </summary>
+        protected CommandResult ExecuteCommandAndLogOutputAsVerbose(CommandLineInvocation invocation)
+            => ExecuteCommandAndLogOutput(invocation, true);
+
+        CommandResult ExecuteCommandAndLogOutput(CommandLineInvocation invocation, bool logOutputAsVerbose)
```
MissedTheMark, line 41: Maybe make `logOutputAsVerbose` optional with a default value of `false` then you don't need the other overload?

source/Calamari/Kubernetes/SetupKubectlAuthentication.cs
```diff
@@ -97,7 +97,7 @@ void SetupContext(string @namespace)
             {
                 SetupGCloudContext(@namespace);
             }
-            else if(variables.IsSet(SpecialVariables.ClusterUrl)) // Most other auth mechanisms require some manual commands to configure
+            else if (variables.IsSet(SpecialVariables.ClusterUrl)) // Most other auth mechanisms require some manual commands to configure
```
MissedTheMark, line 100: At some point we should set up PrettyBot to clean up Calamari automatically - if it's not already.

MissedTheMark (approved):
Looks good 👍

source/Calamari/Kubernetes/Integration/CommandLineTool.cs
APErebus: I thought about that but I kinda wanted to make it explicit, rather than an optional boolean 🤔

PR #1205, opened 2023-12-20
Configure AWS CLI when setting up k8s authn

EddyMoulton:
## Issue

When running the kustomize step (or any step package step) against an AWS target we did not set up the authentication for the AWS CLI.

Other AWS step package steps got around this by authenticating within the steps execution, but that's not an option for this step that targets one of many k8s cloud providers.
Other k8s steps that do work against AWS don't use the step package framework, so they can share the same env var context.

[sc-64525]

Replaces https://github.com/OctopusDeploy/Calamari/pull/1198 after re-applying changes over main. 

## Solution

I've added a check to see if the env is configured and set the AWS CLI up correctly. ~~I've used the profile option provided by the AWS CLI to separate any credentials we set from what may be used manually (eg. if someone ssh'd in and configured it) and any other steps that may be similarly misconfigured in the future (this would have resulted in the wrong credentials being used).~~

Due to the step package framework not sharing the env var context, I've changed the kubeconfig authn mechanism to be a plain token instead of calling the AWS command. This means it can be cleaned up along with the rest of the workspace at the end of the deployment. The kubeconfig contents does get logged in the verbose logs, but the token is redacted automatically from the logs.
This token is valid for 15 minutes based on my testing, I believe this should be fine as it is generated once per step - so there should not be such a large delay between grabbing the token and before starting any processing.

As part of this I did a bit of basic refactoring, essentially pulling the AWS specific bits out of `SetupKubectlAuthentication.cs` into specific classes like the Azure and Google cloud auth is.

### DIFF
--- a/source/Calamari/Kubernetes/Authentication/AwsCliAuth.cs
+++ b/source/Calamari/Kubernetes/Authentication/AwsCliAuth.cs
@@ -66,8 +66,7 @@ namespace Calamari.Kubernetes.Authentication
                     var region = GetEksClusterRegion(clusterUrl);
                     if (!string.IsNullOrWhiteSpace(region))
                     {
-                        var apiVersion = awsCli.GetEksClusterApiVersion(clusterName, region);
-                        SetKubeConfigAuthenticationToAwsCli(user, clusterName, region, apiVersion);
+                        SetKubeConfigAuthenticationToAwsCli(user, clusterName, region);
                         return true;
                     }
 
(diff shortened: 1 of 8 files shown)


source/Calamari/Kubernetes/Authentication/AwsCliAuth.cs
```diff
@@ -0,0 +1,154 @@
...
+
+        bool TrySetKubeConfigAuthenticationToAwsCli(string clusterName, string clusterUrl, string user)
+        {
+            if (!awsCli.TrySetAws())
+            {
+                log.Verbose("Could not find the aws cli, falling back to the aws-iam-authenticator.");
+                return false;
+            }
+
+            ConfigureAwsCli();
```
EddyMoulton, line 58: This is the change from previous behaviour

source/Calamari/Kubernetes/Authentication/AwsCliAuth.cs
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        void ConfigureAwsCli()
+        {
+            if (!environmentVars.ContainsKey("AWS_ACCESS_KEY_ID"))
+            {
+                var awsEnvironmentGeneration =
+                    AwsEnvironmentGeneration.Create(log, deploymentVariables).GetAwaiter().GetResult();
+                environmentVars.AddRange(awsEnvironmentGeneration.EnvironmentVars);
+            }
```
EddyMoulton, line 114: If we are running a kubectl script step, this has already been done so we can skip it the second time.

source/Calamari/Kubernetes/Authentication/AwsCliAuth.cs
```diff
@@ -0,0 +1,154 @@
...
+
+            awsCli.Configure(
+                GetEnvironmentVarOrDefault("AWS_ACCESS_KEY_ID"),
+                GetEnvironmentVarOrDefault("AWS_SECRET_ACCESS_KEY"),
+                GetEnvironmentVarOrDefault("AWS_REGION"),
+                GetEnvironmentVarOrDefault("AWS_SESSION_TOKEN")
+            );
+        }
```
EddyMoulton, line 122: `AwsEnvironmentGeneration` does everything we want it to already, with the only caveat that it simply sets environment variables to configure the AWS CLI instead of configuring it with `aws configure` - however the environment variables aren't set in the context of running the kustomize step.

source/Calamari/Kubernetes/Authentication/AwsCliAuth.cs
zentron: To clarify, what's the scenario where this path gets invoked twice?

EddyMoulton: Discussed over zoom, but for posterity - When we run a kubectl script step, calamari will handle the authentication via the `AwsScriptWrapper` (I think, haven't double checked the exact location) and then again when we authenticate for kubectl. This doesn't get run for a step package framework step.

EddyMoulton:
GitGuardian above is picking up `k8s-aws-v1.token` as a high entropy secret :joy: 
I'd prefer to leave it as is because it describes the form we expect, however if this is going to cause issue in the future too it's an easy change.

source/Calamari/Kubernetes/Authentication/AwsCliAuth.cs
```diff
@@ -0,0 +1,134 @@
...
+            if (string.IsNullOrEmpty(token))
+            {
+                token = awsCli.GetEksClusterToken(clusterName, region);
+            }
+
+            var arguments = new List<string> { "config", "set-credentials", user, $"--token={token}" };
+
+            log.AddValueToRedact(token, "<token>");
+            kubectl.ExecuteCommandAndAssertSuccess(arguments.ToArray());
+        }
```
EddyMoulton, line 131: This is where we change to now set the token statically instead of configuring it to grab the token as part of running `kubectl`

zentron (approved):

PR #1206, opened 2023-12-20
Add warning message about upcoming F# script deprecation

kevjt:
# Background

[sc-66834]

Support for F# scripting is being deprecated in 2024. Warning messages are being added to provide users with some notice before its removal.

# Results
Adds a feature toggled warning message in the task log of F# script executions.
[PR](https://github.com/OctopusDeploy/OctopusDeploy/pull/21945) to add feature toggle

Part of https://github.com/OctopusDeploy/OctopusDeploy/issues/21942

## Before
Screenshot 2023-12-20 at 3 00 10 pm

## After
Screenshot 2023-12-20 at 3 03 05 pm

### DIFF
--- a/source/Calamari.Common/Features/Scripting/ScriptEngine.cs
+++ b/source/Calamari.Common/Features/Scripting/ScriptEngine.cs
@@ -9,7 +9,9 @@ using Calamari.Common.Features.Scripting.Python;
 using Calamari.Common.Features.Scripting.ScriptCS;
 using Calamari.Common.Features.Scripting.WindowsPowerShell;
 using Calamari.Common.Features.Scripts;
+using Calamari.Common.FeatureToggles;
 using Calamari.Common.Plumbing.Extensions;
+using Calamari.Common.Plumbing.Logging;
 using Calamari.Common.Plumbing.Variables;
 
 namespace Calamari.Common.Features.Scripting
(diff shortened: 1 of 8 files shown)


source/Calamari.Common/Features/Scripting/ScriptEngine.cs
```diff
@@ -104,12 +108,21 @@ IScriptExecutor GetScriptExecutor(ScriptSyntax scriptSyntax, bool runDotnetScrip
                 case ScriptSyntax.Bash:
                     return new BashScriptExecutor();
                 case ScriptSyntax.FSharp:
+                    LogFSharpDeprecationWarning(variables);
```
zentron, line 111: Can we pass this down into the `FSharpExecutor` class itself so that this class doesn't need to concern itself with this script specific logic?

source/Calamari.Common/Features/Scripting/ScriptEngine.cs
```diff
@@ -104,12 +108,21 @@ IScriptExecutor GetScriptExecutor(ScriptSyntax scriptSyntax, bool runDotnetScrip
...
                 default:
                     throw new NotSupportedException($"{scriptSyntax} script are not supported for execution");
             }
         }
+
+        void LogFSharpDeprecationWarning(IVariables variables)
+        {
+            if (FeatureToggle.FSharpDeprecationFeatureToggle.IsEnabled(variables))
+            {
+                log.Warn($"Executing FSharp scripts will soon be deprecated. Please read our deprecation {log.FormatLink("https://octopus.com/blog/2024-deprecated-features#f-sharp", "blog post")} for more details.");
```
zentron, line 124: This url should be converted to use a short URL
https://octopushq.atlassian.net/wiki/spaces/MAR/pages/2393899875/Short+URLs+process+and+rules
This ensures that if we move the underlying page in future these links continue to work

zentron (changes requested):
The general message etc sounds good, just a note about moving it down into the FSharp specific class

source/Calamari.Common/Features/Scripting/ScriptEngine.cs
kevjt: I placed it in the `ScriptEngine` class originally to have the log appear as early as possible (before any of the script wrappers in the chain are executed). But I suppose this doesn't really matter much, and you're right in moving logic down to the `FSharpExecutor` because it should make testing a bit easier I think.

zentron (approved):
Thanks for the quick adjustment!

PR #1207, opened 2024-01-04
Better variable replacement error handling

veochen-octopus:
If `IgnoreVariableReplacementErrors` is set we still log a warning of the exception message and its stacktrace. This can cause confusion in some cases. Logging the error as `INFO` and stacktrace as `VERBOSE` provides the same information in a more friendly way.

[sc-67827]

### DIFF
--- a/source/Calamari.Common/Features/ConfigurationVariables/ConfigurationVariablesReplacer.cs
+++ b/source/Calamari.Common/Features/ConfigurationVariables/ConfigurationVariablesReplacer.cs
@@ -48,8 +48,7 @@ namespace Calamari.Common.Features.ConfigurationVariables
             {
                 if (ignoreVariableReplacementErrors)
                 {
-                    log.Warn(ex.Message);
-                    log.Warn(ex.StackTrace);
+                    log.InfoFormat("Ignoring variable replacement error: {0}", ex.Message);
                 }
                 else
                 {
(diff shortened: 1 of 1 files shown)


source/Calamari.Common/Features/ConfigurationVariables/ConfigurationVariablesReplacer.cs
```diff
@@ -48,8 +48,8 @@ public void ModifyConfigurationFile(string configurationFilePath, IVariables var
             {
                 if (ignoreVariableReplacementErrors)
                 {
-                    log.Warn(ex.Message);
-                    log.Warn(ex.StackTrace);
+                    log.InfoFormat("Ignoring variable replacement error: {0}", ex.Message);
+                    log.VerboseFormat(ex.StackTrace);
```
IsaacCalligeros95, line 52: I'm not sure if it matters, but should this just be `log.Verbose(ex.StackTrace);`

IsaacCalligeros95: Happy to approve if it doesn't and keep the green run.

veochen-octopus: Probably doesn't matter but I'll do it right.

IsaacCalligeros95 (approved):
LGTM

PR #1211, opened 2024-01-12
Remove Octopus.Diagnostics dependency

scme0:
A warning came up in Calamari because Octopus.Diagnostics is built for 4.6.1 and Calamari is built for 4.6.2.

I noticed that we are only using the LogCategory Enum and also that the Octopus.Diagnostics repo has been archived. It looks like the functionality has been brought into Octopus Server. Because of this, I've copied the enum out and removed the dependency.


kevjt (approved):
lgtm

PR #1219, opened 2024-01-24
Add check for KUBERNETES_SERVICE_PORT_HTTPS in ambient auth check

scme0:


source/Calamari/Kubernetes/SetupKubectlAuthentication.cs
```diff
@@ -215,8 +215,14 @@ bool HasAmbientKubeContext()
              * https://kubernetes.io/docs/tasks/run-application/access-api-from-pod/#directly-accessing-the-rest-api
              */
             return fileSystem.FileExists("/var/run/secrets/kubernetes.io/serviceaccount/token")
-                   && !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("KUBERNETES_SERVICE_HOST"))
-                   && !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("KUBERNETES_SERVICE_PORT"));
+                   && EnvironmentVariableSet("KUBERNETES_SERVICE_HOST")
```
flin-8, line 218: Let's wait for the results of the [investigation](https://octopusdeploy.slack.com/archives/C05KK1G85J5/p1706054935205829) so we can make the change in one go

scme0:
After investigating the issue that this was trying to fix, we found that this wasn't required.

PR #1221, opened 2024-02-01
New Calamari command for the Kubernetes Kustomize step to run as a conventional step (not using step-api)

kevjt:
# Background

We are migrating the Kustomize action from the step api to a conventional step - more details can be found in this [doc](https://docs.google.com/document/d/1s_b-I-1Tlcu3HmZMyfgORW5tALHPeFLys6T6RymfBAI/edit#heading=h.zdkbl0q6yra6)

[sc-68738]

# Related PRs
[Add action handler in Server](https://github.com/OctopusDeploy/OctopusDeploy/pull/22498)
[UI changes](https://github.com/OctopusDeploy/OctopusDeploy/pull/22446)

### DIFF
--- a/source/Calamari/Kubernetes/Commands/Executors/KustomizeExecutor.cs
+++ b/source/Calamari/Kubernetes/Commands/Executors/KustomizeExecutor.cs
@@ -1,0 +1,23 @@
+#if !NET40
+using System;
+using System.Collections.Generic;
+using Calamari.Common.Commands;
+using Calamari.Common.Plumbing.Logging;
+using Calamari.Kubernetes.Integration;
+using Calamari.Kubernetes.ResourceStatus.Resources;
+
+namespace Calamari.Kubernetes.Commands.Executors
+{
+    class KustomizeExecutor : BaseKubernetesApplyExecutor
+    {
+        public KustomizeExecutor(ILog log, Kubectl kubectl) : base(log, kubectl)
+        {
+        }
+
+        protected override IEnumerable<ResourceIdentifier> ApplyAndGetResourceIdentifiers(RunningDeployment deployment)
+        {
+            throw new NotImplementedException();
+        }
+    }
+}
+#endif
--- a/source/Calamari/Kubernetes/Commands/Executors/KustomizeExecutor.cs
+++ b/source/Calamari/Kubernetes/Commands/Executors/KustomizeExecutor.cs
@@ -36,8 +36,7 @@ namespace Calamari.Kubernetes.Commands.Executors
 
             if (overlayPath == null)
             {
-                log.Warn("Kustomization directory not specified");
-                return Enumerable.Empty<ResourceIdentifier>();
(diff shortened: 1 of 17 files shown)


source/Calamari/Kubernetes/Commands/Executors/KustomizeExecutor.cs
```diff
@@ -0,0 +1,92 @@
...
+        protected override IEnumerable<ResourceIdentifier> ApplyAndGetResourceIdentifiers(RunningDeployment deployment)
+        {
+            var variables = deployment.Variables;
+            var overlayPath = variables.Get(SpecialVariables.KustomizeOverlayPath);
+
+            if (overlayPath == null)
+            {
+                log.Warn("Kustomization directory not specified");
+                return Enumerable.Empty<ResourceIdentifier>();
+            }
```
tleed5, line 41: Do you know what happens in the step package kustomize action if the overlay path isn't set? In this version the deployment will succeed but with warnings and that doesn't feel like correct behaviour to me. We will need to add some validation to the frontend to prevent the overlay path not been set but I think this should throw an exception might be worth double checking with others in the team though.

Screenshot 2024-02-02 at 10 01 22 am

tleed5 (approved):
Awesome work! Just a comment on what we should do if the overlay path isn't set but I don't see that as a blocking issue

source/Calamari/Kubernetes/Commands/Executors/KustomizeExecutor.cs
kevjt: Good question. My thought was that this is already being validated by the [frontend](https://github.com/OctopusDeploy/OctopusDeploy/pull/22446/files#diff-942a70e020f19d3d7fd87a3be24c8920068a26aa80da2fa2b1ac2046bab78099R14) such that it is a required field. From Calamari's perspective, if the overlay path is not specified (not possible unless the user runs the Calamari command directly somehow) then it would be a no-op but not a fatal exception. But at the same time I guess you could argue that if you ran the `kubectl apply -k` command with no directory (or it fails to find the kustomization file) then it results in an error rather than no-op. So I'm more than happy to make it an exception just to be consistent.

tleed5: I think keeping it consistent and adding the exception or removing the check for a null overlay path and relying entirely the error from `kubectl` is a good path forward so we avoid representing a deployment as successful when nothing was actually deployed.

I have a TODO to double check the frontend validation, I think it just displays an error but doesn't prevent you from saving process. Probably need to add a new kind of validator something like the [raw yaml validation](https://github.com/OctopusDeploy/OctopusDeploy/blob/c0e8d0a3eb6a08582a39a52dd9fc68aaa819624a/source/Octopus.Kubernetes/Validation/KubernetesDeployRawYamlActionValidator.cs#L14).

PR #1222, opened 2024-02-05
Remove journalling for Kubernetes steps

flin-8:
[sc-70260]

Kubernetes steps don't need to keep a journal, since the journal is intended for long-lived extracted files that are used externally AFTER deployment has completed.


APErebus (approved):
excellent!

PR #1225, opened 2024-02-14
Update YamlDotNet and handle multiple comments following a block

IsaacCalligeros95:
This PR fixes an issue where multiple comments after a parent block pushes the : after the comment block. This results in failing deployments. This is fixed in the latest version of YamlDotnet. 

Source YAML:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: Test
  labels:
    app: web
mq:
#environment: "12345"
#appId: "12345" 
    user: "admin"
    retry:
      max: "3"
      redelivery-attempts: "3"
#maxLocalMessages: "30"
    auditQueue: "AuditService"
```
    
YAML after deployment: You can see the : was moved from mq: and placed below the commented lines.
```

apiVersion: apps/v1
kind: Deployment
metadata:
  name: Test
  labels:
    app: web
mq
# environment: "12345"
# appId: "12345" 
:
  user: "Dan"
  retry:
    max: "3"
    redelivery-attempts: "3"
    # maxLocalMessages: "30"
  auditQueue: "NewServiceName"
```


source/Calamari.Common/Features/StructuredVariables/YamlFormatVariableReplacer.cs
```diff
@@ -155,7 +155,7 @@ void LogReplacement(string key)
             }
             catch (Exception e) when (e is SyntaxErrorException || e is SemanticErrorException)
             {
-                throw new StructuredConfigFileParseException(e.Message, e);
+                throw new StructuredConfigFileParseException(e.ToString(), e);
```
IsaacCalligeros95, line 158: The exception message no longer includes the Start and End locations covered in the `ParsingErrorPositionIsReportedToUser` test. ToString includes these and has the same string formatting as 8.2.1.

source/Calamari.Tests/Fixtures/StructuredVariables/Approved/YamlVariableReplacerFixture.ShouldPreserveEncodingUtf8UnixNoBom.approved.txt
```diff
@@ -1,6 +1,6 @@
 64 61 67 67 65 72 3A 20   dagger: 
-E2 80 A0 0A 73 75 6E 6E   ???^sunn
-69 65 73 3A 20 22 5C 55   ies: "\U
-30 30 30 31 46 36 30 45   0001F60E
-22 0A 65 75 72 6F 3A 20   "^euro: 
-E2 82 AC 0A               ???^
\ No newline at end of file
+E2 80 A0 0D 0A 73 75 6E   ???^^sun
```
IsaacCalligeros95, line 2: I couldn't really get to the bottom of why the additional Carat characters are being added here.

I'm guessing this is a CRLF change https://stackoverflow.com/questions/13836352/what-is-the-utf-8-representation-of-end-of-line-in-text-file which could be an issue for Unix. I'm waiting to see what happens in the pipeline.

IsaacCalligeros95: It looks like the difference here and for other [tests](https://build.octopushq.com/buildConfiguration/OctopusDeploy_CalamariNetCore_PublishNugetPackagesToMyGet/11165366?buildTab=dependencies&mode=list&state=failed&type=snapshot&hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildTestsSection=true#11165308) is that it adds a carriage return. This looks to be a carriage return (0D) followed by null (00) followed by a line feed (0A)

image

IsaacCalligeros95: Possibly https://github.com/aaubry/YamlDotNet/issues/867

IsaacCalligeros95:
There's a potential issue with how YamlDotnet parses anchor tags in later versions with the introduction of an [Empty] on the deseralized anchor instead of an empty string. This results in a new line when writing the element, which depends on the target OS. I don't think this would actually be a problem, but given the broad range of customer use cases I'm hesitant to push this change through as is.

```
Scalar [anchor = [empty], tag = ?, value = †, style = Plain, isPlainImplicit = True, isQuotedImplicit = False]      (new)
Scalar [anchor = , tag = , value = †, style = Plain, isPlainImplicit = True, isQuotedImplicit = False]              (old)
```

PR #1227, opened 2024-02-20
Add Helm OCI Registry support

scme0:
# Background

I'm using the OCI Registry for the automatic/server driven upgrades of the Kubernetes Agent.

# Results

It's actually pretty easy to add, as we just need to adjust one part of the helm command.

### DIFF
--- a/source/Calamari/Kubernetes/SpecialVariables.cs
+++ b/source/Calamari/Kubernetes/SpecialVariables.cs
@@ -55,9 +55,9 @@ namespace Calamari.Kubernetes
 
             public static class ScriptSource
             {
-                public const string OciRegistry = nameof(OciRegistry);
-                public const string Package = nameof(Package);
-                public const string GitRepository = nameof(GitRepository);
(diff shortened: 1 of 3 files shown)


source/Calamari/Kubernetes/SpecialVariables.cs
```diff
@@ -51,6 +51,14 @@ public static class Helm
             public const string TillerNamespace = "Octopus.Action.Helm.TillerNamespace";
             public const string TillerTimeout = "Octopus.Action.Helm.TillerTimeout";
             public const string ChartDirectory = "Octopus.Action.Helm.ChartDirectory";
+            public const string OciRegistry = "Octopus.Action.Helm.OciRegistry";
+
+            public static class ScriptSource
+            {
+                public const string OciRegistry = nameof(OciRegistry);
```
flin-8, line 58: Just a minor thought - consider make this a string literal, so renaming the variable above won't change this value

scme0:
We are going in a different direction so I'm going to close this.

PR #1233, opened 2024-03-15
Send null instead of empty string when there is no service password

tothegills:
Relates to https://github.com/OctopusDeploy/Issues/issues/8676


susanpann (approved):
noice

PR #1235, opened 2024-03-27
Add Calamari variable to print Kubectl output on parsing error

scme0:
Adding a variable to Calamari to allow customers to turn on verbose logging of kubectl output in case it fails to parse the yaml.

This will hopefully help us to diagnose this issue: https://octopusdeploy.slack.com/archives/CNHBHV2BX/p1710389143225389

### DIFF
--- a/source/Calamari/Kubernetes/Commands/Executors/BaseKubernetesApplyExecutor.cs
+++ b/source/Calamari/Kubernetes/Commands/Executors/BaseKubernetesApplyExecutor.cs
@@ -140,6 +140,10 @@ namespace Calamari.Kubernetes.Commands.Executors
                 log.Verbose(outputJson);
                 log.Verbose("---------------------------");
             }
+            else
+            {
+                log.Verbose($"To get Octopus to print out the JSON string retrieved from kubectl, set Octopus Variable '{SpecialVariables.PrintVerboseKubectlOutputOnError}' to 'true'");
+            }
             log.Error("This can happen with older versions of kubectl. Please update to a recent version of kubectl.");
             log.Error("See https://github.com/kubernetes/kubernetes/issues/58834 for more details.");
             log.Error("Custom resources will not be saved as output variables.");
--- a/source/Calamari/Kubernetes/Commands/Executors/BaseKubernetesApplyExecutor.cs
+++ b/source/Calamari/Kubernetes/Commands/Executors/BaseKubernetesApplyExecutor.cs
@@ -142,7 +142,7 @@ namespace Calamari.Kubernetes.Commands.Executors
             }
             else
             {
-                log.Verbose($"To get Octopus to print out the JSON string retrieved from kubectl, set Octopus Variable '{SpecialVariables.PrintVerboseKubectlOutputOnError}' to 'true'");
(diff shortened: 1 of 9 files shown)


kevjt (approved):

source/Calamari/Kubernetes/Commands/Executors/BaseKubernetesApplyExecutor.cs
```diff
@@ -135,12 +135,15 @@ void WriteResourcesToOutputVariables(IEnumerable<ResourceIdentifier> resources)
...
         
-        void LogParsingError(string outputJson)
+        void LogParsingError(string outputJson, bool logKubectlOutputOnError)
         {
-            log.Error($"\"kubectl apply -o json\" returned invalid JSON:");
-            log.Error("---------------------------");
-            log.Error(outputJson);
-            log.Error("---------------------------");
+            log.Error($"\"kubectl apply -o json\" returned invalid JSON");
+            if (logKubectlOutputOnError)
```
scme0, line 141: I've added this in as it feels like this could have similar security concerns to the other one. Do we want to put in the (visible by default) error message that they can switch this flag on to log the raw kubectl output?

zentron (approved):
Lets make sure we update our documentation to provide details for this useful flag

zentron:
This should be considered a breaking change in the release note since it now requires the variable to return the data. In the RN include some of the reasoning about why we wanted to put it behind a flag.

scme0:
> This should be considered a breaking change

@zentron Even if it's just the logging which has changed? Would that be a major version bump for Calamari?
Are there Release Notes for Calamari? I'll add it to the release notes for Server when we bump the Calamari version.

source/Calamari/Kubernetes/ResourceStatus/ResourceRetriever.cs
```diff
@@ -56,13 +62,42 @@ private IEnumerable<Resource> GetChildrenResources(Resource parentResource, IKub
...
+        {
+            try
+            {
+                return function(jsonString, options);
+            }
+            catch (JsonException)
+            {
+                LogJsonStringError(jsonString, options);
+                throw;
+            }
```
tleed5, line 85: Is there any concern that the `JsonException` has any sensitive information in it? Just thinking for the case when the exception bubbles up and something else logs it out

tleed5 (approved):
Looks good to me, just a small question around the `JsonException` handling

scme0:
After final adjustments I've tested and screen capped a "demo" version:

Because I don't know how to make it fail to parse, I adjusted the code (just for the test) to print the error everytime it runs:
Screenshot 2024-04-04 at 11 53 51
Screenshot 2024-04-04 at 11 52 36


Output in my task log looks like this:

No variable provided:
Screenshot 2024-04-04 at 11 50 52

Variable provided:
Screenshot 2024-04-04 at 11 51 45

Screenshot 2024-04-04 at 11 51 53

source/Calamari/Kubernetes/ResourceStatus/ResourceRetriever.cs
scme0: It doesn't log much of anything to be honest which is part of the reason why we've needed to add this. It does log line and column numbers so you can identify where it's failing to parse as well as a character or two which it doesn't expect, but it doesn't seem to log more than that.

PR #1236, opened 2024-04-02
Add Kubernetes resources to monitor as soon as the each batch of yaml is applied

kevjt:
## Background
This PR puts back the improvement that was originally added to improve the response time of the KOS feature when there are many and/or large batches of YAML that need to be applied. Instead of waiting for all batches to complete before starting the monitoring, the callback is called after every batch operation in the case of `GatherAndApplyRawYamlExecutor` such that it starts getting object statuses sooner.

### DIFF
--- a/source/Calamari.Tests/KubernetesFixtures/Approved/KubernetesContextScriptWrapperLiveFixtureEks.DeployRawYaml_WithMultipleYamlFilesGlobPatterns_YamlFilesAppliedInCorrectBatches.ApplyingBatches.approved.txt
+++ b/source/Calamari.Tests/KubernetesFixtures/Approved/KubernetesContextScriptWrapperLiveFixtureEks.DeployRawYaml_WithMultipleYamlFilesGlobPatterns_YamlFilesAppliedInCorrectBatches.ApplyingBatches.approved.txt
@@ -4,9 +4,15 @@ Matched file: deployments/subfolder/myapp-deployment.yml
 Created Resources:
  - Deployment/nginx-deployment in namespace calamari-testing
  - Deployment/nginx-deployment2 in namespace calamari-testing
+Resource Status Check: 2 new resources have been added:
+ - Deployment/nginx-deployment in namespace calamari-testing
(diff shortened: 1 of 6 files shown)


source/Calamari.Tests/KubernetesFixtures/Approved/KubernetesContextScriptWrapperLiveFixtureEks.DeployRawYaml_WithMultipleYamlFilesGlobPatterns_YamlFilesAppliedInCorrectBatches.ApplyingBatches.approved.txt
kevjt, line 1: This effectively reverts to the assent logs that were there before the refactoring which caused this regression

https://github.com/OctopusDeploy/Calamari/pull/1221/files?diff=unified&w=0#diff-0db08797bc823f1629cdd5af8e58f9cff19d259b717e9529f9e9c784e6dacff3L1-L26

source/Calamari/Kubernetes/Commands/Executors/GatherAndApplyRawYamlExecutor.cs
```diff
@@ -50,6 +51,12 @@ protected override IEnumerable<ResourceIdentifier> ApplyAndGetResourceIdentifier
             foreach (var directory in directories)
             {
                 var res = ApplyBatchAndReturnResourceIdentifiers(directory).ToList();
```
scme0, line 53: I think you can change `.ToList()` to `.ToArray()` here and remove the `.ToArray()` on line 57.

scme0 (approved):
Nice! LGTM just one minor comment.

source/Calamari/Kubernetes/Commands/Executors/GatherAndApplyRawYamlExecutor.cs
kevjt: Yes good catch!

PR #1238, opened 2024-04-03
Migrate Calamari.AzureResourceGroup project to use modern azure sdk

tleed5:
# Background
Since the `Microsoft.Azure.Management.*` aka "azure libraries for net" libraries are now deprecated since 2022 https://github.com/Azure/azure-libraries-for-net we need to migrate our usages over to the new "azure sdk for net" libraries. This was initially done for the Azure App service Project and now is continuing for the rest of the azure related projects where possible 

[SC-74178]

# Results 
- Moved the ARM client wrapper into a shared library `Calamari.Azure` so it can be reused by other projects 
- Updated `Calamari.Azure` to target .net 6.0 and also updated some dependencies in `Calamari` project
- Removed `Microsoft.Azure.Management` dependencies in `Calamari.AzureAppService` projects
- Removed unused code in `Calamari.AzureAppService` that was missed when removing the legacy behaviours 
- Added a new behaviour to `Calamari.AzureResourceGroup` `DeployAzureResourceGroupBehaviour` that uses the modern azure sdk and renamed the existing one to `LegacyDeployAzureResourceGroupBehaviour`. The behaviour that gets used is based on the `ModernAzureSdkFeatureToggle` feature


source/Calamari.Azure/Calamari.Azure.csproj
```diff
@@ -5,10 +5,13 @@
         <OutputType>Library</OutputType>
         <Authors>Octopus Deploy</Authors>
         <Copyright>Octopus Deploy Pty Ltd</Copyright>
-        <TargetFrameworks>net462;netstandard2.1</TargetFrameworks>
+        <RuntimeIdentifiers>win-x64;linux-x64;osx-x64;linux-arm;linux-arm64</RuntimeIdentifiers>
+        <LangVersion>8.0</LangVersion>
+        <TargetFrameworks>net462;net6.0</TargetFrameworks>
```
tleed5, line 10: Is there some consequences of doing this? I kinda did it blindly but the builds were working so I didn't worry about it too much 🤷

source/Calamari.AzureAppService/Calamari.AzureAppService.csproj
```diff
@@ -12,15 +12,9 @@
     <TargetFrameworks>net462;net6.0</TargetFrameworks>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Azure.Identity" Version="1.10.4" />
-    <PackageReference Include="Azure.ResourceManager" Version="1.7.0" />
```
tleed5, line 16: I figure it is better for these dependencies to be managed solely by `Calamari.Azure`

source/Calamari.AzureResourceGroup/DeployAzureResourceGroupBehaviour.cs
```diff
@@ -32,197 +41,139 @@ public DeployAzureResourceGroupBehaviour(TemplateService templateService, IResou
...
-                                previousPollWait = currentPollWait;
-                                currentPollWait = temp + currentPollWait;
-                            }
-
-                            break;
-                    }
-                }
+                var deploymentResult = await AsyncResourceGroupPollingTimeoutPolicy.ExecuteAsync(async timeoutCancellationToken =>
+                                                                                                 {
+                                                                                                     var delayStrategy = DelayStrategy.CreateExponentialDelayStrategy(TimeSpan.FromSeconds(1), PollingTimeout);
```
kevjt, line 136: Is `DelayStrategy.CreateExponentialDelayStrategy` a built-in way for the armClient to retry/poll the state of the operation? If so then you'd want to have a separate timeout for your `AsyncResourceGroupPollingTimeoutPolicy` because I assume you meant that as the absolute timeout. At the moment, they share the 30 second timeout, however it will take much longer than 30 seconds for the exponential delay strategy to actually reach the maximum 30 second polling interval (e.g. with x2 exponential 1+2+4+8+16+30=61). 

I can see that `AzureAppServiceZipDeployBehaviour` has a polling timeout of 3 minutes so perhaps that would be a sensible figure to use.

source/Calamari.AzureResourceGroup/DeployAzureResourceGroupBehaviour.cs
```diff
@@ -32,197 +41,139 @@ public DeployAzureResourceGroupBehaviour(TemplateService templateService, IResou
...
+                                                                                                     var result = await deploymentOperation.WaitForCompletionAsync(delayStrategy, timeoutCancellationToken);
+                                                                                                     return result;
+                                                                                                 },
+                                                                                                 CancellationToken.None);
+                log.Info($"Deployment completed with status: {deploymentResult.Data.Properties?.ProvisioningState}");
             }
+            catch
+            {
+                log.Error("Error polling for deployment completion");
+                throw;
```
kevjt, line 146: How does this exception get handled? Does it need to be wrapped in a `CommandException` for the correct exit status?

source/Calamari.Azure/Calamari.Azure.csproj
kevjt: I don't see an issue but @scme0 would be the guy with a better answer

kevjt (commented):
Minor comments plus still some merge conflicts. On the whole looks good. Nice work!

source/Calamari.AzureResourceGroup/DeployAzureResourceGroupBehaviour.cs
tleed5: Ah good catch, I was trying to replicate a similar behaviour that was used previously i.e. a fibonacci step off.  

Ill update `AsyncResourceGroupPollingTimeoutPolicy` to 3 minutes which will result in a maximum timeout of 3 minutes and it should make around 10 api calls(Assuming the deployment takes that long and my math is right) which I think is all good.

source/Calamari.AzureResourceGroup/DeployAzureResourceGroupBehaviour.cs
tleed5: It bubbles up and hits this [exception handler](https://github.com/OctopusDeploy/Calamari/blob/a5aeca4690dfc5dbde4494fe7c30cb8312c7e1a8/source/Calamari.Common/CalamariFlavourProgram.cs#L82-L85) which logs out the exception and returns exit code 100. I could wrap it in a `CommandException` so it returns exit code 1 but it might potentially lose some detail in the error log

Screenshot 2024-04-12 at 11 13 20 am

kevjt: I think you can pass the actual exception into `CommandException` as the innerException and you would still get the full stack trace of the inner exception. I'm not to concerned either way because I don't know the fine distinction between a command error vs other error exit status code.

source/Calamari.Azure/Calamari.Azure.csproj
scme0: I *think* it should be ok as long as it's dependencies are happy with it.

kevjt (approved):
LGTM

PR #1239, opened 2024-04-04
Fix local build for *nix

scme0:
Adjusting the Flavours that are handled makes the build succeed. Obviously this means the bundle is missing certain flavours so I've made it only do this for local builds.

### DIFF
--- a/build/Build.cs
+++ b/build/Build.cs
@@ -218,9 +218,10 @@ namespace Calamari.Build
                  .DependsOn(Compile)
                  .Executes(async () =>
                  {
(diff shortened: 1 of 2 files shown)


tleed5 (approved):
Awesome, thank you for fixing this. Is the reason some of the flavours aren't cross platform just that they need to be updated to support .net 6.0?

scme0:
I believe some nuget packages are full framework only, though I'm not 100% sure if we just need to use newer versions of nuget packages or whatever. For now I guess this allows developers to build locally but we'll need to adjust these lists if/when we adjust these flavours to become crossplatform. 👍

PR #1242, opened 2024-04-05
Allow kubectl server side apply

tleed5:
# Background
As a part of supporting kubectl server side apply we need to update Calamari to allow it to pass through the required kubectl arguments `--server-side` `--force-conflicts` and `--field-manager`

[SC-51459]

# Results
If the known variables `Octopus.Action.Kubernetes.ServerSideApply.Enabled` and `Octopus.Action.Kubernetes.ServerSideApply.ForceConflicts` are set to `"true"` the `--server-side` `--force-conflicts` and `--field-manager` arguments will be added to the kubectl apply

### DIFF
--- a/source/Calamari.Tests/KubernetesFixtures/Commands/Executors/GatherAndApplyRawYamlExecutorTests.cs
+++ b/source/Calamari.Tests/KubernetesFixtures/Commands/Executors/GatherAndApplyRawYamlExecutorTests.cs
@@ -120,6 +120,145 @@ namespace Calamari.Tests.KubernetesFixtures.Commands.Executors
         }
 
         [Test]
+        public async Task AppliesKubernetesManifestsWithServerSideApply()
+        {
+            // Arrange
+            AddTestFiles();
+            SetupCommandLineRunnerMocks();
+            var variables = new CalamariVariables
+            {
+                [KnownVariables.EnabledFeatureToggles] = FeatureToggle.GlobPathsGroupSupportFeatureToggle.ToString(),
+                [KnownVariables.OriginalPackageDirectoryPath] = StagingDirectory,
+                [SpecialVariables.CustomResourceYamlFileName] = "dirA/*\ndirB/*",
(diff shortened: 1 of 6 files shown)


source/Calamari.Tests/KubernetesFixtures/Commands/Executors/GatherAndApplyRawYamlExecutorTests.cs
```diff
@@ -119,6 +119,145 @@ public async Task AppliesKubernetesManifestsAndAddsResourceIdentifiers()
...
+            commandLineRunner.ReceivedCalls().Count().Should().Be(4);
+            var commandLineArgs = commandLineRunner.ReceivedCalls().SelectMany(call => call.GetArguments().Select(arg => arg.ToString())).ToArray();
+            commandLineArgs[0]
+                .Should()
+                .Contain("apply -f")
+                .And.Contain("--recursive")
+                .And.Contain("-o json")
+                .And.Contain($"{Path.Combine("grouped", "1")}")
+                .And.Contain("--server-side")
+                .And.Contain("--field-manager octopus");
```
kevjt, line 155: It's probably a good idea to check that it also does not contain the `--force-conflicts` flag since this is the only difference between this test and the next one

kevjt (commented):
LGTM

Just wanted to double check on the naming of the field manager as "octopus" before we merge this. I'm worried octopus could be a common word so something like "octopusdeploy" might be better at preventing any possible name collisions.

kevjt:
> LGTM
> 
> Just wanted to double check on the naming of the field manager as "octopus" before we merge this. I'm worried octopus could be a common word so something like "octopusdeploy" might be better at preventing any possible name collisions.

After some discussion in this [thread](https://octopusdeploy.slack.com/archives/C06A5JTBCC8/p1712642115277749), we will be sticking with `octopus` as the name for the field manager.

kevjt (approved):

PR #1247, opened 2024-04-10
Remove obsolete Kubernetes deployment scripts

kevjt:
## Background

The KubernetesDeployment scripts were [originally added](https://github.com/OctopusDeploy/Calamari/pull/1026) for testing the `RawYamlCommand` at the time. The idea was to use the scripts that Octopus Server would normally send during runtime, and run the deployment using the former `RunScriptCommand` method to see if the `RawYamlCommand` produced any differences/regressions.

The feature toggles that were added to [introduce that change](https://github.com/OctopusDeploy/OctopusDeploy/pull/17529/files#diff-906cb3427b61052fdffa7d98975d6722e8e9669d4a90fcd10238689badc6ea15R49-R62) have already been [removed](https://github.com/OctopusDeploy/OctopusDeploy/pull/21036). Thus these files are no longer needed as the raw yaml step does not rely on the script wrapper anymore.

[sc-75366]


scme0 (approved):
LGTM

PR #1249, opened 2024-04-22
Pin AzureRM terraform provider version in tests

APErebus:
We were getting failing tests for some reason due to a change in the AzureRM terraform provider. I have pinned this at the last known good version


flin-8 (approved):
LGTM

PR #1250, opened 2024-04-23
Retry temporary file deletion

tothegills:
A customer has encountered an issue on a dynamic worker deleting the interim file when deploying an Azure App Service. It is difficult to determine what is holding onto the file. I assume it is a transient issue, so hammering it with retries.

```
The process cannot access the file 'C:\Octopus\Tentacle\Files\ABC@S0.0.0-preview.0.9@C648E19E03056448B5851DB873FD5237.zip' because it is being used by another process.
```

Relates to https://github.com/OctopusDeploy/Issues/issues/8727


susanpann (approved):
looks good!

PR #1251, opened 2024-04-24
Pin versions of Terraform and the AzureRM Terraform provider

rain-on:
A number of flakey tests appeared on the 12-April-2024, which corresponds to when new versions of both the azureRM provider and Terraform itself were released.

This change pins both of these dependencies to versions which were available prior to 12-April.

### DIFF
--- a/source/Calamari.Tests/KubernetesFixtures/Terraform/Clusters/AKS-local-access-disabled/providers.tf
+++ b/source/Calamari.Tests/KubernetesFixtures/Terraform/Clusters/AKS-local-access-disabled/providers.tf
@@ -2,7 +2,7 @@ terraform {
   required_providers {
     azurerm = {
       source  = "hashicorp/azurerm"
-      version = "3.99.0" 
+      version = "3.95.0" 
     }
     kubernetes = {
       source  = "hashicorp/kubernetes"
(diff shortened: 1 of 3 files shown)


source/Calamari.Tests/KubernetesFixtures/Terraform/Clusters/AKS-local-access-disabled/providers.tf
```diff
@@ -2,7 +2,7 @@ terraform {
   required_providers {
     azurerm = {
       source  = "hashicorp/azurerm"
-      version = "3.99.0" 
+      version = "3.95.0"
```
APErebus, line 5: Can you also replace this in the other aks providers.tf. Just so they are the same

rain-on:
Its looking like the problem is _not_ terraform or the provider, but rather something that's happening in Azure.

This issue(https://github.com/hashicorp/terraform-provider-azurerm/issues/18268#issuecomment-2064133115) has comments indicating the problem is not tied to a any given version of the azurerm provider, and "randomly" started appearing a few weeks agin.

Suggesting we mute the affected tests for a week, and monitor until azure/terraform/provider are happy once again.

PR #1253, opened 2024-05-07
Update to polly 8.3.1

rain-on:
Currently the azureappservice project is using _quite_ an old version of polly (5.6.1) - this upgrades the dependency to 8.3.1 - which is still supports .netframework 4.6.2.


APErebus (approved):
LGTM! Thanks @rain-on

PR #1255, opened 2024-05-10
Insert all Secret References into External Variables

rain-on:
All external variables now have secret references such that if an env var does not exist, it can instead get the values from 1password.

The values also exist in 1password, but are currently duplicates of the values used in Octopus Server.


source/Calamari.Testing/EnvironmentVariables.cs
```diff
@@ -11,73 +11,75 @@ namespace Calamari.Testing
...
+        [EnvironmentVariable("CALAMARI_AUTHPASSWORD", "op://Calamari Secrets For Tests/MyGet Package Manager/password")]
         MyGetFeedPassword,
 
-        [EnvironmentVariable("GOOGLECLOUD_OCTOPUSAPITESTER_JSONKEY", "GoogleCloud - OctopusAPITester", "op://Calamari Secrets for Tests/Google Cloud Octopus Api Tester JsonKey")]
+        [EnvironmentVariable("GOOGLECLOUD_OCTOPUSAPITESTER_JSONKEY", "op://Calamari Secrets for Tests/Google Cloud Octopus Api Tester/jsonkey")]
         GoogleCloudJsonKeyfile,
 
-        [EnvironmentVariable("GitHub_RateLimitingPersonalAccessToken", "GitHub test account PAT")]
+        //TODO(tmm): not sure about this one - this is a copy of the github account above.
+        [EnvironmentVariable("GitHub_RateLimitingPersonalAccessToken", "op://Calamari Secrets for Tests/GitHub Test Account/PAT")]
```
MissedTheMark, line 82: Is there anyone we can talk to about this? Looks like @IsaacCalligeros95 set this up originally?

MissedTheMark (approved):
Looks good to me, though it would be good to get some clarity on the rate-limiting stuff.

source/Calamari.Testing/EnvironmentVariables.cs
IsaacCalligeros95: This variable is used in the test setup InstallTools to get the aws-iam-authenticator. It was available in LastPass but doesn't look to have made it to 1Password. I just used a separate token to avoid any potential rate limiting on the existing tokens but we can probably just use any of the existing GH PAT variables in teamcity.  It can be any user token and doesn't require any permissions, it's just for that install step.

MissedTheMark: Thanks for the context @IsaacCalligeros95!

PR #1257, opened 2024-05-10
Fix: AzureRM Deprecation feature toggle

rain-on:
The feature toggle was not being passed into underlying ps script correctly.
This change updates the wrapper to explicitly add the enabled feature toggle to the variables passed into the script.

This also includes a few changes to make the calamari tests more platform-agnostic:
* Using Path.PathSeparator for generating path env vars
* Attempting to use pwsh when checking power-shell versioning


MissedTheMark (approved):
Changes look good. 
@rain-on demoed the changes to me over Zoom, showing that it gives the expected results when the toggle is removed/disabled/enabled.

PR #1259, opened 2024-05-14
Fix KOS not running for Kubernetes Agent scripts

APErebus:
The Kubernetes Object Status  was not correctly running for Deploy Containers  via Kubernetes agent because this step is actually executed as a script. The `ResourceStatusReporterWrapper` was being skipped for Kubernetes agents as there is no cluster url or cluster names.

After a discussion with @zentron, we decided it makes sense to just rely on the `ResourceStatusCheck` variable value (once we pass the blue/green & wait deployment checks

Shortcut story: [sc-78819]

### DIFF
--- a/source/Calamari/Kubernetes/KubernetesVariableExtensions.cs
+++ b/source/Calamari/Kubernetes/KubernetesVariableExtensions.cs
@@ -1,0 +1,22 @@
+﻿using System;
+using Calamari.Common.Plumbing.Variables;
+
+namespace Calamari.Kubernetes
+{
+    public static class KubernetesVariableExtensions
+    {
+        public static bool IsKubernetesScript(this IVariables variables)
+        {
+            //A Kubernetes agent won't have any of the other variable set
+            var isKubernetesAgentTarget = string.Equals(variables.Get(MachineVariables.DeploymentTargetType), "KubernetesTentacle", StringComparison.OrdinalIgnoreCase);
+            
+            var hasClusterUrl = !string.IsNullOrEmpty(variables.Get(SpecialVariables.ClusterUrl));
+            
+            var hasClusterName = !string.IsNullOrEmpty(variables.Get(SpecialVariables.AksClusterName)) ||
+                                 !string.IsNullOrEmpty(variables.Get(SpecialVariables.EksClusterName)) ||
+                                 !string.IsNullOrEmpty(variables.Get(SpecialVariables.GkeClusterName));
+            
+            return isKubernetesAgentTarget || hasClusterUrl || hasClusterName;
+        }
+    }
+}
--- a/source/Calamari/Kubernetes/KubernetesVariableExtensions.cs
+++ b/source/Calamari/Kubernetes/KubernetesVariableExtensions.cs
@@ -7,7 +7,7 @@ namespace Calamari.Kubernetes
     {
         public static bool IsKubernetesScript(this IVariables variables)
         {
-            //A Kubernetes agent won't have any of the other variable set
+            //A Kubernetes agent won't have any of the other variables set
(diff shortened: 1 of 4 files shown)


source/Calamari/Kubernetes/KubernetesVariableExtensions.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using Calamari.Common.Plumbing.Variables;
+
+namespace Calamari.Kubernetes
+{
+    public static class KubernetesVariableExtensions
+    {
+        public static bool IsKubernetesScript(this IVariables variables)
```
APErebus, line 8: Moved into a method for clarity

source/Calamari/Kubernetes/KubernetesVariableExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+        {
+            //A Kubernetes agent won't have any of the other variable set
```
scme0, line 10:
```suggestion
            //A Kubernetes agent won't have any of the other variables set
```

source/Calamari/Kubernetes/ResourceStatus/ResourceStatusReportWrapper.cs
```diff
@@ -35,19 +35,18 @@ public class ResourceStatusReportWrapper : IScriptWrapper
...
         public bool IsEnabled(ScriptSyntax syntax)
         {
-            var resourceStatusEnabled = variables.GetFlag(SpecialVariables.ResourceStatusCheck);
-            var isBlueGreen = variables.Get(SpecialVariables.DeploymentStyle) == "bluegreen";
-            var isWaitDeployment = variables.Get(SpecialVariables.DeploymentWait) == "wait";
-            if (!resourceStatusEnabled || isBlueGreen || isWaitDeployment)
+            var isBlueGreen = string.Equals(variables.Get(SpecialVariables.DeploymentStyle), "bluegreen", StringComparison.OrdinalIgnoreCase);
+            var isWaitDeployment = string.Equals(variables.Get(SpecialVariables.DeploymentWait) , "wait", StringComparison.OrdinalIgnoreCase);
+
+            //A blue/green or deployment wait is already waiting, so we don't run the
```
scme0, line 41: so we don't run the what? The world wants to know 😄

source/Calamari/Kubernetes/ResourceStatus/ResourceStatusReportWrapper.cs
```diff
@@ -35,19 +35,18 @@ public class ResourceStatusReportWrapper : IScriptWrapper
...
                 return false;
             }
 
-            var hasClusterUrl = !string.IsNullOrEmpty(variables.Get(SpecialVariables.ClusterUrl));
-            var hasClusterName = !string.IsNullOrEmpty(variables.Get(SpecialVariables.AksClusterName)) ||
-                                 !string.IsNullOrEmpty(variables.Get(SpecialVariables.EksClusterName)) ||
-                                 !string.IsNullOrEmpty(variables.Get(SpecialVariables.GkeClusterName));
-            return hasClusterUrl || hasClusterName;
+            // At this point, we only care about the status of the resource status check
+            // If someone has set this variable manually, then it might blow up, but hey, that's on them
```
scme0, line 48: nit: I don't love casual style code comments.
```suggestion
            // If someone has set this variable manually then it might blow up, but that's not a supported configuration
```

scme0 (approved):
Nice one! Just a few random nit picks.

I was we ported all the k8s steps to Calamari, then this kind of thing wouldn't happen!

source/Calamari/Kubernetes/ResourceStatus/ResourceStatusReportWrapper.cs
APErebus: Fair 😁

APErebus:
Related to https://github.com/OctopusDeploy/Issues/issues/8789

PR #1264, opened 2024-05-17
Remove win-x86 runtime identifiers

APErebus:
These are the only two projects that win-x86 builds are being built for, which seems incorrect.


flin-8 (approved):
x86 is so 1990's

PR #1266, opened 2024-05-22
Support kubernetes file system, retrieve storage space information from env vars

scme0:
# Background

The Kubernetes Script Pods use a PV mount to get access to the shared file system managed by the Kubernetes Tentacle. When Calamari is acquiring packages, it checks if it should do package cleanup first.

Currently the Calamari implementation cannot do package retention when in the Kubernetes agent because it is not able to calculate the amount of storage on the PV.

[sc-76306]

# Results

This PR introduces a FileSystem type for Kubernetes. Instead of trying to calculate the storage usage itself, it relies on Tentacle setting the values as environment variables when launching script pods. The PR to add that is here (https://github.com/OctopusDeploy/OctopusTentacle/pull/941).

With this, the package retention code can correctly calculate if it should clean up old packages when storage space is running low.

# Screen shot:

Screenshot 2024-05-22 at 11 18 00

### DIFF
--- a/source/Calamari.Common/Plumbing/CalamariEnvironment.cs
+++ b/source/Calamari.Common/Plumbing/CalamariEnvironment.cs
@@ -12,6 +12,9 @@ namespace Calamari.Common.Plumbing
         /// </summary>
         public static readonly bool IsRunningOnMono = Type.GetType("Mono.Runtime") != null;
 
+        public static bool IsRunningOnKubernetes =>
+            Environment.GetEnvironmentVariable(KubernetesEnvironmentVariables.KubernetesAgentNamespace) != null;
+
         /// <summary>
         /// Based on some internal methods used my mono itself
         /// https://github.com/mono/mono/blob/master/mcs/class/corlib/System/Environment.cs
--- a/source/Calamari.Common/Plumbing/CalamariEnvironment.cs
+++ b/source/Calamari.Common/Plumbing/CalamariEnvironment.cs
@@ -13,7 +13,7 @@ namespace Calamari.Common.Plumbing
         public static readonly bool IsRunningOnMono = Type.GetType("Mono.Runtime") != null;
 
         public static bool IsRunningOnKubernetes =>
-            Environment.GetEnvironmentVariable(KubernetesEnvironmentVariables.KubernetesAgentNamespace) != null;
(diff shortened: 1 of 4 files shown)


source/Calamari.Common/Plumbing/KubernetesEnvironmentVariables.cs
```diff
@@ -0,0 +1,9 @@
+namespace Calamari.Common.Plumbing
+{
+    public static class KubernetesEnvironmentVariables
+    {
+        public const string KubernetesAgentNamespace = "OCTOPUS__K8STENTACLE__NAMESPACE";
+        public const string KubernetesAgentVolumeFreeBytes = "OCTOPUS__K8STENTACLE__PERSISTENVOLUMEFREEBYTES";
+        public const string KubernetesAgentVolumeTotalBytes = "OCTOPUS__K8STENTACLE__PERSISTENVOLUMETOTALBYTES";
```
APErebus, line 7:
```suggestion
        public const string KubernetesAgentVolumeFreeBytes = "OCTOPUS__K8STENTACLE__PERSISTENTVOLUMEFREEBYTES";
        public const string KubernetesAgentVolumeTotalBytes = "OCTOPUS__K8STENTACLE__PERSISTENTVOLUMETOTALBYTES";
```

APErebus (changes requested):

source/Calamari.Common/Plumbing/KubernetesEnvironmentVariables.cs
scme0: Dang it! Good catch!

source/Calamari.Common/Plumbing/CalamariEnvironment.cs
```diff
@@ -12,6 +12,9 @@ public static class CalamariEnvironment
         /// </summary>
         public static readonly bool IsRunningOnMono = Type.GetType("Mono.Runtime") != null;
 
+        public static bool IsRunningOnKubernetes =>
+            Environment.GetEnvironmentVariable(KubernetesEnvironmentVariables.KubernetesAgentNamespace) != null;
```
APErebus, line 16: The more I think about it, `string.IsNullOrWhitespace` is safer
```suggestion
            !string.IsNullOrWhiteSpace(Environment.GetEnvironmentVariable(KubernetesEnvironmentVariables.KubernetesAgentNamespace));
```

APErebus (approved):
1 suggestion, but not blocking

PR #1268, opened 2024-05-22
Upgrade SharpCompress to 0.37.2

APErebus:
Upgrades [SharpCompress](https://github.com/adamhathcock/sharpcompress) to [0.37.2](https://github.com/adamhathcock/sharpcompress/releases/tag/0.37.2)

which fixes an issue with packages > 4GB in size.

I had to bump `System.Text.Encoding.CodePages` as well as sharpcompress needed > `8.0.0`

Shortcut story: [sc-60060]


source/Calamari.Common/Features/Packages/TarPackageExtractor.cs
```diff
@@ -25,58 +23,61 @@ public TarPackageExtractor(ILog log)
...
+            finally
+            {
+                if (compressionStream != inStream)
+                    compressionStream.Dispose();
+            }
 
             return files;
         }
 
+        static void EnsureTargetDirectoryExists(string directory) => Directory.CreateDirectory(directory);
```
APErebus, line 49: There was a change in SharpCompress `0.29` that it now verifies that the target directory exists before extraction, so we now verify the target directory exists :)

MissedTheMark (approved):
LGTM.

source/Calamari.Tests/AWS/S3/CompressedPackages/TestTarBzip2Package.0.0.1.tar.bz2
APErebus, line 1: For some reason these two tar files were corrupt (according to `SharpCompress`) and couldn't be decompressed. Once I recreated them in WSL, they worked 🤷 

@zentron Should I be concerned about this?

zentron: 🤔 any idea about _why_ its corrupt? We have had some weird cases with compression algorithms in the past handing edge cases better/worse. 
I'd be more comfortable if we do a quick dig into what the decompression logic is doing that's different

APErebus: I'll investigate further. I was just getting EOF errors from SharpCompress.

When I tried to extract it via 7-Zip locally, it failed as well

zentron (approved):
If it doesn't work locally with any other tools and we cant reproduce that format, lets move forward and respond if something comes up.

source/Calamari.Tests/AWS/S3/CompressedPackages/TestTarBzip2Package.0.0.1.tar.bz2
APErebus: @MissedTheMark and I investigated and it looks like those archives are just corrupt and the old version of SharpCompress actually silently failed to decompress these archives.

PR #1269, opened 2024-05-22
Add warning logs if no or not enough packages can be deleted to free up the desired space

scme0:
# Background

Sometimes the desired free space can't be freed up by package retention. This could be if there are no packages to delete and something else is taking up the space, or if packages have been added but because a workspace isn't cleaned up properly, the package journal still has a lock for a specific package. This means package retention cannot delete it.

This could also happen if you have a lot of parallel deployments.

# Results

This PR just adds some warning logs to that it's more obvious to users why packages aren't being removed. The last log just says that it will remove a certain amount of memory's worth of packages but then there are no follow up messages to say what was removed. This will give users an indication of what's happening.


scme0:
Closing as not needed.

PR #1270, opened 2024-05-22
Correctly set WindowsFxVersion for windows container app services

APErebus:
When deploying a container to an app service, we were incorrectly always setting the `LinuxFxVersion` for all app service operating systems. However, if this is a windows app service, we should be setting the `WindowsFxVersion` property.

We can determine what the app service's OS is by looking at the `kind` property.

Shortcut story: [sc-41798]

### DIFF
--- a/source/Calamari.AzureAppService.Tests/AzureAppServiceDeployContainerBehaviourFixture.cs
+++ b/source/Calamari.AzureAppService.Tests/AzureAppServiceDeployContainerBehaviourFixture.cs
@@ -18,136 +18,260 @@ using Octostache;
 
 namespace Calamari.AzureAppService.Tests
 {
-    [TestFixture]
-    public class AzureAppServiceDeployContainerBehaviorFixture : AppServiceIntegrationTest
+    public class AzureAppServiceDeployContainerBehaviourFixture
     {
-        CalamariVariables newVariables;
-        readonly HttpClient client = new HttpClient();
-
-        // For some reason we are having issues creating these linux resources on Standard in EastUS
-        protected override string DefaultResourceGroupLocation => "westus2";
-
-        protected override async Task ConfigureTestResources(ResourceGroupResource resourceGroup)
+        [TestFixture]
+        public class WhenUsingAWindowsAppService : AppServiceIntegrationTest
         {
-            var (_, webSite) = await CreateAppServicePlanAndWebApp(resourceGroup,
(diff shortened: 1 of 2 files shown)


source/Calamari.AzureAppService.Tests/AzureAppServiceDeployContainerBehaviourFixture.cs
```diff
@@ -18,136 +18,260 @@
...
 namespace Calamari.AzureAppService.Tests
 {
-    [TestFixture]
-    public class AzureAppServiceDeployContainerBehaviorFixture : AppServiceIntegrationTest
+    public class AzureAppServiceDeployContainerBehaviourFixture
     {
-        CalamariVariables newVariables;
-        readonly HttpClient client = new HttpClient();
+        [TestFixture]
+        public class WhenUsingAWindowsAppService : AppServiceIntegrationTest
```
MissedTheMark, line 24: This and the linux test are checking the exact same thing, in the same way, just with a different setup, correct?  
Should they share the `Azure[Windows/Linux]ContainerDeploy()` and `Azure[Windows/Linux]ContainerSlotDeploy()` methods in that case?  Generally I'm not a fan of sharing test code, but these seem like they're actually the same tests.  
It's a little unfortunate that the AppServiceIntegrationTest structure is such that you can't use different setups for the same tests.

MissedTheMark (approved):
Overall, looks good.  Just left a comment about the duplicated code performing the same tests for Windows/Linux just with different setup.

source/Calamari.AzureAppService.Tests/AzureAppServiceDeployContainerBehaviourFixture.cs
APErebus: Correct. They have different setups

They could share the code 🤔 I'm probably more inclined to keep the separate for now. I'm not sure we gain _that_ much by sharing the code

MissedTheMark: My main reason for suggesting sharing is that currently it's hard to quickly tell that they're doing the exact same thing.  Adding a comment saying "this is does the same thing as the linux tests with diff setups" would provide a similar outcome.
Up to you though, it's not a huge deal.

APErebus: Added a comment

PR #1272, opened 2024-05-27
Order ScriptProxyFixtureBase tests

APErebus:
Due to test splitting, we need the no proxy tests to run first so it can pre-acquire packages without running through a non-existant proxy


IsaacCalligeros95 (approved):
LGTM, cheers for the pipeline speed up!

APErebus:
Confirming that it fixes the issue when the tests are split

image

PR #1274, opened 2024-05-29
Preserve .tar file attributes when extracting

APErebus:
When extracting .tar files, we preserve the file attributes (such as permissions)


APErebus:
This attribute is not supported/implemented.

PR #1161, opened 2023-10-11
Attempt to clean up the upload file if it was duplicated/renamed during deployment

benPearce1:
[sc-61082]

This is a workaround which probably requires a better solution.

Related to https://github.com/OctopusDeploy/Issues/issues/7382

Related to https://github.com/OctopusDeploy/Calamari/pull/1155

PR #1275, opened 2024-05-29
Fix Async deploy file lock

IsaacCalligeros95:
The Async zip deploy functionality clashed with a change made to [cleanup uploaded files](https://github.com/OctopusDeploy/Calamari/pull/1161/files). This change leaves in place the delete re-try, although that's likely redundant now it may help in other scenarios. This change also extends the file clean-up to the code path where any of the following variables are used.
`
KnownVariables.Features.ConfigurationTransforms,
KnownVariables.Features.StructuredConfigurationVariables,
KnownVariables.Features.SubstituteInFiles
`


hnrkndrssn (approved):
Nice, looks good to me! 👍

PR #1282, opened 2024-05-31
Set source branches to release for releases

IsaacCalligeros95:
We're having issues with GitVersion selecting non release branches & creating incorrect build versions for tags. This change adds the source-branches restriction of releases on release builds. 

https://gitversion.net/docs/reference/configuration
> source-branches
Because Git commits only refer to parent commits (not branches) GitVersion sometimes cannot tell which branch the current branch was branched from.

The approach here would be to:
* Create a PR targeting your release version, in this case, 27.0.10.
* Merge.
~~Create a hotfix tag off of that release branch.~~ TBD whether this is required.
* Trigger the build in TeamCity against the release branch. (Subsequent builds against the same release branch should auto increment the hotfix version)

### DIFF
--- a/GitVersion.yml
+++ b/GitVersion.yml
@@ -4,6 +4,7 @@ branches:
     mode: ContinuousDelivery
     is-release-branch: false
     tag: hotfix
+    source-branches: [ 'release' ]
     increment: None
 ignore:
   sha: [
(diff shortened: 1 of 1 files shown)


scme0 (approved):

GitVersion.yml
```diff
@@ -4,13 +4,13 @@ branches:
     mode: ContinuousDelivery
     is-release-branch: false
     tag: hotfix
+    source-branches: [ 'release' ]
```
hnrkndrssn, line 7: does this need to be `release/*`?

hnrkndrssn (approved):

PR #1284, opened 2024-05-31
Fix output variable values for base kubernetes apply

eddymoulton:
This command is used to populate the output variable in the deployment process with the JSON output of each resource.

While testing something else, I noticed it wasn't working - turns out we were missing the namespace here.


kevjt (approved):
nice. Was this working before but only for non namespaced resources?

PR #1286, opened 2024-06-03
Randomly assign region for Azure WebApp container tests

APErebus:
We've been getting random failures when performing the windows container web app tests. This PR just randomises against a large list of regions to avoid capacity issues.

Shortcut story: [sc-80825]

### DIFF
--- a/source/Calamari.AzureAppService.Tests/RandomAzureRegion.cs
+++ b/source/Calamari.AzureAppService.Tests/RandomAzureRegion.cs
@@ -1,0 +1,30 @@
+﻿using System;
+using System.Linq;
+using JetBrains.Annotations;
+
+namespace Calamari.AzureAppService.Tests
+{
+    public static class RandomAzureRegion
+    {
+        static Random random = new Random();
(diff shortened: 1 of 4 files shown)


source/Calamari.AzureAppService.Tests/RandomAzureRegion.cs
```diff
@@ -0,0 +1,30 @@
...
+            "centralus",
+            "eastus",
+            "eastus2",
+            "westus",
+            "westus2",
+            "westus3",
+            "australiaeast"
+        };
+
+        public static string GetRandomRegion(params string[] excludedRegions)
```
MissedTheMark, line 23: Take it or leave it - but on first read of the calls to this method I assumed it was choosing a random region from the regions provided as params.
```suggestion
        public static string GetRandomRegionExcluding(params string[] excludedRegions)
```

MissedTheMark (approved):
One suggestion, but overall LGTM.  🤞 this solves our issues!

PR #1293, opened 2024-06-19
Return applied k8s manifests as output vars

eddymoulton:
As a generic way to return the desired manifest of resources, we can set output variables for each:
- Step
- Deployment target
- K8S resource

This way each variable contains a single K8S resource and we can combine it/treat it separately as required.

We deserialize each resource and pull out important values to assist in moving the data around and processing it.


flin-8 (approved):
LGTM

flin-8 (changes requested):
Actually wait for the outcome of service msg's vs. output variables

PR #1294, opened 2024-06-24
Test recent Terraform versions and address flaky tests

APErebus:
We are not testing against modern versions of Terraform, instead testing against _very_ old versions.

We now test the last non-1.0 release (0.15.5) and then the latest released build in the 1.0 series.

Also, the Azure tests are using a very outdated provider. I'm hoping updating the provider will improve the reliability of these tests, as they have been quite flaky.

I adjusted the terraform tests on TeamCity to run 4 batches in parallel as the extra testing configurations add a fair bit of time, but with the parallelisation, this doesn't impact the builds.

Also includes a fix to the k8s tests to bump the AzureRM provider to pick up a fix to the sporadic 404's we get
https://github.com/hashicorp/terraform-provider-azurerm/pull/25758


source/Calamari.Terraform.Tests/Azure/example.tf
```diff
@@ -1,24 +1,22 @@
...
 
-variable "random" {
-  description = "A random value"
-}
+resource "random_pet" "prefix" {}
 
 resource "azurerm_resource_group" "resgrp" {
-  name     = "terraformtestrg${var.random}"
-  location = "AustraliaSouthEast"
+  name     = "${random_pet.prefix.id}-rg"
```
APErebus, line 12: swapped to the random pet prefix

source/Calamari.Terraform.Tests/Azure/example.tf
```diff
@@ -1,24 +1,22 @@
...
+  location = "Australia East"
 }
 
 resource "azurerm_app_service_plan" "service_plan" {
-  name                = "terraform_test_service_plan${var.random}"
-  location            = "${azurerm_resource_group.resgrp.location}"
-  resource_group_name = "${azurerm_resource_group.resgrp.name}"
+  name                = "${random_pet.prefix.id}-plan"
+  location            = azurerm_resource_group.resgrp.location
```
APErebus, line 18: we don't need to do the interpolation for plain values anymore

source/Calamari.Terraform.Tests/Azure/example.tf
```diff
@@ -1,24 +1,22 @@
...
-variable "random" {
-  description = "A random value"
-}
+resource "random_pet" "prefix" {}
 
 resource "azurerm_resource_group" "resgrp" {
-  name     = "terraformtestrg${var.random}"
-  location = "AustraliaSouthEast"
+  name     = "${random_pet.prefix.id}-rg"
+  location = "Australia East"
```
APErebus, line 13: Changed to Australia East as it's a bigger region

source/Calamari.Terraform.Tests/CommandsFixture.cs
```diff
@@ -510,7 +517,7 @@ public async Task AzureIntegration()
             var expectedHostName = $"{appName}.azurewebsites.net";
 
             using var temporaryFolder = TemporaryDirectory.Create();
-            CopyAllFiles(TestEnvironment.GetTestPath("Azure"), temporaryFolder.DirectoryPath);
+            CopyAllFiles(TestEnvironment.GetTestPath("Azure"), temporaryFolder.DirectoryPath, terraformCliVersion);
```
APErebus, line 520: Add the ability to have tf files for specific versions (in case the tf syntax changes). We don't use it here, but a safe addition.

source/Calamari.Terraform.Tests/CommandsFixture.cs
```diff
@@ -172,7 +171,7 @@ async Task DownloadCli(string destination, string version)
 
             if (Directory.Exists(destinationDirectoryName))
             {
-                var path = Directory.EnumerateFiles(destinationDirectoryName).FirstOrDefault();
+                var path = Directory.EnumerateFiles(destinationDirectoryName).FirstOrDefault(f => Path.GetFileName(f).Contains("terraform"));
```
APErebus, line 174: They started adding `LICENSE.txt` to the download, which is the first file, which is not the executable

source/Calamari.Terraform.Tests/CommandsFixture.cs
```diff
@@ -26,13 +26,12 @@
 
 namespace Calamari.Terraform.Tests
 {
-    [TestFixture("0.11.15")]
-    [TestFixture("0.13.0")]
-    [TestFixture("1.0.0")]
+    [TestFixture("0.13.7")]
+    [TestFixture("1.8.5")]
```
APErebus, line 30: After chatting with @zentron, our oldest supported version is `0.13` series, as it appears to not be EOL (yet)

source/Calamari.Terraform.Tests/CommandsFixture.cs
```diff
@@ -243,7 +240,7 @@ public void UserDefinedEnvVariables_OverrideDefaultBehaviour()
         [Test]
         public void ExtraInitParametersAreSet()
         {
-            IgnoreIfVersionIsNotInRange("0.11.15", "0.15.0");
+            IgnoreIfVersionIsNotInRange("0.0.0", "1.0.0", "-get-plugins was removed in 0.15.0/1.0.0");
```
APErebus, line 243: adjusted these versions

source/Calamari.Terraform.Tests/CommandsFixture.cs
```diff
@@ -447,21 +444,12 @@ public void UsesTemplateDirectory()
...
             using var temporaryFolder = TemporaryDirectory.Create();
             CopyAllFiles(TestEnvironment.GetTestPath("GoogleCloud"), temporaryFolder.DirectoryPath);
 
-            var environmentJsonKey = Environment.GetEnvironmentVariable(jsonEnvironmentVariableKey);
-            if (environmentJsonKey == null)
-            {
-                throw new Exception($"Environment Variable `{jsonEnvironmentVariableKey}` could not be found. The value can be found in the password store under GoogleCloud - OctopusAPITester");
-            }
-
+            var environmentJsonKey = await ExternalVariables.Get(ExternalVariable.GoogleCloudJsonKeyfile, CancellationToken.None);
```
APErebus, line 452: Updated to use the secret manager

PR #1295, opened 2024-06-24
Allow passing arguments to dotnet-script

hnrkndrssn:
At the moment we don't support passing arguments to the `dotnet-script` executable. This means that a customer that needs to load dependencies in their scripts that are different from those that are used by `dotnet-script` are out-of-luck as they are unable to instruct `dotnet-script` to use isolated assembly load context.

This PR extends our parsing of script arguments passed as part of the deployment process so that arguments that should be passed to `dotnet-script` can be specified and we will pass those through.

Relates to https://github.com/OctopusDeploy/Issues/issues/8885

[sc-82730]


IsaacCalligeros95 (approved):
LGTM, cheers for dealing with a regex issue 😢

PR #1299, opened 2024-06-25
Added logging for setting custom service credentials

benPearce1:
[sc-82975]

PR #1305, opened 2024-07-05
Added logging for setting custom service credentials

benPearce1:
replaces #1299


IsaacCalligeros95 (approved):
LGTM, I haven't looked in this area before do we actually have test coverage here? Not concerned by this change just curious.

PR #1304, opened 2024-07-01
Removed `KubernetesAksKubeloginFeatureToggle`

droyad:
Removed feature toggle as the feature has been rolled out successfully.

Did not test this, leaving it up to the team to decide wether that is needed.

PR #1306, opened 2024-07-08
Adds back the condition to check whether kubelogin is configured when using AKS kube login

mik-ky:
This was accidentally removed as part of the feature toggle removal in #1304


APErebus (approved):
Awesome! Thanks @mik-ky

PR #1308, opened 2024-07-17
Bundle full framework calamari executables

zentron:
As part of the Windows Server 2008 deprecation, we are [planning to](https://docs.google.com/document/d/1nG_L4azBM1lPhLi9YHMe1jUS74ZhGO7TckL1djYTjIo/edit#heading=h.31p7vw1j6nwj) temporarily publish netframework versions of Calamari into a single distributable for customers to manually install on their Windows 2008 machines.

This PR bundles together all the netframework executables into a single Calamari.Legacy zip pacakge. 
The next step will be to update the release process to push it somewhere for easy access

### DIFF
--- a/build/Build.cs
+++ b/build/Build.cs
@@ -354,7 +354,7 @@ namespace Calamari.Build
                   .Executes(() =>
                             {
                                 Log.Verbose($"Compressing Calamari.Legacy");
-                                LegacyCalamariDirectory.ZipTo(ArtifactsDirectory / "Calamari.Legacy.zip");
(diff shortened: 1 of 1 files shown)


APErebus (approved):
LGTM!

Do we also need to update the TeamCity build process to publish the new artifact?

zentron:
> Do we also need to update the TeamCity build process to publish the new artifact?

Yep, wanted to get this is mainline to avoid having to muck around with branch conditions. This zip is otherwise basically ignored in existing pipeline

PR #1309, opened 2024-07-18
skip SCM check if resource group name is missing.

domenicsim1:
skip SCM check if resource group name is not provided to fix the following issue. https://github.com/OctopusDeploy/Issues/issues/8906


tothegills (approved):
I haven't run it, but we've discussed on call and the if statement looks like it's in the right spot.

PR #1311, opened 2024-07-25
Only bundle the legacy Calamari package if running on Windows

zentron:


MissedTheMark (approved):
Just ran this locally on my Mac, and worked great. 🚀

PR #1313, opened 2024-07-29
Retrieve k8s resources from kubectl as single line JSON

APErebus:
We've had reports from customers that they sometimes receive JSON parse errors when Kubernetes Object Status is running.

After receiving some logs, it looks like we are getting truncated JSON back from the kubectl command.

By default, the `kubectl get X -o json` returns formatted, multi-line JSON.

By changing to `-o=jsonpath="{@}"`, we force it to output the entire object as _unformatted_ json, meaning by capturing one line, we should capture the entire json string

I also changed it so the debug output will now show _all_ the command output, not just the info/json string

Shortcut story: [sc-83213]

### DIFF
--- a/source/Calamari/Kubernetes/ResourceStatus/KubectlGet.cs
+++ b/source/Calamari/Kubernetes/ResourceStatus/KubectlGet.cs
@@ -1,3 +1,4 @@
+using System.Collections.Generic;
 using System.Linq;
 using Calamari.Common.Plumbing.Extensions;
 using Calamari.Kubernetes.Integration;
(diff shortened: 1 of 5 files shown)


source/Calamari/Kubernetes/ResourceStatus/KubectlGet.cs
```diff
@@ -6,26 +6,26 @@ namespace Calamari.Kubernetes.ResourceStatus
...
     public class KubectlGet : IKubectlGet
     {
-        public string Resource(string kind, string name, string @namespace, IKubectl kubectl)
+        public ICommandOutput Resource(string kind, string name, string @namespace, IKubectl kubectl)
         {
             return kubectl.ExecuteCommandAndReturnOutput(new[]
             {
-                "get", kind, name, "-o json", string.IsNullOrEmpty(@namespace) ? "" : $"-n {@namespace}"
-            }).Output.InfoLogs.Join(string.Empty);
+                "get", kind, name, "-o=jsonpath='{@}'", string.IsNullOrEmpty(@namespace) ? "" : $"-n {@namespace}"
```
APErebus, line 19: changing to `-o=jsonpath='{@}'` outputs the entire root json object on one line

source/Calamari/Kubernetes/ResourceStatus/ResourceRetriever.cs
```diff
@@ -72,27 +74,36 @@ private IEnumerable<Resource> GetChildrenResources(Resource parentResource, IKub
...
                 log.Error("---------------------------");
                 log.Error(jsonString);
                 log.Error("---------------------------");
+                log.Error("Full command output:");
+                log.Error("---------------------------");
+                foreach (var msg in commandOutput.Messages)
+                {
+                    log.Error($"{msg.Level}: {msg.Text}");
+                }
+                log.Error("---------------------------");
```
APErebus, line 106: Write the entire command output so we can see if anything is erroneously ending up in the error stream

MissedTheMark (approved):
Looks good!

zentron (approved):
Exposing the `ICommandOutput` parameter from the `IKubectlGet` interface feels like the abstraction is a little leaky, but if it helps us resolve the problem and make a change later then lets go with it

APErebus:
> Exposing the `ICommandOutput` parameter from the `IKubectlGet` interface feels like the abstraction is a little leaky, but if it helps us resolve the problem and make a change later then lets go with it

Yeah, that's a fair call. I can change it to return just a formatted string of all the output?

zentron:
> Yeah, that's a fair call. I can change it to return just a formatted string of all the output?

If that still works then that would be cleaner. I can see you later consume the different messages and expose their log levels so don't mind if you to make the call for this small change.

APErebus:
> If that still works then that would be cleaner. I can see you later consume the different messages and expose their log levels so don't mind if you to make the call for this small change.

Updated in [`ef4ccd8` (#1313)](https://github.com/OctopusDeploy/Calamari/pull/1313/commits/ef4ccd86efbe3fec9bbd42db27805be1cef940e5)

PR #1314, opened 2024-07-30
Make sure target directory exists when extracting using Jar tool

mik-ky:
This fixes an issue where extracting an additional package using a Jar tool would fail. When extracting a package that is referenced as an additional package, the package name is appended to the target directory. This does not exist at the time of verifying whether the archive is valid or not, which fails the action. 

When referenced as a main package it extracts to the current working folder, which is why it was not problematic.

**Before**
Pasted Graphic 7

**After**
Pasted Graphic 8

[sc-50612]

Relates to https://github.com/OctopusDeploy/Issues/issues/8181


tothegills (commented):
That `CombinedPackageExtractor` makes things difficult!
Looks good.

tothegills (approved):

PR #1317, opened 2024-08-16
BaseKubernetesApplyExecutor - Composition over inheritance

zentron:
Couple small changes around the kubernetes executor. 

The multiple implementations of the `BaseKubernetesApplyExecutor.cs` is a good example of where we should really be using composition over inheritance. There are parts of this that can be split out, moving the logic spread across multiple levels of this type hierarchy.
Ill expect more to come


source/Calamari/Kubernetes/Commands/Executors/GatherAndApplyRawYamlExecutor.cs
```diff
@@ -62,80 +64,40 @@ protected override async Task<IEnumerable<ResourceIdentifier>> ApplyAndGetResour
             return resourcesIdentifiers;
         }
 
-        IEnumerable<GlobDirectory> GroupFilesIntoDirectories(RunningDeployment deployment, List<string> globs, IVariables variables)
```
zentron, line 65: Pulling this glob processing out into its own class for SRP

source/Calamari/Kubernetes/Commands/Executors/BaseKubernetesApplyExecutor.cs
```diff
@@ -45,37 +41,10 @@ public async Task<bool> Execute(RunningDeployment deployment, Func<ResourceIdent
         }
         
         protected abstract Task<IEnumerable<ResourceIdentifier>> ApplyAndGetResourceIdentifiers(RunningDeployment deployment, Func<ResourceIdentifier[], Task> appliedResourcesCallback = null);
-        
-        protected void CheckResultForErrors(CommandResultWithOutput commandResult, string directory)
```
zentron, line 49: This can be made more generic than in the Base executor

source/Calamari/Kubernetes/Commands/Executors/GatherAndApplyRawYamlExecutor.cs
```diff
@@ -43,12 +41,16 @@ protected override async Task<IEnumerable<ResourceIdentifier>> ApplyAndGetResour
...
-            var directories = GroupFilesIntoDirectories(deployment, globs, variables);
-            
+
+            var globberGrouping = new GlobberGrouping(fileSystem);
+            var globDirectories = globberGrouping.Group(deployment.CurrentDirectory, globs);
+            variables.Set(SpecialVariables.GroupedYamlDirectories, string.Join(";", globDirectories.Select(d => d.Directory)));
+
             var resourcesIdentifiers = new HashSet<ResourceIdentifier>();
-            foreach (var directory in directories)
+            for (int i = 0; i < globDirectories.Count(); i++)
```
zentron, line 50: We seem to use the index number for the message below. Rather than putting this logging specific property on the GlobDirectory object, it has been changed to use this method.

source/Calamari/Kubernetes/Commands/Executors/KustomizeExecutor.cs
```diff
@@ -61,7 +61,12 @@ protected override async Task<IEnumerable<ResourceIdentifier>> ApplyAndGetResour
         void ValidateKubectlVersion(string currentDirectory)
         {
             var commandResult = kubectl.ExecuteCommandAndReturnOutput("version", "--client", "-o", "json");
-            CheckResultForErrors(commandResult, currentDirectory);
+            commandResult.LogErrorsWithSanitizedDirectory(log, currentDirectory);
+            if (commandResult.Result.ExitCode != 0)
```
zentron, line 65: Pulled out of the `CheckResultForErrors` method. The way we are using different exceptions around the place for control flow needs a little more scrutiny.

source/Calamari/Kubernetes/Commands/Executors/GlobberGrouping.cs
```diff
@@ -0,0 +1,72 @@
...
+            this.fileSystem = fileSystem;
+        }
+
+        /// <summary>
+        /// Groups together files from glob patterns into directories.
+        /// </summary>
+        /// <param name="deployment"></param>
+        /// <param name="rootDirectory"></param>
+        /// <param name="globs"></param>
+        /// <returns>Ordered result of directories</returns>
```
eddymoulton, line 24: nit: Just a little clearer about the ordering this spits out

```suggestion
        /// <returns>Result of directories in the same order as globs parameter</returns>
```

eddymoulton (approved):
Nice!

PR #1320, opened 2024-08-19
Perform variable substitutions on template values files from dependencies

APErebus:
When adding the new helm template values sources code, I missed the logic that did variable substitution on the files sourced from dependencies (packages/git repos). It's no longer working because we don't store the paths on the `ValuesFilePaths` on the package dependencies themselves anymore

Shortcut story: [sc-89421]

### DIFF
--- a/source/Calamari/Kubernetes/Commands/HelmUpgradeCommand.cs
+++ b/source/Calamari/Kubernetes/Commands/HelmUpgradeCommand.cs
@@ -30,32 +30,32 @@ namespace Calamari.Kubernetes.Commands
         PathToPackage pathToPackage;
         readonly ILog log;
         readonly IScriptEngine scriptEngine;
-        readonly IDeploymentJournalWriter deploymentJournalWriter;
         readonly IVariables variables;
         readonly ICalamariFileSystem fileSystem;
         readonly ISubstituteInFiles substituteInFiles;
         readonly IExtractPackage extractPackage;
+        readonly HelmTemplateValueSourcesParser templateValueSourcesParser;
         readonly ICommandLineRunner commandLineRunner;
 
         public HelmUpgradeCommand(
             ILog log,
             IScriptEngine scriptEngine,
-            IDeploymentJournalWriter deploymentJournalWriter,
             IVariables variables,
             ICommandLineRunner commandLineRunner,
             ICalamariFileSystem fileSystem,
             ISubstituteInFiles substituteInFiles,
-            IExtractPackage extractPackage
+            IExtractPackage extractPackage,
+            HelmTemplateValueSourcesParser templateValueSourcesParser
         )
         {
             Options.Add("package=", "Path to the NuGet package to install.", v => pathToPackage = new PathToPackage(Path.GetFullPath(v)));
             this.log = log;
             this.scriptEngine = scriptEngine;
-            this.deploymentJournalWriter = deploymentJournalWriter;
             this.variables = variables;
             this.fileSystem = fileSystem;
             this.substituteInFiles = substituteInFiles;
             this.extractPackage = extractPackage;
+            this.templateValueSourcesParser = templateValueSourcesParser;
             this.commandLineRunner = commandLineRunner;
         }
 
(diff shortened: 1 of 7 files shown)


source/Calamari/Kubernetes/Commands/HelmUpgradeCommand.cs
```diff
@@ -138,6 +140,13 @@ IEnumerable<string> ExplicitlySpecifiedValuesFiles()
                 }
             }
         }
+        
+        IList<string> TemplateValuesFiles(RunningDeployment deployment)
+        {
+            //we pass null here for the log as we don't want to log the values files being included twice. They will be properly logged during the deployment
+            return HelmTemplateValueSourcesParser.ParseTemplateValuesFilesFromDependencies(deployment, fileSystem, null)
+                                                 .ToList();
```
APErebus, line 148: As the comment says, if we include the log, it does the `Included files...` log lines twice. Suppressing those log messages here then let's them show up later.

source/Calamari/Kubernetes/Helm/HelmTemplateValueSourcesParser.cs
```diff
@@ -20,6 +20,33 @@ public static IEnumerable<string> ParseTemplateValuesSources(RunningDeployment d
...
+
+            var parsedJsonArray = JArray.Parse(templateValueSources);
+
+            //we are only interested in the values files in external dependencies (chart/package/git repo), so filter this array
+            var relevantTypes = parsedJsonArray.Where(t =>
+                                                      {
+                                                          var type = (TemplateValuesSourceType)Enum.Parse(typeof(TemplateValuesSourceType),t.Value<string>(nameof(TemplateValuesSource.Type)));
+                                                          return type == TemplateValuesSourceType.Chart || type == TemplateValuesSourceType.Package || type == TemplateValuesSourceType.GitRepository;
+                                                      })
+                                               .ToList();
```
APErebus, line 41: We take the same variable and just filter the types to those that we need to perform variable substitution on.

source/Calamari/Kubernetes/Commands/HelmUpgradeCommand.cs
zentron: This feels a little unexpected. Its not clear from either inside or outside the class that this is the intention of nulling a log object, which is usually a cross cutting concern.

Looking at the `HelmTemplateValueSourcesParser` class itself, I'm actually thinking it should perhapos be turned into a class, rather than a static method, passing the `ICalamariFileSystem fileSystem, ILog log` in the ctor. The result of this is that the actual method invocation can take parameters that describe the intention of the call. 

So it might take `Deployment deployment, bool logFoundFiles` or something to that effect. 

It might be a little overhead since its not mutating any state on the object, but it makes the intention a little clearer and less surprising.

zentron (approved):
Approved the general change, but the structure itself could take another look to avoid the null log.

source/Calamari/Kubernetes/Commands/HelmUpgradeCommand.cs
APErebus: @zentron Would you be comfortable if I merge this, then refactor it in a separate PR?
Just so we can get this fix moving through to customers

source/Calamari/Kubernetes/Helm/HelmTemplateValueSourcesParser.cs
```diff
@@ -9,9 +9,30 @@
...
     {
-        public static IEnumerable<string> ParseTemplateValuesSources(RunningDeployment deployment, ICalamariFileSystem fileSystem, ILog log)
+        readonly ICalamariFileSystem fileSystem;
+        readonly ILog log;
+
+        public HelmTemplateValueSourcesParser(ICalamariFileSystem fileSystem, ILog log)
+        {
+            this.fileSystem = fileSystem;
+            this.log = log;
+        }
```
APErebus, line 21: Changed this to a class based on feedback from @zentron

zentron (approved):

PR #1323, opened 2024-08-20
Add tests for import-certificate command

zentron:
Before I go ripping out the Windows Cert logic from Calamari, it's worthwhile adding some tests to cover its current usage.

As I extract the logic these tests should keep staying green. In future when we remove the netframework code entirely from this solution we can remove\readjust these tests.


MissedTheMark (approved):
LGTM

PR #1324, opened 2024-08-20
Remove `CalamariEnvironment.IsRunningOnMono`

zentron:
Previously the `FileBasedSemaphoreManager` was being used [only for mono](https://github.com/OctopusDeploy/Calamari/pull/1324/files#diff-d853f4ba525f6ca6062b6c207ccf42fe9861950279ced1c228412b99063e7b75L10). 
This PR represents the removal of this class, then pulling the thread on all the other now-unused classes that it made use of.

Also [got rid of](https://github.com/OctopusDeploy/Calamari/pull/1324/files#diff-de96705c9b8ccf56cf2a8ac70ba84c4dafd943fb35130a137c0030867140e4d8L13) `IsRUnningOnMono` since who would run Calamari on Mono? :)


source/Calamari.Tests/Fixtures/PackageDownload/PackageDownloadFixture.cs
```diff
@@ -99,9 +99,6 @@ public void ShouldDownloadPackage()
         [RequiresNonFreeBSDPlatform]
         public void ShouldDownloadMavenPackage()
         {
-            if (CalamariEnvironment.IsRunningOnMac && TestEnvironment.IsCI && !CalamariEnvironment.IsRunningOnMono)
```
zentron, line 102: These whole inconclusive tests seem a bit weird. 
We should be able to remove this condition entirely now, but if the tests play up ill take a look at readjusting.

eddymoulton: Seems like these haven't busted up any tests - nice removal!

eddymoulton (approved):
Looks good, I'm under the assumption that we don't need to validate the claim "since who would run Calamari on Mono?" any more than we already have.

I'll get that Azure test working again today so you can merge this in.

PR #1326, opened 2024-08-21
Remove all `#if NET40` compiler directives

zentron:


source/Calamari.Shared/Integration/Packages/Download/OciPackageDownloader.cs
```diff
@@ -373,20 +361,12 @@ AuthenticationHeaderValue CreateAuthenticationHeader(NetworkCredential credentia
 
         HttpResponseMessage SendRequest(HttpRequestMessage request)
         {
-#if NET40
-            return client.SendAsync(request).Result;
-#else
             return client.SendAsync(request).GetAwaiter().GetResult();
-#endif
```
MissedTheMark, line 380: Not saying that it should be in this PR, but it would be nice to get ride of the `.GetAwaiter().GetResult()` calls and make things async.

MissedTheMark (approved):
Sooooo much cleaner! Love it.  🚀

source/Calamari.Shared/Integration/Packages/Download/OciPackageDownloader.cs
zentron: Yeah I was tempted to keep pulling the string and seeing where it led, but then I wouldnt be able to create more PRs :)

PR #1328, opened 2024-08-23
Use the pathed package id instead of package reference name

tothegills:
### Before
image

### After
image

Relates to https://github.com/OctopusDeploy/Issues/issues/9011

### DIFF
--- a/source/Calamari/Kubernetes/Commands/HelmUpgradeCommand.cs
+++ b/source/Calamari/Kubernetes/Commands/HelmUpgradeCommand.cs
@@ -149,7 +149,7 @@ namespace Calamari.Kubernetes.Commands
             var packageRoot = packageReferenceName;
             if (string.IsNullOrEmpty(packageReferenceName))
             {
-                packageRoot = variables.Get(PackageVariables.IndexedPackageId(packageReferenceName ?? ""));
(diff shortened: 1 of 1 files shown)


source/Calamari/Kubernetes/Commands/HelmUpgradeCommand.cs
```diff
@@ -149,7 +149,7 @@ string PackageDirectory(string packageReferenceName)
             var packageRoot = packageReferenceName;
             if (string.IsNullOrEmpty(packageReferenceName))
             {
-                packageRoot = variables.Get(PackageVariables.IndexedPackageId(packageReferenceName ?? ""));
+                packageRoot = PackageName.ExtractPackageNameFromPathedPackageId(variables.Get(PackageVariables.IndexedPackageId(packageReferenceName ?? "")));
```
benPearce1, line 152: nit: empty strings are better represented by string.Empty, shows that is was intentionally and empty string

don't fix this now due because Calamari 👎

benPearce1 (approved):

source/Calamari/Kubernetes/Commands/HelmUpgradeCommand.cs
hnrkndrssn: legacy code :)

benPearce1: > legacy code :)
it is always legacy code

PR #1329, opened 2024-08-23
Use static EKS cluster for authentication tests

eddymoulton:
In order to reduce the time spent spinning up and down resources during tests we will be setting up static k8s clusters with supported cloud providers to run our authentication tests against. Because these tests will be able to run in parallel there will not be any cross interaction if multiple test runs are being completed at once.

How to review:
- Code quality
- Can you access the terraform cloud workspace?
- Have I configured the variables in a logical way?

Related: https://github.com/OctopusDeploy/Calamari/pull/1321


APErebus (approved):
:excited:

PR #1330, opened 2024-08-28
Report manifests applied via service messages

zentron:
In order to support live object status within Octopus server, applied resource manifests will get sent back via service messages as they are applied.

Manifests get converted from YAML to JSON to reduce the size of the documents being sent.


source/Calamari/Kubernetes/ManifestReporter.cs
```diff
@@ -0,0 +1,85 @@
...
+                    foreach (var document in yamlStream.Documents)
+                    {
+                        if (!(document.RootNode is YamlMappingNode rootNode))
+                        {
+                            continue;
+                        }
+
+                        var updatedDocument = YamlSerializer.Serialize(rootNode);
+                        
+                        var ns = GetNamespace(rootNode);
```
zentron, line 69: Getting the namespace on this side means that we dont need to load anything on the server side.
Unlike things like space\project etc, its not particularly sensitive since we would have otherwise trusted what was passed over in the manifest as the default anyway.

source/Calamari.Tests/KubernetesFixtures/ManifestReporterTests.cs
```diff
@@ -0,0 +1,113 @@
...
+            {
+                var mr = new ManifestReporter(variables, CalamariPhysicalFileSystem.GetPhysicalFileSystem(), memoryLog);
+
+                mr.ReportManifestApplied(filePath);
+
+                memoryLog.ServiceMessages.First().Properties.Should().Contain(new KeyValuePair<string, string>("ns", "default")); 
+            }
+        }
+
+        IDisposable CreteFile(string yaml, out string filePath)
```
eddymoulton, line 105: I thought this was a fancy library named after the island :greece:  before I saw it was a local method :joy:

eddymoulton (approved):

zentron:
Thanks @eddymoulton

PR #1337, opened 2024-09-09
Expose WindowsCertStore as an instance

zentron:
As part of the continued work to remove our full framework builds, we will be moving the exclusively full-framework functionality to a separate process. This is one of several PRs that will slowly move us towards this approach. It is hoped that by breaking the change up into self-contained changes will make it more reviewable.

To move towards this, the interaction with the WindowsCertStore should take place through an instance, and not a static utility class.

This PR should have no functional change beyond constructing an instance of the `WindowsX509CertificateStore`

### DIFF
--- a/source/Calamari.Shared/Integration/Certificates/NoOpWindowsX509CertificateStore.cs
+++ b/source/Calamari.Shared/Integration/Certificates/NoOpWindowsX509CertificateStore.cs
@@ -1,0 +1,46 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Security.Cryptography.X509Certificates;
+
+namespace Calamari.Integration.Certificates
+{
+    /// <summary>
+    /// Stand in replacement for IWindowsX509CertificateStore that will be registered for non Windows machines.
+    /// This should never end up being called. If it is, something has gone wrong somewhere else
+    /// </summary>
+    public class NoOpWindowsX509CertificateStore: IWindowsX509CertificateStore
+    {
+        public string? FindCertificateStore(string thumbprint, StoreLocation storeLocation)
+        {
+            throw new System.NotImplementedException();
+        }
+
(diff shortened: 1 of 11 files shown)


source/Calamari.Shared/Integration/Certificates/NoOpWindowsX509CertificateStore.cs
```diff
@@ -0,0 +1,46 @@
...
+using System.Collections.Generic;
+using System.Security.Cryptography.X509Certificates;
+
+namespace Calamari.Integration.Certificates
+{
+    /// <summary>
+    /// Stand in replacement for IWindowsX509CertificateStore that will be registered for non Windows machines.
+    /// This should never end up being called. If it is, something has gone wrong somewhere else
+    /// </summary>
+    public class NoOpWindowsX509CertificateStore: IWindowsX509CertificateStore
```
zentron, line 11: In future we wont want to rely on compiler directives to prevent parts of the code from not compiling. This class will be registered for non-windows calamari.

source/Calamari/Program.cs
```diff
@@ -81,6 +81,13 @@ protected override void ConfigureContainer(ContainerBuilder builder, CommonOptio
             builder.RegisterType<KubectlGet>().As<IKubectlGet>().SingleInstance();
             builder.RegisterType<HelmTemplateValueSourcesParser>().AsSelf().SingleInstance();
 
+            
+#if WINDOWS_CERTIFICATE_STORE_SUPPORT
```
zentron, line 85: Eventually we will get rid of this compiler directive entirely, until the cert logic is externalized, this allows the command classes above to take a dependency on `IWindowsX509CertificateStore`, even though they should never end up invoking it.

APErebus (approved):
LGTM 🚀

PR #1338, opened 2024-09-10
Modify `PrivateKeyAccessRule` to make serialization possible

zentron:
Another small step towards extracting full framework functionality of Calamari.

This change modifies the `PrivateKeyAccessRule` class to remove the `IdentityReference` property as this class will need to be serialized when sent between the separate processes.

Although the class already supports serializing the identity as a string, [this assumes](https://github.com/OctopusDeploy/Calamari/compare/robe/privatekey-identity?expand=1#diff-12a03db61327b49997534d32869cca1e42f0a505a4574572bc02950ada0a1f41L14) that it is in an `NTAccount` format and does not as easily support the `SecurityIdentifier` type which [can be passed](https://github.com/OctopusDeploy/Calamari/compare/robe/privatekey-identity?expand=1#diff-eb75b4cf1e476a44314561886eb2bb5bbb11c504ea7e72dffcde228d44a91e75R73) into this class constructor internally.

### DIFF
--- a/source/Calamari.Shared/Integration/Certificates/PrivateKeyAccessRule.cs
+++ b/source/Calamari.Shared/Integration/Certificates/PrivateKeyAccessRule.cs
@@ -1,7 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Security.AccessControl;
-using System.Security.Principal;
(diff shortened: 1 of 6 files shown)


APErebus (approved):
LGTM

PR #1343, opened 2024-09-12
Fail if customer is using Legacy IIS (IIS6)

zentron:
LegacyIIS support was added [many moon ago](https://github.com/OctopusDeploy/Calamari/pull/7) when IIS6 on Windows2003 was still supported however IIS7 on Windows 2008+ used different API.

As noted in the [AutoDetect code](https://github.com/OctopusDeploy/Calamari/blob/07a75ce2baa77832ac1046c3aaf31f9da8bd18da/source/Calamari/Integration/Iis/WebServerSupport.cs#L25) the auto selection of this process only apples for Windows 2003 and below. Windows 2003 is no longer supported and so no customers should be relying on this code path.

Since the pending removal of this code path has never been warned, a temporary last-ditch support will be temporarily be supplied via an additional force variable.

Given the lack of support for servers that should be running IIS6, It is expected that this code will be removed within a few months.

### DIFF
--- a/source/Calamari/Deployment/Conventions/LegacyIisWebSiteConvention.cs
+++ b/source/Calamari/Deployment/Conventions/LegacyIisWebSiteConvention.cs
@@ -36,7 +36,7 @@ namespace Calamari.Deployment.Conventions
 
             // In situations where the IIS version cannot be correctly determined automatically,
             // this variable can be set to force IIS6 compatibility.
-            var legacySupport = deployment.Variables.GetFlag(SpecialVariables.UseLegacyIisSupport);
+            var legacySupport = LegacySupport(deployment);
 
             var updated = iis.OverwriteHomeDirectory(iisSiteName, webRoot, legacySupport);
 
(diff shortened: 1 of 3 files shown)


source/Calamari/Deployment/Conventions/LegacyIisWebSiteConvention.cs
```diff
@@ -49,6 +49,32 @@ public void Install(RunningDeployment deployment)
...
+
+            // We were previously never warning of its imminent removal so
+            // provide a final legacy support capability with another big warning
+            // This will provide us some cover for later removal.
+            var forceLegacySupport = deployment.Variables.GetFlag(SpecialVariables.UseLegacyIisSupportForce);
+            if (!forceLegacySupport)
+                throw new CommandException($"Support for IIS6 is no longer supported.\r\n"
+                                           + $"Remove the {SpecialVariables.UseLegacyIisSupportForce} variable and ensure you are targeting IIS7+.\r\n"
+                                           + $"To provide a temporary work around set the `{SpecialVariables.UseLegacyIisSupportForce}` to `true`.\r\n"
+                                           + $"This capability will be very shortly removed without further warning.");
```
APErebus, line 68: This doesn't make sense to me. We are telling them to  remove the variable, but then tell them how to set the variable. Should we remove the 2nd line?

APErebus (approved):
Just 1 question about a message

source/Calamari/Deployment/Conventions/LegacyIisWebSiteConvention.cs
zentron: Yeah maybe bad wording. 🤔 ill give it a tweak.
Originally I just failed with no recourse, but decided to add the additional fallback. The phrasing probably got a bit muddled in that change.

PR #1344, opened 2024-09-12
Move `IWindowsX509CertificateStore.FindCertificateStore` out of netfx exclusive codepath

zentron:
As part of chipping away at netfx exclusive code, there are some paths that we can potentially move directly as-is into netcore compiled libraries.

So long as it is running on windows, netcore still supports calling into unmanaged code via `DllImport`

This change moves the one certificate management method that has a return type, which should at a minimum simplify the extraction of the remaining code by avoiding any need to deal with return types.

### DIFF
--- a/source/Calamari.Shared/Integration/Certificates/WindowsCertificateStoreLocator.cs
+++ b/source/Calamari.Shared/Integration/Certificates/WindowsCertificateStoreLocator.cs
@@ -1,0 +1,101 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Runtime.InteropServices;
+using System.Security.Cryptography.X509Certificates;
+
+namespace Calamari.Integration.Certificates
+{
+    public static class WindowsCertificateStoreLocator
+    {
+        static readonly IList<string> EnumeratedStoreNames = new List<string>();
+        static readonly object StoreNamesSyncLock = new object();
+        
+        public static string? FindCertificateStore(string thumbprint, StoreLocation storeLocation)
+        {
+            foreach (var storeName in GetStoreNames(storeLocation))
+            {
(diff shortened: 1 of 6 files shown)


source/Calamari.Shared/Integration/Certificates/WindowsCertificateStoreLocator.cs
```diff
@@ -0,0 +1,101 @@
...
+        {
+            CurrentUser = 1 << 16, // CERT_SYSTEM_STORE_CURRENT_USER
+            LocalMachine = 2 << 16, // CERT_SYSTEM_STORE_LOCAL_MACHINE
+            CurrentService = 4 << 16, // CERT_SYSTEM_STORE_CURRENT_SERVICE
+            Services = 5 << 16, // CERT_SYSTEM_STORE_SERVICES
+            Users = 6 << 16, // CERT_SYSTEM_STORE_USERS
+            UserGroupPolicy = 7 << 16, // CERT_SYSTEM_STORE_CURRENT_USER_GROUP_POLICY
+            MachineGroupPolicy = 8 << 16, // CERT_SYSTEM_STORE_LOCAL_MACHINE_GROUP_POLICY
+            LocalMachineEnterprise = 9 << 16, // CERT_SYSTEM_STORE_LOCAL_MACHINE_ENTERPRISE
+        }
```
APErebus, line 91: This looks like it's been copied from somewhere, but we haven't removed it in the PR?

zentron: :D I noticed that when I just looked at the PR before and wondered if you would pick it up before I got around to making an update.
Good catch

PR #1346, opened 2024-09-13
Remove NetFx Compiler Directives from tests

zentron:
With some recent (and long ago) changes, we can remove some of the compiler directives used in tests resulting in different tests between netfx and netcore builds.
Just another step toward confidence and consolidation.


APErebus (approved):
ever closer

PR #1347, opened 2024-09-20
Support authenticating using AWS Cli and kubectl exec via feature toggle

APErebus:
The change in #1205 caused an issue for at least one customer as the token generated by the AWS CLI has a fixed 14min expiry.

While for most customers, this isn't an issue, if you have a lone running k8s script step (such as a monitoring script after deployment), then the token might expire midway through that script and it can't be refreshed. This worked previously because kubectl was handling the expired token, then re-requesting a token when it expires. This isn't something that Calamari supports.

The solution is to add a feature toggle that reverts the authentication for the AWS CLI to use the kubectl exe, rather than generating a token and passing it back in.

Shortcut story: [sc-86744]

### DIFF
--- a/source/Calamari.Tests/KubernetesFixtures/Authentication/SetupKubectlAuthenticationAwsFixture.cs
+++ b/source/Calamari.Tests/KubernetesFixtures/Authentication/SetupKubectlAuthenticationAwsFixture.cs
@@ -8,6 +8,8 @@ using Calamari.Kubernetes.Authentication;
 using FluentAssertions;
 using NSubstitute;
 using NUnit.Framework;
+using Octopus.CoreUtilities;
+using Octopus.Versioning.Semver;
 
 namespace Calamari.Tests.KubernetesFixtures.Authentication
 {
@@ -77,6 +79,8 @@ namespace Calamari.Tests.KubernetesFixtures.Authentication
             AddLogForWhichAws();
             AddLogForAwsVersion(CurrentAwsVersion);
 
+            kubectl.GetVersion().Returns(Maybe<SemanticVersion>.Some(new SemanticVersion(1, 29, 7)));
+
             variables.SetStrings(KnownVariables.EnabledFeatureToggles,
                                  new[]
                                  {
@@ -89,7 +93,6 @@ namespace Calamari.Tests.KubernetesFixtures.Authentication
             expectedInvocations.AddRange(
                                          new List<(string, string)>
                                          {
-                                             GetEksClusterApiVersionInvocation,
                                              SetKubectlCredentialsViaExecInvocation,
                                              GetNamespaceInvocation
                                          });
@@ -284,9 +287,6 @@ namespace Calamari.Tests.KubernetesFixtures.Authentication
 
         static (string, string) SetKubectlTokenInvocation
             => ("kubectl", "config set-credentials octouser --token=k8s-aws-v1.token");
-
-        static (string, string) GetEksClusterApiVersionInvocation
-            => ("aws", $"eks get-token --cluster-name={EksClusterName} --region={AwsRegion}");
 
         static (string, string) SetKubectlCredentialsViaExecInvocation
             => ("kubectl", $"config set-credentials octouser --exec-command=aws --exec-arg=eks --exec-arg=get-token --exec-arg=--cluster-name={EksClusterName} --exec-arg=--region={AwsRegion} --exec-api-version=client.authentication.k8s.io/v1beta1");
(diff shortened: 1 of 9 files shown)


source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperFixture.cs
```diff
@@ -268,6 +288,37 @@ public async Task ExecutionWithEKS_AwsCLIAuthenticator()
...
+
+            //set the feature toggle
+            variables.SetStrings(KnownVariables.EnabledFeatureToggles,
+                                 new[]
+                                 {
+                                     FeatureToggle.KubernetesAuthAwsCliWithExecFeatureToggle.ToString()
+                                 },
+                                 ",");
+
+            // Overriding the cluster url with a valid url. This is required to hit the aws eks get-token endpoint.
```
MissedTheMark, line 285: Do you mean valid as in 'the format of this url is valid', or valid as in 'you can reach this'?

MissedTheMark (approved):
I'm not completely on top of all the AWS calls, but it looks fine from here, and the tests look good.
LGTM 👍

source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperFixture.cs
```diff
@@ -53,11 +53,31 @@ public void Setup()
...
+        [TestCase("",
+                     "",
+                     "",
+                     "KubernetesTentacle",
+                     true)]
+        [TestCase("",
+                     "",
+                     "",
+                     "",
+                     false)]
```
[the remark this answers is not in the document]
APErebus: Yeah, I fixed those. nfi what happened there

source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperFixture.cs
APErebus: I honestly don't know. I copied the test above this one and tweaked it. The comment came from there.

source/Calamari.Tests/KubernetesFixtures/Authentication/SetupKubectlAuthenticationAwsFixture.cs
```diff
@@ -37,13 +39,13 @@ public void Setup()
...
-                fileSystem,
-                environmentVars,
-                workingDirectory);
+                                           variables,
+                                           log,
+                                           commandLineRunner,
+                                           kubectl,
+                                           fileSystem,
+                                           environmentVars,
+                                           workingDirectory);
```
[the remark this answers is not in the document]
APErebus: I just have the default rider formatting set 🤷‍♂️

eddymoulton (approved):

source/Calamari/Kubernetes/Authentication/AwsCliAuth.cs
```diff
@@ -129,6 +149,34 @@ void SetKubeConfigAuthenticationToAwsCli(string user, string clusterName, string
             log.AddValueToRedact(token, "<token>");
             kubectl.ExecuteCommandAndAssertSuccess(arguments.ToArray());
         }
+        
+        void SetKubeConfigAuthenticationToAwsCliUsingExec(string user, string clusterName, string region)
+        {
+            var oidcJwt = deploymentVariables.Get(AccountVariables.Jwt);
+            
+            var apiVersion = GetKubeCtlAuthApiVersion();
```
APErebus, line 157: We changed to use the kubectl approach for getting the correct api version. This should really only be `v1beta1` as `v1alpha1` was deprecated in `1.24` of Kubernetes (which is not supported on AWS)

source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureEks.cs
```diff
@@ -305,6 +306,32 @@ public void AuthorisingWithAmazonAccount(bool runAsScript)
                 ExecuteCommandAndVerifyResult(TestableKubernetesDeploymentCommand.Name);
             }
         }
+        
+        [Test]
+        [TestCase(true)]
+        [TestCase(false)]
+        [WindowsTest] // We are having an issue with this test running on Linux. The test successfully executes on Windows.
```
APErebus, line 313: I was having massive issues getting these tests to pass on Debian on TeamCity. The tests pass fine correct, but the AWS CLI on Debian is having problems.

We also realized the other CLI tests in `KubernetesContextScriptWrapperFixture.cs` had the `[WindowsTest] attribute applied to them, with a comment saying they have issues on Linux.

After discussion with @eddymoulton, we decided that the code is working (as proved by the windows test), but to get unblocked, just not run this test on Linux

eddymoulton: > the code is working (as proved by the windows test)

Confidence was also increased by this same test working locally on my ubuntu based machine.

PR #1348, opened 2024-09-20
Upgrade Bouncy Castle to latest non Octopus version

zentron:
The change in our for https://github.com/bcgit/bc-csharp/compare/master...OctopusDeploy:bc-csharp:octopus
Looks to have long merged with the upstream repo (https://github.com/bcgit/bc-csharp/issues/74)

As such we should be able to retarget the source.

### DIFF
--- a/source/Calamari.Shared/Integration/Certificates/WindowsX509CertificateStore.cs
+++ b/source/Calamari.Shared/Integration/Certificates/WindowsX509CertificateStore.cs
@@ -516,11 +516,15 @@ namespace Calamari.Integration.Certificates
             }
         }
 
-        static IList<Org.BouncyCastle.X509.X509Certificate> GetCertificatesToImport(byte[] pfxBytes, string password)
+        static IList<Org.BouncyCastle.X509.X509Certificate> GetCertificatesToImport(byte[] pfxBytes, string? password)
         {
             using (var memoryStream = new MemoryStream(pfxBytes))
             {
-                var pkcs12Store = new Pkcs12Store(memoryStream, password?.ToCharArray() ?? "".ToCharArray());
+                // The latest version of BouncyCastle fails if the cert doesn't require a key, but we pass an empty array key.
+                // Will issue a PR to make this configurable at least in a way that doesn't require writing environment variables.
+                Environment.SetEnvironmentVariable("Org.BouncyCastle.Pkcs12.IgnoreUselessPassword", "true");
+                var pkcs12Store = new Pkcs12StoreBuilder().Build();
+                pkcs12Store.Load(memoryStream, password?.ToCharArray() ?? "".ToCharArray());
 
                 if (pkcs12Store.Count < 1)
                     throw new Exception("No certificates were found in PFX");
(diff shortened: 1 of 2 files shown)


source/Calamari.Shared/Integration/Certificates/WindowsX509CertificateStore.cs
```diff
@@ -516,11 +516,15 @@ static CryptoKeySecurity GetCspPrivateKeySecurity(SafeCertContextHandle certific
...
+        static IList<Org.BouncyCastle.X509.X509Certificate> GetCertificatesToImport(byte[] pfxBytes, string? password)
         {
             using (var memoryStream = new MemoryStream(pfxBytes))
             {
-                var pkcs12Store = new Pkcs12Store(memoryStream, password?.ToCharArray() ?? "".ToCharArray());
+                // The latest version of BouncyCastle fails if the cert doesn't require a key, but we pass an empty array key.
+                // Will issue a PR to make this configurable at least in a way that doesn't require writing environment variables.
+                Environment.SetEnvironmentVariable("Org.BouncyCastle.Pkcs12.IgnoreUselessPassword", "true");
+                var pkcs12Store = new Pkcs12StoreBuilder().Build();
+                pkcs12Store.Load(memoryStream, password?.ToCharArray() ?? "".ToCharArray());
```
zentron, line 527: Not entirely sure why they [used an environment variable](https://github.com/bcgit/bc-csharp/blob/master/crypto/src/pkcs/Pkcs12Store.cs#L219) rather than something with more control like an instance or static property.

In the older version of BouncyCastle this password check did not take place.

source/Calamari.Shared/Integration/Certificates/WindowsX509CertificateStore.cs
```diff
@@ -516,11 +516,15 @@ static CryptoKeySecurity GetCspPrivateKeySecurity(SafeCertContextHandle certific
...
 
-        static IList<Org.BouncyCastle.X509.X509Certificate> GetCertificatesToImport(byte[] pfxBytes, string password)
+        static IList<Org.BouncyCastle.X509.X509Certificate> GetCertificatesToImport(byte[] pfxBytes, string? password)
         {
             using (var memoryStream = new MemoryStream(pfxBytes))
             {
-                var pkcs12Store = new Pkcs12Store(memoryStream, password?.ToCharArray() ?? "".ToCharArray());
+                // The latest version of BouncyCastle fails if the cert doesn't require a key, but we pass an empty array key.
+                // Will issue a PR to make this configurable at least in a way that doesn't require writing environment variables.
+                Environment.SetEnvironmentVariable("Org.BouncyCastle.Pkcs12.IgnoreUselessPassword", "true");
```
APErebus, line 525:
```suggestion
                Environment.SetEnvironmentVariable(Org.BouncyCastle.Pkcs.Pkcs12Store.IgnoreUselessPasswordProperty, "true");
```
Can we use the public const prop rather that hardcode the string?

APErebus (approved):
1 minor thing, but LGTM

PR #1349, opened 2024-09-23
Log errors rather than writing to console

zentron:
A few places in the BashFixture we were logging to `Console.Error` rather than `ILog`. This means that during testing the output would not end up in our test output.

Also a change to ensure that we run _both_ assertions in the BashFixture to avoid the first one failing and resulting in missing context about the second assertion.


APErebus (approved):
👍

PR #1352, opened 2024-09-26
Remove as much of the WINDOWS_CERT compiler directives as possible

zentron:
## Background
This PR essentially moves away a bunch of `#if WINDOWS_CERTIFICATE_STORE_SUPPORT` compiler directives to pinpoint the code that is currently (as of this PR) unable to be compiled for .netcore.

The [follow up PR](https://github.com/OctopusDeploy/Calamari/pull/1353) takes the next leap and attempts to provide capabilities for netcore to involve the few places that are not supported as discovered by these changes.

## Whats Changed
What this turns out to be is essentially just the code that deals with assigning access to windows certificates. Although now then exposes in netcore classes that uses unmanaged code (windows dlls etc), these code paths should only ever exist on tasks running on windows machines (or they would have otherwise been failing in the past anyway, who's trying to run IIS steps on linux?)

## How to review
Hopefully just a sanity check, nothing functionally should really change for users.


source/Calamari.Common/CalamariFlavourProgram.cs
```diff
@@ -81,7 +81,7 @@ protected virtual int Run(string[] args)
             }
             catch (Exception ex)
             {
-                return ConsoleFormatter.PrintError(ConsoleLog.Instance, ex);
+                return ConsoleFormatter.PrintError(Log, ex);
```
zentron, line 84: This change allows us to capture the output properly when debugging tests

source/Calamari.Shared/Integration/Certificates/WindowsNative/CertificatePal.cs
```diff
@@ -115,6 +114,8 @@ public static byte[] GetCspPrivateKeySecurity(SafeCspHandle cspHandle)
             return buffer;
         }
 
+        // At the moment these models are not in the Calamari Solution. This is the next step.
+#if WINDOWS_CERTIFICATE_STORE_SUPPORT
         public static byte[] GetCngPrivateKeySecurity(SafeNCryptKeyHandle hObject)
```
zentron, line 119: The `SafeNCryptKeyHandle` structures are not currently defined in the Octopus solution like some of the others, and so this will need to be introduced as netcore does not have access to them from mscorlib like natframework does.

To keep the changes in multiple stages, this will be introduced in a followup PR

source/Calamari.Shared/Integration/Certificates/WindowsNative/WindowsX509Native.cs
```diff
@@ -69,14 +68,7 @@ internal static extern bool CryptAcquireCertificatePrivateKey(SafeCertContextHan
                                                                       [Out] out int dwKeySpec,
                                                                       [Out, MarshalAs(UnmanagedType.Bool)] out bool pfCallerFreeProvOrNCryptKey);
 
-        [DllImport("Crypt32.dll", SetLastError = true)]
```
zentron, line 72: Shuffled down to put between the other unmanaged invocations that are currently only available in netfx for this PR

source/Calamari.Shared/Integration/Certificates/WindowsX509CertificateStore.cs
```diff
@@ -18,14 +17,26 @@
...
         const string IntermediateAuthorityStoreName = "CA";
         public static readonly string RootAuthorityStoreName = "Root";
 
+        public WindowsX509CertificateStore(ILog log)
+        {
+            this.log = log;
+        }
+
+        // This should only be used in tests
+        public WindowsX509CertificateStore(): this(ConsoleLog.Instance)
```
zentron, line 35: We want to pass through the `ILog` to help with tests looking for specific output, but other logs dont care and so this default constructor makes it a little easier for those scenarios.

source/Calamari.Shared/Integration/Certificates/WindowsX509CertificateStore.cs
```diff
@@ -126,41 +137,12 @@ public void AddPrivateKeyAccessRules(string thumbprint, StoreLocation storeLocat
...
                     throw new Exception("Certificate does not have a private-key");
 
-                AddPrivateKeyAccessRules(privateKeyAccessRules, certificate);
+                CryptoKeySecurityAccessRules.AddPrivateKeyAccessRules(privateKeyAccessRules, certificate);
 
                 store.Close();
             }
         }
 
-        public CryptoKeySecurity GetPrivateKeySecurity(string thumbprint, StoreLocation storeLocation, string storeName)
```
zentron, line 135: Moved below to place together with other netfx only calls to be wrapped in the compiler directive.

source/Calamari.Shared/Integration/Certificates/WindowsX509CertificateStore.cs
```diff
@@ -406,7 +390,55 @@ static IList<SafeCertContextHandle> GetCertificatesFromPfx(byte[] pfxBytes, stri
             }
         }
 
-        static void AddPrivateKeyAccessRules(ICollection<PrivateKeyAccessRule> accessRules, SafeCertContextHandle certificate)
+        static IList<Org.BouncyCastle.X509.X509Certificate> GetCertificatesToImport(byte[] pfxBytes, string? password)
```
zentron, line 393: I think most of this diff is just GitHub getting confused about moving methods around. There shouldnt be any particular changes here.

source/Calamari.Tests/Fixtures/Certificates/ImportCertificateCommandFixture.cs
```diff
@@ -1,18 +1,19 @@
...
 using Calamari.Testing.Helpers;
+using Calamari.Testing.Requirements;
 using Calamari.Tests.Helpers;
 using Calamari.Tests.Helpers.Certificates;
 using NUnit.Framework;
 using Octostache;
 
 namespace Calamari.Tests.Fixtures.Certificates
 {
+    [Category(TestCategory.CompatibleOS.OnlyWindows)]
```
zentron, line 16: Previously this whole test wouldnt even exist in netcore, now netcore can run some of these tests so long as its on windows

source/Calamari.Tests/Fixtures/Certificates/ImportCertificateCommandFixture.cs
```diff
@@ -53,12 +54,15 @@ public void AddingCertToStore_AddsCert()
             cert.EnsureCertificateNotInStore(storeName, certificateStoreLocation);
         }
 
+#if WINDOWS_CERTIFICATE_STORE_SUPPORT
```
zentron, line 57: This test still relies on some of the capabilities not yet available to netcore (see above), it will be exposed again in a later PR

source/Calamari.Shared/Integration/Certificates/PrivateKeyAccessRuleExtensionMethods.cs
```diff
@@ -1,6 +1,4 @@
-﻿#if WINDOWS_CERTIFICATE_STORE_SUPPORT
```
zentron, line 1: A bunch of the file changes are just this, removing the `#if WINDOWS_CERTIFICATE_STORE_SUPPORT` compiler directive, meaning it compiles into netcore, even if it shouldnt actually be getting invoked by anything (since the `Calamari.netfx` flavour is used for anything on Windows that would have the IIS/WindowsCert variables provided)

source/Calamari.Shared/Integration/Certificates/CryptoKeySecurityAccessRules.cs
```diff
@@ -0,0 +1,166 @@
...
+using System.Security.AccessControl;
+using System.Security.Cryptography;
+using System.Security.Cryptography.X509Certificates;
+using static Calamari.Integration.Certificates.WindowsNative.WindowsX509Native;
+#endif
+
+namespace Calamari.Integration.Certificates
+{
+#if WINDOWS_CERTIFICATE_STORE_SUPPORT
+    public static class CryptoKeySecurityAccessRules
```
zentron, line 16: This class is just the extraction of a bunch of CryptoKey logic that was moved out of `WindowsX509CertificateStore` that cannot _currently_ compile in netcore. The functionality is essentially unchanged, with the addition of a compiler directive returning the "failing" implementation to be used by netcore (not actually used-used, but compiled for)

source/Calamari.Shared/Integration/Certificates/WindowsX509CertificateStore.cs
```diff
@@ -331,14 +314,14 @@ static void AddCertificateToStore(CertificateSystemStoreLocation storeLocation,
...
                         {
-                            Log.Info($"Certificate '{subjectName}' already exists in store '{storeName}'.");
+                            log.Info($"Certificate '{subjectName}' already exists in store '{storeName}'.");
                             return;
                         }
 
                         throw new CryptographicException(error);
                     }
 
-                    Log.Info($"Imported certificate '{subjectName}' into store '{storeName}'");
```
zentron, line 341: The use of these static `Log` methods made testing using the in-process Calamari fail (as opposed to many tests which actually invoke `Calamari.exe` and make debugging impossible)

source/Calamari.Shared/Integration/Certificates/WindowsX509CertificateStore.cs
```diff
@@ -406,116 +389,6 @@ static IList<SafeCertContextHandle> GetCertificatesFromPfx(byte[] pfxBytes, stri
             }
         }
 
-        static void AddPrivateKeyAccessRules(ICollection<PrivateKeyAccessRule> accessRules, SafeCertContextHandle certificate)
```
zentron, line 409: Moved out to `CryptoKeySecurityAccessRules`

source/Calamari.Tests/Fixtures/Certificates/ImportCertificateCommandFixture.cs
```diff
@@ -74,13 +78,14 @@ public void NoStoreLocationProvided_StoresInUserName()
...
-        
+#endif        
+
         CalamariResult Invoke(VariableDictionary variables)
         {
             using (var variablesFile = new TemporaryFile(Path.GetTempFileName()))
             {
                 variables.Save(variablesFile.FilePath);
-                return Invoke(Calamari()
+                return InvokeInProcess(Calamari()
```
zentron, line 88: This simply runs the test in the same process as the test harness, making debugging possible. 
I'd like to move all Calamari tests to this format now that we have a bit more of a handle on the different runtimes that we are actually running

source/Calamari.Tests/Fixtures/Certificates/WindowsX509CertificateStoreFixture.cs
```diff
@@ -221,12 +227,22 @@ public void ImportExistingCertificateShouldNotOverwriteExistingPrivateKeyRights(
                 Convert.FromBase64String(sampleCertificate.Base64Bytes()), sampleCertificate.Password,
                 storeLocation, storeName, sampleCertificate.HasPrivateKey);
 
-            var privateKeySecurity = new WindowsX509CertificateStore().GetPrivateKeySecurity(sampleCertificate.Thumbprint,
-                storeLocation, storeName);
+            var privateKeySecurity = CryptoKeySecurityAccessRules.GetPrivateKeySecurity(sampleCertificate.Thumbprint,
```
zentron, line 230: Now that the `CryptoKeySecurityAccessRule` method is extracted, we can actually consolidate some of the test assertions that check for certificate access rules (See Below)

APErebus (approved):
LGTM

PR #1353, opened 2024-09-26
Support using NetCore for all Windows Cert functionality

zentron:
## Background
This PR builds on top of https://github.com/OctopusDeploy/Calamari/pull/1352 to isolate the changes to focus on just enabling the capabilities required to expose the final windows certificate capabilities for netcore.

## What's changed
It turns out a lot of the structures defined in `System.Security.AccessControl.CryptoKeyAccessRule` were not ported to netcore. Since the existing code essentially calls down into unmanaged windows OS dlls,  functionally it shouldn't matter if the calls are being made from netcore or fullframework. 

This PR then introduces a bunch of these new classes (alongside pre-existing structures that have been used for some time for the same purpose). At the moment some `if #NETFX` compiler directives allow full framework to still make use of its existing class structure, however eventually it will all just be netcore and this will go away.

There are some tests that have had to be tweaked to account for now running in netcore and netfx but as before, functionally there should be little impact to end users.

## What's next
With full IIS and Windows Cert capability in theory available in netcore, future changes will take place in Octopus Server to start making use of the netcore version of Calamari rather than full framework. A feature flag will provide some safety to allow us to switch back if this becomes a problem.


source/Calamari.Shared/Integration/Certificates/CryptoKeySecurityAccessRules.cs
```diff
@@ -1,18 +1,260 @@
...
 using static Calamari.Integration.Certificates.WindowsNative.WindowsX509Native;
-#endif
+
 
 namespace Calamari.Integration.Certificates
 {
-#if WINDOWS_CERTIFICATE_STORE_SUPPORT
+    #if !NETFX
+       [Flags]
+    public enum CryptoKeyRights
```
zentron, line 19: These classes are all missing from the netcore libraries and so need to be C+P so that we can continue to use through our unmanaged code calls

APErebus: not gonna lie, took me longer than I care to get `C+P` 😆

APErebus (approved):
LGTM

PR #1354, opened 2024-10-01
Fix issue when Helm chart is extracted to root directory

APErebus:
@FinnianDempsey identified an issue with git sourced helm charts, when the helm chart is in the root of the repository. Due to how git source packaging extraction occurs, this results in the chart being extracted to the root staging directory.

This then triggered a bug when searching for the chart directory where it would incorrectly append `Chart.yaml` to the path when it found a `Chart.yaml` in the directory.

I have removed this and added a new test covering this scenario.

Fixes https://github.com/OctopusDeploy/Issues/issues/9037


NickJosevski (approved):
👍

PR #1355, opened 2024-10-02
Update Calamari.AzureWebApp to support .NET 6.0

APErebus:
Adds .NET 6.0 as a target framework for `Calamari.AzureWebApp`.

This removes the use of `Microsoft.Azure.Management.*` libraries (which are deprecated in favour of the `Azure.ResourceManager.*` libraries).

Also, we were using the `Microsoft.Web.Deployment` library which is the DLL underpinning the Web Deploy "framework" for performing web deployments. This DLL has not been updated for Net Core/.NET Standard, so can't be used by the .NET 6 version of the app.

To solve this, I wrote a new .NET Framework app, `Calamari.AzureAppService.NetCoreShim`, which wraps `Microsoft.Web.Deployment` library and only exposes cli params for the things we actually care about. This shim application is then bundled with Calamari (approx. 6mb)

I have also updated TeamCity to run the tests against .NET Core

NOTE: When merged, update the teamcity configuration to remove the conditional artifact dependency (`?:`) for the netcore tests build configuration

Shortcut story: [sc-85595]


source/Calamari.CloudAccounts/AzureAccountFactory.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Calamari.Common.Plumbing.Variables;
+
+namespace Calamari.CloudAccounts
+{
+    public static class AzureAccountFactory
+    {
+        public static IAzureAccount Create(IVariables variables)
+            => !string.IsNullOrEmpty(variables.Get(AzureAccountVariables.OpenIDJwt)) //if we have an OpenIdConnect JWT, return an AzureOidcAccount
+                ? (IAzureAccount)new AzureOidcAccount(variables)
+                : new AzureServicePrincipalAccount(variables);
```
APErebus, line 10: We were doing this check-JWT logic in a few places, so consolidated here under a single factory

source/Calamari.ConsolidateCalamariPackages/MigratedCalamariFlavours.cs
```diff
@@ -10,7 +10,6 @@ public class MigratedCalamariFlavours
 
         public static List<string> FullFrameworkOnlyFlavours = new()
         {
-            "Calamari.AzureWebApp",
             "Calamari.AzureCloudService",
```
APErebus, line 13: Only 1 to go... and by that, I suspect this will be a straight _delete_

source/Calamari.AzureWebApp/NetFxWebDeploymentExecutor.cs
```diff
@@ -0,0 +1,181 @@
...
+using Calamari.AzureWebApp.Integration.Websites.Publishing;
+using Calamari.AzureWebApp.Util;
+using Calamari.Common.Commands;
+using Calamari.Common.Plumbing.Logging;
+using Calamari.Common.Plumbing.Variables;
+using Microsoft.Web.Deployment;
+
+namespace Calamari.AzureWebApp
+{
+    public class NetFxWebDeploymentExecutor : IWebDeploymentExecutor
```
APErebus, line 15: This uses the `Microsoft.Web.Deployment` libraries, which there aren't NetCore versions of

source/Calamari.AzureWebApp/NetCoreWebDeploymentExecutor.cs
```diff
@@ -0,0 +1,297 @@
...
+using Calamari.AzureWebApp.Util;
+using Calamari.Common.Commands;
+using Calamari.Common.Features.Processes;
+using Calamari.Common.Plumbing.FileSystem;
+using Calamari.Common.Plumbing.Logging;
+using Calamari.Common.Plumbing.Variables;
+
+namespace Calamari.AzureWebApp
+{
+    public class NetCoreWebDeploymentExecutor : IWebDeploymentExecutor
```
APErebus, line 19: This executes via `msdeploy.exe`

PR #1357, opened 2024-10-08
Update Calamari.AzureServiceFabric to support .NET 6.0

APErebus:
Due to the requirement on the Service Fabric SDK, is set to only build the `windows` build of .NET 6.0

PR #1358, opened 2024-10-08
Update Calamari.AzureServiceFabric to support .NET 6.0

APErebus:
Supersedes #1357

Due to the requirement on the Service Fabric SDK, is set to only build the `windows` build of .NET 6.0

Updates the main `Microsoft.ServiceFabric` library to `10.1.2448`. I can't find any specific release notes, but I _think_ the versions match the versions of Service Fabric service.

Looking at the code, we only use the SDK above in the health check behaviour (to test connecting to the SF cluster)


zentron:
Nice. Have you been able to cobble together a test with a "real" service fabric app to verify. I'd imagine setting it up will probably end up taking more time than it took to make this change, but at least we can consider updating the test since the current SF Tests are pretty anemic. 
I think long term SF is in our sights for removal as its usage gets lower.

APErebus:
> Have you been able to cobble together a test with a "real" service fabric app to verify.

I haven't no 🤔 
I've never done anything with SF. Let me investigate

APErebus:
@zentron I have added automated tests for the two service fabric commands. They test the health check command and deploy a small application to a static, known cluster.

APErebus:
Just to prove the tests run locally
{0A5BBD24-70E9-48EE-834E-3BB461E67799}

source/Calamari.AzureServiceFabric/Calamari.AzureServiceFabric.csproj
```diff
@@ -5,12 +5,14 @@
...
     <OutputType>Exe</OutputType>
-    <TargetFramework>net462</TargetFramework>
+    <TargetFrameworks>net462;net6.0-windows</TargetFrameworks>
+    <RuntimeIdentifiers>win-x64</RuntimeIdentifiers>
+    <LangVersion>8.0</LangVersion>
   </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Identity.Client" Version="4.48.1" />
-    <PackageReference Include="Microsoft.ServiceFabric" Version="5.4.145" />
```
APErebus, line 13: It is very important that this version of the NuGet package exactly matches the version of the Runtime & SDK that you have installed locally. If they are different, you will likely receive COM errors.

Also, the runtime and SDK installed on TeamCity build agents also need to match this value.

zentron: So are we saying that customers will need specifically `10.1.2338` installed locally?
It might be worth adding a comment in here to make sure future us are aware that we dont break things.

zentron (approved):
Its great that you have managed to get some tests put in place. As painful as it is that should make things a little safer for the next person coming to do maintenance.

source/Calamari.AzureServiceFabric/Calamari.AzureServiceFabric.csproj
APErebus: @zentron I _think_ so. I think it's going to depend on what changes there are between the SDK & the runtime. The frustrating issue with this is that there is some magic COM stuff going on, so it _might_ work, but I guess if the runtime changes don't match, then you might get COM issues

source/Calamari.AzureServiceFabric/Calamari.AzureServiceFabric.csproj
```diff
@@ -5,12 +5,24 @@
...
+    
+    <!-- IMPORTANT NOTE
+    The version of this NuGet package is important. 
+    As the SDK makes native COM calls to the Service Fabric Runtime, if there are any divergence between this package version
+    and the versions of the installed Runtime & SDK, you may receive issues.
+    If this package version is updated, the TeamCity build steps will need updating to install the matching versions.
```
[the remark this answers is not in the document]
APErebus: @akirayamamoto ah good idea! I'll do that

PR #1359, opened 2024-10-10
Preserve URL-encoded slashes in filenames during path variable expansion for Unix-like systems

HuyPhanNguyen:
# Preserve URL-encoded slashes in filenames during path expansion on Linux

## Problem
On Linux systems, the current path expansion logic can incorrectly modify filenames containing URL-encoded forward slashes (%2F or %2f). This leads to issues with file paths that legitimately include these encodings, such as `~/Octopus/Files/Data%2FFileTransferService%2FFileTransferService@S11.0.zip`

[sc-94859](https://app.shortcut.com/octopusdeploy/story/94859/could-not-find-package-error-when-deploying-to-a-linux-webapp-using-a-package-from-an-external-generic-artifactory-feed)
[Fixes 9048](https://github.com/OctopusDeploy/Issues/issues/9048)

## Solution
Added a check in the `ExpandPathEnvironmentVariables` method specifically for Linux systems to detect the presence of "%2F" or "%2f" in the path. If found, the method now returns the original path without further modification, preserving the intended filename structure.

## Changes
- Added a condition to check for "%2F" or "%2f" in the path, only on Linux systems
- If these URL-encoded slashes are present on Linux, the method returns the original path
- This preserves the integrity of filenames that legitimately include these encodings on Linux deployments
- No changes to behavior on Windows or macOS systems

### DIFF
--- a/source/Calamari.Tests/Fixtures/Util/CrossPlatformFixture.cs
+++ b/source/Calamari.Tests/Fixtures/Util/CrossPlatformFixture.cs
@@ -44,8 +44,8 @@ namespace Calamari.Tests.Fixtures.Util
         [TestCase("$MARIO_BROTHERZZ/blah", "/blah", Description = "Variables terminate at last non alpha numeric character")]
         [TestCase("IMA$MARIO_BROTHER", "IMALUIGI", Description = "Variables begin from dollar character")]
         [TestCase("\\$MARIO_BROTHER/blah", "\\$MARIO_BROTHER/blah", Description = "Escaped dollar preserved")]
-        [TestCase("$MARIO_BROTHER/blah%2fblah.zip", "LUIGI/blah%2fblah.zip", Description = "URL-encoded forward slash (%2f) is preserved")]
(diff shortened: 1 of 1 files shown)


source/Calamari.Common/Plumbing/Extensions/CrossPlatformExtensions.cs
```diff
@@ -14,6 +14,20 @@ public static string ExpandPathEnvironmentVariables(string path)
 
                 path = Regex.Replace(path, @"(?<!\\)\$([a-zA-Z0-9_]+)", "%$1%");
                 path = Environment.ExpandEnvironmentVariables(path);
+                
+                // Special case: If the path contains "%2F", which represents an encoded forward slash,
+                // we assume it's part of a URL-encoded filename and not an unexpanded environment variable.
+                // In this case, we return the path as is to avoid incorrectly modifying the filename.
+                // Example: "~/Octopus/Files/Data%2FFileTransferService%2FFileTransferService@S11.0.zip"
+                // Here, %2F is part of the filename and should not be treated as an environment variable.
+                if (path.Contains("%2F") || path.Contains("%2f"))
```
hnrkndrssn, line 23:
```suggestion
                if (path.Contains("%2F", StringComparison.OrdinalIgnoreCase))
```

PR #1360, opened 2024-10-11
don't early exit if the git dependency name is empty

benPearce1:
[sc-94444]

Before:
image

After:
This is failing due to the missing chart.yaml file. The values file is noted as being found in the log: 

`Performing variable substitution on '/etc/octopus/Work/hRiQvkDl6EePTIZJRabcJg/staging/values.yaml'`

image

### DIFF
--- a/source/Calamari.Tests/Fixtures/Integration/Packages/DockerImagePackageDownloaderFixture.cs
+++ b/source/Calamari.Tests/Fixtures/Integration/Packages/DockerImagePackageDownloaderFixture.cs
@@ -114,7 +114,7 @@ namespace Calamari.Tests.Fixtures.Integration.Packages
             var exception = Assert.Throws<CommandException>(() => downloader.DownloadPackage("octopus-echo",
                 new SemanticVersion("1.1"), "docker-feed",
                 new Uri(AuthFeedUri),
-                FeedUsername, "SuperDooper",
+                "Nonexistantuser", "SuperDooper",
                 true, 
                 //we don't want to perform too many of these otherwise jfrog / artifactory gets sad at us
                 2,
--- a/source/Calamari/Kubernetes/Helm/GitRepositoryValuesFileWriter.cs
+++ b/source/Calamari/Kubernetes/Helm/GitRepositoryValuesFileWriter.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using Calamari.Common.Commands;
 using Calamari.Common.Plumbing.FileSystem;
 using Calamari.Common.Plumbing.Logging;
+using Calamari.Common.Plumbing.Variables;
 
 namespace Calamari.Kubernetes.Helm
 {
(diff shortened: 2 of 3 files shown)


source/Calamari/Kubernetes/Helm/GitRepositoryValuesFileWriter.cs
```diff
@@ -15,12 +16,12 @@ public static IEnumerable<string> FindChartValuesFiles(RunningDeployment deploym
                                                                ILog log,
                                                                string valuesFilePaths,
                                                                bool logIncludedFiles = true)
-            => FindGitDependencyValuesFiles(deployment,
```
IsaacCalligeros95, line 18: I'm confused on the naming here now. The HelmTemplateValueSourcesParser is checking if the script source is a GitRepository and if it is calling GitRepositoryValuesFileWriter.FindChartValuesFiles. Am I right in understanding this is a normal project with a git sourced helm chart and the values file is brought in as a part of the git repository? My understanding might be way off here, really all I'm trying to check is have we checked all combinations of CAC, git sourced, and normal projects with this change?

IsaacCalligeros95 (commented):
I think this will need to be backported to 2024.3

source/Calamari/Kubernetes/Helm/GitRepositoryValuesFileWriter.cs
benPearce1: I haven't changed to flow here, just removed the check for the git dependency name for the "value from chart" path

The specific scenario is a CAC project where the helm chart comes from the Project git repo and the values file is source from the chart.

I have also documented a workaround (confirmed as working by the customer) which may negate the need to backport.

source/Calamari.Tests/Fixtures/Integration/Packages/DockerImagePackageDownloaderFixture.cs
```diff
@@ -114,7 +114,7 @@ public void PackageWithWrongCredentials_FailsWithRetry()
             var exception = Assert.Throws<CommandException>(() => downloader.DownloadPackage("octopus-echo",
                 new SemanticVersion("1.1"), "docker-feed",
                 new Uri(AuthFeedUri),
-                FeedUsername, "SuperDooper",
+                "Nonexistantuser", "SuperDooper",
```
benPearce1, line 117: To fix an issue with a flakey test
https://octopusdeploy.slack.com/archives/C01HH8T16G3/p1728877229231419

IsaacCalligeros95 (approved):
LGTM

source/Calamari/Kubernetes/Helm/GitRepositoryValuesFileWriter.cs
benPearce1: The naming was confusing before anyway

PR #1361, opened 2024-10-14
Removed the azurecloudservice projects and update solution

rain-on:
As Azure Cloud Services has been retired by Microsoft, the ability to execute Cloud-Service steps is no longer required.

Server will be updated to fail steps which attempt to use this flavour of calarmari - at which point this version of calamari will be packaged into server (i.e. so server has 1 nice change which disables azure-cloud-services).


APErebus (approved):
Begone foul code

PR #1362, opened 2024-10-14
Execute app service tests against static app service plans

APErebus:
We have flaky issues in Calamari tests due to capacity issues creating app service plans.

To avoid this, we now have 3 static app service plans that all the test app services are created against. Each test now creates it's own app service against the static plan.

The plans are in the `calamari-testing-static-rg` resource group, but the per-test resources are created in their own resource groups, which are torn-down at the end of the test (not test fixture)

### DIFF
--- a/source/Calamari.AzureAppService.Tests/AppServiceIntegrationTest.cs
+++ b/source/Calamari.AzureAppService.Tests/AppServiceIntegrationTest.cs
@@ -30,22 +30,23 @@ namespace Calamari.AzureAppService.Tests
         protected string ClientSecret { get; private set; }
         protected string TenantId { get; private set; }
         protected string SubscriptionId { get; private set; }
-        protected string ResourceGroupName { get; private set; }
-        protected string ResourceGroupLocation { get; private set; }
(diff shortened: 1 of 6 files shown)


source/Calamari.AzureAppService.Tests/AppServiceIntegrationTest.cs
```diff
@@ -83,12 +79,27 @@ public async Task Setup()
...
+            
+            TestContext.WriteLine($"Creating resource group {ResourceGroupName}");
+            
+            var response = await subscriptionResource
                                  .GetResourceGroups()
                                  .CreateOrUpdateAsync(WaitUntil.Completed,
-                                                      ResourceGroupName,
-                                                      new ResourceGroupData(new AzureLocation(ResourceGroupLocation))
+                                                      name,
+                                                      new ResourceGroupData(AzureLocation.AustraliaEast)
```
APErebus, line 102: This needs to be `AustraliaEast` as the static resource groups are in `AustraliaEast`

PR #1365, opened 2024-10-21
Add new KubernetesObjectManifestInspectionFeatureToggle

APErebus:
Adds a new `KubernetesObjectManifestInspectionFeatureToggle` which enables the k8s manifests being sent back to Server. This just enables the same code path as the `KubernetesLiveObjectStatusFeatureToggle`

### DIFF
--- a/source/Calamari.Common/FeatureToggles/OctopusFeatureToggle.cs
+++ b/source/Calamari.Common/FeatureToggles/OctopusFeatureToggle.cs
@@ -4,8 +4,14 @@ namespace Calamari.Common.FeatureToggles
 {
     public static class OctopusFeatureToggles
     {
+        public static class KnownSlugs
+        {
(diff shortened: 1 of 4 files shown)


source/Calamari.Common/FeatureToggles/OctopusFeatureToggle.cs
```diff
@@ -4,8 +4,14 @@ namespace Calamari.Common.FeatureToggles
 {
     public static class OctopusFeatureToggles
     {
+        public static class KnownSlugs
+        {
+            public const string KubernetesObjectManifestInspection = "kubernetes-object-manifest-inspection";
```
APErebus, line 9: So it can be used in a test case

kevjt (approved):
LGTM

PR #1366, opened 2024-10-21
Increase default timeout for Azure resource group behaviour

IsaacCalligeros95:
ARM templates vary widely in how many resources they are spinning up and how long they will take. The old behaviour used an incremental backoff with a maximum polling wait time of 30 seconds and did not timeout. I've re-introduced the maximum 30 second poll interval and increased the overall timeout from 3 minutes to 30, this should be long enough to allow all ARM templates to apply successfully.

### DIFF
--- a/source/Calamari.AzureResourceGroup/DeployAzureResourceGroupBehaviour.cs
+++ b/source/Calamari.AzureResourceGroup/DeployAzureResourceGroupBehaviour.cs
@@ -25,7 +25,7 @@ namespace Calamari.AzureResourceGroup
 {
     class DeployAzureResourceGroupBehaviour : IDeployBehaviour
     {
-        static readonly TimeSpan PollingTimeout = TimeSpan.FromMinutes(3);
+        static readonly TimeSpan PollingTimeout = TimeSpan.FromMinutes(20);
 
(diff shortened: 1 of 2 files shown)


hnrkndrssn (approved):

source/Calamari.AzureResourceGroup/DeployAzureResourceGroupBehaviour.cs
```diff
@@ -128,14 +123,19 @@ async Task<ArmOperation<ArmDeploymentResource>> CreateDeployment(ArmClient armCl
             }
         }
 
-        async Task PollForCompletion(ArmOperation<ArmDeploymentResource> deploymentOperation)
+        async Task PollForCompletion(ArmOperation<ArmDeploymentResource> deploymentOperation, IVariables variables)
         {
+            var pollingTimeoutVariableValue = variables.Get(SpecialVariables.Action.Azure.ArmDeploymentTimeout, "30");
+            int.TryParse(pollingTimeoutVariableValue, out var pollingTimeoutValue);
```
hnrkndrssn, line 129: what if I configure my variable with a non-integer value? do we need an `if` here otherwise `pollingTimeoutValue` will be `null`?

hnrkndrssn (approved):
LGTM 👍

PR #1367, opened 2024-10-23
Send desired K8s manifests as YAML instead of JSON

flin-8:
[#project-md-k8s-live-object-status](https://octopusdeploy.slack.com/archives/C06HYUY4XLH)

We want to store the desired resource manifests as YAML for manifest inspection. Also, the JSON serialization had a bug where a number would get quoted when serialized.

### DIFF
--- a/source/Calamari.Tests/KubernetesFixtures/ManifestReporterTests.cs
+++ b/source/Calamari.Tests/KubernetesFixtures/ManifestReporterTests.cs
@@ -45,8 +45,10 @@ namespace Calamari.Tests.KubernetesFixtures
             //Test that quotes are preserved, especially for numbers
             var yaml = @"name: George Washington
 alphafield: ""fgdsfsd""
-age: 89
-height_in_inches: ""5.75""
+unquoted_int: 89
+quoted_int: ""89""
+unquoted_float: 5.75
+quoted_float: ""5.75""
 ".ReplaceLineEndings();
             
             using (CreateFile(yaml, out var filePath))
--- a/source/Calamari.Tests/KubernetesFixtures/ManifestReporterTests.cs
+++ b/source/Calamari.Tests/KubernetesFixtures/ManifestReporterTests.cs
@@ -41,7 +41,7 @@ namespace Calamari.Tests.KubernetesFixtures
             var variables = new CalamariVariables();
             variables.Set(KnownVariables.EnabledFeatureToggles, enabledFeatureToggle);
             
-            var yaml = "name: George Washington\nalphafield: \"fgdsfsd\"\nage: 89\nheight_in_inches: \"5.75\"\n";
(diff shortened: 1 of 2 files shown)


source/Calamari.Tests/KubernetesFixtures/ManifestReporterTests.cs
```diff
@@ -40,16 +41,21 @@ public void GivenValidYaml_ShouldPostSingleServiceMessage(string enabledFeatureT
...
             var variables = new CalamariVariables();
             variables.Set(KnownVariables.EnabledFeatureToggles, enabledFeatureToggle);
-
-            var yaml = @"foo: bar";
-            var expectedJson = "{\"foo\": \"bar\"}";
+            
+            //Test that quotes are preserved, especially for numbers
+            var yaml = @"name: George Washington
+alphafield: ""fgdsfsd""
+age: 89
```
eddymoulton, line 48: nit: George Washington is either 67 or 292 - depending on whether you consider death to stop aging :stuck_out_tongue:

source/Calamari.Tests/KubernetesFixtures/ManifestReporterTests.cs
```diff
@@ -40,16 +41,21 @@ public void GivenValidYaml_ShouldPostSingleServiceMessage(string enabledFeatureT
             var memoryLog = new InMemoryLog();
             var variables = new CalamariVariables();
             variables.Set(KnownVariables.EnabledFeatureToggles, enabledFeatureToggle);
-
-            var yaml = @"foo: bar";
-            var expectedJson = "{\"foo\": \"bar\"}";
+            
+            //Test that quotes are preserved, especially for numbers
```
eddymoulton, line 45: Should we have ints and floats both in and out of quotes here for extra security?

eddymoulton (approved):

source/Calamari.Tests/KubernetesFixtures/ManifestReporterTests.cs
flin-8: Good idea

source/Calamari/Kubernetes/ManifestReporter.cs
```diff
@@ -55,7 +51,7 @@ public void ReportManifestApplied(string filePath)
             if (!FeatureToggle.KubernetesLiveObjectStatusFeatureToggle.IsEnabled(variables) && !OctopusFeatureToggles.KubernetesObjectManifestInspectionFeatureToggle.IsEnabled(variables))
                 return;
 
-            using (var yamlFile = fileSystem.OpenFile(filePath, FileAccess.ReadWrite))
+            using (var yamlFile = fileSystem.OpenFile(filePath, FileAccess.Read, FileShare.Read))
```
flin-8, line 54: @eddymoulton I've changed this to only require read access - should be ok yeah?

eddymoulton: Correct, that's a good change

PR #1368, opened 2024-10-23
Report helm manifests for live status

eddymoulton:
Sends manifests for helm installations back to server _after_ the helm installation is completed.

This is the first pass at implementing helm support, but isn't yet ideal because we won't get any information back during deployments if `--wait` or `--atomic` flags are used. This is known and we will be coming back to solve for this.

I have added a `HelmCli` class to capture the logic used for helm (similar to what we have done with other CLI tools). For now this is purely for `helm get manifest` but there is no reason that it shouldn't be used for `helm upgrade` once we've proven it out. I opted not to do any actual refactoring here for the sake of time, but this is a step in the right direction imo. 
There are so many ways of running things in Calamari (eg. the existing helm convention uses the `ScriptEngine`, is there a reason for that?) so I'm hoping that creating a `HelmCli` is the right approach. - (28-10-24) Discussed this with Rob and we think it's a decent approach

### DIFF
--- a/source/Calamari/Kubernetes/Integration/CaptureCommandOutput.cs
+++ b/source/Calamari/Kubernetes/Integration/CaptureCommandOutput.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using Calamari.Common.Plumbing.Commands;
 
@@ -8,13 +9,17 @@ namespace Calamari.Kubernetes.Integration
     {
         Message[] Messages { get; }
         IEnumerable<string> InfoLogs { get; }
+        string MergeInfoLogs();
     }
+
     public class CaptureCommandOutput : ICommandInvocationOutputSink, ICommandOutput
     {
         private readonly List<Message> messages = new List<Message>();
         public Message[] Messages => messages.ToArray();
 
         public IEnumerable<string> InfoLogs => Messages.Where(m => m.Level == Level.Info).Select(m => m.Text).ToArray();
+
+        public string MergeInfoLogs() => string.Join(Environment.NewLine, InfoLogs);
 
         public void WriteInfo(string line)
         {
--- a/source/Calamari/Kubernetes/Integration/HelmCli.cs
+++ b/source/Calamari/Kubernetes/Integration/HelmCli.cs
@@ -1,0 +1,68 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Calamari.Common.Features.Processes;
+using Calamari.Common.Plumbing.Logging;
+using Calamari.Common.Plumbing.Variables;
+
+namespace Calamari.Kubernetes.Integration
+{
+    public class HelmCli : CommandLineTool
+    {
+        public HelmCli(ILog log, ICommandLineRunner commandLineRunner, string workingDirectory, Dictionary<string, string> environmentVars)
+            : base(log, commandLineRunner, workingDirectory, environmentVars)
+        {
+            ExecutableLocation = "helm";
+        }
+
+        readonly List<string> builtInArguments = new List<string>();
+
+        public HelmCli WithExecutable(string customExecutable)
+        {
+            ExecutableLocation = customExecutable;
+            return this;
+        }
+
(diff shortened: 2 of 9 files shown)


source/Calamari/Kubernetes/Integration/HelmCli.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        readonly List<string> builtInArguments = new List<string>();
+
+        public HelmCli WithExecutable(string customExecutable)
+        {
+            ExecutableLocation = customExecutable;
+            return this;
+        }
+
+        public HelmCli WithExecutable(IVariables variables)
```
eddymoulton, line 27: Same logic used in `HelmUpgradeConvention` to get the executable - now with tests!

The only thing that the existing convention does is a `chmod +x` in the script before running it - I'm going to do some testing to make sure that the `CommandLineRunner` doesn't need that to happen before I merge this.

eddymoulton: I've tested this on linux and it works as well as it does without - there are some permission issues with the existing step if you use helm from a package and it's got the wrong permissions.

I'll ask someone to test it out in windows.

source/Calamari/Kubernetes/Integration/CaptureCommandOutput.cs
```diff
@@ -8,14 +9,18 @@ public interface ICommandOutput
...
     }
+
     public class CaptureCommandOutput : ICommandInvocationOutputSink, ICommandOutput
     {
         private readonly List<Message> messages = new List<Message>();
         public Message[] Messages => messages.ToArray();
 
         public IEnumerable<string> InfoLogs => Messages.Where(m => m.Level == Level.Info).Select(m => m.Text).ToArray();
 
+        public string MergeInfoLogs() => string.Join(Environment.NewLine, InfoLogs);
```
flin-8, line 22: What are your thoughts on having the newline character depend on the OS where Calamari is running?

source/Calamari/Kubernetes/Integration/HelmCli.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            var helmExecutable = variables.Get(SpecialVariables.Helm.CustomHelmExecutable);
+            if (string.IsNullOrWhiteSpace(helmExecutable))
+            {
+                return this;
+            }
+
+            if (variables.GetIndexes(PackageVariables.PackageCollection)
```
flin-8, line 35: Any other variables we need to handle? Or is this all?

flin-8 (approved):
Generally looks good, just a few Q's

source/Calamari/Kubernetes/Integration/CaptureCommandOutput.cs
eddymoulton: I don't have a lot of them - there are existing examples of joining these log messages with `Environment.NewLine` so my assumption is that it works fine in all cases we're using it at the moment.

Sticking with Unix newlines might be a good idea to be consistent - perhaps there is a bug we just haven't seen come up yet?
By the same token, perhaps there is no bug because windows people tend to have everything on windows so we're just getting lucky...

I don't have any good answers aside from sticking with the status quo

source/Calamari/Kubernetes/Integration/HelmCli.cs
eddymoulton: Successfully tested on a windows machine with a windows worker: https://octopusdeploy.slack.com/archives/C0431AXP230/p1730072949591379

source/Calamari/Kubernetes/Integration/HelmCli.cs
eddymoulton: The variables being passed here are all the variables that were passed down in the deployment, which I believe should be the lot of them.

This is eq to what we are currently doing for the helm upgrade step

source/Calamari/Kubernetes/Integration/CaptureCommandOutput.cs
flin-8: cool, status quo is good

eddymoulton:
Replaced by https://github.com/OctopusDeploy/Calamari/pull/1374

PR #1372, opened 2024-10-31
Revert "Update Calamari.AzureWebApp to support .NET 6.0"

APErebus:
Reverts OctopusDeploy/Calamari#1355

There is a packing issue around Newtonsoft.Json and the AzureWebApp .NET Core project which causes deployment failures


IsaacCalligeros95 (approved):
LGTM

PR #1374, opened 2024-10-31
Enable Kubernetes Resource Status for Helm step

APErebus:
We are adding support for Kubernetes Resource Status (KOS) for Helm deployments.

To do this, we retrieve the deployed manifest from Helm using `helm get manifest` in a background thread that is running side-by-side of the main helm deployment thread.

As this new feature is behind a feature flag, I have created a new `HelmUpgradeWithKOSConvention` and use the conditional install convention to use the existing code if the feature toggle is not set.

The other major change is that the helm installation now uses the new `HelmCli`, rather than creating a small script and executing that. This is why there is more conventions needed to set up authentication


build/Build.cs
```diff
@@ -685,8 +685,8 @@ void DoPackage(string project, string framework, string version, string? runtime
         void SetOctopusServerCalamariVersion(string projectFile)
         {
             var text = File.ReadAllText(projectFile);
-            text = Regex.Replace(text, @"<PackageReference Include=""Calamari.Consolidated"" Version=""([\S])+"" />",
-                                 $"<PackageReference Include=\"Calamari.Consolidated\" Version=\"{NugetVersion.Value}\" />");
```
APErebus, line 689: Fixed build-local not updating server.csproj file due to changes in that file

source/Calamari/Kubernetes/Conventions/HelmUpgradeWithKOSConvention.cs
```diff
@@ -0,0 +1,554 @@
...
+        {
+            var isKOSEnabled = deployment.Variables.GetFlag(SpecialVariables.ResourceStatusCheck);
+
+            var releaseName = GetReleaseName(deployment.Variables);
+
+            var helmCli = new HelmCli(log, commandLineRunner, deployment);
+
+            kubectl.SetKubectl();
+
+            var currentRevisionNumber = helmCli.GetCurrentRevision(releaseName);
```
APErebus, line 65: We retrieve the current revision number, because we use the incremented value to retrieve the manifests

source/Calamari/Kubernetes/Conventions/HelmUpgradeWithKOSConvention.cs
```diff
@@ -0,0 +1,554 @@
...
+
+                    //if the resource doesn't have a namespace set in the manifest set it to the helm namespace.
+                    //This is because namespaced resources that don't have the namespace defined in the manifest will be set in the namespace set in the helm command
+                    //if this is null, we'll fall back on the namespace defined for the kubectl tool (which is the default target namespace)
+                    //we aren't changing the manifest here, just changing where the kubectl looks for our resource.
+                    //We also try and filter out known non-namespaced resources
+                    if (string.IsNullOrWhiteSpace(@namespace) && !KubernetesApiResources.NonNamespacedKinds.Contains(kind))
+                    {
+                        @namespace = deployment.Variables.Get(SpecialVariables.Helm.Namespace)?.Trim();
+                    }
```
APErebus, line 160: We have a fixed list of items that are known to not be namespaced, so we don't include the namespace if we _know_ they aren't namespaced.

For all other types, we just fallback to using the helm namespaced variable

source/Calamari.Tests/KubernetesFixtures/GoDurationParserFixture.cs
eddymoulton, line 1: Might be worth adding some `ParseDuration` tests in here too

source/Calamari/Kubernetes/ResourceStatus/ResourceUpdateReporter.cs
```diff
@@ -82,24 +83,24 @@ private void SendServiceMessage(Resource resource, bool removed, int checkCount)
             
             var parameters = new Dictionary<string, string>
```
eddymoulton, line 84: Some of these parameters are inferred from the deployment context now - which hasn't been updated here.

If you're up for it, should be a quick fix
https://github.com/OctopusDeploy/OctopusDeploy/blob/main/source/Octopus.Core/Features/Kubernetes/ServiceMessages/Observability/KubernetesDeploymentStatusServiceMessageMapper.cs#L15-L18

eddymoulton (approved):
Looking good!

source/Calamari/Kubernetes/ResourceStatus/ResourceUpdateReporter.cs
APErebus: Hrm, I might leave them as a separate change

source/Calamari.Tests/KubernetesFixtures/GoDurationParserFixture.cs
APErebus: Added in [b655396](https://github.com/OctopusDeploy/Calamari/pull/1374/commits/b6553963039488ecb672506fc178711fc34e72d2)

source/Calamari/Kubernetes/Conventions/HelmUpgradeWithKOSConvention.cs
```diff
@@ -0,0 +1,554 @@
...
+    {
+        readonly ILog log;
+        readonly ICommandLineRunner commandLineRunner;
+        readonly ICalamariFileSystem fileSystem;
+        readonly HelmTemplateValueSourcesParser valueSourcesParser;
+        readonly IResourceStatusReportExecutor statusReporter;
+        readonly IManifestReporter manifestReporter;
+        readonly Kubectl kubectl;
+
+        public HelmUpgradeWithKOSConvention(ILog log,
```
zentron, line 38: This has become a bit of a mamoth class. Could we break it up into orchestrating the two concerns, execution vs polling\parsing?
It doesnt _look_ like there should be shared state between them, but it was hard to tell :D

source/Calamari/Kubernetes/Conventions/HelmUpgradeWithKOSConvention.cs
zentron: This sounds related to the work @liam-mackie was doing around dealing with non-namespaced resources. 

It's not _likely_ that customers will run into it here, but it is possible I guess for them to be creating a non-namespaced resource that we don't know about.

I'm wondering if we should fallback to the similar logic we were _going_ to go with these resources and check with the known resource types for their namespaced-ness.

Not something I'd block the PR over, but id like to make sure we are happy with the potential gap

flin-8: What about custom resources that are cluster scoped (not namespaced)?

APErebus: because we use `kubectl`, it will actually ignore it if it knows it's not namespaced.

However, we actually ignore non-namespaced resources completely

This code above mainly just affects the UI and what resources we tell you we are tracking.

This is not a complete solution, but as I said, non-namespaced resources aren't supported anyway, so we need a better solution here.

source/Calamari/Kubernetes/ResourceStatus/RunningResourceStatusCheck.cs
```diff
@@ -60,15 +61,32 @@ public RunningResourceStatusCheck(
...
+            //when the passed cancellation token is cancelled, ask the task to stop
+            shutdownCancellationToken.Register(() =>
+                                       {
+                                           log.Verbose("Resource Status Check: Stopping after next status check.");
+                                           statusCheckTask.StopAfterNextResourceCheck();
+                                       });
+            
+            // we use CancellationToken.None as we don't want to use the passed cancellation token
+            // as it causes the status check task to abort early without retrieving the statuses.
+            await taskLock.WaitAsync(CancellationToken.None);
```
zentron, line 78: Why not use the 0-parameter overloaded method?

source/Calamari/Kubernetes/ResourceStatus/RunningResourceStatusCheck.cs
```diff
@@ -116,8 +134,9 @@ public async Task AddResources(ResourceIdentifier[] newResources)
         {
             taskCancellationTokenSource = new CancellationTokenSource();
             resources.UnionWith(newResources);
-            return await statusCheckTaskFactory()
-                .Run(resources, options, timeout, taskCancellationTokenSource.Token);
+
+            statusCheckTask = statusCheckTaskFactory();
+            return await statusCheckTask.Run(resources, options, timeout, log, taskCancellationTokenSource.Token);
```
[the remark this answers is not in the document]
zentron: Something is off here with the names 😵

source/Calamari/Kubernetes/ResourceStatus/ResourceStatusCheckTask.cs
```diff
@@ -72,13 +80,28 @@ public async Task<Result> Run(IEnumerable<ResourceIdentifier> resources, Options
...
                 return result;
             }, cancellationToken);
         }
 
+        /// <summary>
+        /// Stops the status check loop after the next resource status checks occur.
+        /// </summary> 
+        public void StopAfterNextResourceCheck()
+        {
+            stopAfterNextStatusCheck = true;
```
zentron, line 102: Is this overcomplicating? Why not just cancel the current status check in progress, then run "one more" before exiting?

zentron (changes requested):
Couple of questions for clarification around the approach.

source/Calamari/Kubernetes/ResourceStatus/ResourceStatusCheckTask.cs
APErebus: Because the current status check is a background loop, so I could cancel the current loop, but then re-running it would just trigger the loop again.

I'll take another look at it

source/Calamari/Kubernetes/ResourceStatus/RunningResourceStatusCheck.cs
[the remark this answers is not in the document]
APErebus: yeah, it's a bit ugly.. i'll see if I can tweak the names to make it clearer

source/Calamari/Kubernetes/ApiResourceIdentifier.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+
+namespace Calamari.Kubernetes
+{
+    public class ApiResourceIdentifier : IEquatable<ApiResourceIdentifier>
```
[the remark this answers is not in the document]
APErebus: yeah

source/Calamari/Kubernetes/Commands/HelmUpgradeCommand.cs
```diff
@@ -74,34 +85,62 @@ public override int Execute(string[] commandLineArguments)
...
-                // Any values files explicitly specified by the user will also have variable substitution applied
+                // Any values file explicitly specified by the user will also have variable substitution applied
                 new DelegateInstallConvention(d => substituteInFiles.Substitute(d.CurrentDirectory, ExplicitlySpecifiedValuesFiles().ToList(), true)),
-                new DelegateInstallConvention(d => substituteInFiles.Substitute(d.CurrentDirectory, TemplateValuesFiles(d)  , true)),
+                new DelegateInstallConvention(d => substituteInFiles.Substitute(d.CurrentDirectory, TemplateValuesFiles(d), true)),
                 new ConfiguredScriptConvention(new DeployConfiguredScriptBehaviour(log, fileSystem, scriptEngine, commandLineRunner)),
-                new HelmUpgradeConvention(log, scriptEngine, commandLineRunner, fileSystem, templateValueSourcesParser),
+
+                //if the feature toggle _is not_ enabled, use the old convention
+                new ConditionalInstallationConvention<HelmUpgradeConvention>(d => !OctopusFeatureToggles.KOSForHelmFeatureToggle.IsEnabled(d.Variables),
```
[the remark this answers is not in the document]
APErebus: ok, gotcha

source/Calamari/Kubernetes/Conventions/HelmUpgradeWithKOSConvention.cs
APErebus: I should be able to yeah

source/Calamari/Kubernetes/ResourceStatus/ResourceStatusCheckTask.cs
APErebus: I looked into this some more and I'm not sure I can easily make a change here without making more wholesale changes to the structure of those classes and the background loop.

source/Calamari/Kubernetes/Commands/HelmUpgradeCommand.cs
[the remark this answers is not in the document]
APErebus: Addressed in [cb8c691](https://github.com/OctopusDeploy/Calamari/pull/1374/commits/cb8c6911e21de93b83b7e105dc302fbf580ac21b)

source/Calamari/Kubernetes/Conventions/HelmUpgradeWithKOSConvention.cs
APErebus: Addressed in [2889b8a](https://github.com/OctopusDeploy/Calamari/pull/1374/commits/2889b8a9e0435b4187f109d158ebe50d7cc9b191) and [14025aa](https://github.com/OctopusDeploy/Calamari/pull/1374/commits/14025aae75ca84cd4884e4c7d7e85eff9ad99ea1)

source/Calamari/Kubernetes/ResourceStatus/RunningResourceStatusCheck.cs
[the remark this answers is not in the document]
APErebus: Addressed in [81c770e](https://github.com/OctopusDeploy/Calamari/pull/1374/commits/81c770e2fb45b2f18d8fc07be799e235e08ce1ec)

APErebus:
@zentron I think I've addressed all your feedback 🙇

source/Calamari/Kubernetes/ResourceStatus/ResourceStatusCheckTask.cs
zentron: 👍 all good, we will probably review once LOS is available

zentron (approved):

PR #1379, opened 2024-11-07
update `dotnet-script` to `1.5.0`

zentron:


IsaacCalligeros95 (approved):
Looks like dotnet-script 1.5.0 still runs on targets with just net6.0 installed. I used the Windows Worker tools 6.1.1 for this. That was my main concern. There is still the case where a script is written with some net6.0 functionality, the target has net8.0 installed and it unexpectedly changes the behaviour. It's an edge case but one worth considering.

PR #1382, opened 2024-11-11
Remove Helm V2

APErebus:
As of Octopus Server 20215.1, we no longer support Helm V2.

As a result, any attempt to install using the Helm V2 CLI will now throw an exception.

I chose not to bother with a test to validate this exception is thrown, because at a certain point, we kinda don't care if it behaves strangely for users. They've had enough time to migrate off Helm V2

Fixes https://github.com/OctopusDeploy/Issues/issues/9124


zentron (approved):
I agree that im not super happy about leaving the variable in the deployment process. We might come back to it but in the meantime this PR cleans up the code itself so we can treat the variable as ignored until\if we decide to remove it

PR #1385, opened 2024-11-12
Force Zip file extraction to use UTF-8 for netfx

rain-on:
If a Zip file does not explicitly indicate to use UTF-8 for decoding (via the EFS flag) - the file/directory entry will be extracted using the system-default encoding (which may not support multi-byte chars).

When using netcore - if no encoding is specified, it ALWAYS uses UTF-8 - as such, this issue only affects netfx users of calamari.


APErebus:
Are we looking to backport this? If so, probably easier to target the release branch first then forward merge to `main`

Also, what happens if the _zip_ filename has accent chars in it? Might be a good second test

rain-on:
I would like to backport.
I had assumed we'd update calamari as a standard commit - then apply the new calamari version to the old octopus - then forward merge?

source/Calamari.Tests/Fixtures/Integration/Packages/Samples/utf8.Filenámes-1.0.0.zip
kevjt, line 1: What's this file used for? I see the test only references the other one. Are you missing a test case maybe?

source/Calamari.Tests/Fixtures/Integration/Packages/PackageExtractorFixture.cs
```diff
@@ -47,6 +47,27 @@ public void ExtractPumpsFilesToFilesystem(Type extractorType, string extension,
...
+        
+        
+        [Test]
+        public void CanExtractZipFileContainingSpecialCharacters()
+        {
+            var fileName = GetFixtureResource("Samples", string.Format("utf8.Filenames-1.0.0.zip"));
+
+            Type extractorType = typeof(ZipPackageExtractor);
+            
+            var extractor = (IPackageExtractor) Activator.CreateInstance(extractorType, ConsoleLog.Instance);
```
kevjt, line 59: Is there any reason why the extractor can't be instantiated directly with `new`? That way you don't need to have the additional public constructor just for this test.

kevjt (commented):
Looks mostly good. Some questions

flin-8:
Changes look good to me, Trent will rebase off the Calamari branch for 2024.3

source/Calamari.Tests/Fixtures/Integration/Packages/Samples/utf8.Filenámes-1.0.0.zip
rain-on: missed a file - the same test now can unzip either file - just to confirm that utf-8 chars in the zip filename are handled.

source/Calamari.Tests/Fixtures/Integration/Packages/PackageExtractorFixture.cs
rain-on: way better solution, 🙏

kevjt (approved):
LGTM

rain-on:
Closed, and reimplemented on an old branch, which will be merged forward.

PR #1392, opened 2024-11-19
Bump Nuget dependencies

zentron:
`NuGet.Versioning` -> `5.11.6`
`NuGet.Commands` -> `5.11.5`
`System.Security.Cryptography.Cng` -> `5.0.0`

A breaking change to `NuGet.Commands` required some refactoring of one of the downloader classes

### DIFF
--- a/source/Calamari.Shared/Integration/Packages/NuGet/NuGetV3LibDownloader.cs
+++ b/source/Calamari.Shared/Integration/Packages/NuGet/NuGetV3LibDownloader.cs
@@ -8,10 +8,10 @@ using NuGet.Common;
 using NuGet.Protocol;
 using NuGet.Protocol.Core.Types;
 using System.IO;
-using Calamari.Common.Plumbing.Logging;
+using System.Threading.Tasks;
 using Calamari.Common.Plumbing.Extensions;
-using NuGet.DependencyResolver;
-using NuGet.LibraryModel;
+using NuGet.Commands;
+using NuGet.Packaging.Core;
 using Octopus.Versioning;
 
 namespace Calamari.Integration.Packages.NuGet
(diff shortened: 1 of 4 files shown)


source/Calamari.Shared/Integration/Packages/NuGet/NuGetV3LibDownloader.cs
```diff
@@ -25,31 +25,27 @@ public static void DownloadPackage(string packageId, IVersion version, Uri feedU
...
-                                                        4096,
-                                                        true))
+                var packageDownloader =  providers.GetPackageDownloaderAsync(new PackageIdentity(packageId, version.ToNuGetVersion()), sourceCacheContext, logger, CancellationToken.None)
+                                                      .GetAwaiter()
+                                                      .GetResult();
+                var fileCopied = packageDownloader.CopyNupkgFileToAsync(targetTempNupkg, CancellationToken.None).GetAwaiter().GetResult();
+                
+                if (!fileCopied) //I would expect any actual standard exception to be thrown above and not returned as a bool
                 {
-                    providers.CopyToAsync(libraryIdentity, nupkgStream, CancellationToken.None)
```
zentron, line 50: This `CopyToAsync` method appears to have been replaced by the `CopyNupkgFileToAsync` on the right

source/Calamari.Shared/Integration/Packages/NuGet/NuGetV3LibDownloader.cs
```diff
@@ -25,31 +25,27 @@ public static void DownloadPackage(string packageId, IVersion version, Uri feedU
             if (feedCredentials != null)
             {
                 var cred = feedCredentials.GetCredential(feedUri, "basic");
-                sourceRepository.PackageSource.Credentials = new PackageSourceCredential("octopus", cred.UserName, cred.Password, true);
+                sourceRepository.PackageSource.Credentials = new PackageSourceCredential("octopus", cred.UserName, cred.Password, true, null);
```
zentron, line 28: Based on comment in decompiled source for this class (cannot find link in GitHub, sorry), passing now treats all authentication types as valid, presumably treated the same as the missing value in the original signature to the right.

        /// <summary>
        /// Converts an authentication type filter string from the config file syntax to a list of valid authentication
        /// types
        /// </summary>
        /// <param name="str">
        /// Comma-delimited list of authentication types the credential is valid for as stored in the config file.
        /// If null or empty, all authentication types are valid. Example: 'basic,negotiate'
        /// </param>
        /// <returns>
        /// Enumeration of valid authentication types. If empty, all authentication types are valid.
        /// </returns>

PR #1394, opened 2024-11-27
Deprecate AzureRM - PowerShellContext

rain-on:
The AzureRM deprecation flag is required to be included in the AzurePowershellContext.ps1 to show a warning if users
are still using AzureRM module (Rather than Azure CLI).

This feature flag means users are presented with a warning atm - which will become an error at some point in the future.

### DIFF
--- a/source/Calamari/Kubernetes/Scripts/AzurePowershellContext.ps1
+++ b/source/Calamari/Kubernetes/Scripts/AzurePowershellContext.ps1
@@ -153,7 +153,14 @@ function ConnectAzAccount
         }
         elseif (Get-AzureRmModuleInstalled)
         {
-            Initialize-AzureRmContext
+            if($OctopusAzureRmIsDeprecated -like [Boolean]::TrueString) {
+                Write-Error "Azure Resource Manager modules are no longer available for authenticating with Azure, you are required to move to Azure CLI or the Az powershell modules."
+                exit 2
+            }
+            else {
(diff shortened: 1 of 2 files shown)


source/Calamari/Kubernetes/Scripts/AzurePowershellContext.ps1
```diff
@@ -11,6 +11,9 @@ $OctopusAzureEnvironment = $OctopusParameters["Octopus.Action.Azure.Environment"
 $OctopusOpenIdJwt = $OctopusParameters["Octopus.OpenIdConnect.Jwt"]
 $OctopusUseOidc = ![string]::IsNullOrEmpty($OctopusOpenIdJwt)
 
+## The script is passed the following parameters.
```
flin-8, line 14: Which version of Server are you targeting? I've made a breaking change in Calamari (https://github.com/OctopusDeploy/Calamari/pull/1405) so you'll need to target a 28.x.x branch if this is going to **2024.4**

rain-on: I think it'll need to be 2025.1 - so main is fine (i.e. I don't really want to push this out to on-prem).
Aside from that - the feature toggle should probably be flipped to prevent the usage of AzureRM by default (but with an escape hatch).

zentron (approved):

zentron:
@rain-on  it looks like this PR can now be closed as the code itself has since been removed?

PR #1395, opened 2024-11-27
Use GVK for Kubernetes Object Status

kevjt:
## Background
Kubernetes resource objects are distinguished by a combination of their API group, resource type (referred to as _kind_  throughout this PR), namespace (for namespaced resources), and name [[source]](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names). 

Although the resource API version is not needed to identify an individual object, the version does play a part in defining the schema of the resource kind. For example, `extensions/v1beta1` had a _Deployment_ before it was moved to `apps/v1` and it had slightly different semantics and supported options.

The `ResourceFactory` that is responsible for choosing the type of resource to parse into currently does not check the group or version of the API, it only checks the kind. As a result, this could lead to runtime parser exceptions in the scenario that one deploys a [Custom Resource](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/) that shares the same kind name as one which we currently support, as the custom resource could have a different schema.

[sc-98554]

## Result
- Introduced concept of GroupVersionKind (GVK)
- Changed `IResourceIdentity` to include the GVK
- Updated the KOS ResourceFactory to only parse the manifest into existing Resource types if the GVK is something we support
- ChildKind relationships have also been replaced with ChildGVK relationships
- Reports the resource _group_ to Server via a new service message property. This will be needed for [correlating](https://github.com/OctopusDeploy/OctopusDeploy/blob/main/source/Octopus.Core/Features/Kubernetes/GetKubernetesTaskResourceStatusRequestV1Handler.cs#L42) deployment resource manifests with KOS resources.
 
This change makes KOS more robust, and we can easily extend the supported list of GVKs when there are newer versions etc.


source/Calamari/Kubernetes/ResourceLoggingExtensionMethods.cs
```diff
@@ -11,7 +11,7 @@ public static void LogResources<T>(this ILog log, IEnumerable<T> resourceToLog)
             foreach (var resourceIdentifier in resourceToLog)
             {
                 var hasNamespace = !string.IsNullOrEmpty(resourceIdentifier.Namespace);
-                log.Verbose($" - {resourceIdentifier.Kind}/{resourceIdentifier.Name} {(hasNamespace ? $"in namespace {resourceIdentifier.Namespace}" : "")}");
+                log.Verbose($" - {resourceIdentifier.GroupVersionKind}/{resourceIdentifier.Name} {(hasNamespace ? $"in namespace {resourceIdentifier.Namespace}" : "")}");
```
APErebus, line 14: what does this do to the logs, do you get something like `apps/v1/Deployment/Blah`?

source/Calamari/Kubernetes/ResourceStatus/Resources/ResourceFactory.cs
```diff
@@ -6,6 +7,23 @@ namespace Calamari.Kubernetes.ResourceStatus.Resources
 {
     public static class ResourceFactory
     {
+        static Dictionary<ResourceGroupVersionKind, Func<JObject, Options, Resource>> resourceFactories = new Dictionary<ResourceGroupVersionKind, Func<JObject, Options, Resource>>
```
APErebus, line 10: nice

source/Calamari/Kubernetes/ResourceStatus/Resources/ResourceFactory.cs
```diff
@@ -16,38 +34,8 @@ public static IEnumerable<Resource> FromListJson(string json, Options options)
...
-                    return new ConfigMap(data, options);
-                case "Secret":
-                    return new Secret(data, options);
-                case "PersistentVolumeClaim":
-                    return new PersistentVolumeClaim(data, options);
-                default:
-                    return new Resource(data, options);
-            }
+            var gvk = data.ToResourceGroupVersionKind();
+            return resourceFactories.ContainsKey(gvk) ? resourceFactories[gvk](data, options) : new Resource(data, options);
```
APErebus, line 38:
```suggestion
            return resourceFactories.TryGetValue(gvk, out var factory) ? factory(data, options) : new Resource(data, options);
```
1 less hash check in the dictionary

source/Calamari/Kubernetes/ResourceStatus/Resources/ResourceGroupVersionKind.cs
```diff
@@ -0,0 +1,106 @@
...
+            }
+        }
+
+        public override string ToString()
+        {
+            return !Group.IsNullOrEmpty() ? $"{Group}/{Version}/{Kind}" : $"{Version}/{Kind}";
+        }
+    }
+
+    public static class ResourceGroupVersionKindExtensionMethods
```
APErebus, line 50: nit: probably can move to a separate file

APErebus (approved):
LGTM!

Couple of minor things, but not blocking. Great stuff 👏

source/Calamari/Kubernetes/ResourceLoggingExtensionMethods.cs
kevjt: Yep. The GroupVersionKind `ToString` is used automatically. It ends up with logs like in https://github.com/OctopusDeploy/Calamari/pull/1395/files#diff-0db08797bc823f1629cdd5af8e58f9cff19d259b717e9529f9e9c784e6dacff3

PR #1397, opened 2024-12-02
Report applied Kubernetes manifests for the Kustomize command

kevjt:
Adds manifest reporting capability (via service messages) to the Kustomize command.

The command for applying the Kustomization is broken into a two stage process:
1. `kubectl kustomize` to build the Kustomization into a hydrated manifest file which is written to the working directory.
2. `kubectl apply -f` to apply the hydrated manifest.

The manifest reporter is able to report using the hydrated manifest file.

This process is similar to what was previously done for the (deprecated) [kustomize step-package](https://github.com/OctopusDeploy/step-package-kustomize/blob/main/steps/kustomize-v2/src/executor.ts).


APErebus (approved):
very nice!

PR #1399, opened 2024-12-04
Authenticate gcloud with generic oidc jwt token

grace-rehn:
When running a gcloud script with an generic oidc account, authenticate using the account's generated jwt token.
[#project-generic-oidc](https://octopusdeploy.slack.com/archives/C081K9EPXT9)

[sc-98276]

generic account:

image

gcloud script:

`GCP is a project variable set to the generic account`

image
image

script run:

image

### DIFF
--- a/source/Calamari.GoogleCloudScripting/GoogleCloudContextScriptWrapper.cs
+++ b/source/Calamari.GoogleCloudScripting/GoogleCloudContextScriptWrapper.cs
@@ -38,9 +38,13 @@ namespace Calamari.GoogleCloudScripting
             Dictionary<string, string>? environmentVars)
         {
             var workingDirectory = Path.GetDirectoryName(script.File)!;
+            
+            // Create temporary files for JWT authentication
+            using var jwtFile = new TemporaryFile(Path.Combine(workingDirectory, "jwt_token.txt"));
+            using var jsonAuthFile = new TemporaryFile(Path.Combine(workingDirectory, "auth_config.json"));
 
             environmentVars ??= new Dictionary<string, string>();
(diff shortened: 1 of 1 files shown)


source/Calamari.GoogleCloudScripting/GoogleCloudContextScriptWrapper.cs
```diff
@@ -222,6 +205,129 @@ CommandResult ExecuteCommand(params string[] arguments)
                 return result;
             }
 
+            bool SetGcloudExecutable()
```
benPearce1, line 208: The convention for returning a boolean to indicate success would be `TrySetGcloudExecutable`

source/Calamari.GoogleCloudScripting/GoogleCloudContextScriptWrapper.cs
```diff
@@ -87,26 +96,10 @@ public CommandResult Execute()
...
-                {
-                    if (!File.Exists(gcloud))
-                    {
-                        log.Error($"The custom gcloud location of {gcloud} does not exist. Please make sure gcloud is installed in that location.");
-                        return errorResult;
-                    }
-                }
-                else
+                var isGcloudSet = SetGcloudExecutable();
+                if (!isGcloudSet)
```
benPearce1, line 100:
```suggestion
                if (!SetGcloudExecutable())
```

benPearce1 (approved):
Looks good. Love the refactoring to extract each of the auth methods to individual methods.
Small nitpick about the naming of the function. Also you can just nest the each of the calls in the `if` statement and avoid the assignment, but this is also just a preference, if you already have a green build, don't change anything.

PR #1401, opened 2024-12-09
Update KOS to watch for new secret and configmap manifest files

kevjt:
This change is needed for https://github.com/OctopusDeploy/OctopusDeploy/pull/29574 to ensure KOS continues to work.

It supports the changes to convert the Deploy Secret & Config map actions to use `kubectl apply -f` instead of `kubectl create --from-file`.

[sc-98555]


eddymoulton (approved):
Would you mind fixing the formatting?
Otherwise all good!

source/Calamari.Tests/KubernetesFixtures/ResourceStatus/assets/manifests/unknown.yml
kevjt, line 1: This is used to test the resource finder doesn't just find any files that aren't in its list of known filenames

PR #1402, opened 2024-12-10
Do not attempt to retrieve the applied Helm manifest when ResourceStatusCheck is disabled

kevjt:
Part of https://github.com/OctopusDeploy/Issues/issues/9160

The Helm upgrade command should not try to retrieve the applied manifest when the ResourceStatusCheck flag is not set (assuming none of the Manifest Inspection and Live Object Status feature toggles are enabled either).

This PR additionally includes more verbose logging around the Helm release name and revision number that Calamari tries to retrieve, to help with debugging cases where the `PollForManifest` method times out and fails.

[sc-99264]

### DIFF
--- a/source/Calamari/Kubernetes/Conventions/Helm/HelmManifestAndStatusReporter.cs
+++ b/source/Calamari/Kubernetes/Conventions/Helm/HelmManifestAndStatusReporter.cs
@@ -41,7 +41,9 @@ namespace Calamari.Kubernetes.Conventions.Helm
         {
             await Task.Run(async () =>
                            {
-                               if (deployment.Variables.GetFlag(SpecialVariables.ResourceStatusCheck)
+                               var resourceStatusCheckIsEnabled = deployment.Variables.GetFlag(SpecialVariables.ResourceStatusCheck);
+                               
+                               if (resourceStatusCheckIsEnabled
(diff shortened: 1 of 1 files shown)


source/Calamari/Kubernetes/Conventions/Helm/HelmManifestAndStatusReporter.cs
```diff
@@ -40,15 +41,20 @@ public async Task StartBackgroundMonitoringAndReporting(
         {
             await Task.Run(async () =>
                            {
-                               var manifest = await PollForManifest(deployment, helmCli, releaseName, revisionNumber);
+                               if (deployment.Variables.GetFlag(SpecialVariables.ResourceStatusCheck)
```
APErebus, line 44: nit: can we get this into a variable and then use it, rather than parsing/searching variables twice?

APErebus (approved):
1 minor nit, but thanks!

source/Calamari/Kubernetes/Conventions/Helm/HelmManifestAndStatusReporter.cs
kevjt: yea I should have done that 😅

PR #1403, opened 2024-12-11
Index.json in Consolidated zip stores full path, not just hash

rain-on:
At the moment the index.json in consolidated calamari contains a list of hashes (aka directories) which must be consolidated in order for a given flavour of calamari to execute.

Thus when octopus is trying to construct a calamari to run, it copies ALL files from each of the listed directories into a single zip (with pathing) - and sends that over to the target.

Up until now, that works correctly.

As part of the migration to netcore - AzureWebApp required a "shim app" to wrap the capabilities in Microsoft.Deployments.Web (which does not have a netcore counterpart) - this shim-app is a netfx app - and uses newtonsoft.

When a consolidated calamari is created - the AzureWebApp now list in its hashes:
* Directories containing required files for the netcore capabilities
* Directories containing required files for the shim-app.

Unfortunately, this means that Newtonsoft is pulled in 3 times:
* 👍 Once, in root directory for netcore
* 👍 Once, in shim-app subdirectory for netfx
* 👎 Once again in root directory, due to nature of "copy everything in hash-dir"/.

Thus - the netcore version of newtonsoft in the root-directory is overwritten with the netfx version - meaning Calamari.AzureWebApp fails to start when it is reconstructed from a consolidated calamari.

================

The solution, at this stage, is to replace the hashes, with a full-blown file-reference (See example below).

Only real concern - depending on the platform creating the index.json - you may get different path separators :/

<code>
"Packages": {
    "Calamari": {
      "PackageId": "Calamari",
      "Version": "2018.4.3",
      "IsNupkg": true,
      "PlatformHashes": {
        "netfx": [
          "00e0765eb7aceea4cb85cfeeafe702d0\\Azure.Identity.dll",
          "00e0765eb7aceea4cb85cfeeafe702d0\\Cloud\\Azure.Identity.dll",
          "01656ee57f64634592a6f674815ad8b6\\Cloud\\Microsoft.Azure.Management.Network.Fluent.dll",
          "01656ee57f64634592a6f674815ad8b6\\Microsoft.Azure.Management.Network.Fluent.dll",
          "05d1b950c470ea8b0aa357f9a59cf264\\Cloud\\System.Resources.Writer.dll",
          "05d1b950c470ea8b0aa357f9a59cf264\\System.Resources.Writer.dll",
          "06d000552ed6785988ae188fc35d1b86\\Cloud\\System.Security.Cryptography.X509Certificates.dll",
          "06d000552ed6785988ae188fc35d1b86\\System.Security.Cryptography.X509Certificates.dll",
          "081d9558bbb7adce142da153b2d5577a\\Cloud\\Newtonsoft.Json.dll",
          "081d9558bbb7adce142da153b2d5577a\\Newtonsoft.Json.dll",

</code>


source/Calamari.ConsolidateCalamariPackages/ConsolidatedPackageCreator.cs
```diff
@@ -56,7 +56,7 @@ Dictionary<string, string[]> GroupByPlatform(IEnumerable<SourceFile> filesForPac
                     .GroupBy(f => f.Platform)
                     .ToDictionary(
                         g => g.Key,
-                        g => g.Select(f => f.Hash).OrderBy(h => h).ToArray()
+                        g => g.Select(f => Path.Combine(f.Hash, f.FullNameInDestinationArchive.Replace('/', Path.DirectorySeparatorChar))).OrderBy(h => h).ToArray()
```
APErebus, line 59: We should hardcode to use `/` as it's supported on windows, but `\` is not supported on unix-based systems

IsaacCalligeros95:
We're looking to include dotnet-script in the NetFramework builds of Calamari. I missed the use case of pre/post deploy scripts on steps like deploy a package step. I've included this change on a [spike pr](https://github.com/OctopusDeploy/Calamari/pull/1407), at this stage I'm assuming it'll help the [failures we saw](https://build.octopushq.com/buildConfiguration/TeamModernDeployments_Calamari_ChainFull/15362219?showLog=15362161_12921_1603.11272&logFilter=debug&logView=flowAware) if it does is there any concerns with us backporting the change? (I'll obviously wait for this to be approved and merged first)

APErebus:
> We're looking to include dotnet-script in the NetFramework builds of Calamari. I missed the use case of pre/post deploy scripts on steps like deploy a package step. I've included this change on a [spike pr](https://github.com/OctopusDeploy/Calamari/pull/1407), at this stage I'm assuming it'll help the [failures we saw](https://build.octopushq.com/buildConfiguration/TeamModernDeployments_Calamari_ChainFull/15362219?showLog=15362161_12921_1603.11272&logFilter=debug&logView=flowAware) if it does is there any concerns with us backporting the change? (I'll obviously wait for this to be approved and merged first)

What version do you need to backport this into?

APErebus (commented):
This looks good to me, will wait for more info from @IsaacCalligeros95 on if we need to backport this to an LTS calamari

IsaacCalligeros95:
> What version do you need to backport this into?

The customer issue I'm looking at is on Server 2024.3, which is on the `release/27.3.5` stream of Calamari. Server 2024.4 is on `28.2.2`. If it's a pain to backport these changes + the corresponding server updates I'm happy to backport it with the other changes I'm making.

rain-on:
It's sounding like this should actually be put onto the 27.3.5 branch - then merged forward?

APErebus:
> It's sounding like this should actually be put onto the 27.3.5 branch - then merged forward?

That would probably be easiest yep 👍

rain-on:
Closed in favour of #1416

PR #1406, opened 2024-12-12
Enable dotnetcore for framework

bec-callow-oct:
Avoid throwing an exception when attempting to run .NET scripts in the .NET Framework build of Calamari. This is for Pre-Deploy scripts on Deploy a Package as this step does not support .NET Core

[SC-99263]

### DIFF
--- a/source/Calamari.Common/Features/Scripting/DotnetScript/DotnetScriptBootstrapper.cs
+++ b/source/Calamari.Common/Features/Scripting/DotnetScript/DotnetScriptBootstrapper.cs
@@ -84,9 +84,6 @@ namespace Calamari.Common.Features.Scripting.DotnetScript
 
         public static string FindBundledExecutable()
         {
-            if (ScriptingEnvironment.IsNetFramework())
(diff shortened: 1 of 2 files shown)


IsaacCalligeros95 (approved):
LGTM

IsaacCalligeros95:
There's a lot of moving parts on this one. I've had a chat with Trent and we'll need to wait for this change to [fix the Calamari Consolidation process](https://github.com/OctopusDeploy/Calamari/pull/1403) when bundling a .Net6 package in a netframework build. Without this change we can overwrite dlls and break a lot of other things. This also requires Server changes to handle the reassembly of the calamari bundle which will have to be backported to Server. 

The alternative to this is that the deploy a package step as of 2024.4 supports running against the netcore versions of Calamari and netframework will be deprecated entirely. This is behind a feature toggle, but if the customer was willing to wait for 2024.4 they'll be able to fix this issue by just enabling that feature toggle.

benPearce1 (approved):

PR #1408, opened 2024-12-12
Add EKS Pod Identity Support

liam-mackie:
This PR enables EKS Pod Identity Support to Calamari. This is done by leveraging the `GenericContainerCredentials` type in the AWS SDK, which does all the work for us.

This has been smoke tested using the `S3 Upload to S3 step` and the `AWS CLI Step` on the Kubernetes Worker, set to use the pod role assigned. This worked perfectly:
CleanShot 2024-12-13 at 16 42 34@2x

Because of the way that the EKS pod identity manager binds to a virtual IP, it's not feasible for us to use `kubectl port-forward` to test this, and there aren't any alternatives for good tests yet. As such, the tests currently use WireMock to mock out a fake EKS Pod Identity service, so that we can test that our code returns the correct authentication details when the response on the wire is well-known. 

**Note:** This will _not_ fix the error when trying to download packages from the S3 feed with machine credentials

### DIFF
--- a/source/Calamari.Aws/Calamari.Aws.csproj
+++ b/source/Calamari.Aws/Calamari.Aws.csproj
@@ -24,10 +24,10 @@
   <ItemGroup>
     <PackageReference Include="AWSSDK.EKS" Version="3.7.13.24" />
     <PackageReference Include="AWSSDK.CloudFormation" Version="3.7.9.14" />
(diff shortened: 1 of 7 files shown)


APErebus (approved):
Yeah awesome! 🥳

PR #1409, opened 2024-12-12
Allow new Helm upgrade command to be executed via script

kevjt:
Part of https://github.com/OctopusDeploy/Issues/issues/9168

The new Helm Upgrade command (introduced as part of the KOS for Helm feature) changed the way Helm was being executed. It was previously being executed via a Bash (or Powershell) script, whereas now the Helm command is being invoked directly with the arguments.

Customers who were using Bash-specific formatting in their arguments, e.g. quotes around literals, have experienced errors in their Helm step as these arguments are no longer being processed by Bash.

This PR introduces a feature-flagged workaround that enables affected customers to continue deploying using the new Helm command by reverting to the script-based deployment method.

[sc-99264]

### DIFF
--- a/source/Calamari.Tests/KubernetesFixtures/Integration/HelmCliTests.cs
+++ b/source/Calamari.Tests/KubernetesFixtures/Integration/HelmCliTests.cs
@@ -3,12 +3,14 @@ using System.Collections.Generic;
 using System.IO;
 using Calamari.Common.Commands;
 using Calamari.Common.Features.Processes;
+using Calamari.Common.FeatureToggles;
 using Calamari.Common.Plumbing.FileSystem;
 using Calamari.Common.Plumbing.Variables;
 using Calamari.Kubernetes;
 using Calamari.Kubernetes.Integration;
 using Calamari.Testing.Helpers;
 using Calamari.Testing.Requirements;
+using Calamari.Tests.Fixtures.Integration.FileSystem;
 using FluentAssertions;
 using FluentAssertions.Execution;
 using NSubstitute;
@@ -19,6 +21,9 @@ namespace Calamari.Tests.KubernetesFixtures.Integration
     [TestFixture]
     public class HelmCliTests
     {
+        const string BashScriptFilename = "Calamari.HelmUpgrade.sh";
+        readonly CalamariVariables executeViaScriptFeatureToggle = new CalamariVariables { [KnownVariables.EnabledFeatureToggles] = OctopusFeatureToggles.KnownSlugs.ExecuteHelmUpgradeCommandViaShellScript };
+        
         [Test]
         public void ExecutesWithDefaultExecutable()
         {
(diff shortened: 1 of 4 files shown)


source/Calamari.Tests/KubernetesFixtures/Integration/HelmCliTests.cs
```diff
@@ -77,7 +82,6 @@ public void UsesCustomHelmExecutableFromPackage()
 
             var (helm, commandLineRunner, workingDirectory, _) = GetHelmCli(expectedExecutable, new CalamariVariables
             {
-                [SpecialVariables.Helm.Packages.CustomHelmExePackageKey] = expectedPackageKey,
```
kevjt, line 80: I couldn't find where this variable was being used. The current tests pass even without this here.

source/Calamari/Kubernetes/Integration/HelmCli.cs
```diff
@@ -130,5 +146,60 @@ void ChmodExecutable()
...
+            var command = string.Join(" ", cmd);
+            log.Verbose(command);
+            return command;
+        }
+        
+        void SetExecutable(List<string> commandBuilder, ScriptSyntax syntax)
+        {
+            if (isCustomExecutable)
+            {
+                commandBuilder.Add(syntax == ScriptSyntax.PowerShell ? ". " : $"chmod +x \"{ExecutableLocation}\";");
```
kevjt, line 181: A lot of this is copied over from https://github.com/OctopusDeploy/Calamari/blob/main/source/Calamari/Kubernetes/Conventions/HelmUpgradeConvention.cs

APErebus (approved):
Yep, look good to me. @zentron might have an opinion on the warning message

source/Calamari/Kubernetes/Integration/HelmCli.cs
```diff
@@ -102,6 +112,12 @@ public CommandResult Upgrade(string releaseName, string packagePath, IEnumerable
             buildArgs.Add(releaseName);
             buildArgs.Add(packagePath);
 
+            if (OctopusFeatureToggles.ExecuteHelmUpgradeCommandViaShellScriptFeatureToggle.IsEnabled(variables))
+            {
+                log.Warn("Backwards compatibility for Helm command execution via bash scripts is temporary and will be removed in a future release. Provide arguments compatible with direct Helm invocation, avoiding bash-specific formatting.");
```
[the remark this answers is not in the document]
zentron: Usually I like to give some indication of when a feature might be removed. Since I'd argue this isn't a standard deprecation, but a bugfix that the customer was just relying on, we probably don't need a full 12 month notice.

Lets aim for `2025.3` (6 months) before we strip it out. We can always take longer if we need, but this gives us a target timeframe that, if we hear no (good) objections we can actually take the next step.

When we have a public issue that links to this change its probably worth including a reference to this as well.

zentron (approved):
Minor comment around the warning message, otherwise 👍 
Thanks for going over this one again.

PR #1410, opened 2024-12-13
Backports 28.2.2 into release/28.2.0

APErebus:
28.2.2 was used in Octopus Server 2024.4 in https://github.com/OctopusDeploy/OctopusDeploy/pull/29025

Ideally, the fix that was needed would have been made on the `release/28.2.0` branch and then forward merged to `main`. That was missed, so we are backporting the relevant commits.

We need to hotfix it so updating this branch to the actual version in use


rain-on (approved):
LGTM - the branch is identical to 28.2.2

PR #1411, opened 2024-12-13
Add feature toggle to disable FSharp script execution

APErebus:
In discussion with @zentron, we missed adding the "soft" disable of FSharp scripts in 2024.3, so adding this here via a feature toggle `DisableFSharpScriptExecutionFeatureToggle`

Just throws a `CommandExecption` when feature toggle is enabled


zentron (approved):
Quick and simple. If we have a link somewhere i'd add that as well. But otherwise :shipit:

APErebus:
> Quick and simple. If we have a link somewhere i'd add that as well. But otherwise :shipit:

Just a link to the deprecation notice?

PR #1412, opened 2024-12-13
Fix account priority in AWS script step

octopus-hideaki:
The AWS Script step ignores OIDC accounts when 2 accounts for a variable exist.

The script always chooses to authenticate using access keys even if the OIDC account is higher in priority according to the scoping rule. 

This PR will make Calamari select the authentication method based on the account type, which is selected by scoping priority rather than the presence of access key variables.

https://github.com/OctopusDeploy/Issues/issues/9120
[[sc-97348](https://app.shortcut.com/octopusdeploy/story/97348/sev-3-oidc-accounts-being-ignored-in-deployments-when-having-2-aws-accounts-in-a-variable-one-aws-account-unscoped-and-one)]

### DIFF
--- a/source/Calamari.CloudAccounts/AwsEnvironmentGeneration.cs
+++ b/source/Calamari.CloudAccounts/AwsEnvironmentGeneration.cs
@@ -173,14 +173,19 @@ namespace Calamari.CloudAccounts
         {
             switch (accountType)
             {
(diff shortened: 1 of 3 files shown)


mik-ky:
We should add the "fixes ... [issue]" when we are updating the calamari dependency in Server. In Calamari, just mention the issue :)

mik-ky (approved):

PR #1414, opened 2024-12-16
Add retry to Semaphore acquisition

APErebus:
We have had instances of failures to acquire a semaphore when running Windows tests.

We _believe_ this could be due to multiple Calamari instances trying to create/acquire the same global, named Semaphore.

To try and alleviate this, I've added a simple retry of 100ms (plus a bit of jitter) for up to 5 attempts. I also changed the logic so that if we acquire the semaphore, but fail to update it's access control, we don't fail out (as the next step is to acquire the semaphore again anyway), rather just to log and continue onwards.

I've also added the same SystemSemaphore tests for the *nix platforms as it uses a `Mutex`, not a `Semaphore`.


source/Calamari.Common/Features/Processes/Semaphores/SystemSemaphoreManager.cs
```diff
@@ -90,24 +95,32 @@ IDisposable AcquireMutex(string name, string waitMessage)
...
+
+            try
+            {
+                semaphore.SetAccessControl(semaphoreSecurity);
+            }
+            catch (Exception e)
+            {
+                log.Verbose($"Failed to set access controls on semaphore '{name}': {e.PrettyPrint()}");
+                //at this point we have a valid semaphore, which is what we'd end up doing anyway
+            }
```
APErebus, line 122: If we had created the semaphore on line 112 and this failed, then exception handler would have attempted to create/acquire the exact same semaphore again, just without this access control call.
I figured if this SetAccessControl failed, then we still have a valid semaphore, or at least the exact same semaphore we'd get in the exception handler

source/Calamari.Tests/Fixtures/Integration/Process/Semaphores/NixSystemSemaphoreFixture.cs
```diff
@@ -0,0 +1,11 @@
+using Calamari.Testing.Helpers;
+using NUnit.Framework;
+
+namespace Calamari.Tests.Fixtures.Integration.Process.Semaphores
+{
+    [TestFixture]
+    [Category(TestCategory.CompatibleOS.OnlyNixOrMac)]
+    public class NixSystemSemaphoreFixture : SemaphoreFixtureBase
```
APErebus, line 8: Let's test the mutex as well

source/Calamari.Common/Features/Processes/Semaphores/SystemSemaphoreManager.cs
```diff
@@ -5,24 +5,37 @@
...
         {
             log = ConsoleLog.Instance;
             initialWaitBeforeShowingLogMessage = (int)TimeSpan.FromSeconds(3).TotalMilliseconds;
-        }
 
-        public SystemSemaphoreManager(ILog log, TimeSpan initialWaitBeforeShowingLogMessage)
-        {
-            this.log = log;
-            this.initialWaitBeforeShowingLogMessage = (int)initialWaitBeforeShowingLogMessage.TotalMilliseconds;
+            semaphoreAcquisitionPipeline = new ResiliencePipelineBuilder()
```
APErebus, line 24: Uses Polly's retry logic to do it for us

source/Calamari.Common/Features/Processes/Semaphores/ISemaphore.cs
APErebus, line 1: Unused

source/Calamari.Common/Features/Processes/Semaphores/SystemSemaphoreManager.cs
flin-8: I wonder if there's any side effect from `SetAccessControl` failing...

Otherwise you might as well do this operation in 2 stages:
1. Create the Semaphore in `AcquireSemaphore()` in a retry loop (does this ever fail?)
2. Try to set the access control in a retry loop

APErebus: > Otherwise you might as well do this operation in 2 stages:

Yeah, good point.

> Create the Semaphore in AcquireSemaphore() in a retry loop (does this ever fail?)

This is the bit that is throwing the exception, not the `SetAccessControl`

source/Calamari.Common/Features/Processes/Semaphores/SystemSemaphoreManager.cs
flin-8: I'm not sure what the consequences of not setting access control is, we'll either end up:

- All retries fail, there's no access control, we wait 500ms <-- **how likely is this to occur?**
- Succeeds on 1st/2nd try with or without access control, we wait 100ms

Are we better off with a shorter overall delay by performing less retries?

APErebus: Honestly I don't know how likely it is because I've not been able to reliably get it to fail.

I don't mind changing the delay to 50ms or drop the retry count to 3? If you considering a 500ms delay to be too long. My feeling in the context of Calamari starting up, that 500ms isn't the worst if it eventually fails anyway

source/Calamari.Common/Features/Processes/Semaphores/SystemSemaphoreManager.cs
APErebus: @flin-8 I looked into this, `SetAccessControl` calls through to underlying Win32 APIs to set ACL's on the semaphore

It uses the `NativeObjectSecurity.Persist` method.

I think, considering that all the errors we see are around creating/acquiring the semaphore, that we can probably do the create/acquire in a retry loop, then just handle the access control separately

APErebus: Changed in [045e62b](https://github.com/OctopusDeploy/Calamari/pull/1414/commits/045e62bcbfee1af5fa152a58fb517869f1c7484f)

source/Calamari.Common/Features/Processes/Semaphores/SystemSemaphoreManager.cs
APErebus: Changed to 50ms delay in [74552d6](https://github.com/OctopusDeploy/Calamari/pull/1414/commits/74552d682b016308928da243ef38be7db3c15c42)

source/Calamari.Common/Features/Processes/Semaphores/SystemSemaphoreManager.cs
```diff
@@ -90,24 +94,33 @@ IDisposable AcquireMutex(string name, string waitMessage)
...
         {
             var semaphoreSecurity = new SemaphoreSecurity();
             var everyone = new SecurityIdentifier(WellKnownSidType.WorldSid, null);
             var rule = new SemaphoreAccessRule(everyone, SemaphoreRights.FullControl, AccessControlType.Allow);
 
             semaphoreSecurity.AddAccessRule(rule);
 
-            bool createdNew;
+            //we try and create/acquire a global semaphore with some retry
+            var semaphore = semaphoreAcquisitionPipeline.Execute(() => new Semaphore(1,1, name));
```
flin-8, line 112: You can move this up to the caller of this method

APErebus: done in [7a92457](https://github.com/OctopusDeploy/Calamari/pull/1414/commits/7a92457fab930a967c63fd150d1d617121dc10f1)

flin-8 (approved):
LGTM

APErebus:
I created a Server PR that contains this PR build and ran a full chain. It all went green, so comfortable merging this.
https://github.com/OctopusDeploy/OctopusDeploy/pull/29843

PR #1415, opened 2024-12-17
Use ECR feed for OCI and docker

eddymoulton:
ECR supports more than just container (docker) images as it's a fully featured OCI registry.

To allow support for this we can query the manifest before we download/pull the image to switch between our pre-existing support for OCI feeds and docker feeds.

This approach means that users must have some pre-existing knowledge about the package when selecting from an ECR feed as it could be a helm chart, container image or any other [OCI artifact type](https://github.com/opencontainers/artifacts/blob/main/artifact-authors.md#defining-oci-artifact-types).
It is worth nothing that currently we do nothing to filter out non-docker images from the search and any helm charts in ECR would appear as options when deploying a container - but would simply fail to deploy.

[[sc-75291]]


source/Calamari.Shared/Integration/Packages/Download/Oci/OciArtifactManifestRetriever.cs
```diff
@@ -0,0 +1,50 @@
...
+                var isHelmChart =
+                    jsonManifest.HasConfigMediaTypeContaining(OciConstants.Manifest.Config.OciImageMediaTypeValue)
+                    || jsonManifest.HasLayersMediaTypeContaining(OciConstants.Manifest.Layers.HelmChartMediaTypeValue);
+
+                if (isHelmChart)
+                    return OciArtifactTypes.HelmChart;
+
+                var isDockerImage = jsonManifest.HasMediaTypeContaining(OciConstants.Manifest.DockerImageMediaTypeValue)
+                                    || jsonManifest.HasConfigMediaTypeContaining(OciConstants.Manifest.Config.DockerImageMediaTypeValue)
+                                    || jsonManifest.HasLayersMediaTypeContaining(OciConstants.Manifest.Layers.DockerImageMediaTypeValue);
```
eddymoulton, line 39: Note for myself before merging, I need to confirm these are all correct and expected.

PR #1417, opened 2024-12-18
Don't run Helm manifest retrieval for older versions of Helm

APErebus:
Helm versions less than 3.13.0 did not have the `get metadata` command, this would result in an error, this giving a current revision of `1`.

This causes an issue in two ways:

1. If revision `1` exists, then we are returning an old manifest
2. Revision `1` might have been removed due to the `--history-max` `upgrade` flag defaulting to `10`, so it would fail to find a revision.

This change now checks the version of Helm and verifies it's greater than `3.13`. If the version of helm is < `3.13`, we warn the user and _disable_ the manifest retrieval. This will mean KOS and LOS won't run, but I think that's acceptable as we aren't using KOS data to determine success or failure of the step.

There is also a second big fix which is that a helm manifest may include empty documents due to helm templating generating an empty file. In these cases, they have no root node and we _were_ warning them. An empty document receives a `null` `rootNode.Tag`, so now, in the cases where the rootNode _has_ a tag, then we log a verbose message.

### DIFF
--- a/source/Calamari.Tests/KubernetesFixtures/Helm3UpgradeFixture.cs
+++ b/source/Calamari.Tests/KubernetesFixtures/Helm3UpgradeFixture.cs
@@ -102,7 +102,7 @@ namespace Calamari.Tests.KubernetesFixtures
 
                 Assert.AreEqual(ReleaseName.ToLower(), result.CapturedOutput.OutputVariables["ReleaseName"]);
 
-                result.AssertOutputMatches("Helm manifest retrieval requires Helm v3.13 or newer.");
+                result.AssertOutputMatches("Octopus needs Helm v3.13 or later to display object status and manifests.");
 
                 result.CapturedOutput.ServiceMessages
                       .Where(sm => sm.Name == SpecialVariables.ServiceMessages.ResourceStatus.Name)
--- a/source/Calamari.Tests/KubernetesFixtures/Helm3UpgradeFixture.cs
+++ b/source/Calamari.Tests/KubernetesFixtures/Helm3UpgradeFixture.cs
@@ -1,13 +1,10 @@
-﻿using System.Collections.Generic;
-using System.Linq;
+﻿using System.Linq;
 using System.Threading.Tasks;
 using Calamari.Common.FeatureToggles;
 using Calamari.Common.Plumbing.Variables;
 using Calamari.Kubernetes;
 using Calamari.Testing.Helpers;
 using Calamari.Testing.Requirements;
-using Calamari.Tests.Fixtures;
-using Calamari.Tests.Helpers;
 using FluentAssertions;
 using NUnit.Framework;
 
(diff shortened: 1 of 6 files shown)


source/Calamari.Tests/KubernetesFixtures/Helm3UpgradeFixture.cs
```diff
@@ -79,12 +76,39 @@ public void ReportsObjectStatus()
...
-                                 sm.Properties[SpecialVariables.ServiceMessages.ResourceStatus.Attributes.Status] == Kubernetes.ResourceStatus.Resources.ResourceStatus.Successful.ToString());
+                  .Contain(sm => sm.Properties[SpecialVariables.ServiceMessages.ResourceStatus.Attributes.Name] == $"mychart-configmap-{ReleaseName}" && sm.Properties[SpecialVariables.ServiceMessages.ResourceStatus.Attributes.Namespace] == Namespace && sm.Properties[SpecialVariables.ServiceMessages.ResourceStatus.Attributes.Group] == "" && sm.Properties[SpecialVariables.ServiceMessages.ResourceStatus.Attributes.Version] == "v1" && sm.Properties[SpecialVariables.ServiceMessages.ResourceStatus.Attributes.Kind] == "ConfigMap" && sm.Properties[SpecialVariables.ServiceMessages.ResourceStatus.Attributes.Status] == Kubernetes.ResourceStatus.Resources.ResourceStatus.Successful.ToString());
+        }
+
+        [Test]
+        [RequiresNonFreeBSDPlatform]
+        [RequiresNon32BitWindows]
+        [RequiresNonMac]
+        [Category(TestCategory.PlatformAgnostic)]
+        public async Task HelmVersionOlderThanMinimumVersion_DoesNotRunObjectStatus()
```
kevjt, line 87: nice

kevjt (approved):
LGTM

source/Calamari/Kubernetes/Conventions/Helm/HelmManifestAndStatusReporter.cs
```diff
@@ -42,12 +44,19 @@ public async Task StartBackgroundMonitoringAndReporting(
...
+                                   resourceStatusCheckIsEnabled
                                    || FeatureToggle.KubernetesLiveObjectStatusFeatureToggle.IsEnabled(deployment.Variables)
                                    || OctopusFeatureToggles.KubernetesObjectManifestInspectionFeatureToggle.IsEnabled(deployment.Variables))
                                {
-                                   var manifest = await PollForManifest(deployment, helmCli, releaseName, revisionNumber);
+                                   if (!DoesHelmCliSupportManifestRetrieval(out var helmVersion))
+                                   {
+                                       log.Warn($"Octopus needs Helm v3.13 or later to display object status and manifests. Your current version is {helmVersion}. Please update your Helm executable or container to enable our new Kubernetes capabilities. Learn more in our {log.FormatShortLink("KOS", "documentation")}.");
+                                       return;
+                                   }
```
APErebus, line 57: @kevjt I realised this _had_ to go inside this `if` statement, otherwise we would be doing the version check even when manifest retrieval was disabled

kevjt: oh yea, good catch

PR #1418, opened 2024-12-18
Support deploying empty or hooks-only Helm charts

APErebus:
Some customers have helm charts that are either empty or have only hooks in them.
This results in an empty `helm get manifest` call, which currently breaks manifest retrieval.

The code has now been updated to track if a successful `helm get manifest` call has occurred, so we know if the `manifest` variable is `null` _and_ we had a successful cli call.

Added two new tests for a hooks only chart and an empty chart (no hooks, no manifests)


source/Calamari/Kubernetes/Conventions/Helm/HelmManifestAndStatusReporter.cs
```diff
@@ -108,9 +120,15 @@ async Task<string> PollForManifest(RunningDeployment deployment,
...
                 }
 
-            if (string.IsNullOrWhiteSpace(manifest))
+            //it's possible that some manifests doesn't
+            if (string.IsNullOrWhiteSpace(manifest) && !didSuccessfullyExecuteCliCall)
                 throw new CommandException("Failed to retrieve helm manifest in a timely manner");
 
+            //Log if we found a manifest, or not
+            log.Verbose(string.IsNullOrWhiteSpace(manifest)
+                            ? $"Retrieve an empty manifest for {releaseName}, revision {revisionNumber}."
```
kevjt, line 129:
```suggestion
                            ? $"Retrieved an empty manifest for {releaseName}, revision {revisionNumber}."
```
Will also need to update the test assertions

source/Calamari/Kubernetes/Conventions/Helm/HelmManifestAndStatusReporter.cs
```diff
@@ -108,9 +120,15 @@ async Task<string> PollForManifest(RunningDeployment deployment,
                     await Task.Delay(TimeSpan.FromSeconds(1), ct.Token);
                 }
 
-            if (string.IsNullOrWhiteSpace(manifest))
+            //it's possible that some manifests doesn't
+            if (string.IsNullOrWhiteSpace(manifest) && !didSuccessfullyExecuteCliCall)
```
kevjt, line 124:
```suggestion
            if (!didSuccessfullyExecuteCliCall)
```
If the cli call failed for some reason, this should probably throw regardless of the cli output.

source/Calamari.Tests/KubernetesFixtures/Helm3UpgradeFixture.cs
```diff
@@ -111,6 +111,66 @@ public async Task HelmVersionOlderThanMinimumVersion_DoesNotRunObjectStatus()
             }
         }
 
+        [Test]
+        [RequiresNonFreeBSDPlatform]
+        [RequiresNon32BitWindows]
+        [RequiresNonMac]
+        [Category(TestCategory.PlatformAgnostic)]
```
kevjt, line 118: What does it mean for a test to be platform agnostic but still require a 64-bit Windows environment?

APErebus: I have no idea 🤣

kevjt (approved):
LGTM

PR #1420, opened 2025-01-08
Added hash to every flavour

rain-on:
The Calamari binaries go through various translations prior to being executed on a target.

Source Code --> Local project Binary --> Publish Directory --> Zip archive --> ConsolidatedBinary --> Reconstituted On Server.

It is important that the reconstituted binary on the server is the same as the binaries built, and tested as part of the Calamari build process.

This is generally assumed to have been correct due to existing integration tests etc in both Calamari an Server

This PR proposes that a new file be added to each Calamari binary (one for each flavour and platform) which contains a hash of the binary directories content.

This hash-file will form part of the Consolidated bundle which is embedded in Server, such that server can reconsitute the required Calamari flavour, calculate the hash of this application - and confirm it is the same (or not) as the value stored _in_ the binary.

This does not seek to overcome malicious behaviour (whereby the provider of a ConsolidatedBundle could provide illegal binaries, with matching hash).

It is only to ensure that the end-to-end process of Calamari Build --> Reconstituting does not introduce any errors.

NOTE: The Hash of each flavour will be confirmed in an octopusServer TEST, not at runtime.
This is sufficient to ensure an illegal ConsolidatedBundle cannot get onto master.

NOTE: Not yet tested in OctopusServer - but there is a risk that the hashes will not match! The current process potentially puts _too many_ files into the reconstituted directory! [which is actually safe, as the application just won't load/invoke those excess files]


build/Build.cs
```diff
@@ -33,62 +35,43 @@ partial class Build : NukeBuild
         readonly Configuration Configuration =
             IsLocalBuild ? Configuration.Debug : Configuration.Release;
 
-        [Required]
-        readonly Solution Solution = SolutionModelTasks.ParseSolution(SourceDirectory / "Calamari.sln");
+        [Required] readonly Solution Solution = SolutionModelTasks.ParseSolution(SourceDirectory / "Calamari.sln");
```
flin-8, line 38: Why has this formatting changed?

build/Build.cs
```diff
@@ -118,63 +101,63 @@ public Build()
...
-                  {
-                      Log.Information("Compiling Calamari v{CalamariVersion}", NugetVersion.Value);
-
-                      DotNetBuild(_ => _.SetProjectFile(Solution)
-                                        .SetConfiguration(Configuration)
-                                        .SetNoRestore(true)
-                                        .SetVersion(NugetVersion.Value)
-                                        .SetInformationalVersion(GitVersionInfo?.InformationalVersion));
-                  });
+            _ => //_.DependsOn(CheckForbiddenWords)
```
flin-8, line 149: comments?

build/Build.cs
```diff
@@ -696,6 +682,27 @@ string GetNugetVersion()
...
+        void CreateHashFileForProject(string directory)
+        {
+            byte[] GetHashOfFileContent(string f)
+            {
+                using var md5 = MD5.Create();
+                using var stream = File.OpenRead(f); 
+                return md5.ComputeHash(stream);
+            }
+
+            var allFiles = Directory.GetFiles(directory, "*.*", SearchOption.AllDirectories);
```
flin-8, line 713: QQ is this recursive?

rain-on: yeah it is - I figured we should to drill into all sub directories to collect the required hashes - might be slow? or is there another concern?
(btw - this results in the same file count as seen from the terminal, so logically - it does what I thought it should).

build/Build.cs
```diff
@@ -696,6 +701,27 @@ string GetNugetVersion()
...
+
+            var concatFileHashes = allFiles.SelectMany(GetHashOfFileContent);
```
flin-8, line 715: This only has the hash of files, not their filenames & path... I think you probably want something like:
- filepath1/hash1
- filepath2/hash2

then sort and concat that result.

rain-on: yeah - could, figured this would be safe, without going too far down the rabbit hole.

rain-on: :shrug: its a cheap change, and better represents the application "shape" - see if I can do it.

build/Build.cs
flin-8: Recursive is good :)

build/Build.cs
rain-on: reverted (i blame rider)

build/Build.cs
rain-on: reverted - was a shortcut to speed up manual testing

rain-on:
This should not be required once the consolidated calamari library is put together

PR #1423, opened 2025-01-15
Improved namespace handling for manifest reporting

APErebus:
Currently, for non-namespaced resource types (such as a `PersistentVolume`, or `ClusterRoleBinding`), they are reported with the implicit namespace.

This is a less than ideal solution as it gives an incorrect impression of the namespace of the resources, especially for manifest inspection.

This PR updates the manifest reporting to make a kubectl call to `api-resources` which returns an output that contains all the know resource types and if they are namespaced are not. Unfortunately it's a shitty format and can't be outputted as `json`, so we have a specific parser.

If we can't load that info, we then have a preset list of known values as a fallback.

We then use this lookup to work out if we need to _remove_ any namespace information.

*We only remove namespace information for resources where we _know_ they are non-namespaced*

A large chunk of this PR is cribbed from @liam-mackie's previous work.

EDIT:

As the helm KOS code performs a similar action for resolving the namespace, I have coalesced this logic into a new class, the `KubernetesManifestNamespaceResolver`. This code is responsible for the logic around what namespace to resolve based on the manifest, the variables and if we know if it's a namespaced resource or not (via the new cli call above)

## Further info

There are a number of scenario's that can occur. In the following scenario's, "known" or "unknown" refers to if the cluster knows about the resource type and can tell us if it's namespaced or not.

#### A namespaced resource _has_ the namespace in the manifest
✔️ The namespace in the manifest is used

#### A namespaced resouce _does not have_ a namespace in the manifest
✔️ The implicit namespace is used

#### A unknown namespaced resource _has_ a namespace in the manifest
✔️ The namespace in the manifest is used

#### A unknown namespaced resource _does not have_ a namespace in the manifest
✔️ The implicit namespace is used

#### A known non-namespaced resource _does not have_ a namespace in the manifest
✔️ The namespace is `null`

#### A known non-namespaced resource _has_ a namespace in the manifest
✔️ The namespace in the manifest is ignored and the namespace is `null`

#### A unknown non-namespaced resource _has_ a namespace in the manifest
❓ The namespace in the manifest is used. This is incorrect, but we can't know any different.

#### A unknown non-namespaced resource _does not have_ a namespace in the manifest
❓ The implicit namespace in the manifest is used. This is incorrect, but we can't know any different.


source/Calamari.Tests/KubernetesFixtures/ApiResourceOutputParserFixture.cs
```diff
@@ -0,0 +1,49 @@
...
+{
+    [TestFixture]
+    public class ApiResourceOutputParserFixture
+    {
+        [Test]
+        public void ShouldParseCorrectly()
+        {
+            List<string> outputLines = new List<string>
+            {
+"NAME                                SHORTNAMES   APIVERSION                        NAMESPACED   KIND                               VERBS                                                        CATEGORIES",
```
APErebus, line 16: Sample output

source/Calamari.Tests/KubernetesFixtures/ManifestReporterTests.cs
```diff
@@ -149,16 +164,106 @@ public void GivenValidYamlString_ShouldPostSingleServiceMessage(string enabledFe
...
-            var mr = new ManifestReporter(variables, CalamariPhysicalFileSystem.GetPhysicalFileSystem(), memoryLog);
+            var mr = new ManifestReporter(variables, CalamariPhysicalFileSystem.GetPhysicalFileSystem(), memoryLog, apiResourceScopeLookup);
 
             mr.ReportManifestApplied(yaml);
 
             var expected = ServiceMessage.Create(SpecialVariables.ServiceMessages.ManifestApplied.Name, ("ns", "default"), ("manifest", expectedYaml));
             memoryLog.ServiceMessages.Should().BeEquivalentTo(new List<ServiceMessage> { expected });
         }
 
+        [TestCase(nameof(FeatureToggle.KubernetesLiveObjectStatusFeatureToggle))]
```
APErebus, line 179: All these tests were copied from Liam's work :)

source/Calamari/Kubernetes/ApiResourceOutputParser.cs
```diff
@@ -0,0 +1,81 @@
...
+                throw new InvalidOperationException("kubectl api-resources did not have a valid header row");
+            }
+
+            //determine the size of each column based on the header row
+            var columns = new List<Column>();
+
+            var activeColumn = new Column();
+
+            var previousChar = SpaceChar;
+            for (var i = 0; i < headerRow.Length; i++)
```
APErebus, line 30: we use the width of the headers as to dictate how we substring each row

source/Calamari/Kubernetes/ApiResourceScopeLookup.cs
```diff
@@ -0,0 +1,119 @@
...
+            {
+                var apiResourceLines = kubectl.ExecuteCommandAndReturnOutput("api-resources", "-o", "wide");
+                apiResourceLines.Result.VerifySuccess();
+
+                return ApiResourceOutputParser.ParseKubectlApiResourceOutput(apiResourceLines.Output.InfoLogs.ToList());
+            }
+            catch
+            {
+                log.Warn("Unable to retrieve resource scoping using kubectl api-resources. Using default resource scopes.");
+                return DefaultResourceScopeLookup;
```
APErebus, line 44: Any error here is warned, but we fallback to our defaults

source/Calamari/Kubernetes/ManifestReporter.cs
```diff
@@ -24,20 +24,25 @@ public class ManifestReporter : IManifestReporter
...
+            this.resourceScopeLookup = resourceScopeLookup;
         }
 
         string GetNamespace(YamlMappingNode yamlRoot)
         {
-            var implicitNamespace = variables.Get(SpecialVariables.Namespace) ?? "default";
+            //we check to see if there is an explicit helm namespace defined first
+            //then fallback on the action/target default namespace
+            //otherwise fallback on default
+            var implicitNamespace = variables.Get(SpecialVariables.Helm.Namespace) ?? variables.Get(SpecialVariables.Namespace) ?? "default";
```
APErebus, line 45: Helm steps can have their own default namespace, so fallback on that first, then the target/action namespace, then the cluster default

source/Calamari.Tests/KubernetesFixtures/ManifestReporterTests.cs
APErebus: No longer relevant

source/Calamari/Kubernetes/KubernetesManifestNamespaceResolver.cs
```diff
@@ -0,0 +1,70 @@
...
+            if (!string.IsNullOrWhiteSpace(ns))
+            {
+                return ns;
+            }
+            
+            //if we don't, then fallback on variables
+            //we check to see if there is an explicit helm namespace defined first
+            //then fallback on the action/target default namespace
+            //otherwise fallback on default
+            return variables.Get(SpecialVariables.Helm.Namespace) ?? variables.Get(SpecialVariables.Namespace) ?? "default";
```
APErebus, line 60: Helm steps can have their own default namespace, so fallback on that first, then the target/action namespace, then the cluster default

source/Calamari/Kubernetes/KubernetesManifestNamespaceResolver.cs
```diff
@@ -0,0 +1,70 @@
...
+using YamlDotNet.RepresentationModel;
+
+namespace Calamari.Kubernetes
+{
+    public interface IKubernetesManifestNamespaceResolver
+    {
+        string ResolveNamespace(YamlMappingNode rootManifestNode, IVariables variables);
+    }
+
+    public class KubernetesManifestNamespaceResolver : IKubernetesManifestNamespaceResolver
```
APErebus, line 12: This is the new class that contains all the logic for resolving a namespace for a given manifest root node.

source/Calamari.Tests/KubernetesFixtures/ManifestReporterTests.cs
```diff
@@ -83,63 +92,31 @@ public void GivenInValidManifest_ShouldNotPostServiceMessage(string enabledFeatu
 
         [TestCase(nameof(FeatureToggle.KubernetesLiveObjectStatusFeatureToggle))]
         [TestCase(OctopusFeatureToggles.KnownSlugs.KubernetesObjectManifestInspection)]
-        public void GivenNamespaceInManifest_ShouldReportManifestNamespace(string enabledFeatureToggle)
```
APErebus, line 86: We don't test this kind of logic in the manifest reporter as this is the responsibility of the `KubernetesManifestNamespaceResolver`

source/Calamari.Tests/KubernetesFixtures/ApiResourceOutputParserTests.cs
```diff
@@ -0,0 +1,49 @@
...
+"validatingwebhookconfigurations                  admissionregistration.k8s.io/v1   false        ValidatingWebhookConfiguration     create,delete,deletecollection,get,list,patch,update,watch   api-extensions",
+"customresourcedefinitions           crd,crds     apiextensions.k8s.io/v1           false        CustomResourceDefinition           create,delete,deletecollection,get,list,patch,update,watch   api-extensions",
+"apiservices                                      apiregistration.k8s.io/v1         false        APIService                         create,delete,deletecollection,get,list,patch,update,watch   api-extensions",
+"controllerrevisions                              apps/v1                           true         ControllerRevision                 create,delete,deletecollection,get,list,patch,update,watch   "
+            };
+
+            var parsedResult = ApiResourceOutputParser.ParseKubectlApiResourceOutput(outputLines);
+
+            parsedResult.Should()
+                        .NotBeEmpty();
```
eddymoulton, line 46: Would be worth asserting that the output is as expected - I'd be happy with a really cut down list compared to what you've got to save the typing.

eddymoulton (approved):
Just a test I think could be better, but otherwise looks solid.

source/Calamari.Tests/KubernetesFixtures/ApiResourceOutputParserTests.cs
APErebus: ah yeah, I meant to do that. I think the original reason it was like this was I was using the test to debug and fix it... Good catch!

APErebus: Actually tested in [30b70a4](https://github.com/OctopusDeploy/Calamari/pull/1423/commits/30b70a47b90731774233af3c34ff789ba46b7d18)

source/Calamari.Tests/KubernetesFixtures/ApiResourceOutputParserTests.cs
```diff
@@ -0,0 +1,59 @@
...
+"validatingadmissionpolicybindings                admissionregistration.k8s.io/v1   false        ValidatingAdmissionPolicyBinding   create,delete,deletecollection,get,list,patch,update,watch   api-extensions",
+"validatingwebhookconfigurations                  admissionregistration.k8s.io/v1   false        ValidatingWebhookConfiguration     create,delete,deletecollection,get,list,patch,update,watch   api-extensions",
+"customresourcedefinitions           crd,crds     apiextensions.k8s.io/v1           false        CustomResourceDefinition           create,delete,deletecollection,get,list,patch,update,watch   api-extensions",
+"apiservices                                      apiregistration.k8s.io/v1         false        APIService                         create,delete,deletecollection,get,list,patch,update,watch   api-extensions",
+"controllerrevisions                              apps/v1                           true         ControllerRevision                 create,delete,deletecollection,get,list,patch,update,watch   "
+            };
+
+            var parsedResult = ApiResourceOutputParser.ParseKubectlApiResourceOutput(outputLines);
+
+            parsedResult.Should()
```
APErebus, line 45: test a bunch of the variations such as with/without shortnames, namespaced and different api versions

PR #1429, opened 2025-01-24
Fix variable substitution for Helm Template Value Source types

kevjt:
## Background

Part of https://github.com/OctopusDeploy/Issues/issues/9224

A previous [update](https://github.com/OctopusDeploy/Calamari/pull/1426) to fix an issue with unescaped quotes in `InlineYamlTemplateValuesSource` introduced a bug where Octopus variables in other TemplateValuesSource types are not being evaluated.

## Result

- Perform variable evaluation on the properties of all the TemplateValuesSource class types
- Refactored the code to make testing this a bit easier


APErebus (approved):
Excellent stuff 👏 
Thanks for handling this one

PR #1433, opened 2025-02-05
fix: don't assign our oidc jwt token to the kubectl token config value

hnrkndrssn:
We were passing the in the Octopus OIDC JWT when configuring credentials for `kubectl` this causes authentication using AWS OIDC accounts for EKS clusters to fail due to the token not matching the expected format, `k8s-aws-v1.`, which is the token generated when calling `aws eks get-token`.

Relates to OctopusDeploy/Issues#9243

[sc-102071]

### Before
image

### After
image


IsaacCalligeros95 (approved):
LGTM, my understanding of the flow here is that the AWS Environment Generation handles the OIDC authentication and sets the AWS_SESSION_TOKEN and that's what's used automatically for the get token command now.

I'm just curious about how this was tested and whether it's been tested without the cloud target discovery step running before hand. I think AWS persists the creds file to disk which AFAIK should be removed between steps & shouldn't be an issue but I just want to sanity check that this isn't working because of the previous authentication step.

benPearce1 (approved):
nice, delete all the code!

hnrkndrssn:
> I'm just curious about how this was tested and whether it's been tested without the cloud target discovery step running before hand

Good point, I have tested this with both cloud target discovery and the cluster as a deployment target without using cloud target discovery and both scenarios work as expected.

PR #1441, opened 2025-02-13
Add variable enumeration for bash

IsaacCalligeros95:
Support enumerating variable in bash scripts. This change is based off this [2018 change](https://github.com/OctopusDeploy/Calamari/pull/337/files#diff-6cf26d1135d46ee2458052461a45a0843e0c0e86224290de4512618d594e781e) and will add support for accessing variables in bash scripts as associative arrays. In Powershell this is supported by the `OctopusParameters` property, in bash this will be backed by `octopus_parameters`. Details on iterating associative arrays are [available here](https://phoenixnap.com/kb/bash-associative-array). As an example this can be used with 

The implementation for this is based off of the PowerShell implementation of OctopusParameters `Decrypt-Variables`. 

For PowerShell we do the following in **C#**
* Base 64 encode the variable name and value with a `$` delimiter between them. (This is to make sure all special characters are removed)
* Encrypt the combined string of all variables (We can a base string and IV)
* Base64 encode the base string and convert the IV to hex.

These are then string replaced in the PowerShell bootstrap.ps1 script.

The PowerShell bootstrap script does the following:
* Decodes the IV and Hex
* Decrypts the whole string
* Decodes one by one the variable name and values

This works well in Powershell ~3000 variables takes about 1 second in this process.
In base we use Openssl to decode base64 strings, this spins up a process and overall repeating this same process for ~3000 variables takes ~60 seconds.

To work around this I've settled on Hex encoding the variable names, unlike base64 hex encoded strings can be concatenated and decrypted all at once, this also doesn't depend on the openssl process being invoked. In this version I've settled on xxd, and if it's not available we do not populate the octopus_parameters variable. We can handle this ourselves but the implementation is messy and I did run into some issues with encoding/decoding emojis.

Accessing keys

```
echo ${!example_array[@]}
```

Accessing values
```
echo ${example_array[@]}
```

Accessing a value
```
echo ${example_array["key1"]}
```

Combined example.
```
for key in "${!octopus_parameters[@]}"; do
    value="${octopus_parameters[$key]}"
    echo "Key: $key, Value: $value"
done
```

After:
image

image

PR #1443, opened 2025-02-19
Bash variable name iteration

IsaacCalligeros95:
This is an alternative to #1441 If we can use the cli command `xxd` this is not an issue and simplifies the implementation significantly.

This change adds the ability to iterate variable names in base. The implementation for this is based off of the PowerShell implementation of OctopusParameters `Decrypt-Variables`. 

For PowerShell we do the following in **C#**
* Base 64 encode the variable name and value with a `$` delimiter between them. (This is to make sure all special characters are removed)
* Encrypt the combined string of all variables (We can a base string and IV)
* Base64 encode the base string and convert the IV to hex.

These are then string replaced in the PowerShell bootstrap.ps1 script.

The PowerShell bootstrap script does the following:
* Decodes the IV and Hex
* Decrypts the whole string
* Decodes one by one the variable name and values

This works well in Powershell ~3000 variables takes about 1 second in this process.
In base we use Openssl to decode base64 strings, this spins up a process and overall repeating this same process for ~3000 variables takes ~60 seconds.

To work around this I've settled on Hex encoding the variable names, unlike base64 hex encoded strings can be concatenated and decrypted all at once, this also doesn't depend on the openssl process being invoked. We have to handle the hex to string ourselves, we could use [strtonum](https://stackoverflow.com/questions/8199934/strtonum-in-os-x-not-found) but it's not available everywhere. 

image

Accessing keys

```
echo ${!example_array[@]}
```

### DIFF
--- a/source/Calamari.Common/Features/Scripting/Bash/Bootstrap.sh
+++ b/source/Calamari.Common/Features/Scripting/Bash/Bootstrap.sh
@@ -266,7 +266,7 @@ function log_environment_information
 log_environment_information
 
 if (( ${BASH_VERSINFO[0]} > 4 || (${BASH_VERSINFO[0]} == 4 && ${BASH_VERSINFO[1]} > 2) )); then
-  declare -grA example_array=(#### VariableNamesArrayDeclarations ####)
+  declare -grA octopus_parameters=(#### VariableNamesArrayDeclarations ####)
 else
   echo "Bash version 4.2 or later is required to use octopus_parameters"
 fi
--- a/source/Calamari.Common/Features/Scripting/Bash/Bootstrap.sh
+++ b/source/Calamari.Common/Features/Scripting/Bash/Bootstrap.sh
@@ -66,16 +66,6 @@ function get_octopusvariable
     esac
 }
 
-# -----------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


source/Calamari.Common/Features/Scripting/Bash/Bootstrap.sh
```diff
@@ -263,4 +263,92 @@ function log_environment_information
...
+    rm -f "$tmp_file"
+
+    for key in "${decoded_keys[@]}"; do
+        # Value is empty for now, since we write the hex encoded key/values to disk we don't want to write the values
+        # https://github.com/OctopusDeploy/Calamari/pull/1441/files is an example implementation that includes the values
+        octopus_parameters["$key"]=""
+    done
+}
+
+hex_to_string() {
```
IsaacCalligeros95, line 311: This can be removed if we can depend on `xxd`. This is generally supported but not available everywhere, we could check if it is available and only add the octopus_parameters if it is.

IsaacCalligeros95: `hexdump` seems generally available on Nix, but is not installed by default on mac

PR #1445, opened 2025-03-03
ServiceFabric missing DLL fix/AAD token changes. 

MissedTheMark:
### Background
Prior to the introduction of Net6 AzureServiceFabric, there was only NetFx - so Octopus would be forced to use the NetFx build of Calamari.

Octopus defaults to choosing NetCore when available, not NetFx (based on ordering in [DeploymentPackageResolver](https://github.com/OctopusDeploy/OctopusDeploy/blob/main/source/Octopus.Server/Orchestration/ServerTasks/Deploy/DeploymentPackage/DeploymentPackageResolver.cs)) - therefore, anyone who installed Calamari 28.2.2 or newer will get the Net6 version of AzureServiceFabric in Calamari..

This version was unable to authenticate using Azure AD with username/password (Certs etc work fine), because it spawns a PowerShell script which in turn attempts to do an Add-Type -Path Microsoft.Identity.Client which fails.

Why?

Well, the Net6 Calamari is a standalone app - i.e. it wants NOTHING installed on the box - everything is self-contained.
Unfortunately, Powershell 5 is a NetFx app, and relies on the Global Assembly Cache (GAC) - which is a global repository of libraries that it wants to use.

So when we try and load Microsoft.Identity.Client in PS5, it tries to load its dependencies, and it fails because they're not in the GAC like it expects.

### Changes in this PR
To solve the above problem, I've moved the Azure AD token retrieval into C# code, and pass it to the Powershell script. This means we don't have to rely on Powershell trying to find Microsoft.Identity.Client (and failing).

In doing this, I've modified and consolidated the existing health check code for auth token retrieval. It's worth noting that because Calamari is a remote tool, it's not able to trigger an interactive MFA auth flow. This was an existing restriction (and a good reason to use another auth type), but I've improved the messaging around this for both health checks and deployments.

Old messaging:
SCR-20250305-kjeu
New messaging:
SCR-20250305-kxhf

The retained "invalid username or password' message on the end is unfortunately just the message that Azure AD returns when interaction is required for the auth flow.

In moving the token retrieval to C# code, I've also removed the existing 'support' for AzureAD SecureClientCertificate, but AFAIK Service Fabric (still) doesn't support that anyway, and we filter that out at our endpoint too: [AzureServiceFabricClusterEndpointValidator](https://github.com/OctopusDeploy/OctopusDeploy/blob/3ec629d30a8d820449460c4a05bd96a2fe63502f/source/Octopus.Server.Extensibility.Sashimi.AzureServiceFabric/Endpoints/AzureServiceFabricClusterEndpointValidator.cs#L28)
Our front-end code doesn't support it either. 

### How to review
I've tested this on a Win11 vm, connecting to a Service Fabric cluster.  Unfortunately I can't test the _full_ process due to our Azure MFA requirements, so it would be good if you have a personal cluster without MFA to test it on.  I _have_, however, been able to demonstrate that we can get past the DLL issue.
I can give you the details for connecting to my cluster if needed.

I'd also appreciate a general sanity check to make sure that I haven't missed something here.

### Related links
Fixes https://github.com/OctopusDeploy/Issues/issues/8023
Shortcut story [#102627](https://app.shortcut.com/octopusdeploy/story/102627/sev-2-service-fabric-deployments-broken-due-to-calamari-missing-microsoft-identity-client-dll-since-upgrade-to-2024-4-6995)


source/Calamari.AzureServiceFabric/Integration/AzureServiceFabricPowerShellContext.cs
```diff
@@ -57,31 +61,39 @@ public CommandResult ExecuteScript(Script script,
...
             if (securityMode == AzureServiceFabricSecurityMode.SecureClientCertificate.ToString())
             {
-                var certificateVariable = GetMandatoryVariable(SpecialVariables.Action.ServiceFabric.ClientCertVariable);
+                var certificateVariable = variables.GetMandatoryVariable(SpecialVariables.Action.ServiceFabric.ClientCertVariable);
                 clientCertThumbprint = variables.Get($"{certificateVariable}.{CertificateVariables.Properties.Thumbprint}");
             }
+            else if (securityMode == AzureServiceFabricSecurityMode.SecureAzureAD.ToString())
+            {
+                if (string.IsNullOrWhiteSpace(aadUserCredUsername))  return new CommandResult("Failed to find a value for the AAD username.", 0);
+                if (string.IsNullOrWhiteSpace(aadUserCredPassword))  return new CommandResult("Failed to find a value for the AAD password.", 0);
```
MissedTheMark, line 80: This validation was originally performed on line 74-81 of `AzureServiceFabricContext.ps1`
Now that we're getting the AAD token in C#, we need to validate it in this file instead.

source/Calamari.AzureServiceFabric/Integration/AzureServiceFabricPowerShellContext.cs
```diff
@@ -57,31 +61,39 @@ public CommandResult ExecuteScript(Script script,
...
             SetOutputVariable("OctopusFabricCertificateFindValueOverride", variables.Get(SpecialVariables.Action.ServiceFabric.CertificateFindValueOverride));
             SetOutputVariable("OctopusFabricCertificateStoreLocation", variables.Get(SpecialVariables.Action.ServiceFabric.CertificateStoreLocation, "LocalMachine"));
             SetOutputVariable("OctopusFabricCertificateStoreName", variables.Get(SpecialVariables.Action.ServiceFabric.CertificateStoreName, "MY"));
-            SetOutputVariable("OctopusFabricAadCredentialType", variables.Get(SpecialVariables.Action.ServiceFabric.AadCredentialType));
-            SetOutputVariable("OctopusFabricAadClientCredentialSecret", variables.Get(SpecialVariables.Action.ServiceFabric.AadClientCredentialSecret));
-            SetOutputVariable("OctopusFabricAadUserCredentialUsername", variables.Get(SpecialVariables.Action.ServiceFabric.AadUserCredentialUsername));
-            SetOutputVariable("OctopusFabricAadUserCredentialPassword", variables.Get(SpecialVariables.Action.ServiceFabric.AadUserCredentialPassword));
```
MissedTheMark, line 84: Doing Azure AD token retrieval in C# means we don't need to pass these to powershell.  Note that (as mentioned in the in-code comment below), we don't support the AAD `SecureClientCertificate` auth type, mostly because (as far as I can tell) Azure doesn't support it for Service Fabric, but also because we filter it out in `AzureServiceFabricClusterEndpointValidator.cs`

source/Calamari.AzureServiceFabric/Scripts/AzureServiceFabricContext.ps1
```diff
@@ -15,11 +15,7 @@
 ##   OctopusFabricCertificateFindValueOverride               // The type of FindValue for searching the certificate in the Azure certificate store (use this if you specify a FindType different to 'FindByThumbprint' and do NOT wish to use the client certificate thumbprint value)
 ##   OctopusFabricCertificateStoreLocation                   // The certificate store location (should be 'LocalMachine' by default)
 ##   OctopusFabricCertificateStoreName                       // The certificate store name (should be 'MY' by default)
-##   OctopusFabricAadCredentialType                          // The credential type for authentication
-##   OctopusFabricAadClientCredentialSecret                  // The client application secret for authentication
-##   OctopusFabricAadUserCredentialUsername                  // The username for authentication
-##   OctopusFabricAadUserCredentialPassword                  // The password for authentication
-##   OctopusFabricActiveDirectoryLibraryPath                 // The path to Microsoft.Identity.Client.dll
```
MissedTheMark, line 22: These have been replaced with the token below, and used in the C# token retrieval.

source/Calamari.AzureServiceFabric/Scripts/AzureServiceFabricContext.ps1
```diff
@@ -67,97 +63,9 @@ function ValidationMessageForAzureADParameters() {
...
-        }
-        if (!$OctopusFabricAadUserCredentialPassword) {
-            return "Failed to find a value for the AAD password."
-        }
-    }
-
     return $null
 }
 
-function GetAzureADAccessToken() {
```
MissedTheMark, line 86: This is the original Powershell code to retrieve the token, now in C#.

source/Calamari.AzureServiceFabric/HealthCheckBehaviour.cs
```diff
@@ -133,24 +122,5 @@ public async Task Execute(RunningDeployment context)
                 fabricClient.Dispose();
             }
         }
-
-        #region Auth helpers
-        static string GetAccessToken(AzureActiveDirectoryMetadata aad, string aadUsername, string aadPassword)
```
MissedTheMark, line 138: Moved into common class.

source/Calamari.AzureServiceFabric/AzureADUsernamePasswordTokenRetriever.cs
```diff
@@ -0,0 +1,45 @@
...
+                    .GetResult();
+
+                return authResult.AccessToken;
+            }
+            catch (MsalUiRequiredException ex)
+            {
+                log.Error($"Unable to retrieve authentication token: User interaction is required to connect to the Service Fabric cluster with the provided account. Please change the account's settings, or use a different account.");
+                log.Error($"Details: {ex.PrettyPrint()}");
+                return "BAD_TOKEN"; // You cannot return null or an empty value here or the Azure lib fails trying to call a lib that doesn't exist "System.Fabric.AzureActiveDirectory.Client" 
+            }
```
MissedTheMark, line 37: If you connect with MFA enabled for the account, or some other _thing_ that requires interaction (eg expired password), Calamari won't be able to connect.

source/Calamari.AzureServiceFabric/Integration/AzureServiceFabricPowerShellContext.cs
```diff
@@ -90,6 +102,45 @@ public CommandResult ExecuteScript(Script script,
             }
         }
 
+        async Task<string> GetAzureADAccessToken(string serverCertThumbprint, string connectionEndpoint, string aadUserCredentialUsername, string aadUserCredentialPassword)
```
rain-on, line 105: I _feel_ like this could be pulled out and made common between the health-check and here - maybe even pulling out the whole switch-statement around auth-type.

source/Calamari.AzureServiceFabric/Calamari.AzureServiceFabric.csproj
```diff
@@ -23,6 +23,7 @@
     SDK: https://download.microsoft.com/download/b/8/a/b8a2fb98-0ec1-41e5-be98-9d8b5abf7856/MicrosoftServiceFabricSDK.7.1.2338.msi
     -->
     <PackageReference Include="Microsoft.ServiceFabric" Version="10.1.2338" />
+    <PackageReference Include="Microsoft.ServiceFabric.Client.Http" Version="5.0.1" />
```
rain-on, line 26: if we lean into what was done in the health-check is this extra dependency required?

source/Calamari.Tests/Helpers/StubScriptWrapper.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace Calamari.Tests.Helpers
 {
-    class StubScriptWrapper : IScriptWrapper
+    public class StubScriptWrapper : IScriptWrapper
```
rain-on, line 9: ? now sure where this is being used that required public-ness.

source/Calamari.AzureServiceFabric/Integration/AzureServiceFabricPowerShellContext.cs
rain-on: or maybe not - you are doing slightly different things

rain-on (approved):
LGTM

rain-on (changes requested):
Needs to target 2025.1 - then let it roll forward

rain-on (approved):
Approved now on correct branch

PR #1448, opened 2025-03-07
K8s: An unparsable ApiVersion string now throws, rather than returning "null, null"

rain-on:


APErebus (commented):
Should we also throw if there is no apiVersion? (line 32)

APErebus (approved):

PR #1451, opened 2025-03-12
Fix some warnings/nullable inconsistencies.

MissedTheMark:
### Background
Calamari has a whole bunch of build warnings, and nullable things that are inconsistent/misleading.

The PR endeavors to fix some of these problems.

### How to review
For once, my commits are well named, and each contains a specific change across files, or changes in a single file.  It would probably be best to go through each commit one at a time, using the commit message for context, rather than attempting to review the whole PR without context.

### DIFF
--- a/source/Calamari.Common/Features/Deployment/Journal/DeployedPackage.cs
+++ b/source/Calamari.Common/Features/Deployment/Journal/DeployedPackage.cs
@@ -10,6 +10,7 @@ namespace Calamari.Common.Features.Deployment.Journal
 {
     public class DeployedPackage
     {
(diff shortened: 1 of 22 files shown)


source/Calamari.Common/Features/Deployment/Journal/DeployedPackage.cs
```diff
@@ -10,6 +10,7 @@ namespace Calamari.Common.Features.Deployment.Journal
 {
     public class DeployedPackage
     {
+#pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor - Handled by guard method.
```
rain-on, line 13: I'm confused - the API allows null - but then we bounce it via the Guard?

PR #1452, opened 2025-03-12
Fix Calamari multi target build

MissedTheMark:
Only run the `CopyDotnetScriptFilesAfterBuild` target for non-cross-targeting builds (i.e. only run for each framework build).

See explanation here: https://octopusdeploy.slack.com/archives/C06FXB5PT/p1741672943262309

Also cleans up the build.cs file a bit.

### DIFF
--- a/build/Build.cs
+++ b/build/Build.cs
@@ -117,8 +117,8 @@ namespace Calamari.Build
         Lazy<string> NugetVersion { get; }
 
         Target CheckForbiddenWords =>
-            _ => _.Executes(() =>
-            {
+            d => d.Executes(() =>
+            {         
                 Log.Information("Checking codebase for forbidden words");
 
                 const string arguments =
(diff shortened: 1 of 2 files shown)


build/Build.cs
```diff
@@ -117,8 +117,8 @@ public Build()
         Lazy<string> NugetVersion { get; }
 
         Target CheckForbiddenWords =>
-            _ => _.Executes(() =>
-            {
+            d => d.Executes(() =>
+            {
```
rain-on, line 121: whitespace - here and elsewhere

rain-on (approved):
Love to fix whitespace issues - but otherwise looks good.

build/Build.cs
MissedTheMark: I'll sort that out. Thanks!

PR #1459, opened 2025-03-23
Use region endpoint for assume role

bec-callow-oct:
[SC-101215]

When attempting to run a `Run an AWS CLI Script` or `Plan to apply a Terraform template` step with `Assume a different AWS service role` enabled, the AWS global STS endpoint is used. This endpoint is not supported by AWS `opt-in` regions, resulting in an AuthFailure when attempting to run the step. This PR ensures that the region endpoint is used for the region specified within the step process.

Fixes: https://github.com/OctopusDeploy/Issues/issues/9216

### DIFF
--- a/source/Calamari.CloudAccounts/AwsEnvironmentGeneration.cs
+++ b/source/Calamari.CloudAccounts/AwsEnvironmentGeneration.cs
@@ -358,7 +358,7 @@ namespace Calamari.CloudAccounts
         {
             if ("True".Equals(assumeRole, StringComparison.OrdinalIgnoreCase))
             {
-                var client = new AmazonSecurityTokenServiceClient(AwsCredentials, AwsRegion);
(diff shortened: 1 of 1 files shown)


source/Calamari.CloudAccounts/AwsEnvironmentGeneration.cs
```diff
@@ -358,7 +358,7 @@ async Task AssumeRole()
         {
             if ("True".Equals(assumeRole, StringComparison.OrdinalIgnoreCase))
             {
-                var client = new AmazonSecurityTokenServiceClient(AwsCredentials);
+                var client = new AmazonSecurityTokenServiceClient(AwsCredentials, AwsRegion);
```
mik-ky, line 361: What happens if the environment variable for `AwsRegion` is not set?

bec-callow-oct: > What happens if the environment variable for `AwsRegion` is not set?

Good point. The field is required in Octopus Server, but is not validated in Calamari. I've added a fallback

mik-ky (approved):

PR #1461, opened 2025-04-01
Adjust ApiResources error log level to be Verbose

APErebus:
We've had feedback from users that the `log.Warn` here is confusing.

It's not something the user can fix and isn't really a big issue as there is a default, so I'm comfortable changing this back to `Verbose`


rain-on (approved):
LGTM

PR #1464, opened 2025-04-03
Update default region for s3 bucket to the AWS global endpoint region

hnrkndrssn:
The default region for S3 buckets, according to AWS, is `us-east-1` so this PR updates the region we use by default from `us-west-1` to `us-east-1`.

[sc-107220]

Relates to OctopusDeploy/Issues#9338


source/Calamari.Tests/KubernetesFixtures/Terraform/Clusters/AKS-local-access-disabled/aks.tf
```diff
@@ -8,7 +8,7 @@ resource "azurerm_kubernetes_cluster" "default" {
   resource_group_name = azurerm_resource_group.default.name
   location            = "Australia East"
   dns_prefix          = "${random_pet.prefix.id}-k8s"
-  kubernetes_version  = "1.28"
+  kubernetes_version  = "1.29"
```
hnrkndrssn, line 11: This change is to fix AKS tests.

benPearce1 (approved):

PR #1471, opened 2025-04-10
Support OIDC Auth for GCR Feeds

grace-rehn:

Updates Calamari to take contributed JWT and Audience variables to do an OIDC auth exchange for Google Container Registry Feeds.

Part of [#project-oidc-external-feeds](https://octopusdeploy.slack.com/archives/C08GX2ANBRP)

[sc-107861]

### DIFF
--- a/source/Calamari.CloudAccounts/GoogleAuthenticationProvider.cs
+++ b/source/Calamari.CloudAccounts/GoogleAuthenticationProvider.cs
@@ -1,0 +1,95 @@
+using System;
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using Calamari.Common.Plumbing.Logging;
+using Calamari.Common.Plumbing.Variables;
+using Newtonsoft.Json;
+using NetWebRequest = System.Net.WebRequest;
+
+
+namespace Calamari.CloudAccounts
+{
+    public class GoogleAuthenticationProvider
+    {
+        readonly HttpClient httpClient;
+        const string StsEndpoint = "https://sts.googleapis.com/v1/token";
+        const string GcrScope = "https://www.googleapis.com/auth/devstorage.read_write";
+        const string OidcUsername = "oauth2accesstoken";
+        
+        public GoogleAuthenticationProvider()
+        {
+            httpClient = new HttpClient(new HttpClientHandler {Proxy = NetWebRequest.DefaultWebProxy});
+        }
+        
+        public (string Username, string Password, Uri RegistryUri) GetGcrUserNamePasswordCredentials(string username, string password, Uri registryUri)
+        {
+            return (username, password, registryUri);
+        }
+        
+        public async Task<(string Username, string Password, Uri RegistryUri)> GetGcrOidcCredentials(IVariables variables, Uri registryUri)
+        {
+            try
+            {
+                Log.Verbose("Starting GCR OIDC credential retrieval process");
+                var jwt = variables.Get(AuthenticationVariables.Jwt);
+                var audience = variables.Get(AuthenticationVariables.Google.Audience)?.Replace("https:", "");
+                var gcrToken = await ExchangeJwtForGcrToken(jwt, audience);
+                Log.Verbose($"Successfully retrieved credentials for {registryUri}");
+                return (OidcUsername, gcrToken, registryUri);
+            }
+            catch (Exception ex)
+            { 
+                throw new Exception($"GCR-LOGIN-ERROR: Failed to verify OIDC credentials. Error: {ex.Message}", ex);
+            }
+        }
+        
+        async Task<string> ExchangeJwtForGcrToken(string jwt, string audience)
+        {
+            
+            var requestContent = new FormUrlEncodedContent(new Dictionary<string, string>
+            {
+                { "grant_type", "urn:ietf:params:oauth:grant-type:token-exchange" },
+                { "audience", audience },
+                { "scope", GcrScope },
+                { "requested_token_type", "urn:ietf:params:oauth:token-type:access_token" },
+                { "subject_token", jwt },
+                { "subject_token_type", "urn:ietf:params:oauth:token-type:jwt" }
+            });
+    
+            var response = httpClient.PostAsync(StsEndpoint, requestContent, CancellationToken.None)
+                                     .GetAwaiter().GetResult();
+            if (!response.IsSuccessStatusCode)
+            {
+                var errorContent = await response.Content.ReadAsStringAsync();
+                throw new Exception($"Failed to exchange JWT for GCR token. Status: {response.StatusCode}, Error: {errorContent}");
+            }
+
+            var responseJson = await response.Content.ReadAsStringAsync();
+            var tokenResponse = JsonConvert.DeserializeObject<TokenResponse>(responseJson);
+    
+            if (string.IsNullOrEmpty(tokenResponse?.AccessToken))
+            {
+                throw new Exception("Access token was not present in the response");
+            }
+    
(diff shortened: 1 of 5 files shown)


source/Calamari.CloudAccounts/GoogleAuthenticationProvider.cs
```diff
@@ -0,0 +1,95 @@
...
+                { "grant_type", "urn:ietf:params:oauth:grant-type:token-exchange" },
+                { "audience", audience },
+                { "scope", GcrScope },
+                { "requested_token_type", "urn:ietf:params:oauth:token-type:access_token" },
+                { "subject_token", jwt },
+                { "subject_token_type", "urn:ietf:params:oauth:token-type:jwt" }
+            });
+    
+            var response = httpClient.PostAsync(StsEndpoint, requestContent, CancellationToken.None)
+                                     .GetAwaiter().GetResult();
```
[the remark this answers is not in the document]
grace-rehn: Yep, I'll update this. Thanks :)

source/Calamari.CloudAccounts/GoogleAuthenticationProvider.cs
```diff
@@ -0,0 +1,94 @@
...
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using Calamari.Common.Plumbing.Logging;
+using Calamari.Common.Plumbing.Variables;
+using Newtonsoft.Json;
+using NetWebRequest = System.Net.WebRequest;
+
+
```
wlthomson, line 11: Super nit: whitespace.

source/Calamari.CloudAccounts/GoogleAuthenticationProvider.cs
```diff
@@ -0,0 +1,94 @@
...
+    public class GoogleAuthenticationProvider
+    {
+        readonly HttpClient httpClient;
+        const string StsEndpoint = "https://sts.googleapis.com/v1/token";
+        const string GcrScope = "https://www.googleapis.com/auth/devstorage.read_write";
+        const string OidcUsername = "oauth2accesstoken";
+        
+        public GoogleAuthenticationProvider()
+        {
+            httpClient = new HttpClient(new HttpClientHandler {Proxy = NetWebRequest.DefaultWebProxy});
```
wlthomson, line 23: Probably not an issue, but I recall that directly constructing `HttpClient` can be problematic, depending on how frequently this class is instantiated?

Is it possible/a good idea to add the client as a constructor argument and inject a shared one in?

source/Calamari.CloudAccounts/GoogleAuthenticationProvider.cs
```diff
@@ -0,0 +1,94 @@
...
+            return (username, password, registryUri);
+        }
+        
+        public async Task<(string Username, string Password, Uri RegistryUri)> GetGcrOidcCredentials(IVariables variables, Uri registryUri)
+        {
+            try
+            {
+                Log.Verbose("Starting GCR OIDC credential retrieval process");
+                var jwt = variables.Get(AuthenticationVariables.Jwt);
+                var audience = variables.Get(AuthenticationVariables.Google.Audience)?.Replace("https:", "");
```
wlthomson, line 37: Another minor: is it worth parsing the `Audience` as a `Uri` as a sanity check? 

e.g. `new Uri(variables.Get(AuthenticationVariables.Google.Audience)).Host;`

I think that would allow you to get the stripped [Uri.Host](https://learn.microsoft.com/en-us/dotnet/api/system.uri.host?view=net-9.0) (although looks like you have intentionally left `//` in there, so may not be what you're after).

wlthomson (commented):
Couple of minors, completely optional.

Would approve if I knew more about what's going on 😆 , but will leave that to others with more context in this area!

source/Calamari.Shared/Integration/Packages/Download/DockerImagePackageDownloader.cs
```diff
@@ -75,7 +75,8 @@ public PackagePhysicalFileMetadata DownloadPackage(string packageId,
         {
             var feedType = variables.Get(AuthenticationVariables.FeedType);
             if (feedType == FeedType.AwsElasticContainerRegistry.ToString()
```
wlthomson, line 77: Probably overkill, but it might be worth adding a little mapper for this which just takes a string and matches it to the associated `FeedType` (this would also let you map case insensitively etc. if you so wanted)?

source/Calamari.CloudAccounts/GoogleAuthenticationProvider.cs
wlthomson: From a quick search in the sln, it looks like this is a pattern we use in a few places already, so I think you can ignore this one!

source/Calamari.CloudAccounts/GoogleAuthenticationProvider.cs
grace-rehn: Yeah I think Google expects `//` to be present in the audience. I could use a hybrid approach and use a URI and add back the `//`.

Might be useful so I can ensure both `https` and `http` and stripped out.

source/Calamari.CloudAccounts/GoogleAuthenticationProvider.cs
benPearce1: @wlthomson The execution model for calamari would mostly make HttpClient reuse redundant. The lifetime of the code is a single execution per step (Calamari is actually a CLI)

source/Calamari.Shared/Integration/Packages/Download/GcrFeedLoginDetailsProvider.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace Calamari.Integration.Packages.Download
+{
+    public class GcrFeedLoginDetailsProvider : IFeedLoginDetailsProvider
+    {
+        public async Task<(string Username, string Password, Uri FeedUri)> GetFeedLoginDetails(IVariables variables, string? username, string? password, Uri feedUri)
+        { 
+            var gcrAuth = new GoogleAuthenticationProvider();
+            var usingOidc = !string.IsNullOrWhiteSpace(variables.Get(AuthenticationVariables.Jwt));
+            Log.Verbose(usingOidc ? "Gcr Feed - OIDC token detected - using token-based authentication flow." : "Gcr Feed - Using json key authentication flow.");
+            return usingOidc ? (await gcrAuth.GetGcrOidcCredentials(variables, feedUri)) : ( await Task.FromResult(gcrAuth.GetGcrUserNamePasswordCredentials(username, password, feedUri)));
```
benPearce1, line 16:
```suggestion
            if (usingOidc) {
                 Log.Verbose("Gcr Feed - OIDC token detected - using token-based authentication flow.");
                 return await gcrAuth.GetGcrOidcCredentials(variables, feedUri);
            } 
            
            Log.Verbose("Gcr Feed - Using json key authentication flow.");
            return await Task.FromResult(gcrAuth.GetGcrUserNamePasswordCredentials(username, password, feedUri));
```

source/Calamari.Shared/Integration/Packages/Download/DockerImagePackageDownloader.cs
grace-rehn: Thanks Will, I took another look at this to try clean it up.

I'm now using `Enum.TryParse(contributedFeedType, ignoreCase: true, out FeedType feedType)` and have our supported feed types stored in `SupportedLoginDetailsFeedTypes`.

So this should clean things up a little and also handle case insensitivity.

benPearce1 (approved):

PR #1472, opened 2025-04-11
Change to use Server-like versioning

APErebus:


zentron (approved):
Nice, good to see the change to Octoversion as well.

PR #1474, opened 2025-04-16
Add script functions for reporting Kubernetes manifests from scripts

APErebus:

Adds new functions to both the PowerShell and Bash bootstrap scripts to allow for reporting of kubernetes manifests.

There are two functions

1. `Report-KubernetesManifest`/`report_kubernetes_manifest` which takes two parameters, the manifest content and optionally, the namespace.

This function splits the manifest on `---` (with/without trailing spaces) and registers each document separately using the service message.

If a manifest contains multiple documents and a namespace is provided, all reported manifests get the same namespace

2. `Report-KubernetesManifestFile`/`report_kubernetes_manifest_file` which takes two parameters, the full path to the manifest file and optionally, the namespace.

This function reads the contents of the file and then calls the function above

### DIFF
--- a/source/Calamari.Common/Features/Scripting/Bash/Bootstrap.sh
+++ b/source/Calamari.Common/Features/Scripting/Bash/Bootstrap.sh
@@ -358,15 +358,15 @@ function report_kubernetes_manifest
 
 function report_kubernetes_manifest_file 
 {
-  PATH="$1"
+  FULL_PATH="$1"
   NAMESPACE="$2"
   
-  if [ ! -f "$PATH" ]; then
-    write_verbose "No Kubernetes manifest file was found at '$PATH'."
+  if [ ! -f "$FULL_PATH" ]; then
+    write_verbose "No Kubernetes manifest file was found at '$FULL_PATH'."
     return
   fi
   
-  MANIFEST=$(cat "$PATH")
+  MANIFEST=$(cat "$FULL_PATH")
   
   report_kubernetes_manifest "$MANIFEST" "$NAMESPACE"
 }
--- a/source/Calamari.Common/Features/Scripting/Bash/Bootstrap.sh
+++ b/source/Calamari.Common/Features/Scripting/Bash/Bootstrap.sh
@@ -307,6 +307,72 @@ function decrypt_and_parse_variables {
     exec 3<&-
 }
 
+# -----------------------------------------------------------------------------
+# Functions to report Kubernetes Manifests via service message
+#	Accepts 2 arguments:
+#	  string: the namespace
+#	  string: the manifest
+# -----------------------------------------------------------------------------
+function report_kubernetes_manifest 
+{
+  MANIFEST="$1"
+  NAMESPACE="$2"
+  
+  
+  LINES=()
+
+	# Read lines from file into the array
+	while IFS= read -r LINE; do
+		LINES+=("$LINE")
+	done <<< "$MANIFEST"
+
+  MANIFESTS=()
+  
+  CURRENT=""
+	for LINE in "${LINES[@]}"; do
+		if [ "$LINE" = "---" ];
+		then
+			MANIFESTS+=("$CURRENT")
+			CURRENT=""
+		else
+			CURRENT="$CURRENT$LINE\n"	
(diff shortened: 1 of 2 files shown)


source/Calamari.Common/Features/Scripting/Bash/Bootstrap.sh
```diff
@@ -307,6 +307,70 @@ function decrypt_and_parse_variables {
...
+  
+  CURRENT=""
+	for LINE in "${LINES[@]}"; do
+		if [ "$LINE" = "---" ]; then
+			MANIFESTS+=("$CURRENT")
+			CURRENT=""
+		else
+			CURRENT="$CURRENT$LINE\n"	
+		fi
+  done
```
APErebus, line 338: If someone has a better way to split a variable on a specific set of chars (or even a regex), please let me know, my bash is pretty weak

source/Calamari.Tests/Fixtures/Bash/Scripts/report-kubernetes-manifest.sh
```diff
@@ -0,0 +1,19 @@
...
+\"labels\":
+    \"name\": \"example\"
+---    
+\"apiVersion\": \"v1\"
+\"kind\": \"Namespace\"
+\"metadata\":
+  \"name\": \"diffs\"
+\"labels\":
+    \"name\": \"diffs\"
+"""
```
eddymoulton, line 16: Using Heredoc can clean up those escapes (disclaimer, I didn't test this)

```
read -r -d '' manifest << EOM
"apiVersion": "v1"
"kind": "Namespace"
"metadata":
  "name": "example"
"labels":
    "name": "example"
---    
"apiVersion": "v1"
"kind": "Namespace"
"metadata":
  "name": "diffs"
"labels":
    "name": "diffs"
EOM
```

source/Calamari.Common/Features/Scripting/Bash/Bootstrap.sh
```diff
@@ -307,6 +307,70 @@ function decrypt_and_parse_variables {
...
+      MANIFESTS+=("$CURRENT")
+      CURRENT=""
+    else
+      CURRENT="$CURRENT$LINE\n"	
+    fi
+  done
+    
+  MANIFESTS+=("$CURRENT")
+  
+  for i in "${MANIFESTS[@]}"; do
```
eddymoulton, line 342: naming nit since it's not an index

```suggestion
  for manifest in "${MANIFESTS[@]}"; do
```

source/Calamari.Common/Features/Scripting/WindowsPowerShell/Bootstrap.ps1
eddymoulton, line 1: nit for next time: consider applying the formatting in a separate commit before the change

eddymoulton (approved):
Changes look good to me, just a few nits for your consideration.

Love to see the testing - it's got me wondering if we should add tests for some of the poorly crafted yaml scenarios you identified in the PR description (eg. extra spaces before document separators, excess newlines, double document separators without content between, etc.). 
I'd be content knowing you've tested these manually if you think adding them as test cases is overkill.

source/Calamari.Common/Features/Scripting/WindowsPowerShell/Bootstrap.ps1
APErebus: yeah, I didn't mean to re-format but it did it :shakes-fist:

source/Calamari.Tests/Fixtures/Bash/Scripts/report-kubernetes-manifest.sh
APErebus: I might leave it as it's only a test script

source/Calamari.Common/Features/Scripting/Bash/Bootstrap.sh
APErebus: Fixed in [c97b400](https://github.com/OctopusDeploy/Calamari/pull/1474/commits/c97b40030024beab06c3eef03b273687266986da)

PR #1475, opened 2025-04-19
Support overriding poll period used in Cloud Formation Templates

zentron:
This PR provides a new variable `Octopus.Action.Aws.CloudFormationPollSeconds`  which will override the default period that we poll Cloud Formation when checking for changes.

This was introduced to resolve a customer issue where many concurrent projects are ocassionally causing throttling issues with AWS ([Internal Ticket](https://app.shortcut.com/octopusdeploy/story/108210/make-the-deploy-an-aws-cloudformation-template-polling-interval-configurable#activity-108218))

At the moment to keep effort down, the intention is to consider this variable something that isnt explicitly exposed on the step. We can revisit this if more customers have this requirement

### Alternative
I had originally opted for a backoff mechanism to avoid throttling but it is unclear what the correct value should be. For some customers such as this they may be hitting it so frequently that a small backoff does not help. On the other hand we dont want to penalize the majority of deployments where a couple of quick poll periods are all that is needed to continue.

Ideally we would be able to interrogate the response back from AWS to control the backoff period, however these values do not seem to be exposed by the APIs used, and re-engineering this whole process is considered excessive for this, currently single, customer request.


source/Calamari.Aws/Deployment/Conventions/DeleteCloudFormationStackConvention.cs
```diff
@@ -50,6 +41,7 @@ public override void Install(RunningDeployment deployment)
 
         private async Task InstallAsync(RunningDeployment deployment)
         {
+            deployment.Variables.Get("");
```
flin-8, line 44: What's this?

source/Calamari.Aws/Deployment/Conventions/ExecuteCloudFormationChangeSetConvention.cs
```diff
@@ -69,22 +69,23 @@ await clientFactory().DeleteChangeSetAsync(new DeleteChangeSetRequest
                 return;
             }
 
-            await ExecuteChangeset(clientFactory, stack, changeSet);
+            var timeout = PollPeriod(deployment);
```
flin-8, line 72: `timeout` -> `pollPeriod`

flin-8 (approved):
LGTM - just a variable rename and remove unused code?

source/Calamari.Aws/Deployment/Conventions/DeleteCloudFormationStackConvention.cs
zentron: Nice catch :) artifact from some testing

PR #1476, opened 2025-04-19
Move away from the statis `Log` class

zentron:

### DIFF
--- a/source/Calamari.Aws/Deployment/Conventions/CloudFormationInstallationConventionBase.cs
+++ b/source/Calamari.Aws/Deployment/Conventions/CloudFormationInstallationConventionBase.cs
@@ -1,6 +1,4 @@
 ﻿using System;
-using System.Collections.Generic;
-using System.Linq;
 using System.Threading.Tasks;
 using Amazon.CloudFormation;
 using Amazon.CloudFormation.Model;
@@ -10,9 +8,7 @@ using Calamari.Aws.Integration.CloudFormation;
 using Calamari.Common.Commands;
 using Calamari.Common.Plumbing.Logging;
 using Calamari.Common.Plumbing.Variables;
-using Calamari.Deployment;
 using Calamari.Deployment.Conventions;
-using Calamari.Integration.Processes;
 using Octopus.CoreUtilities;
 using Octopus.CoreUtilities.Extensions;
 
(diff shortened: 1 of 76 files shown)


source/Calamari.Common/CalamariFlavourProgram.cs
```diff
@@ -27,11 +27,11 @@ namespace Calamari.Common
 {
     public abstract class CalamariFlavourProgram
     {
-        protected readonly ILog Log;
+        protected readonly ILog log;
```
flin-8, line 30: this can be private now

source/Calamari.Tests/KubernetesFixtures/HelmUpgradeFixture.cs
```diff
@@ -95,7 +95,7 @@ public virtual void SetUp()
             Environment.SetEnvironmentVariable("TentacleJournal",
                                                Path.Combine(StagingDirectory, "DeploymentJournal.xml"));
 
-            Variables = new VariablesFactory(FileSystem).Create(new CommonOptions("test"));
+            Variables = new VariablesFactory(FileSystem, new SilentLog()).Create(new CommonOptions("test"));
```
flin-8, line 98: Is it worth passing in `ConsoleLog` instead of `SilentLog`? Would that get us more useful output on a test failure?

source/Calamari.Aws/Deployment/Conventions/CloudFormationInstallationConventionBase.cs
```diff
@@ -10,21 +8,21 @@
...
 using Octopus.CoreUtilities;
 using Octopus.CoreUtilities.Extensions;
 
 namespace Calamari.Aws.Deployment.Conventions
 {
     public abstract class CloudFormationInstallationConventionBase: IInstallConvention
     {
-        protected readonly StackEventLogger Logger;
+        protected readonly StackEventLogger StackEventLogger;
+        protected readonly ILog log;
```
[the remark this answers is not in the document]
zentron: Yeah i think this was part of the find and replace. It could be confusing but ill just lean into the convention for now.

source/Calamari.Tests/KubernetesFixtures/HelmUpgradeFixture.cs
zentron: these logs would have been previously just consumed, so to avoid having to rethink about these tests im leaving behavior as-is for now.

PR #1481, opened 2025-04-30
Add retry for failed Kubernetes status check 

octopus-hideaki:
# Background
After a deployment to a Kubernetes cluster, we poll the Kubernetes control plane using the kubectl CLI. The response from the CLI is used to update the state. 

Currently, there is no error handling if any of the responses from the CLI cannot be parsed. This will result in the deployment failing, even though the deployment task might succeed.

This is especially the case for the first status check, where the resource might not be created yet, and the CLI command will return an error. 
# Results
This PR adds better error handling for failed CLI commands and parsing. Instead of throwing immediately, the error is returned as a result. 

The deployment will only fail if all retries of the status check for a resource fail to verify that the resource was updated successfully. 

Fixes: https://github.com/OctopusDeploy/Issues/issues/9360
[sc-107940]


mik-ky (approved):

APErebus (approved):
LGTM! Thanks :)

PR #1482, opened 2025-05-01
Pass in logger when constructing the java step

hnrkndrssn:
An earlier [change](https://github.com/OctopusDeploy/Calamari/pull/1476) that updated the constructors for the java step types missed passing in the new log parameter.


benPearce1 (approved):
👍 That is a hard one to catch. No type safety.

zentron (approved):
Thanks for the find!

PR #1483, opened 2025-05-07
Rework the Helm manifest polling cancellation logic

APErebus:

If the helm install command fails, we continue to attempt to retrieve the manifest until the configured timeout (or 5min) elapses.

This PR adds another CancellationTokenSource that tracks if the helm command fails. If the command fails, it cancels the manifest retrieval polling (if it's still running)


kevjt (approved):
LGTM

PR #1484, opened 2025-05-13
Update script bootstrappers to retain underscores

wlthomson:
[sc-106834]

Fixes https://github.com/OctopusDeploy/Issues/issues/9324.

When a script modules is loaded from source, any underscores are stripped. This can cause resolution issues when the source is imported into a deployment, e.g.

```
Writing script module 'Test_Module' as bash script TestModule.sh. Import this via `source TestModule.sh`.

/home/Octopus/Work/88a3gaHB9km9FHqAHM8Q/Script.sh: line 1: Test_Module.sh: No such file or directory
```

These changes update _all_ script bootstrappers to no longer omit underscores when formatting script module names.


wlthomson:
Keeping in draft for now.

Have reached out to SaST to double check that these changes do not break any intended functionality ([slack thread](https://octopusdeploy.slack.com/archives/C04GAH78491/p1747109525725659)).

source/Calamari.Common/Plumbing/Variables/ScriptVariables.cs
```diff
@@ -26,6 +26,11 @@ public static string GetLibraryScriptModuleName(string variableName)
         {
             return variableName.Replace("Octopus.Script.Module[", "").TrimEnd(']');
         }
+        
+        public static string FormatScriptName(string scriptName)
+        {
+            return new string(scriptName.Where(c => char.IsLetterOrDigit(c) || c == '_').ToArray());
```
APErebus, line 32: is there any other "normal" chars we should support? `-` for instance?

wlthomson: That's the kind of thing I'm trying to understand as I think that largely depends on intended behaviour and why we originally only limited this to letters/digits. Its one of those [Chesterton's Fence](https://fs.blog/chestertons-fence/) situations.

If this is only an arbitrary limitation, arguably we should include `-`, but then theres the question of whitespace and other characters (don't quote me on it, but I believe `!@#$%^&()` are all valid on Win/Unix systems).

APErebus (approved):
This looks good to me!

PR #1485, opened 2025-05-13
Catch YAML parse errors when parsing Helm manifests

sathvikkumar-octo:

[sc-109895]

While deploying certain Helm charts we would get failures in the step when parsing the manifest due to invalid YAML syntax. However this syntax is not invalid in all cases and marks the deployment as failed when Helm has succeeded.

This fix extends our existing approach to catch these YAML exceptions and log them for now. A longer term fix will be to allow YamlDotNet to successfully parse this valid YAML.

Have tested this locally with deploying a Helm chart with a ConfigMap that fails YamlDotNet parsing

As part of this change I've refactored some very similar manifest parsing code into the shared `ManifestParser` class.


eddymoulton (commented):
Looking good!

Would you mind
- Linking the shortcut card in the PR description + provide a sentence or two for the background of why we're making the change
- Adding a test to cover the case that YamlDotNet is failing to parse, ensuring we don't get an exception for it (I don't think it's there in existing tests?)

There is also a CLA you'll need to sign for this as it's a public repo

eddymoulton (approved):
Nice! Thanks for the extra test 🙇

PR #1487, opened 2025-05-26
Fix issue downloading tools for K8s tests

APErebus:
It appears that Cloudfront, which sits in-front of the hashicorp checkpoint API (what we use to work out what version of Terraform to download), requires a user agent string in the HTTP call.

We are now Chrome


flin-8 (approved):
😂

PR #1489, opened 2025-05-26
Update BashFixture tests to remove line ending for windows tests

IsaacCalligeros95:
The new Bash Associative array changes run tests across windows and linux. When running bash on windows with a `\n` line return this is returned in test logs as a CRLF. This is a known potential issue that I'm still working through, this change just prevents test failures in the mean time.


APErebus (approved):
Thanks

PR #1491, opened 2025-05-27
fix: return complete result set in GetResourcesByType() 

HuyPhanNguyen:
## Background
Customer is using Cloud Target Discovery and found their ServicePrincipal wouldn't return all the WebApps in their Azure Account and would cut off after 1000 results so some targets weren't being included in the Deployment.

## Result
[sc-111441](https://app.shortcut.com/octopusdeploy/story/111441/sev-3-cloud-target-discovery-returned-web-apps-are-limited-to-1000-requested-by-finnian-dempsey)

## Before
Only return first 1000 result

## After 
Return all result

### DIFF
--- a/source/Calamari.AzureAppService/Behaviors/TargetDiscoveryBehaviour.cs
+++ b/source/Calamari.AzureAppService/Behaviors/TargetDiscoveryBehaviour.cs
@@ -76,7 +76,7 @@ namespace Calamari.AzureAppService.Behaviors
             });
             try
             {
-                var resources = await armClient.GetResourcesByType( WebAppType, WebAppSlotsType);
+                var resources = await armClient.GetResourcesByType(WebAppType, WebAppSlotsType);
                 var discoveredTargetCount = 0;
                 Log.Verbose($"Found {resources.Length} candidate web app resources.");
                 foreach (var resource in resources)


source/Calamari.AzureAppService/Behaviors/TargetDiscoveryBehaviour.cs
```diff
@@ -210,21 +210,35 @@ public static ServiceMessage CreateWebAppTargetCreationServiceMessage(string? re
...
+            {
+                var page = (await tenant.GetResourcesAsync(query, CancellationToken.None)).Value;
+                var batch = JsonConvert.DeserializeObject<AzureResource[]>(page.Data.ToString())!;
+                allResults.AddRange(batch);
+
+                if (string.IsNullOrEmpty(page.SkipToken))
+                    break;
+
+                query.Options.SkipToken = page.SkipToken;
+            }
```
benPearce1, line 240: I'm sure this will probably just work, but relying on external data as our exit condition is risky.
I would suggest a maximum loop counter
```suggestion
           var attempt = 0
           var maxAttempts = 10
            while (attempt++ < maxAttempts)
            {
                var page = (await tenant.GetResourcesAsync(query, CancellationToken.None)).Value;
                var batch = JsonConvert.DeserializeObject<AzureResource[]>(page.Data.ToString())!;
                allResults.AddRange(batch);

                if (string.IsNullOrEmpty(page.SkipToken))
                    break;

                query.Options.SkipToken = page.SkipToken;
            }
```

source/Calamari.AzureAppService/Behaviors/TargetDiscoveryBehaviour.cs
```diff
@@ -76,7 +76,7 @@ public async Task Execute(RunningDeployment runningDeployment)
             });
             try
             {
-                var resources = await armClient.GetResourcesByType(WebAppType, WebAppSlotsType);
+                var resources = await armClient.GetResourcesByType( WebAppType, WebAppSlotsType);
```
hnrkndrssn, line 79:
```suggestion
                var resources = await armClient.GetResourcesByType(WebAppType, WebAppSlotsType);
```

hnrkndrssn (approved):
This looks good to me 👍

benPearce1 (approved):

PR #1493, opened 2025-05-27
Do not strip spaces from script modules

wlthomson:
Fixes https://github.com/OctopusDeploy/Issues/issues/9428.


source/Calamari.Common/Plumbing/Variables/ScriptVariables.cs
```diff
@@ -31,7 +31,7 @@ public static string GetLibraryScriptModuleName(string variableName)
         public static string FormatScriptNameForPhysicalFilesystem(string scriptName)
         {
             var fileSystem = CalamariPhysicalFileSystem.GetPhysicalFileSystem();
-            return fileSystem.RemoveInvalidFileNameChars(scriptName); 
+            return fileSystem.RemoveInvalidFileNameChars(scriptName).Replace(" ", "");
```
benPearce1, line 34: Do we expect there could be spaces in any place in the name, or only at the end?
Should it be `.TrimEnd()`?

PR #1497, opened 2025-05-30
Calamari Parallelise Build, Restore Publish & Package

IsaacCalligeros95:
This PR brings the average Build & Compile time for Calamari down from ~21 minutes to ~14 minutes. About half of this improvement is from using the PackInParallel behaviour by default and the other half is from parallelising parts of Restore, Build and Publish for Calamari Flavours. In general the issues I ran into are summarised [here](https://www.damirscorner.com/blog/posts/20241122-ParallelBuildsForDotNetProjects.html), we can't just make the DotnetPublish with Restore & Build enabled parallel, even on a per project or framework basis. This change separates out the Build, Restore and Publish stages and makes them parallel where possible. Because there's a lot of File IO operations, this has limits at a relatively low level of parallelism, in general 3/4 processes seems to give the best run time.
[Server PR with E2E tests using Calamari Consolidated from this PR](https://github.com/OctopusDeploy/OctopusDeploy/pull/33716)

## Build times
image

## Bundles are reconstructed as expected in Server
image

## Index.json architectures for each calamari flavour are the same
image

### DIFF
--- a/build/Build.cs
+++ b/build/Build.cs
@@ -320,7 +320,9 @@ namespace Calamari.Build
 
                                                                              var projectName = calamariPackageMetadata.Project?.Name ?? throw new Exception("Could not find project name");
                                                                              var projectSemaphore = semaphores.GetOrAdd(projectName, _ => new SemaphoreSlim(1, 1));
-                                                                             var architectureSemaphore = semaphores.GetOrAdd(calamariPackageMetadata.Architecture ?? throw new Exception($"Unknown Architecture {calamariPackageMetadata.Project?.Name}"), _ => new SemaphoreSlim(1, 1));
+
+                                                                             // for NetFx target frameworks, we use "netfx" as the architecture, and ignore defined runtime identifiers, here we'll just block on all netfx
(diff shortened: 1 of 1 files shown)


build/Build.cs
```diff
@@ -246,90 +245,173 @@ public Build()
...
+                                                                             if (!OperatingSystem.IsWindows() && !calamariPackageMetadata.IsCrossPlatform)
+                                                                             {
+                                                                                 Log.Warning($"Not Building {calamariPackageMetadata.Framework}: can only publish netfx on a Windows OS");
+                                                                                 return;
+                                                                             }
+
+                                                                             var projectName = calamariPackageMetadata.Project?.Name ?? "UnknownProject";
+                                                                             var projectSemaphore = semaphores.GetOrAdd(projectName, _ => new SemaphoreSlim(1, 1));
+                                                                             var architectureSemaphore = semaphores.GetOrAdd(calamariPackageMetadata.Architecture ?? "UnknownArchitecture", _ => new SemaphoreSlim(1, 1));
```
APErebus, line 323: Same here...

APErebus (approved):
This all looks good to me!
I appreciate the validation you've done to make sure it's producing the same packages.

PR #1500, opened 2025-06-03
Fix package cache retention message

octopus-hideaki:
Fix "no pacakge will be deleted" message when keeping all packages

Related: https://github.com/OctopusDeploy/Issues/issues/9421
[sc-112263]


mik-ky (approved):
🌮

PR #1502, opened 2025-06-10
Removed unneeded reference to `Octopus.Data`

droyad:
This seems like some detritus copied over when this was merged together. Likely the reference was added by accident. That it's only for one of the two frameworks also points to it being an accident.

`Octopus.Data` contains `Result` and storage interfaces, plus a reference to `Octopus.Server.MessageContracts`. Nothing that indicates some kind of hidden usage in AzureAppService.

This is one of the last usages of `Octopus.Data`, which would allow us to axe that project.

### DIFF
--- a/build/Build.cs
+++ b/build/Build.cs
@@ -67,6 +67,7 @@ namespace Calamari.Build
         [Parameter] readonly string? TargetRuntime;
 
         const string CiBranchNameEnvVariable = "OCTOVERSION_CurrentBranch";
+
         [Parameter($"The name of the current git branch. OctoVersion will use this to calculate the version number. This can be set via the environment variable {CiBranchNameEnvVariable}.", Name = CiBranchNameEnvVariable)]
         string? BranchName { get; set; }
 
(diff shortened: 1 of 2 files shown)


droyad:
> It looks like removing the ref has changed the build order and we've run into a parallel build issue

image

enf0rc3 (approved):
I am not familiar with this area, But in terms of `Octopus.Data` I am 100% on board!

In regards to the other changes, if we are getting issues running in parallel it makes sense to bring them all into a single action and run them synchronously. 

If this won't cause any major performance issues by making it synchronous, happy to merge.

droyad:
@APErebus It looks like the [other branch is doing the same](https://build.octopushq.com/buildConfiguration/TeamModernDeployments_Calamari_VNext_Build_BuildCompile/17586315)?

[Kicked off a build on main to check](https://build.octopushq.com/buildConfiguration/TeamModernDeployments_Calamari_VNext_Chains_ChainFull/17627742)

droyad:
[Looks like it fails intermittently in the past](https://build.octopushq.com/buildConfiguration/TeamModernDeployments_Calamari_VNext_Build_BuildCompile)

enf0rc3:
@droyad I am happy with this from BFT's perspective. I guess you would need final approvals from team-modern-deployments

PR #1505, opened 2025-07-09
Improve deployment status detection to handle progress deadline exceeded scenarios

octonautcal:
### Context

https://app.shortcut.com/octopusdeploy/story/104135/kos-doesn-t-correctly-mark-a-deployment-resource-as-failed-when-the-progress-deadline-expires

The current deployment status logic uses a simple comparison of replica counts to determine if a deployment is successful or in progress. This approach fails to detect when a Kubernetes deployment has exceeded its progress deadline and is stuck in a failed state.
When a deployment encounters issues (e.g., image pull failures, resource constraints, or other blocking conditions), it may remain indefinitely in a state where the replica counts don't match the desired state, but the deployment is marked as "Progressing" with reason "ProgressDeadlineExceeded". The original logic would incorrectly report this as InProgress rather than Failed.

### Solution

Enhanced the deployment status detection logic to:

1. Check generation vs observedGeneration: Only evaluate status when the deployment controller has observed the current generation, ensuring we're not checking stale status information.
2. Detect progress deadline exceeded: Explicitly check for the "Progressing" condition with reason "ProgressDeadlineExceeded" and mark the deployment as Failed when this occurs.
3. Improved progress detection: More granular checks for in-progress states:

When desired replicas > 0 but not all replicas are up-to-date
When total replicas exceed up-to-date replicas (rollback/scaling scenarios)
When available replicas are less than up-to-date replicas (readiness issues)

4. Fallback for ongoing updates: Mark as InProgress when the deployment controller hasn't yet observed the current generation.

### Testing evidence

**Setup**

- Progression Deadline in seconds set to 1
- Kubernetes Object Status Check set to `Check that Kubernetes objects are running successfully
`
- Deploying an nginx container with version `doesnotexist` to intentionally trigger a failure when the pod attempts to pull it

**Before**

Even though the Progress Deadline of 1 second has been exceeded, Calamari still hangs since the it still thinks that the Deployment is still "Progressing"

https://github.com/user-attachments/assets/bbda6f0f-f564-458b-a2fd-97c87e5e3fcd
Screenshot 2025-07-09 at 11 21 57 am

**After**

Once the Progress Deadline of 1 second has been exceeded, Calamari fails the deployment because it can now correctly identify that the Deployment should be failed.

https://github.com/user-attachments/assets/b8271382-4eb5-44be-8776-8177adae4785

### DIFF
--- a/source/Calamari/Kubernetes/ResourceStatus/Resources/Deployment.cs
+++ b/source/Calamari/Kubernetes/ResourceStatus/Resources/Deployment.cs
@@ -32,6 +32,7 @@ namespace Calamari.Kubernetes.ResourceStatus.Resources
             var generation = FieldOrDefault("$.metadata.generation", 0);
             var observedGeneration = FieldOrDefault("$.status.observedGeneration", 0);
             
+            // Note that deployment status logic aligns with ArgoCD's gitops-engine, which is also used in the Kubernetes Monitor
             if (generation <= observedGeneration)
             {
                 var conditions = json.SelectToken("$.status.conditions") as JArray;
--- a/source/Calamari/Kubernetes/ResourceStatus/Resources/Deployment.cs
+++ b/source/Calamari/Kubernetes/ResourceStatus/Resources/Deployment.cs
@@ -29,12 +29,39 @@ namespace Calamari.Kubernetes.ResourceStatus.Resources
             Available = FieldOrDefault("$.status.availableReplicas", 0);
             UpToDate = FieldOrDefault("$.status.updatedReplicas", 0);
             
-            ResourceStatus = TotalReplicas == Desired
-                             && UpToDate == Desired
-                             && Available == Desired 
-                             && ReadyReplicas == Desired
-                ? ResourceStatus.Successful 
-                : ResourceStatus.InProgress;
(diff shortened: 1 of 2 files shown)


source/Calamari/Kubernetes/ResourceStatus/Resources/Deployment.cs
```diff
@@ -29,12 +29,39 @@ public Deployment(JObject json, Options options) : base(json, options)
...
+                }
+                else
+                {
+                    ResourceStatus = ResourceStatus.Successful;
+                }
+            }
+            else
+            {
+                ResourceStatus = ResourceStatus.InProgress;
+            }
```
eddymoulton, line 64: Did we pull this logic from the gitops-engine?

If so, worth mentioning that and linking to where it came from in a comment so the next person coming along doesn't change it without consideration.
You can mention that we're aligning it with the Kubernetes monitor logic too as justification

octonautcal: updated now in [c3034bb](https://github.com/OctopusDeploy/Calamari/pull/1505/commits/c3034bbdcb9e74fa25d2f4d5a9090bb03e57806b)

PR #1511, opened 2025-07-14
Don't report helm manifests when dry run is enabled

eddymoulton:
[sc-115648]

When Helm dry run is enabled, we never make a real update to the cluster. This ends up having Calamari attempting to fetch an updated revision that doesn't exist - eventually timing out because it will never be created.

If users are validating their Helm applications with dry run, then the features we have created that leverage the reported manifests aren't super useful.
There is a case to be made that they may want to see the (soon to be) applied manifests, however I think we are safe to wait for customer feedback on this as the number of customers using dry run must be low since this is the first time we've heard about this issue.

I've tried to write this in a way that could be built upon if there are other reasons we want to skip the manifest reporting. I considered moving the `--dry-run` flag to be a configurable option in the step, however, again, it seems like no one is really asking for that. Instead we're left with checking for the existance of `--dry-run` in the additional args.

Part of the reason this got through is becuase we are still testing a lot without KOS for helm enabled - I've just enabled it for all tests now since we default the feature toggle to on.

### DIFF
--- a/source/Calamari.Tests/KubernetesFixtures/HelmUpgradeFixture.cs
+++ b/source/Calamari.Tests/KubernetesFixtures/HelmUpgradeFixture.cs
@@ -8,6 +8,7 @@ using Calamari.Common.Features.Deployment;
 using Calamari.Common.Features.Packages;
 using Calamari.Common.Features.Processes;
 using Calamari.Common.Features.Scripts;
+using Calamari.Common.FeatureToggles;
 using Calamari.Common.Plumbing;
 using Calamari.Common.Plumbing.Commands;
 using Calamari.Common.Plumbing.FileSystem;
@@ -21,7 +22,6 @@ using Calamari.Tests.Helpers;
 using Calamari.Util;
 using FluentAssertions;
 using NUnit.Framework;
-using Octopus.Versioning.Semver;
 
 namespace Calamari.Tests.KubernetesFixtures
 {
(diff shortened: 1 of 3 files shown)


source/Calamari.Tests/KubernetesFixtures/HelmUpgradeFixture.cs
```diff
@@ -112,6 +112,9 @@ public virtual void SetUp()
             Variables.Set(SpecialVariables.Account.AccountType, "Token");
             Variables.Set(SpecialVariables.Account.Token, ClusterToken);
 
+            // KOSForHelm is on by default
+            Variables.Set(KnownVariables.EnabledFeatureToggles, Variables.Get(KnownVariables.EnabledFeatureToggles) + "," + OctopusFeatureToggles.KnownSlugs.KOSForHelm);
```
octonautcal, line 116: do these variables ever need to be null checked or are they fairly safe?

i.e.
```
var currentToggles = Variables.Get(KnownVariables.EnabledFeatureToggles);
var newToggle = OctopusFeatureToggles.KnownSlugs.KOSForHelm;

Variables.Set(
    KnownVariables.EnabledFeatureToggles,
    string.IsNullOrEmpty(currentToggles) ? newToggle : currentToggles + "," + newToggle
);
```

eddymoulton: Technically this is probably a good idea, but practically overkill for a test imo.

The way we check feature toggles in Calamari is to check if the slug exists in the `KnownVariables.EnabledFeatureToggles` variable - so we can put any old random stuff in there, so long as the feature toggles are named in some fashion.

octonautcal: ok got it

PR #1512, opened 2025-07-15
Fix Bash Create Artifact tests

APErebus:
We now have a second mac build agent `cloudmac1`, which has a different build agent path

`cloudmac1` -> `/opt/TeamCity/BuildAgent1/work/62728692c7c35200/Binaries/subdir/anotherdir/myfile`
`cloudmac` -> `/opt/TeamCity/BuildAgent/work/62728692c7c35200/Binaries/subdir/anotherdir/myfile`

This is a problem because this path was hardcoded into the test, causing a failure on this build agent.

The test now does a regex check for the service message without caring about the contents of the path. We then extract the path and decode the base64 path and checks the path _ends_ with the expected path

### DIFF
--- a/source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureEks.cs
+++ b/source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureEks.cs
@@ -134,6 +134,7 @@ namespace Calamari.Tests.KubernetesFixtures
         [Test]
         [TestCase(true)]
         [TestCase(false)]
+        [Ignore("These tests will be transitioned to using a local Kind cluster, rather than a remote cluster")]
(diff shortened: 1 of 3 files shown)


flin-8 (approved):
LGTM

source/Calamari.Tests/KubernetesFixtures/KubernetesContextScriptWrapperLiveFixtureEks.cs
```diff
@@ -134,6 +134,7 @@ protected override async Task<Dictionary<string, string>> GetEnvironmentVars(Can
         [Test]
         [TestCase(true)]
         [TestCase(false)]
+        [Ignore("These tests will be transitioned to using a local Kind cluster, rather than a remote cluster")]
```
APErebus, line 137: After discussing with @eddymoulton, we realized these tests were running deployments in a single namespace, meaning there was conflicts between test runs.

These actual deployment tests should be run against a local, isolated cluster (such as a Kind cluster). I'll do this work in another PR

PR #1514, opened 2025-07-16
Execute app service tests against static app service plans

eddymoulton:


APErebus:
Closing this as it turns out this slows down the test process _massively_ (on the order of an additional hour).

We are going to implement an automated cleanup mechanism, as opposed to worrying about these static resources

PR #1521, opened 2025-07-18
Add Azure test subscription cleanup script

eddymoulton:
When running tests against Azure we cannot reliably ensure that the cleanup task runs (tests may be cancelled, etc.) so we end up with resources hanging around costing money.

These scripts replicate what we do in our sandbox environment, but instead explictly target calamari resources in the automated test subscription.

We'll set this up to run daily via a Runbook.


scripts/azure-test-subscription-cleanup/Invoke-SandboxCleanup.Tests.ps1
```diff
@@ -0,0 +1,159 @@
...
+    Import-Module (Join-Path $PSScriptRoot "helpers/cleanup.psm1")
+
+    Remove-Module "logging" -Force -ErrorAction SilentlyContinue
+    Import-Module (Join-Path $PSScriptRoot "helpers/logging.psm1")
+
+    Remove-Module "octopus" -Force -ErrorAction SilentlyContinue
+    Import-Module (Join-Path $PSScriptRoot "helpers/octopus.psm1")
+
+    Remove-Module "slack" -Force -ErrorAction SilentlyContinue
+    Import-Module (Join-Path $PSScriptRoot "helpers/slack.psm1")
```
APErebus, line 15: delete?

scripts/azure-test-subscription-cleanup/Invoke-SandboxCleanup.ps1
```diff
@@ -0,0 +1,60 @@
+Import-Module "${PSScriptRoot}\helpers\azure.psm1"
+Import-Module "${PSScriptRoot}\helpers\logging.psm1"
+Import-Module "${PSScriptRoot}\helpers\octopus.psm1"
+Import-Module "${PSScriptRoot}\helpers\slack.psm1"
```
APErebus, line 4: Remove?

APErebus (approved):
I'm wondering if we should remove/change the use of "Sandbox". It's a bit confusing.

PR #1522, opened 2025-07-20
Update static EKS + GKE cluster versions

eddymoulton:

### DIFF
--- a/source/Calamari.Tests/KubernetesFixtures/Terraform/StaticClusters/EKS/eks.kubernetes.tf
+++ b/source/Calamari.Tests/KubernetesFixtures/Terraform/StaticClusters/EKS/eks.kubernetes.tf
@@ -1,9 +1,9 @@
 data "aws_eks_cluster" "default" {
-  name = aws_eks_cluster.default.name
+  name = "${var.static_resource_prefix}-eks"
 }
 
 data "aws_eks_cluster_auth" "default" {
-  name = aws_eks_cluster.default.name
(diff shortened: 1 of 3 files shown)


source/Calamari.Tests/KubernetesFixtures/Terraform/StaticClusters/EKS/eks.kubernetes.tf
```diff
@@ -1,9 +1,9 @@
 data "aws_eks_cluster" "default" {
-  name = aws_eks_cluster.default.name
+  name = "${var.static_resource_prefix}-eks"
```
APErebus, line 2: hehe, that seems like it shouldn't work 😂

APErebus (approved):

source/Calamari.Tests/KubernetesFixtures/Terraform/StaticClusters/EKS/eks.kubernetes.tf
eddymoulton: @APErebus Liam pointed out the issue - essentially the provider was relying on a data resource, which relied on the cluster resource. But the provider needs to be evaluated first (before anything else get's created).

Setting the name statically gets us around this because the cluster does actually exist and can be referenced - but this is a bit of a hack that means we'll need to do some manual applies if we need to bootstrap a new cluster.

If we need to change this in the future, modules are likely the way to go.

PR #1526, opened 2025-07-28
Add test for module name __main__

wlthomson:
[sc-116319].

Follow up to https://github.com/OctopusDeploy/Calamari/pull/1517.


source/Calamari.Tests/Fixtures/Python/Scripts/checkname.py
```diff
@@ -0,0 +1,4 @@
+if __name__ == "__main__":
+    print("__name__ is __main__")
+else:
+    print("__name__ is {}".format(__name__))
```
benPearce1, line 4: just a thought, this could do a non-zero exit here too

benPearce1 (approved):

PR #1532, opened 2025-08-11
Variables are now loaded from file in a new format

APErebus:
As part of work for the Argo CD integration, we need to know if a variable is sensitive or not.
To do this, we need a new format for transferring the variables from Octopus Server to Calamari.

The existing format is to serialize a Octostache `VariableDictionary` (which is `Dictionary<string,string>`).

The new data format is a list/collection of `CalamariExecutionVariable`s, which have a `Key`, `Value`, and if it's sensitive or not.

We also change the calamari command parameter from `sensitiveVariables` to just plain `variables` as we no longer support non-sensitive variables. I also adjusted the `sensitiveVariablesPassword` to just `variablesPassword` for consistency.

UPDATE: Due to how the V2 execution pipeline runs, it's possible that there is cached calamari command that has the deprecated `sensitiveVariables` and `sensitiveVariablesPassword` arguments. I've re-added these and it just hooks into the same process.

Related Server PR: https://github.com/OctopusDeploy/OctopusDeploy/pull/35662


APErebus:
^ This is a testing secret

PR #1539, opened 2025-08-21
CommitToGit Step errors

rain-on:
From gold-path testing there was a few errors that were identified:

1. The Deployment _must_ have a staging directory set, or else the package-path gets added to the path multiple times
2. Did not handle the idea of "HEAD" - this is now resolved - oops.

### DIFF
--- a/source/Calamari/ArgoCD/Git/RepositoryWrapper.cs
+++ b/source/Calamari/ArgoCD/Git/RepositoryWrapper.cs
@@ -50,7 +50,7 @@ namespace Calamari.ArgoCD.Git
             }
         }
         
-        public void PushChanges(bool requiresPullRequest, string branchName)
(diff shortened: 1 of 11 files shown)


source/Calamari/ArgoCD/Git/RepositoryWrapper.cs
```diff
@@ -67,6 +67,7 @@ public void PushChanges(bool requiresPullRequest, string branchName)
         
         public void PushChanges(string branchName)
         {
+            Log.Info($"Literally pushing to {branchName}");
```
flin-8, line 70: You're logging this?

flin-8: "Literally pushing ..."

PR #1544, opened 2025-08-29
Remove Manifest Inspection and Live Object Status feature toggles

APErebus:

The manifest inspection feature toggle has been enabled in cloud & defaulted to `true` in Server, so can be removed and the code defaulted to use the new path.

As we now _always_ report manifests, the LiveObjectStatus feature toggle is also now not required, so removed.

### DIFF
--- a/source/Calamari.Tests/KubernetesFixtures/ManifestReporterTests.cs
+++ b/source/Calamari.Tests/KubernetesFixtures/ManifestReporterTests.cs
@@ -20,11 +20,10 @@ namespace Calamari.Tests.KubernetesFixtures
     public class ManifestReporterTests
     {
         [Test]
-        public void GivenValidYaml_ShouldPostSingleServiceMessage(string enabledFeatureToggle)
+        public void GivenValidYaml_ShouldPostSingleServiceMessage()
         {
             var memoryLog = new InMemoryLog();
             var variables = new CalamariVariables();
-            variables.Set(KnownVariables.EnabledFeatureToggles, enabledFeatureToggle);
 
             var namespaceResolver = Substitute.For<IKubernetesManifestNamespaceResolver>();
             namespaceResolver.ResolveNamespace(Arg.Any<YamlMappingNode>(), Arg.Any<IVariables>())
@@ -51,11 +50,10 @@ quoted_float: ""5.75""
         }
 
         [Test]
-        public void GivenInvalidManifest_ShouldNotPostServiceMessage(string enabledFeatureToggle)
+        public void GivenInvalidManifest_ShouldNotPostServiceMessage()
         {
             var memoryLog = new InMemoryLog();
             var variables = new CalamariVariables();
-            variables.Set(KnownVariables.EnabledFeatureToggles, enabledFeatureToggle);
 
             var namespaceResolver = Substitute.For<IKubernetesManifestNamespaceResolver>();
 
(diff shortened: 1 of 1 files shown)


source/Calamari/Kubernetes/Conventions/Helm/HelmManifestAndStatusReporter.cs
```diff
@@ -47,43 +47,37 @@ await Task.Run(async () =>
                            {
                                var resourceStatusCheckIsEnabled = deployment.Variables.GetFlag(SpecialVariables.ResourceStatusCheck);
 
-                               if (
-                                   resourceStatusCheckIsEnabled
-                                   || FeatureToggle.KubernetesLiveObjectStatusFeatureToggle.IsEnabled(deployment.Variables)
-                                   || OctopusFeatureToggles.KubernetesObjectManifestInspectionFeatureToggle.IsEnabled(deployment.Variables))
+                               if (!DeploymentSupportsManifestReporting(deployment, out var reason))
                                {
```
APErebus, line 54: We now _always_ report the manifest as MI is now always enabled, so this check isn't needed anymore

source/Calamari.Common/FeatureToggles/FeatureToggle.cs
```diff
@@ -8,7 +8,6 @@
     public enum FeatureToggle {
         SkunkworksFeatureToggle,
         OidcAccountsFeatureToggle,
-        KubernetesLiveObjectStatusFeatureToggle,
```
APErebus, line 11: this can also be removed as we now _always_ report manifests, no matter if LOS is enabled or not

source/Calamari/Kubernetes/ManifestReporter.cs
```diff
@@ -39,9 +39,6 @@ public ManifestReporter(IVariables variables, ICalamariFileSystem fileSystem, IL
 
         public void ReportManifestFileApplied(string filePath)
         {
-            if (!FeatureToggle.KubernetesLiveObjectStatusFeatureToggle.IsEnabled(variables)
-                && !OctopusFeatureToggles.KubernetesObjectManifestInspectionFeatureToggle.IsEnabled(variables))
-                return;
```
APErebus, line 44: as MI feature toggle is `true`, this is _always_ false, so the check can be removed

source/Calamari/Kubernetes/Conventions/Helm/HelmManifestAndStatusReporter.cs
```diff
@@ -47,43 +47,37 @@ await Task.Run(async () =>
                            {
                                var resourceStatusCheckIsEnabled = deployment.Variables.GetFlag(SpecialVariables.ResourceStatusCheck);
 
-                               if (
-                                   resourceStatusCheckIsEnabled
```
eddymoulton, line 51: I've actually looked at making a similar refactoring to this file before - but realised that removing this `resourceStatusCheckIsEnabled` condition means that we are prone to log out messages for all the other conditions that we check.

The verbose log (line 52) is probably much of a muchness, but the warning (line 58) might cause some frustration to customers.

It feels like yelling at people to use a release newer than 2 years old is probably ok at this point though.
I'm alright with risking a complaint or two if you are.

eddymoulton (approved):
LGTM - one side effect for you to consider before merging though.

source/Calamari/Kubernetes/Conventions/Helm/HelmManifestAndStatusReporter.cs
eddymoulton: Do we try and get them anywhere else aside from https://github.com/OctopusDeploy/Calamari/pull/1544/files#diff-93633c3f4bed8f073fc29d72d49a09cd76ce0175d358713855b9f7ff4000b1fdL67?

In the old version that doesn't run run if `resourceStatusCheckIsEnabled` is false

APErebus: It does because of the logical `||`

PR #1545, opened 2025-08-29
Frank/customproperties

flin-8:


source/Calamari.Tests/ArgoCD/Git/TestGitConnection.cs
```diff
@@ -4,9 +4,9 @@
 
 namespace Calamari.Tests.ArgoCD.Git
 {
-    public class GitConnection : IGitConnection
+    public class TestGitConnection : IGitConnection
```
rain-on, line 7: is this still needed, or should we just use the GitConnection struct from the prod code?

rain-on (approved):
minor comments:
* Loading the properties files outside of the intsallConvention should make testing easier
* Is the TestGitConnection needed - or can we just use the GitConnection in tests too?

PR #1549, opened 2025-09-03
CommitToGit can only use non-sensitive values in commit messages

rain-on:
It is possible that the user will try and build an octostache variable from sensitive-values (or just a directly sensitively value) - and use it as a commit summary or description - which would enshrine the sensitive value in the git repository (which would be bad).

This change adds the non-sensitive values to the RunningDeployment, such that the commit summary/message can be extracted from a guaranteed safe list of values.


source/Calamari/ArgoCD/Conventions/ArgoCommitToGitConfigFactory.cs
```diff
@@ -1,4 +1,5 @@
 using System.Collections.Generic;
+using Amazon.CloudFormation.Model;
```
flin-8, line 2: CloudFormation?!

rain-on: 🤷 gone now :)

source/Calamari.Common/Commands/RunningDeployment.cs
```diff
@@ -7,20 +7,27 @@ namespace Calamari.Common.Commands
 {
     public class RunningDeployment
     {
-        public RunningDeployment(IVariables variables, Dictionary<string, string>? environmentVariables = null) : this(
-            null, variables, environmentVariables)
+        public RunningDeployment(IVariables variables, INonSensitiveVariables? nonSensitiveVariables = null, Dictionary<string, string>? environmentVariables = null)
```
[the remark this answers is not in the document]
flin-8: I think this is good - makes it more explicit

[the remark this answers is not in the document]
rain-on: we like explicit :)

source/Calamari.Common/Commands/RunningDeployment.cs
```diff
@@ -7,20 +7,27 @@ namespace Calamari.Common.Commands
...
+                                 INonSensitiveVariables nonSensitiveVariables,
+                                 Dictionary<string, string>? environmentVariables = null)
         {
             if (!string.IsNullOrEmpty(packageFilePath))
             {
                 PackageFilePath = new PathToPackage(packageFilePath);
             }
 
             Variables = variables;
+            NonsensitiveVariables = nonSensitiveVariables ?? new NonSensitiveCalamariVariables();
```
flin-8, line 30: Probably won't need the null coalescing here

source/Calamari.Aws/Commands/ApplyCloudFormationChangesetCommand.cs
```diff
@@ -53,7 +53,7 @@ public override int Execute(string[] commandLineArguments)
                 new CloudFormationOutputsAsVariablesConvention(ClientFactory, stackEventLogger, StackProvider, log)
             };
             
-            var deployment = new RunningDeployment(packageFile, variables);
+            var deployment = new RunningDeployment(packageFile, variables, new NonSensitiveCalamariVariables());
```
APErebus, line 56: This seems like the wrong fix? Wouldn't we want to actually have it populated?

flin-8 (changes requested):
https://github.com/OctopusDeploy/Calamari/pull/1549/files/8c5209044001efedfb09993f92e53eab8bf61fc7#r2329146636

PR #1554, opened 2025-09-09
Wrap paths in quotes

benPearce1:
[sc-119773]

For script steps, where the customers script is sourced from a subdirectory containing a space, the files written to support the gcloud auth are inside the subdirectory. 
This PR ensures that these paths are quoted to avoid errors with paths containing spaces.

Relates to https://github.com/OctopusDeploy/issues/issues/9581


mik-ky (approved):
LGTM.

nit: Not sure what the appetite is around extension methods, but could be worthwhile adding it there if we use it frequently enough?

PR #1576, opened 2025-09-21
Update Argo Image Tag updater to handle nested helm charts

rain-on:
Argo treats _each subdirectory_ as a separate 'Argo Source' - which means that "basic source" in the Application yaml _could_ actually contain a helm chart (determined implicitly by existence of a Chart.Yaml file.

This change updates the basic-repo handling to determine if the repo is actually helm rather than plain yaml - then treats said repo as such.

While fixing this it was found:
1. The package data for the step actually comes in via the IndexedImage (not ImageId)
2. The RepositoryIndex in the InstallContext wasn't being updated following each checkout


Jtango18 (approved):
LOKTM

PR #1585, opened 2025-09-24
Update Image Update step to show rich warnings for missing annotations

rain-on:
It is possible that an Argo CD application has a 'Basic Source' (as identified by its yaml) - which actually contains a helm chart (based on the content of the repo it references).

Octopus _can_ update such a situation, but only if the appropriate annotations have been added to the application.

This PR tries to put in a valuable log message to aid the users to self-serve a solution.

In retrospect, linking to our Argo Applications page is _not_ great - as we won't necessarily be showing the "Generate Helm Annotations" button...

image

### DIFF
--- a/source/Calamari/ArgoCD/Conventions/UpdateArgoCDAppImagesInstallConvention.cs
+++ b/source/Calamari/ArgoCD/Conventions/UpdateArgoCDAppImagesInstallConvention.cs
@@ -294,7 +294,7 @@ namespace Calamari.ArgoCD.Conventions
 
         void GenerateHelmAnnotationLogMessages(Application app, BasicSource source)
         {
-            var docsURL = "https://octopus.com/docs";
+            var docsURL = "https://oc.to/argo-cd-helm-image-annotations";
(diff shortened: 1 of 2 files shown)


Jtango18 (approved):
Ship it!

source/Calamari/ArgoCD/Conventions/UpdateArgoCDAppImagesInstallConvention.cs
```diff
@@ -263,6 +292,15 @@ void PushToRemote(RepositoryWrapper repository,
             }
         }
 
+        void GenerateHelmAnnotationLogMessages(Application app, BasicSource source)
+        {
+            var docsURL = "https://octopus.com/docs";
+            log.WarnFormat("Argo CD Application '{0}' contains a helm chart ({1}), however the application is missing Octopus-specific annotations required for image-tag updating in Helm.",
+                           app.Metadata.Name, Path.Combine(source.Path, ArgoCDConstants.HelmChartFileName));
+            log.WarnFormat("Annotation creation documentation can be found {0}.", log.FormatLink(docsURL, "here"));
```
APErebus, line 300:
```suggestion
            log.WarnFormat("Annotation creation documentation can be found {0}.", log.FormatShortLink("argo-cd-helm-image-annotations", "here"));
```

PR #1587, opened 2025-09-24
Log permissions context if it exists

eddymoulton:


eddymoulton:
Going to replace this with troubleshooting docs

PR #1589, opened 2025-09-25
Argo Manifests: Support Purge output

rain-on:
There is a variable on the Argo Manifest step which is responsible for removing files from the output directory prior to updating it.


source/Calamari/ArgoCD/Conventions/UpdateGitRepositoryInstallConvention.cs
```diff
@@ -74,6 +74,11 @@ public void Install(RunningDeployment deployment)
                     Log.Info($"Copying files into repository {applicationSource.RepoUrl}");
                     var subFolder = applicationSource.Path ?? String.Empty;
                     Log.VerboseFormat("Copying files into subfolder '{0}'", subFolder);
+                    
+                    if (deploymentConfig.PurgeOutputDirectory)
+                    {
+                        repository.StageFilesForRemoval(subFolder, deploymentConfig.RecurseInputPath);
+                    }
```
tleed5, line 81: Should we prevent purging the output directory if the input path is within the output? To avoid a scenario where someone accidentally nukes their template?

tleed5 (approved):
Looks good just one question

PR #1591, opened 2025-09-25
Argo Template step to handle Input Path

rain-on:
During testing it was found that if an input path, not containing a "/" was used - but a directory of that name existed - the entire directory would be templated.

This change change checks in the InputPath's closing character is a  "/" - and if so - the resulting file-system entity _must_ be a directory and vice versa.

If the an filesystem entity of the correct type, with the correct name does not exist, the command will fail.


rain-on:
Closing this as the InputPath should not enforce file/directory - if the user has specified a name as the base path, it should be used regardless of if its a file or a directory

tleed5 (approved):

PR #1595, opened 2025-09-30
Disable TreatWarningsAsErrors in Calamari.Common

APErebus:
It's causing issues in `main` builds due to the `AlphaFS` pre-release package


flin-8 (approved):
LGTM

PR #1596, opened 2025-09-30
Argo Template step to include all files, not just Yaml

rain-on:
It was found during testing that the template step did not overwrte the helper.tpl style files in a helm repo.

The behavior should be, that any file in the input templates _should_ be copied to the output.

This change removes the "*.yml/*.yaml" filename filter, and replaces it with "*".


flin-8 (approved):
LGTM

PR #1600, opened 2025-10-02
Fail updating Argo CD app manifests the application has multiple sources

APErebus:
NOTE: The step used to be called `CommitToGitCommand`. I've renamed this to `UpdateArgoCDAppManifestsCommand` for consistency and clarity.

This is a temporary limitation that an application cannot have it's manifests updated if it has more than one source.

There will be future work to enable a scenario where you can target the source you want to update out of many.


rain-on (approved):
LGTM

PR #1601, opened 2025-10-02
Change SubstituteInFiles to support custom file matching/discovery

APErebus:
Changes the SubstituteInFiles to have a default `ISubstituteFileMatcher`, which is the existing glob code.

However, it now optionally takes a custom matcher. This is used in the Argo CD App Manifest update command as the file matching does not use globs


Jtango18 (approved):
Ship it!

PR #1602, opened 2025-10-02
fix values file collation

rain-on:
When iterating over the Applications & Applicatuion Sources, the "detected" helm charts do not get cleaned between applications - meaning a HelmChart from App1 will get swept up for specialisation in App2 - which is not what is wanted.

Still trying to work out if this is easily testable.


flin-8 (approved):
LGTM

PR #1603, opened 2025-10-06
Remove the sourceDTO which isn't being used anyway

rain-on:
There was duplication in the Argo steps regarding:
* Data coming from the yaml vs
* Data coming from the 'package custom DTO'.

Given the business logic has moved to using _just_ the yaml - it seems the SourceDTO typ should be removed, and all references to it should be removed from the tests.


APErebus:
@rain-on can this be closed?

PR #1695, opened 2025-12-15
Placeholder

Jtango18:
Placeholder PR for implementing test logic mentioned in: https://github.com/OctopusDeploy/OctopusDeploy/pull/39063


Jtango18:
On closer inspection of what the deleted test in https://github.com/OctopusDeploy/OctopusDeploy/pull/39063 actually did was 

1. Create a Powershell script
2. Deploy it using Azure Script Step
3. Check that the resourcse deployed into Azure matched.

This doesn't really test anything related to Octopus. We already have tests that ensure Scripts are executed properly. This is basically testing "did the script I write produces what I think it should produce". That's actually between the script author and the Azure API regardless of Octopus/Calamari changes/implementations. 

Closing as will not implement

PR #1730, opened 2026-01-23
Add selective preloading of Bash script modules

kube89:
This PR introduces an opt-in way to preload Bash script modules before the user’s Bash script runs.

New variable: **Octopus.Action.Script.PreloadScriptModules**

If that variable is set, Calamari will source specific script modules in the bootstrapper first.
Then it sources the actual user script as usual.

### Usage 
Set the project variable and scope it to a process. Can scope to specific steps for different requirements, and maintain an entry not scoped at step-level if you want to define a default across the process.

### Test scenario
**Modules defined:**
image

**Project variables:**
image

**Deployment process with 2 steps (both running same inline script):**
image

**Logs (verbose) Step 1:**
image

**Logs (verbose) Step 2:**
image


zentron (changes requested):
Thanks @kube89.
Im interested to know what problem this is solving that using the [existing script modules](https://octopus.com/docs/deployments/custom-scripts/script-modules) doesn't provide?

HuyPhanNguyen:
Hi @kube89, the PR diff is noisy right now as it includes commits from for the ArgoCD refactor. This makes it hard to review your actual feature.
Could you rebase onto current main? That should clean up the diff to just your preload feature.

zentron:
Hi @kube89 we appreciate the effort you have gone to with this proposal.
After talking it through with out Product Manager at the moment we have decided not to merge this PR.

We plan to focus on some of the ways that Octopus executes steps soon, and we want to avoid adding additional hidden variable behaviour in the meantime as that might make any transition that little bit harder or unexpected. 

Although there is some overhead in invoking some `initization` method across the script steps that use this, I'm hoping that the inconvenience of this workaround is at least not a blocker for your work.
 
If you have other requests we encourage you to reach out via support to talk through what you are after so we can ensure it fits into the plans for the future. 

Thanks again.

PR #1744, opened 2026-01-30
Argo Image Update Step: Remove package purpose from package filtering

rain-on:
Currently, the Argo Image Update step accepts a list of packages which are to be updated.

Calamari filters this list, removing any packages which are not flagged with a purpose of "DockerImageReference".

What is the risk of removing this?

Someone creates a step which has a non-docker package - this package therefore will not exist within the k8s yaml which is to be updated. So we may perform extra iterations for packages which cannot be updated (thus the step takes longer), but it will not functionally change the behaviour.

It should be noted that the step-UI prevents the inclusion of any packages which are _not_ containers - thus 'bad' packages must be added to the step via other means (eg Terraform, via the API etc.).


APErebus (commented):
Is this hurting us by keeping it?

PR #1840, opened 2026-03-23
Retry terraform init on transient 502/503 errors

HuyPhanNguyen:
# Background

`terraform init` frequently fails with `502 Bad Gateway` when downloading providers from GitHub (e.g., dnsimple/dnsimple). This is a transient CDN/network error that resolves on retry, but currently Calamari runs `terraform init` exactly once and fails immediately on any non-zero exit code.

Reported by Matt Richardson — the ReleaseBot project on deploy.octopus.app hits this "fairly frequently."

# Results

`terraform init` now retries up to 3 times with exponential backoff (5s, 10s, 20s) when a transient HTTP error (502/503) is detected in stderr. Non-transient errors still fail immediately.

fixes HPY-1283

## Before
`terraform init` fails on first 502 Bad Gateway. Deployment fails with a stack trace. User must manually re-deploy.

## After
`terraform init` detects the transient 502/503 in stderr, logs a warning, backs off, and retries. If the error clears (which it usually does), init succeeds and the deployment continues. If all retries are exhausted, the original error is thrown as before.

### DIFF
--- a/source/Calamari.Common/CalamariFlavourProgram.cs
+++ b/source/Calamari.Common/CalamariFlavourProgram.cs
@@ -12,6 +12,7 @@ using Calamari.Common.Features.EmbeddedResources;
 using Calamari.Common.Features.FunctionScriptContributions;
 using Calamari.Common.Features.Packages;
 using Calamari.Common.Features.Processes;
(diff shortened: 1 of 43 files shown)


hnrkndrssn (approved):
LGTM 👍

PR #1867, opened 2026-04-07
Fix spurious partial script isolation warning when only TentacleHome env var is set

rhysparry:
`PartiallyConfigured` was checking `TentacleHome` alongside the CLI-supplied Level and MutexName options. Since `TentacleHome` is always populated from the environment on a Tentacle host, this caused a misleading 'Some script isolation options were provided, but the following required options are missing' log message on every invocation where script isolation was not explicitly configured.

Fix by removing `TentacleHome` from the `PartiallyConfigured` check — it is an environment prerequisite, not a user-supplied option, and should not be treated as evidence of partial configuration intent.

Resolves EFT-3170


gb-8 (approved):
✅ LGTM

PR #1868, opened 2026-04-07
Enable script isolation alert service messages

rhysparry:
To support logging occurrences of script isolation changing levels due to insufficient support of the underlying system, we are now recording a service message which can be handled by Octopus Server so that we can alert/monitor the impact of this change.

Resolves EFT-3167


gb-8 (approved):
✅ LGTM

PR #1869, opened 2026-04-07
Allow control of whether we generate output variables for S3 commands

david-staniec-octopus:
This PR introduces a new option in the `UploadAwsS3Command` to allow the caller to disable creation of output variables. This is via the new enum `S3OutputVariablesStrategy`, which can be set to `AllFiles` or `NoFiles`.

For deployments with thousands of files, this can create hundreds of thousands of output variables. For performance, the caller can choose to disable the creation of these output variables for performance purposes.

This defaults to `S3OutputVariablesStrategy.AllFiles` to maintain the existing behaviour of creating these variables.


Jtango18 (approved):
Sensible addition. LGTM

PR #1872, opened 2026-04-09
Use explicit region for AWS auth

sathvikkumar-octo:
As part of preparing to update to AWS SDK v4 we need to specify a region when authenticating with AWS

- We remove the regionless fallback for AWS step auth, all steps should have a region already.
- K8s step EKS discovery OIDC auth now uses the first region configured for discovery and then uses these credentials for all configured discovery regions.
- K8s step using EKS API target is unchanged as no region is used.
- ECR feed OIDC auth now uses the region set in the feed credentials.
- Prevents warning messages from `aws-iam-authenticator`

This also means the optional region set on AWS account credentials in server is not used for anything on the Calamari side. Full details below.

# Affected flows
| Calamari Path                                             | Region Source                                                               |
|-----------------------------------------------------------|-----------------------------------------------------------------------------|
| AWS Steps (CloudFormation, S3, Run Script, Terraform)     | `Octopus.Action.Aws.Region` (step, always required)                         |
| K8s step targeting EKS API target (AWS CLI)               | Extracted from EKS cluster URL, passed as `--region` to `aws eks get-token` |
| K8s step targeting EKS API target (aws-iam-authenticator) | Resolved by aws-iam-authenticator                                           |
| K8s step using AWS target discovery (OIDC)                | `Regions.First()` from discovery context                                    |
| ECR Feed (OIDC) on worker                                 | `AuthenticationVariables.Aws.Region` (feed config, always present)          |

# Region resolution in aws-iam-authenticator

`aws-iam-authenticator` uses Go SDK v2 and resolves its STS region independently:

1. Explicit `--region` option (not passed in our flow)
2. IMDS (EC2 instance metadata)
3. Hardcoded fallback to `us-east-1`

Source: https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/pkg/token/token.go

```go
if options.Region != "" {
	cfg.Region = options.Region
}
// The SDK requires a region for clients
// https://docs.aws.amazon.com/sdk-for-go/v2/developer-guide/configure-gosdk.html
if cfg.Region == "" {
	// Attempt to get the region from IMDS (applicable if run on an EC2 instance)
	imdsClient := imds.NewFromConfig(cfg)
	region, err := imdsClient.GetRegion(context.Background(), &imds.GetRegionInput{})
	if err != nil {
		// Default to the global region
		logrus.Infof("failed to get region from IMDS for token generation, defaulting to us-east-1. imds error: %v", err)
		cfg.Region = "us-east-1"
	} else {
		cfg.Region = region.Region
	}
}
```

Ref MD-1714
Fixes MD-1723

### DIFF
--- a/source/Calamari.Aws/Kubernetes/Discovery/AwsAuthenticationDetails.cs
+++ b/source/Calamari.Aws/Kubernetes/Discovery/AwsAuthenticationDetails.cs
@@ -1,5 +1,7 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
+using Amazon;
 using Amazon.Runtime;
 using Amazon.SecurityToken;
 using Amazon.SecurityToken.Model;
@@ -51,7 +53,8 @@ namespace Calamari.Aws.Kubernetes.Discovery
                 var sessionDuration = Credentials.Account.SessionDuration;
                 var jwt = Credentials.Account.Jwt;
 
-                var assumeRoleWithWebIdentityResponse = new AmazonSecurityTokenServiceClient(new AnonymousAWSCredentials()).AssumeRoleWithWebIdentityAsync(new AssumeRoleWithWebIdentityRequest
+                var regionEndpoint = RegionEndpoint.GetBySystemName(Regions.First());
+                var assumeRoleWithWebIdentityResponse = new AmazonSecurityTokenServiceClient(new AnonymousAWSCredentials(), regionEndpoint).AssumeRoleWithWebIdentityAsync(new AssumeRoleWithWebIdentityRequest
                 {
                     RoleArn = roleArn,
                     DurationSeconds = int.TryParse(sessionDuration, out var seconds) ? seconds : 3600,
--- a/source/Calamari.CloudAccounts/AwsEnvironmentGeneration.cs
+++ b/source/Calamari.CloudAccounts/AwsEnvironmentGeneration.cs
@@ -22,7 +22,7 @@ namespace Calamari.CloudAccounts
         const string RoleUri = "http://169.254.169.254/latest/meta-data/iam/security-credentials/";
         const string MetadataHeaderToken = "X-aws-ec2-metadata-token";
         const string MetadataHeaderTTL = "X-aws-ec2-metadata-token-ttl-seconds";
-        private const string DefaultSessionName = "OctopusAwsAuthentication";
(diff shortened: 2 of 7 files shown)


source/Calamari.CloudAccounts/AwsEnvironmentGeneration.cs
```diff
@@ -166,9 +164,15 @@ async Task<bool> VerifyLogin()
         /// </summary>
         void PopulateCommonSettings()
         {
+            if (string.IsNullOrWhiteSpace(region))
+            {
+                throw new Exception("AWS-LOGIN-ERROR-0007: "
+                                    + "No AWS region was specified. Please set the region in the AWS account or the step configuration.");
```
sathvikkumar-octo: Nobody should ever see this message but this should only mention the step and not account

zentron: nice

source/Calamari.Aws/Kubernetes/Discovery/AwsAuthenticationDetails.cs
```diff
@@ -51,7 +53,8 @@ public bool TryGetCredentials(ILog log, out AWSCredentials credentials)
                 var sessionDuration = Credentials.Account.SessionDuration;
                 var jwt = Credentials.Account.Jwt;
 
-                var assumeRoleWithWebIdentityResponse = new AmazonSecurityTokenServiceClient(new AnonymousAWSCredentials()).AssumeRoleWithWebIdentityAsync(new AssumeRoleWithWebIdentityRequest
+                var regionEndpoint = RegionEndpoint.GetBySystemName(Regions.First());
```
zentron, line 56: This is the scenario where for the purposes of discovery we were considering falling back to the account region, but now just defaulting to "whatever" since its just to get the token that we use to later look through all regions?

zentron (approved):

source/Calamari.Aws/Kubernetes/Discovery/AwsAuthenticationDetails.cs
sathvikkumar-octo: Yep, I've updated my PR description as well to provide more info on the flows.

Jtango18 (approved):
This looks like it nicely handles a whole bunch of these dge cases and get's us closer to no breaking changes for AWS

PR #1873, opened 2026-04-09
Force a basic auth header when cloning & pushing to self-hosted ADO

rain-on:
Self-hosted ADO wants all auth to use NTLM, so even if provide username and password, git negotiates to the most secure available credential solution - meaning git tries to use NTLM - and fails given we only have a username/password.

Argo gets around this by forcing a basic-auth header into the git requests (see this issue https://github.com/argoproj/argo-cd/issues/11914).

This code is then implemented in argo in client.go (runCredentialedCmd).

Thus - this change adds a custom basic-auth header to all clone and push operations.

OctopusServer does something similar with a SmartSubTransport handler.
As such, this handler has been copied to calamari, and registered with libgit2sharp

Fixes MD-1663

### DIFF
--- a/source/Calamari/ArgoCD/ArgoCDModule.cs
+++ b/source/Calamari/ArgoCD/ArgoCDModule.cs
@@ -3,6 +3,7 @@ using Calamari.ArgoCD.Conventions;
 using Calamari.ArgoCD.Git;
 using Calamari.ArgoCD.Git.PullRequests;
 using Calamari.ArgoCD.Git.PullRequests.Vendors.GitLab;
+using LibGit2Sharp;
 using Microsoft.Extensions.Caching.Memory;
 using Microsoft.Extensions.Options;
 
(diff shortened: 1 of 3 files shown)


source/Calamari/ArgoCD/Git/GitHttpSmartSubTransport.cs
```diff
@@ -0,0 +1,301 @@
...
+    /// without breaking other existing authentication or exhausting sockets. Read
+    /// more here: https://github.com/libgit2/libgit2sharp/issues/1894
+    ///
+    /// For now, we're have implemented this sub-transport to get around these issues.
+    /// This code is pretty much taken wholesale from the LibGit2Sharp ManagedHttpSmartSubtransport
+    /// https://github.com/OctopusDeploy/libgit2sharp/blob/master/LibGit2Sharp/Core/ManagedHttpSmartSubtransport.cs
+    /// with a few changes (and simplifications) around authentication. We currently only
+    /// support basic auth, but it seems to work nicely for what we need.
+    /// </summary>
+    public class GitHttpSmartSubTransport : RpcSmartSubtransport
```
APErebus, line 24: Should we also include a note here that this code almost directly comes from Octopus Server's codebase?

source/Calamari/ArgoCD/ArgoCDModule.cs
```diff
@@ -23,15 +38,16 @@ protected override void Load(ContainerBuilder builder)
...
             builder.RegisterType<ArgoCDManifestsFileMatcher>().As<IArgoCDManifestsFileMatcher>().InstancePerLifetimeScope();
             builder.RegisterType<ArgoCDFilesUpdatedReporter>().As<IArgoCDFilesUpdatedReporter>().InstancePerLifetimeScope();
-            
+
             builder.RegisterType<SelfHostedGitLabInspector>().AsSelf().InstancePerLifetimeScope();
-            
-            RegisterMemoryCache(builder);;
+
+            RegisterMemoryCache(builder);
+            ;
```
APErebus, line 45:
```suggestion
            RegisterMemoryCache(builder);
```

APErebus (approved):
LGTM! 1 minor nit

PR #1877, opened 2026-04-13
Enable Envirnment Variable for WaitForDebugger

Jtango18:
Currently Calamari.WaitForDebugger can be set as part of a Project variables passed by Server, but there is no way to enable it for non-project Calamari scenarios (i.e. Healthchecks of Targets etc).

This PR centralises the logic for checking that variables either from Variables passed or via an Environment variable specified in Octopus Server env `_CALAMARI_WAIT_FOR_DEBUGGER` inline with other Debug relevant Env Variable naming

CleanShot 2026-04-13 at 11 45 13@2x

### DIFF
--- a/source/Calamari.Common/CalamariFlavourProgram.cs
+++ b/source/Calamari.Common/CalamariFlavourProgram.cs
@@ -14,7 +14,6 @@ using Calamari.Common.Features.Packages;
 using Calamari.Common.Features.Processes;
 using Calamari.Common.Features.Processes.ScriptIsolation;
 using Calamari.Common.Features.Scripting;
-using Calamari.Common.Features.Scripting.DotnetScript;
 using Calamari.Common.Features.StructuredVariables;
 using Calamari.Common.Features.Substitutions;
 using Calamari.Common.Plumbing;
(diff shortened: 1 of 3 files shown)


source/Calamari.Common/CalamariFlavourProgram.cs
```diff
@@ -77,7 +73,6 @@ protected virtual int Run(string[] args)
                         Thread.Sleep(1000);
                     }
                 }
-#endif
```
APErebus, line 80: hrm, I'm inclined to say keep the `#if DEBUG` around this entire block to make it clear that this functionality is a debug thing only, not just the value retrieval

APErebus (changes requested):
Let's keep the debug preprocessor around the entire block

source/Calamari.Common/CalamariFlavourProgram.cs
Jtango18: Strongly disagree, but resolved in https://github.com/OctopusDeploy/Calamari/pull/1877/commits/0303cef040aa16ffc8d1bc241218ab9b5d04dfb3

APErebus (approved):

PR #1880, opened 2026-04-13
Improve Calamari compile times

sathvikkumar-octo:
Improve parallelism in the build by:

* Doing one solution wide restore upfront
* Build the whole solution so projects that aren't runtime specific are compiled.
* Build each runtime in parallel but each runtime is built sequentially
* Remove semaphores and per package compilation

Compile step improvement from ~15min to ~10min on CI

[New](https://build.octopushq.com/buildConfiguration/TeamModernDeployments_Calamari_VNext_Build_BuildCompile/21821517)
[Old](https://build.octopushq.com/buildConfiguration/TeamModernDeployments_Calamari_VNext_Build_BuildCompile/21821297)

### DIFF
--- a/build/Build.PackageCalamariProjects.cs
+++ b/build/Build.PackageCalamariProjects.cs
@@ -1,6 +1,5 @@
-﻿using System.Collections.Generic;
-using System.Collections.Concurrent;
-using System.Threading;
+﻿using System.Collections.Concurrent;
+using System.Collections.Generic;
 using Nuke.Common.ProjectModel;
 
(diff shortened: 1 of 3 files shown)


Jtango18 (commented):
This looks promising to me. Will leave for the 👀 of @APErebus to make the final approve verdict though as there's definitely dragons he's familiar with here.

APErebus (approved):
Looks good! I did a build locally and no issues so let's ship it and see how it plays out.

PR #1883, opened 2026-04-14
Calamari .NET Framework cleanup

sathvikkumar-octo:
Noticed some .NET framework fixes and ended up in a rabbit hole of cleaning up bits that are no longer relevant in .NET 8 land.

  - Remove obsolete polyfill NuGet packages not required for .NET 8
  - Remove TLS 1.2 workarounds (default on net8.0)
  - Remove `RunResolvePackageDependencies` MSBuild workaround
  - Remove dead MSBuild conditions, code, and variables
  - Delete `source/Directory.Build.props` (moved sole usage inline)
  - Resolve TODOs that were previously blocked by .NET Framework constraints
  - Update stale comments referencing .NET Framework


source/Calamari.Azure/Kubernetes/Discovery/AzureKubernetesDiscoverer.cs
```diff
@@ -44,7 +44,7 @@ public override IEnumerable<KubernetesCluster> DiscoverClusters(string contextJs
             var discoveredClusters = new List<KubernetesCluster>();
 
             // There appears to be an issue where the azure client returns stale data
-            // We need to upgrade this to use the newer SDK, but we need to upgrade to .NET 4.6.2 to support that.
```
APErebus, line 47: I believe Rob has a PR floating around that does this

APErebus (approved):
Look good to me. I do think we should run a full chain server branch with this PR before merging, just to be sure

Jtango18 (commented):
This looks legit to me. ~Similar to other PR would like a 2nd set of eyes on it @APErebus~

This also prompts the question - should just kill the AzureWebAppShim project all togteher now that we're pure net8?

sathvikkumar-octo:
> Look good to me. I do think we should run a full chain server branch with this PR before merging, just to be sure

Found `WhenRunningAdHocScript` in server but this uses a mock script service so it doesn't execute Calamari
Added a consolidation test to check if the `dotnet-script/` path is in the package but unsure if its valuable.

> This also prompts the question - should just kill the AzureWebAppShim project all togteher now that we're pure net8?

We need the shim because the MS web deploy library is framework only unfortunately.

sathvikkumar-octo:
Full chain build passing here
https://build.octopushq.com/buildConfiguration/OctopusDeploy_OctopusServer_TeamCorePlatform_ChainFull/21833813

PR #1884, opened 2026-04-14
Return replaced/patched files when creating PR and don't send service message

APErebus:
Related Server PR: https://github.com/OctopusDeploy/OctopusDeploy/pull/42189

Related to MD-1767

When creating a pull request, we no longer want to send the `argocd-files-updated` service message. This is because this message causes an update of the tracked git state in Octopus Server.
For a pull request, this git state should only change when the PR is merged.

We now don't send this service message when the git commit parameters `IsPr` is `true`.
We also include the replaced & patched files in new output variables, which are then captured in Octopus Server and attached to the pull request tracking entity.

### DIFF
--- a/source/Calamari.Tests/ArgoCD/ArgoCDFilesUpdatedReporterTests.cs
+++ b/source/Calamari.Tests/ArgoCD/ArgoCDFilesUpdatedReporterTests.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Text.Json;
 using Calamari.ArgoCD;
+using Calamari.ArgoCD.Conventions;
 using Calamari.ArgoCD.Models;
 using Calamari.Testing.Helpers;
 using FluentAssertions;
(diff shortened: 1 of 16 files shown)


source/Calamari.Tests/ArgoCD/ArgoCDFilesUpdatedReporterTests.cs
```diff
@@ -13,18 +14,27 @@ namespace Calamari.Tests.ArgoCD
     [TestFixture]
     public class ArgoCDFilesUpdatedReporterTests
     {
+        GitCommitParameters gitParams = new(string.Empty, string.Empty, false);
```
flin-8: Did you add a test case for when we are creating a PR? I couldn't find it

APErebus: I don't think I'd got that far yet. Just did enough to get it building

APErebus: test added in [370e603](https://github.com/OctopusDeploy/Calamari/pull/1884/commits/370e603ad88ae9d88fe66436a76a7ec96112b11b)

flin-8 (approved):
LGTM

PR #1886, opened 2026-04-15
fix: Erroneous Commit message for Argo Manifest Template step

rain-on:
It was found that the Manifest inspection step would errantly insert "--No Images updated" as part of its commit message - this was introduced during a recent calamari argo-step refactor.

This change introduces a step-specific "CommitMessageGenerator" such that each step can generate a message independently of each other.

A test was inserted to prevent regression.

* No server change is required

### DIFF
--- a/source/Calamari.Tests/ArgoCD/Commands/Conventions/UpdateArgoCDApplicationManifestsInstallConventionTests.cs
+++ b/source/Calamari.Tests/ArgoCD/Commands/Conventions/UpdateArgoCDApplicationManifestsInstallConventionTests.cs
@@ -159,7 +159,7 @@ namespace Calamari.Tests.ArgoCD.Commands.Conventions
             fileSystem.FileExists(Path.Combine(resultPath, fourthFilename)).Should().BeTrue();
 
             using var resultRepo = new Repository(resultPath);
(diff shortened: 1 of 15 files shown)


source/Calamari.Tests/ArgoCD/Commands/Conventions/UpdateArgoCDApplicationManifestsInstallConventionTests.cs
```diff
@@ -158,6 +158,9 @@ public void ExecuteCopiesFilesOfAnyNameFromPackageIntoRepo()
             fileSystem.FileExists(Path.Combine(resultPath, thirdFilename)).Should().BeTrue();
             fileSystem.FileExists(Path.Combine(resultPath, fourthFilename)).Should().BeTrue();
 
+            using var resultRepo = new Repository(resultPath);
```
flin-8, line 161: What's this?

flin-8: Ah it's testing the message

flin-8: Add an equivalent for the image tag test?

flin-8 (approved):
LGTM

PR #1889, opened 2026-04-16
Git safe clone handles no protocol on uri

rain-on:
If an argo source is a helm-source, referencing a helm repository - it _may_ not include the protocol in the repoURL.

Thus, in Calamari, when deserialising the application yaml, an error is raised when attempting to map the pro=vided URI into an "ssh-mapped uri"

This change adds a check, such that if no protocol is specified, "oci://" is prepended prior to converting to a URI.

Also - it was found that the value calculated during yaml mapping _wasn't being used_ - and instead we re-calculated the URL at repo-clone time.
This was considered a better solution because:
1. Serialisation will pass even if the provided repoURL is not actually a URI
2. We only "translate/validate" the URLs we want to modify - and ones without a protocol should not be scoped for update (as its clearly _not_ a git url).

Fixes MD-1765

### DIFF
--- a/source/Calamari/ArgoCD/Git/GitCloneSafeUrl.cs
+++ b/source/Calamari/ArgoCD/Git/GitCloneSafeUrl.cs
@@ -25,6 +25,10 @@ public static class GitCloneSafeUrl
     /// </summary>
     /// <param name="uri">A, potentially invalid <see cref="Uri"/> object</param>
     /// <returns>A URI that, if is SSH, is well formed<see cref="Uri"/> object</returns>
+    /// <remarks>
+    /// This is invoked during yaml deserialisation, and may be applied to repoURLs which will never actually be cloned
+    /// during step execution (eg sources which have not been scoped to the step).
+    /// </remarks>
     public static Uri FromString(string uri)
     {
         var validProtocols = new[] { "http://", "https://", "oci://" };
--- a/source/Calamari/ArgoCD/Git/GitCloneSafeUrl.cs
+++ b/source/Calamari/ArgoCD/Git/GitCloneSafeUrl.cs
@@ -31,11 +31,11 @@ public static class GitCloneSafeUrl
     /// </remarks>
     public static Uri FromString(string uri)
     {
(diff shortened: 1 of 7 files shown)


source/Calamari/ArgoCD/Git/GitCloneSafeUrl.cs
```diff
@@ -26,8 +27,14 @@ public static class GitCloneSafeUrl
     /// <returns>A URI that, if is SSH, is well formed<see cref="Uri"/> object</returns>
     public static Uri FromString(string uri)
     {
+        var validProtocols = new[] { "http://", "https://", "oci://" };
         if (!uri.StartsWith(StandardSshScpPrefix))
         {
+            // argo does not (always?) add a protocol to helm chart sources.
+            if (!validProtocols.Any(prefix => uri.StartsWith(prefix)))
```
flin-8: We probably want to check if the uri contains `Uri.SchemeDelimiter`, only add `oci://` if the delimiter isn't there

Jtango18 (approved):
This looks legit based on my understanding of the issue.

Jtango18 (changes requested):
Fix build

source/Calamari/ArgoCD/Git/GitCloneSafeUrl.cs
```diff
@@ -24,10 +25,20 @@ public static class GitCloneSafeUrl
     /// </summary>
     /// <param name="uri">A, potentially invalid <see cref="Uri"/> object</param>
     /// <returns>A URI that, if is SSH, is well formed<see cref="Uri"/> object</returns>
+    /// <remarks>
+    /// This is invoked during yaml deserialisation, and may be applied to repoURLs which will never actually be cloned
+    /// during step execution (eg sources which have not been scoped to the step).
+    /// </remarks>
     public static Uri FromString(string uri)
     {
+        var me = Uri.SchemeDelimiter;
```
flin-8: Unused var

Jtango18 (approved):
Seems to do the trick AND it builds :-)

PR #1890, opened 2026-04-16
fix: KustomizeFile patching executes even if no actual change occured.

rain-on:
It was found that a bug existed in the KustomizeUpdater whereby the newTag field would ALWAYS be updated, even if it was already up to date.

This bug was introduced in f4309c463570449da165875aa01314ec716f8ee3 when the comparison of ContainerImageRefernces was updated.

In fixing this behaviour, it was found the test `KustomizeSource_ImageAlreadyAtTargetTag_TracksSourceWithNullCommitSha` was no longer passing.

The test expects to see a single patched file, with a "no-op" json patch, rather than the current behaviour which states "no files were changed" (and therefore, the patched file list is empty).

### DIFF
--- a/source/Calamari.Tests/ArgoCD/Commands/Conventions/UpdateArgoCDAppImagesInstallConventionTest.cs
+++ b/source/Calamari.Tests/ArgoCD/Commands/Conventions/UpdateArgoCDAppImagesInstallConventionTest.cs
@@ -1038,11 +1038,9 @@ namespace Calamari.Tests.ArgoCD.Commands.Conventions
 
             var sourceDetails = actual.TrackedSourceDetails.First();
             sourceDetails.CommitSha.Should().BeNull("no commit was made");
-            // The placeholder-substitution technique locates the tag value in the YAML and produces a patch entry,
-            // but because the before/after content is identical the resulting patch has zero operations.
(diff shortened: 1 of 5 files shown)


source/Calamari.Tests/ArgoCD/Commands/Conventions/UpdateArgoCDAppImagesInstallConventionTest.cs
```diff
@@ -257,6 +257,68 @@ public void DirectorySource_ImageMatches_Update()
...
+                      containers:
+                        - name: nginx
+                          image: nginx:1.27.1 # update to nginx:1.27.1
+                """;
+
+            var clonedRepoPath = RepositoryHelpers.CloneOrigin(tempDirectory, OriginPath, argoCDBranchName);
+            AssertFileContents(clonedRepoPath, yamlFilename, updatedYamlContent);
+
+            var committedContent = fileSystem.ReadFile(Path.Combine(clonedRepoPath, yamlFilename));
+            committedContent.Should().NotContain(":__CALAMARI_PLACEHOLDER__",
```
flin-8: Remove the leading `:`

flin-8 (approved):
LGTM + comment

PR #1892, opened 2026-04-17
Add new Calamari.Contracts library

APErebus:


Jtango18 (commented):
Happy to ✅ this once it's finished. It does make me wonder why we don't use this approach more often with Calamari communication. Feels like we could strip out a biunch of duplication and Magic Strings.

APErebus:
> Happy to ✅ this once it's finished. It does make me wonder why we don't use this approach more often with Calamari communication. Feels like we could strip out a biunch of duplication and Magic Strings.

This is ready to go as I've got server working with this new library

Jtango18 (approved):

PR #1893, opened 2026-04-17
Helm with step-based tags can fail patching

rain-on:

### DIFF
--- a/source/Calamari.Tests/ArgoCD/Commands/Conventions/UpdateArgoCDAppImagesInstallConventionHelmTests.cs
+++ b/source/Calamari.Tests/ArgoCD/Commands/Conventions/UpdateArgoCDAppImagesInstallConventionHelmTests.cs
@@ -1496,6 +1496,176 @@ service:
             AssertFileContents(clonedRepoPath, existingYamlFile, updatedYamlContent);
         }
 
+        [Test]
+        public void RefSource_ImageAlreadyAtTargetTag_Annotation_TracksSourceWithNullCommitSha()
+        {
+            // Arrange
+            var updater = CreateConvention();
+            var variables = new CalamariVariables
+            {
+                [PackageVariables.IndexedImage("nginx")] = "index.docker.io/nginx:1.27.1",
+                [PackageVariables.IndexedPackagePurpose("nginx")] = "DockerImageReference",
+                [ProjectVariables.Slug] = ProjectSlug,
+                [DeploymentEnvironment.Slug] = EnvironmentSlug,
+            };
+            var runningDeployment = new RunningDeployment(null, variables);
+            runningDeployment.CurrentDirectoryProvider = DeploymentWorkingDirectory.StagingDirectory;
+            runningDeployment.StagingDirectory = tempDirectory;
+
+            var existingYamlFile = Path.Combine("otherRepoPath", "values.yaml");
+            originRepo.AddFilesToBranch(argoCDBranchName, (existingYamlFile, @"
+image:
+  repository: index.docker.io/nginx
+  tag: ""1.27.1""
+"));
+
+            var argoCDAppWithHelmSource = new ArgoCDApplicationBuilder()
+                                          .WithName("App1")
+                                          .WithAnnotations(new Dictionary<string, string>()
+                                          {
+                                              [ArgoCDConstants.Annotations.OctopusProjectAnnotationKey(new ApplicationSourceName("ref-source"))] = ProjectSlug,
+                                              [ArgoCDConstants.Annotations.OctopusEnvironmentAnnotationKey(new ApplicationSourceName("ref-source"))] = EnvironmentSlug,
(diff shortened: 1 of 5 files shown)


source/Calamari.Tests/ArgoCD/Commands/Conventions/UpdateArgoCDAppImagesInstallConventionHelmTests.cs
```diff
@@ -1660,6 +1660,56 @@ public void HelmSource_ImageAlreadyAtTargetTag_TracksSourceWithNullCommitSha()
                          ]);
         }
 
+        [Test]
+        public void HelmSource_ImageAlreadyAtTargetTag_UsingStepVariable_TracksSourceWithNullCommitSha()
```
flin-8, line 1742: Also create test case for using annotations with the name and tag split across 2 properties

source/Calamari/ArgoCD/Conventions/UpdateImageTag/AbstractHelmUpdater.cs
```diff
@@ -78,10 +83,42 @@ HelmRefUpdatedResult UpdateHelmImageValues(
...
+    JsonPatchDocument? CreateJsonPatchWithPlaceholders(
+        string content,
+        IReadOnlyCollection<ContainerImageReferenceAndHelmReference> imagesToUpdate,
+        Func<string, IReadOnlyCollection<ContainerImageReferenceAndHelmReference>, ImageReplacementResult> replacerFactory)
+    {
+        var placeholderImages = imagesToUpdate
+            .Select(ir =>
+            {
+                var friendlyName = ir.ContainerReference.FriendlyName();
+                var colonIdx = friendlyName.LastIndexOf(':');
```
scme0: Can we assume that the image always has a tag here? throw error if not? Check what we do in other places.

flin-8: Looks like `DeploymentConfigFactory` is the only place that creates these for real, so it will always have a tag.

To get this in, I reckon we can keep the code and just add a comment to say the tag should always exist

source/Calamari/ArgoCD/Conventions/UpdateImageTag/AbstractHelmUpdater.cs
```diff
@@ -78,10 +83,42 @@ HelmRefUpdatedResult UpdateHelmImageValues(
...
+                var placeholderRef = colonIdx >= 0
+                    ? friendlyName[..colonIdx] + ":__CALAMARI_PLACEHOLDER__"
+                    : friendlyName + ":__CALAMARI_PLACEHOLDER__";
+                return ir with { ContainerReference = ContainerImageReference.FromReferenceString(placeholderRef, defaultRegistry) };
```
flin-8, line 105: For peace of mind, can you add test case for when the container isn't from docker.io? And see that its registry is preserved?

source/Calamari/ArgoCD/Conventions/UpdateImageTag/AbstractHelmUpdater.cs
```diff
@@ -78,10 +83,42 @@ HelmRefUpdatedResult UpdateHelmImageValues(
...
+    /// <summary>
+    /// Creates a JSON patch by running a replacer factory with placeholder tags to produce a "before",
+    /// then running with real tags against the "before" to produce an "after", and diffing the two.
+    /// </summary>
+    JsonPatchDocument? CreateJsonPatchWithPlaceholders(
+        string content,
+        IReadOnlyCollection<ContainerImageReferenceAndHelmReference> imagesToUpdate,
+        Func<string, IReadOnlyCollection<ContainerImageReferenceAndHelmReference>, ImageReplacementResult> replacerFactory)
+    {
+        var placeholderImages = imagesToUpdate
```
flin-8, line 101: The way we generate placeholder images should be the same across source types right? Extract a shared function out?

scme0: Good point, done.

source/Calamari/ArgoCD/Conventions/UpdateImageTag/AbstractHelmUpdater.cs
scme0: done

source/Calamari/ArgoCD/Conventions/UpdateImageTag/AbstractHelmUpdater.cs
scme0: I've added a comment. the other option is that we just throw an exception if there is no tag but as you said, we can leave it for now I guess.

flin-8 (approved):
LGTM

PR #1894, opened 2026-04-17
Refactor helm image replacement handlers

scme0:

We had two distinct processing paths for Helm - one via annotations and one via variables. This refactor unifies them - the only difference is the annotations are preprocessed into image paths so the image replacer can do its thing!

I realised as well that the AbstractHelmUpdater didn't have an "is-a" relationship with BaseUpdater anymore, so I removed the inheritance and move the one static helper method to a static utils class so it can be shared. Also renamed to BaseHelmUpdater because it's now the base.


source/Calamari/ArgoCD/HelmValuesImageReplaceStepVariables.cs
scme0, line 1: This is effectively the `HelmValuesImageReplaceStepVariables.cs` but moved and renamed. The original `HelmContainerImageReplacer.cs` was deleted. Github diff makes this difficult to review 😒

source/Calamari.Tests/ArgoCD/Commands/Conventions/UpdateArgoCDAppImagesInstallConventionHelmTests.cs
```diff
@@ -1557,7 +1557,7 @@ public void RefSource_ImageAlreadyAtTargetTag_UsingStepVariable_TracksSourceWith
             AssertNotUpdatedWithExpectedPatch(getCapturedResults,
                 expectedPatchPointer: "/0/image/tag",
                 expectedPatchValue: "1.27.1",
-                expectedPatchedFilePath: Path.Combine("otherRepoPath", "values.yaml"));
```
scme0, line 1561: We had a weird round trip in the original code where we would take file paths from relative -> absolute -> relative which meant that the paths were normalised to the machine (windows and nix). We don't actually need to do this because calamari handles that down stream automatically. This refactor removes the round-trip so the path here can be a normal string with explicit file path separator.

source/Calamari/ArgoCD/HelmValuesImageReplaceStepVariables.cs
scme0: I've renamed it back to the old name so that github diff picks it up as a move. we can do the renames in a separate PR so it's clear what has changed.

source/Calamari/ArgoCD/Conventions/UpdateImageTag/HelmHelpers.cs
```diff
@@ -1,44 +0,0 @@
-using System;
-using System.Collections.Generic;
-using Calamari.ArgoCD.Models;
-using Calamari.Common.Plumbing.Logging;
-
-namespace Calamari.ArgoCD.Conventions.UpdateImageTag;
-
-public static class HelmHelpers
-{
-    public static void LogHelmSourceConfigurationProblems(ILog log, IReadOnlyCollection<HelmSourceConfigurationProblem> helmSourceConfigurationProblems)
```
scme0, line 10: After refactor, it was only used in AbstractHelmUpdater once so I made it a private static method in that file.

source/Calamari/ArgoCD/HelmValuesImageReplaceStepVariables.cs
```diff
@@ -42,12 +42,12 @@ public ImageReplacementResult UpdateImages(IReadOnlyCollection<ContainerImageRef
             {
                 if (valueToUpdate == newImageTag.ContainerReference.Tag)
                 {
-                    alreadyUpToDate.Add(newImageTag.ContainerReference.Tag);
```
flin-8: Can you see if this is a real bug? Add test?

scme0: I've extracted the tests I added to find this and the fix into this PR: https://github.com/OctopusDeploy/Calamari/pull/1896

source/Calamari/ArgoCD/Conventions/UpdateImageTag/AbstractHelmUpdater.cs
```diff
@@ -12,109 +12,166 @@
 
 namespace Calamari.ArgoCD.Conventions.UpdateImageTag;
 
-public abstract class AbstractHelmUpdater : BaseUpdater
+public abstract class AbstractHelmUpdater : ISourceUpdater
```
APErebus: Nit (not with this PR): I don't love abstract in the class name

PR #1895, opened 2026-04-18
Extract Repository updating from RepositoryAdapter

rain-on:
This is a precursor to CommitToGit step - which should try and use the same mechanisms when updating the git repository.

This also adds a "stageallchanges" to the repositorywrapper - which will be required for CommitToGit as the transform script could add/remove files which we can't track other than via the filesystem


flin-8 (approved):
LGTM

PR #1896, opened 2026-04-20
fix: ensure image references contain the full image and not just the image tag

scme0:

I was adding tests for `HelmValuesImageReplaceStepVariables` class in this pr: https://github.com/OctopusDeploy/Calamari/pull/1894 and I think I found a bug. I've extracted the tests and the fix for the bug into this PR for separate review.

The test failure looks like:
```
Expected result.UpdatedImageReferences[0] to be the same string, but they differ at index 0:
   ↓ (actual)
  "1.27.1"
  "nginx:1.27.1"
   ↑ (expected).
```

These tests fail before fix is applied:
Screenshot 2026-04-20 at 13 15 49

### DIFF
--- a/source/Calamari/ArgoCD/HelmValuesImageReplaceStepVariables.cs
+++ b/source/Calamari/ArgoCD/HelmValuesImageReplaceStepVariables.cs
@@ -42,12 +42,12 @@ public class HelmValuesImageReplaceStepVariables : IContainerImageReplacer
             {
                 if (valueToUpdate == newImageTag.ContainerReference.Tag)
                 {
-                    alreadyUpToDate.Add(newImageTag.ContainerReference.Tag);
+                    alreadyUpToDate.Add(newImageTag.ContainerReference.FriendlyName());
                 }
                 else
                 {
                     updatedYaml = HelmValuesEditor.UpdateNodeValue(updatedYaml, helmReference, newImageTag.ContainerReference.Tag);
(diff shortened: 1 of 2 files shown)


source/Calamari/ArgoCD/HelmValuesImageReplaceStepVariables.cs
```diff
@@ -42,12 +42,12 @@ public ImageReplacementResult UpdateImages(IReadOnlyCollection<ContainerImageRef
             {
                 if (valueToUpdate == newImageTag.ContainerReference.Tag)
                 {
-                    alreadyUpToDate.Add(newImageTag.ContainerReference.Tag);
+                    alreadyUpToDate.Add(newImageTag.ContainerReference.FriendlyName());
```
rain-on, line 45: Wondering if the imagesToUpdate and alreadyUpToDate should be lists of ContainerImageReference rather than string lists - that way we'd have better type protections.

Otherwise, the fix looks good.

rain-on (approved):
We really should use types but this looks good

source/Calamari/ArgoCD/HelmValuesImageReplaceStepVariables.cs
flin-8: Then are you thinking `ImageReplacementResult` will reference `ContainerImageReference` instead of strings @rain-on ?

That can be done in a separate PR

PR #1898, opened 2026-04-22
Fix published library versions

APErebus:
These DLL's should be property versioned, and they weren't

This is the new Contracts and the Consolidation libraries consumed by Octopus Server


NickJosevski (approved):
LGTM

PR #1899, opened 2026-04-22
Add Calamari-side coverage for package transfer + delta reconstruction

julietrb1:

Companion to OctopusDeploy/OctopusDeploy#42443, which asserts that Server sends the correct Calamari command for each delta-compression configuration. That test covers the server decision; these tests cover what Calamari actually does when invoked.

## What's Added

- A new `TransferAndApplyDeltaFixture` chaining `transfer-package` and `apply-delta` end-to-end: transfers a basis package, generates an OctoDiff signature and delta against a new version, then verifies `apply-delta` reconstructs a package whose hash matches the new version. This closes the gap left by the removed `TentaclePackageTransferTestScript` E2E for the delta-enabled path.
- A parameterised `ShouldCopyPackageRegardlessOfFormat` case on `TransferPackageFixture` covering `.nupkg` and `.zip`, so `transfer-package` is no longer only exercised against a single package format.

Cache/skip-cache and delta-disabled scenarios are intentionally not added here. Those are Server-decision paths and are covered by the OctopusDeploy test.

### DIFF
--- a/source/Calamari.Tests/Fixtures/ApplyDelta/TransferAndApplyDeltaFixture.cs
+++ b/source/Calamari.Tests/Fixtures/ApplyDelta/TransferAndApplyDeltaFixture.cs
@@ -1,0 +1,123 @@
+using System;
+using System.IO;
+using Calamari.Common.Plumbing.Extensions;
+using Calamari.Common.Plumbing.FileSystem;
+using Calamari.Common.Plumbing.Variables;
+using Calamari.Testing;
+using Calamari.Testing.Helpers;
+using Calamari.Tests.Fixtures.Deployment.Packages;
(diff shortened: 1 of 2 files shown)


acodrington (approved):
This looks good to me, but I'm not familiar with Calamari etiquette so let's run it by Modern Deployments for the following reasons:

- Does this test provide value? (If not, we should abandon it)
- Does the test follow Calamari testing conventions?
- What's up with [those build errors](https://build.octopushq.com/buildConfiguration/TeamModernDeployments_Calamari_VNext_Chains_ChainFull/21991215)?

APErebus (approved):
LGTM! Thanks @julietrb1

PR #1903, opened 2026-04-24
test: basic auth headers are sent for git requests

eddymoulton:
https://github.com/OctopusDeploy/Calamari/pull/1873 forces basic auth headers to be sent for username/password requests - this tests that fact.

Relates to MD-1806

PR #1904, opened 2026-04-24
Lazily register git transport

eddymoulton:
Built on top of https://github.com/OctopusDeploy/Calamari/pull/1903

When we add support for SSH authentication, we create a dependency on OpenSSL 3.
This is broadly supported at this point by our support platforms, however we can't be sure that our customers are up to date.

The dependency is required only when we access LibGit2Sharp, which only needs to happen for ArgoCD steps at the moment. So I think it's a fairly safe requirement that users using brand new functionality must have up to date workers/targets.

This assumption changed in https://github.com/OctopusDeploy/Calamari/pull/1873 when we forced LibGit2Sharp to load the native binaries during Calamari startup so we can register the transports.

This change makes that registration lazy, so we can return to only requiring any dependencies for actions that need it.

Relates to MD-1806


rain-on (approved):
Remove the static constructor in the test if possible - otherwise LGTM

PR #1905, opened 2026-04-27
Add an empty Commit-to-Git Step

rain-on:
This adds a new command to Calamari, but does not define any behavour - it is a placeholder for future work.


rain-on:
unrequired

PR #1907, opened 2026-04-28
Add null reference guard to DescribeCloudFormationChangeSetConvention

kat-ms:

A recent bump to AWS SDK v4 has meant that `response.Changes` can now be null when there are no changes (instead of an empty list).

As per title, this PR adds a null reference guard to DescribeCloudFormationChangeSetConvention so this is handled without throwing an exception.

Fixes FD-273


benPearce1 (approved):
Looks good. Thanks.

PR #1911, opened 2026-05-05
Allow stage convention to be selective in its packages to extract

rain-on:
The upcoming CommitToGitStep treats the packages in the step as 2 separate lists:
1. Those which support the script to execute
2. Those that should be copied into the repository (but are not required for script support)

As such, the StageDependencyConvention needs to be updated to be able to selectively extract packages based on arbitrary logic (not defined here).

This change results in no functional changes.

### DIFF
--- a/source/Calamari.Common/Util/IExtractionChecker.cs
+++ b/source/Calamari.Common/Util/IExtractionChecker.cs
@@ -1,0 +1,9 @@
+using System;
+using Calamari.Common.Commands;
+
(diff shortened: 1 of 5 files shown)


rain-on:
This has been worked around in the new_commit_to_git branch - whereby all packages are unpacked in a flat directory structure then variables are substituted either sensitively/nopnsensitively based on if the package appears in the metadata property

PR #1912, opened 2026-05-05
Create CommitToGit Command

rain-on:
The commitToGit step is basically the script step, but is also able to copy _some_ of the steps packages into the target git repository.

Thus, there is a single variable which contains a json blob defining which packages should be used for copying (and thus should only be substituted with non-sensitive variables).

The step executes the following in order:
Phase 1:
* Clone target repository into known location, and create variable containing its working directory

Phase 2:
* Unpack primary package, and all associated packages
* Unpack all git dependencies
* Substitute all variables into the script if it is inline.

Phase 3:
* Substitute all variables into packages which are not listed in the "input file metadata"

Phase 4:
* Substitute non-sensitive values into all packages and git-deps in the "input file metadata"
* Copy the specified files into the repository created in Phase 1

Phase 5:
* Execute the script provided (run from deployment directory, but can access respository via variable)

Phase 6:
* Commit changes to repo
* Push to remote and create a PR if required (as per Argo step)
* Raise output variables.


source/Calamari.Shared/Deployment/SpecialVariables.cs
```diff
@@ -353,6 +353,25 @@ public static class Server
...
                 }
             }
+            
+            public static class Git
+            {
+                public const string Url = "Octopus.Action.Git.TargetRepositoryUrl";
+                public const string Username = "Octopus.Action.Git.TargetRepositoryUsername";
+                public const string Password = "Octopus.Action.Git.TargetRepositoryPassword";
+                public const string Reference = "Octopus.Action.Git.TargetRepositoryBranch";
+                public const string DestinationPath = "Octopus.Action.Git.DestinationPath";
```
Jtango18: The mix between target and destination confuzzled me...might be just my brain tho.

source/Calamari/CommitToGit/CommitToGitDependencyMetadataParser.cs
```diff
@@ -0,0 +1,164 @@
...
+
+            return ParseReferenceNames(deployment, relevantTypes);
+        }
+
+        IEnumerable<string> ParseReferenceNames(RunningDeployment deployment, IEnumerable<JToken> parsedJsonArray)
+        {
+            var dependencyNames = new List<string>();
+            // we reverse the order of the array so that we maintain the order that sources at the top take higher precendences (i.e. are adding to the --values list later),
+            // however, within a source, the file path order must be maintained (for consistency) so that later file paths take higher precendence 
+            foreach (var jToken in parsedJsonArray.Select((json, index) => json))
```
Jtango18, line 43: Where is the reversal mentioned in the comment? Am I missing something?

source/Calamari/Commands/CommitToGitCommand.cs
```diff
@@ -0,0 +1,330 @@
...
+        this.gitVendorPullRequestClientResolver = gitVendorPullRequestClientResolver;
+        this.fileSystem = fileSystem;
+        this.variables = variables;
+        this.commandLineRunner = commandLineRunner;
+        this.scriptEngine = scriptEngine;
+        this.deploymentJournalWriter = deploymentJournalWriter;
+        this.configFactory = configFactory;
+    }
+
+    public override int Execute(string[] commandLineArguments)
```
Jtango18, line 72: 💭 This feels like a massive method. Seems like there are 4 pieces perhaps

1. The file copy
2. the transformation
3. the commit and 
4. The script substitution. 

Are they worth separating out?

source/Calamari/Commands/CommitToGitCommand.cs
```diff
@@ -0,0 +1,330 @@
...
+                variables.Set(SpecialVariables.Action.Script.ScriptParameters, scriptParametersArg);
+            }
+        }
+
+        var clock = new SystemClock();
+
+        var deployment = new RunningDeployment(pathToPackage, variables);
+        var baseWorkingDirectory = deployment.CurrentDirectory;
+        var repositoryConfig = configFactory.CreateRepositoryConfig(deployment);
+        var repositoryFactory = new RepositoryFactory(log, fileSystem, baseWorkingDirectory, gitVendorPullRequestClientResolver, clock);
```
Jtango18: Can this be injected?

source/Calamari/Commands/CommitToGitCommand.cs
```diff
@@ -0,0 +1,330 @@
...
+        };
+
+        var commitToRemote = new List<IConvention>
+        {
+            new DelegateInstallConvention(d =>
+                                          {
+                                              var commitParams = repositoryConfig!.CommitParameters;
+                                              var updater = new RepositoryUpdater(commitParams, log, new UserDefinedCommitMessageGenerator(commitParams.Description));
+
+                                              //TODO(tmm): This is a smell, shouldn't need the FileUpdateResult here :/
```
Jtango18: Does this need to be To-Done?

source/Calamari/Commands/CommitToGitCommand.cs
```diff
@@ -0,0 +1,330 @@
...
+        var repositoryFactory = new RepositoryFactory(log, fileSystem, baseWorkingDirectory, gitVendorPullRequestClientResolver, clock);
+        using var clonedRepository = repositoryFactory.CloneRepository(UniqueRepoNameGenerator.Generate(), repositoryConfig.GitConnection);
+        deployment.Variables.Set("Octopus.Calamari.Git.RepositoryPath", clonedRepository.WorkingDirectory);
+        var metadataParser = new CommitToGitDependencyMetadataParser(fileSystem);
+        WriteVariableScriptToFile(deployment);
+        
+        var extractAllPackages = new List<IConvention>
+        {
+            //we only want to include files which are NOT explicitly referenced as dependencies (i.e. we have files which are to be copied into the repo (referenced in variable), and some which should just be used for script dependencies. 
+            new StageDependenciesConvention(pathToPackage, fileSystem, new CombinedPackageExtractor(log, fileSystem, variables, commandLineRunner), new PackageVariablesFactory()),
```
Jtango18, line 120: `PackageVariablesFactory()` is created here and again below. Is that necessary?

Jtango18 (approved):
Mostly questions from me, but nothing too dire blocking here.

source/Calamari/CommitToGit/CommitToGitDependencyMetadataParser.cs
rain-on: yup - stolen from another step - this is not true in this context - removed.

source/Calamari/Commands/CommitToGitCommand.cs
rain-on: Its super light - and now that this broken out to function-per-phase, it make the dependency less simple

source/Calamari/Commands/CommitToGitCommand.cs
rain-on: yup - broke this up into separate functiuons to represent each phase

source/Calamari.Shared/Deployment/SpecialVariables.cs
rain-on: yeah, we'' consolidate on the TargetRepository naming convention

source/Calamari/Commands/CommitToGitCommand.cs
rain-on: don't _think_ so - as we don't know the baseWorking directory until after the deployment is created

source/Calamari/Commands/CommitToGitCommand.cs
rain-on: I've cleaned it up - its not great, but its better.

PR #1914, opened 2026-05-06
Move Argo file changes stuff into Calamari contracts

flin-8:
This is so we can share the service message names etc. with Server

Relates to [MD-1901: Support the same application name in different namespaces](https://linear.app/octopus/issue/MD-1901/support-the-same-application-name-in-different-namespaces)


source/Calamari/ArgoCD/ArgoCDFilesUpdatedReporter.cs
```diff
@@ -34,33 +35,26 @@ public void ReportFilesUpdated(GitCommitParameters gitCommitParameters, IReadOnl
...
-                                                 SpecialVariables.ServiceMessages.ArgoCDFilesUpdated.Name,
+                                                 ServiceMessages.ArgoCDFilesUpdated.Name,
                                                  parameters);
 
                 log.WriteServiceMessage(message);
             }
         }
 
-        List<TrackedSourceDetail> ConvertPathsToPosix(List<TrackedSourceDetail> inputs)
+        static SourceFileChanges MapSource(TrackedSourceDetail trackedSourceDetail)
```
flin-8, line 49: Decouple the service message source details from the `TrackedSourceDetail` (we will add more stuff to `TrackedSourceDetail` in a later PR)

source/Calamari.Tests/ArgoCD/Commands/Conventions/UpdateArgoCDAppImagesInstallConventionHelmTests.cs
```diff
@@ -26,6 +26,7 @@
 using NSubstitute;
 using NUnit.Framework;
 using Octopus.Calamari.Contracts.ArgoCD;
+using FileJsonPatch = Calamari.ArgoCD.FileJsonPatch;
```
zentron: Why do we need to set this using? shouldnt the change result in only one `FileJsonPatch` in the solution?

APErebus: Why use the old one and not the new contracts one?

zentron (approved):
Just a non-blocking question about the using statement

PR #1917, opened 2026-05-07
Add support for ssh git credentials

eddymoulton:
Adds support for SSH credentials when using git, but does not use them until Server gets the same treatment.
We don't have known host support yet, I'm going to do that separately for PR conciseness. In the meantime, we accept all hosts.

Closes MD-1680
Closes MD-1715


source/Calamari/ArgoCD/Git/SshHostKeyVerificationBypass.cs
```diff
@@ -0,0 +1,7 @@
+namespace Calamari.ArgoCD.Git;
+
+internal static class SshHostKeyVerificationBypass
+{
+    // TODO(eddy): Implement proper host key verification
```
eddymoulton, line 5: Future PR

source/Calamari.Tests/ArgoCD/Git/PullRequests/GitVendorApiAdapter_PullRequestTests.cs
```diff
@@ -104,6 +104,7 @@ async Task TestPullRequest(string repositoryUrl, string defaultBranch, string cl
             using var temporaryFolder = TemporaryDirectory.Create();
 
             CredentialsHandler credentialsHandler = (url, usernameFromUrl, types) => new UsernamePasswordCredentials { Username = cloneUsername, Password = clonePassword};
+            LibGit2SharpTransportRegistration.EnsureRegistered();
```
eddymoulton, line 107: We enter git usage in a few weird places in the tests, so for now I've just called this method so we get decent errors if the test runner doesn't meet the dependencies.

Normally this gets done in the repo wrapper and it's a bit cleaner.

source/Calamari/ArgoCD/Git/SshHostKeyVerificationBypass.cs
eddymoulton: https://github.com/OctopusDeploy/Calamari/pull/1958

eddymoulton: NB this doesn't include the known hosts handling for ease of review

source/Calamari/ArgoCD/Git/AuthenticatingRepositoryFactory.cs
```diff
@@ -11,25 +13,50 @@ public class AuthenticatingRepositoryFactory
...
-        Dictionary<string, IGitCredentialDto> gitCredentials,
+        IReadOnlyCollection<IGitCredentialDto> gitCredentials,
         IRepositoryFactory repositoryFactory,
         ILog log)
     {
-        this.gitCredentials = gitCredentials;
+        // Takes the first git credential per URL, with a preference for username/password credentials (they are more broadly useful as they can be used for PR creation)
+        this.gitCredentials = gitCredentials
+                              .GroupBy(c => c.Url)
+                              .ToDictionary(g => g.Key, g => g.OfType<GitCredentialDto>().FirstOrDefault<IGitCredentialDto>() ?? g.First());
```
APErebus, line 23: Are we actually preferencing here like the comment suggests?

source/Calamari/ArgoCD/Git/GitConnection.cs
```diff
@@ -56,4 +56,24 @@ static Uri ParseAsHttpsUri(string repositoryUrl)
             return uri;
         }
     }
+
+    public class SshKeyGitConnection : IGitConnection
```
APErebus: could be a Record?

source/Calamari/ArgoCD/Git/RepositoryFactory.cs
```diff
@@ -122,6 +116,11 @@ RepositoryWrapper CheckoutGitRepository(IGitConnection gitConnection, string che
                 ? gitVendorPullRequestClientResolver.TryResolve(httpsGitConnection, log, CancellationToken.None).Result
                 : null;
 
+            if (gitConnection is SshKeyGitConnection)
+            {
+                log.Verbose("Git is using SSH authentication, Git vendor functionality will not be available");
```
APErebus: does this mean PR's? Should we be more explicit in what this means?

APErebus (approved):
Only a couple of relatively minor questions or comments

source/Calamari/ArgoCD/Git/RepositoryFactory.cs
eddymoulton: Yeah, good call

source/Calamari/ArgoCD/Git/AuthenticatingRepositoryFactory.cs
eddymoulton: Yeah we are `GitCredentialDto` is the username/password type (I didn't rename it for contract reasons, but it might be worth revisiting how we'd do that safely).

So the logic is: For each url, get the username password cred. If we don't find that, get any cred (ie. ssh).

This is going to change with https://linear.app/octopus/issue/MD-1898/match-ssh-keys-to-ssh-urls-but-also-pass-in-usernamepasswords-for-api, but it matches the logic in server at the moment.

Probably also worth noting that server _shouldn't_ be sending down multiple creds at the moment, but keeping this logic in place allows us to have confidence that the credential will be selected deterministically if we do start sending multiple credentials per url to calamari.

PR #1919, opened 2026-05-08
Run find package & register and download & register as single command

davidstaniec:
The find package and download package commands do not register the package; this is done via a separate register command.

It is possible for other deployments to remove the packages via package retention prior to them being registered. Therefore, this PR introduces Find and Register Package and Download and Register package commands. There has been some refactoring to make the code common between the original commands and the new command, in order to address risk of divergence.

Server changes are required to use the new commands, but the existing commands will continue working without corresponding Server changes.


Jtango18 (approved):
Nothing particular scary or problematic in here to my 👀

source/Calamari.Tests/Fixtures/FindPackage/FindAndRegisterPackageFixture.cs
```diff
@@ -0,0 +1,391 @@
...
+using NUnit.Framework;
+using Octopus.Versioning;
+
+namespace Calamari.Tests.Fixtures.FindPackage
+{
+    [TestFixture]
+    public class FindAndRegisterPackageFixture : CalamariFixture
+    {
+        readonly static string tentacleHome = TestEnvironment.GetTestPath("temp", "FindAndRegisterPackage");
+        readonly static string downloadPath = Path.Combine(tentacleHome, "Files");
```
LukeButters, line 26: I have a visceral reaction to statics in tests

source/Calamari.Tests/Fixtures/FindPackage/FindAndRegisterPackageFixture.cs
```diff
@@ -0,0 +1,391 @@
...
+        readonly string packageVersion = "1.0.0";
+        readonly string newpackageVersion = "1.0.1";
+
+        private readonly string mavenPackage = TestEnvironment.GetTestPath("Java", "Fixtures", "Deployment", "Packages", "HelloWorld.0.0.1.jar");
+        private string mavenPackageHash;
+
+        [OneTimeSetUp]
+        public void TestFixtureSetUp()
+        {
+            Environment.SetEnvironmentVariable("TentacleHome", tentacleHome);
```
LukeButters, line 38: OMG, 😢

well I guess this is how it is.

source/Calamari.Tests/Fixtures/FindPackage/FindAndRegisterPackageFixture.cs
```diff
@@ -0,0 +1,391 @@
...
+                var result = FindAndRegisterPackage(packageId, packageVersion, acmeWeb.Hash);
+
+                result.AssertSuccess();
+                result.AssertOutput("Package {0} {1} hash {2} has already been uploaded", packageId, packageVersion, acmeWeb.Hash);
+
+                result.AssertCalamariFoundPackageServiceMessage(Is.True);
+                result.AssertFoundPackageServiceMessage();
+
+                // Verify package was registered with the journal
+                result.AssertOutput("Registered package use/lock for {0} v{1} and task ServerTasks-12345", packageId, packageVersion);
```
LukeButters, line 109: Is it possible to instead assert that the package is indeed locked?

source/Calamari.Tests/Fixtures/PackageDownload/DownloadAndRegisterPackageFixture.cs
```diff
@@ -0,0 +1,671 @@
...
+        {
+            AuthFeedUri = await ExternalVariables.Get(ExternalVariable.MyGetFeedUrl, cancellationToken);
+            AuthFeedUsername = await ExternalVariables.Get(ExternalVariable.MyGetFeedUsername, cancellationToken);
+            AuthFeedPassword = await ExternalVariables.Get(ExternalVariable.MyGetFeedPassword, cancellationToken);
+        }
+
+        [SetUp]
+        public void SetUp()
+        {
+            if (!Directory.Exists(TentacleHome))
```
LukeButters: Should we clear this in the setup so it is cleared on every test run?

LukeButters:
It looks like the race condition remains between download/find and the following registration. What this PR does is significantly reduce the time between those two operations reducing the changes of someone cleaning up the package we just got.

If we had more time we would also ensure that the cleanup operation never deletes files downloaded within the last few seconds.

source/Calamari.Tests/Fixtures/FindPackage/FindAndRegisterPackageFixture.cs
davidstaniec: See latest commit, I had some issues with reading from the repository (we don't seem to have similar tests for the existing register package command). I've fallen back on reading the JSON from the package cache I can see being written to the file system - thoughts?

source/Calamari.Tests/Fixtures/PackageDownload/DownloadAndRegisterPackageFixture.cs
davidstaniec: Sure, not a bad idea. It's in the latest commit.

source/Calamari.Tests/Fixtures/FindPackage/FindAndRegisterPackageFixture.cs
LukeButters: nice

source/Calamari/Commands/FindAndRegisterPackageCommand.cs
```diff
@@ -0,0 +1,75 @@
...
+            {
+                log.ErrorFormat("Failed to find and register package {0} v{1} hash {2}",
+                    options.PackageId, options.PackageVersion, options.PackageHash);
+                return ConsoleFormatter.PrintError(log, ex);
+            }
+        }
+
+        void RegisterPackageUse(PackagePhysicalFileMetadata pkg, IVersion version)
+        {
+            if (string.IsNullOrEmpty(pkg.FullFilePath))
```
LukeButters, line 66: Add a note here about how if the package has no path its not a package that can be registered and note that this checked used to be done in octopus but now it is done here.

LukeButters (approved):

source/Calamari/Commands/FindAndRegisterPackageCommand.cs
davidstaniec: Done, will merge once this update has built

PR #1921, opened 2026-05-11
CommitToGit to accept target repo username/password from customproperties

rain-on:
Currently the CommitToGit accepts the username and password for the target repository to come from variables - however, given git credentials are not scoped to an env/tenant and are instead 'global' it was deemed safer to send these over in a separate dataset which cannot be interfacted with via comand line script etc - it is data which is only available to the calamari codebase.

This requires a change on server to create required custom properties - which is coming, however as this whole step is behind a feature toggle, there is no risk to production systems.

Fixes MD-1629


source/Calamari.Contracts/CommitToGit/CommitToGitCustomPropertiesDto.cs
```diff
@@ -0,0 +1,3 @@
+namespace Octopus.Calamari.Contracts.CommitToGit;
+
+public record CommitToGitCustomPropertiesDto(string CredentialName, string RepositoryUrl, string Username, string Password);
```
flin-8: You probably want a separate class/record for the git credential?

The top level custom properties is for EVERYTHING

source/Calamari.Contracts/ArgoCD/ArgoCDCustomPropertiesDto.cs
```diff
@@ -1,4 +1,4 @@
-using System;
+ using System;
```
flin-8: Undo

flin-8 (approved):
LGTM

PR #1923, opened 2026-05-11
Check OctopusDisableAzureCLI before reading, for strict mode

mik-ky:
## Background
`AzureContext.ps1` script wraps user's scripts supplied in Azure steps. Because this was dot sourced (and I'm guessing there's a good reason to do so), the strict-mode applied to the finally block too.

The finally block reads the `OctopusDisableAzureCLI` variable, which is _optional_. In non-strict mode, this is harmless as an undefined variable evaluates to null. But in strict mode, an undefined variable will throw an error of:
```
InvalidOperation: The variable '$OctopusDisableAzureCLI' cannot be retrieved because it has not been set.
```

This PR is to check whether the variable exists before evaluating it, to prevent deployments resulting in a failed state.

Related to https://github.com/OctopusDeploy/Issues/issues/5489
Related to HPY-54

### DIFF
--- a/source/Calamari.AzureScripting.Tests/AzurePowershellCommandFixture.cs
+++ b/source/Calamari.AzureScripting.Tests/AzurePowershellCommandFixture.cs
@@ -76,6 +76,26 @@ az group list";
 
         [Test]
         [RequiresPowerShell5OrAbove]
+        public async Task ExecuteAnInlinePowerShellCoreScriptWithStrictMode()
+        {
+            var psScript = @"
+Set-StrictMode -Version Latest
+az group list";
+
+            await CommandTestBuilder.CreateAsync<RunScriptCommand, Program>()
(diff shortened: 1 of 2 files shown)


source/Calamari.AzureScripting.Tests/AzurePowershellCommandFixture.cs
```diff
@@ -74,6 +74,26 @@ await CommandTestBuilder.CreateAsync<RunScriptCommand, Program>()
                                     .Execute();
         }
 
+        [Test]
+        [RequiresPowerShell5OrAbove]
```
mik-ky, line 78: I'm not sure if we need this attribute, but following precedence of the existing tests in this file

benPearce1 (approved):
Nice find.

PR #1924, opened 2026-05-11
Don't fail builds for vulns

eddymoulton:
https://octopusdeploy.slack.com/archives/C03SG0LFJHX/p1778460892802739?thread_ts=1777933912.502139&cid=C03SG0LFJHX

[A CVE was published](https://advisories.gitlab.com/nuget/sharpcompress/CVE-2026-44788/) for SharpCompress, which stops us shipping.

This configuration is used by Server too, because there is not much point blocking builds for something that has not changed.


Jtango18 (changes requested):
I would limit this to the 1 specific package that is currently affected by this: 

i.e. `<PackageReference <PackageReference Include="SharpCompress" Version="0.37.2"><NoWarn>NU1902</NoWarn></PackageReference>`

Jtango18 (approved):
Noice!

PR #1930, opened 2026-05-12
refactor: Replace antiquated options functionality

Jtango18:

Removes a bunch overly verbose legacy code for parsing Calamari options. 
Simplifies and adds testing to option parsing


APErebus:
Was all this code not used?

Jtango18:
> Was all this code not used?

That is correct. It wreaked of "we'll put this here because we might need it someday" but that day never came. If you really dig in you'll see the "descriptions" of the command options are never actually manifested anywhere, but I left them in for "compatibility"

APErebus:
We have https://github.com/OctopusDeploy/Shellfish which might be a good replacement one day

APErebus (approved):
removed code is good code

PR #1933, opened 2026-05-13
feature: creates a stubbed command for ECS Target Discovery

Jtango18:


sathvikkumar-octo (approved):

source/Calamari/Program.cs
```diff
@@ -57,6 +59,23 @@ public int Execute(params string[] args)
...
         protected override int ResolveAndExecuteCommand(IContainer container, CommonOptions options)
         {
+            // Handle Pipeline commands such as Target Discovery
+            if (container.IsRegisteredWithName<PipelineCommand>(options.Command))
+            {
+                try
+                {                
+                    var pipeline = container.ResolveNamed<PipelineCommand>(options.Command);
+                    var variables = container.Resolve<IVariables>();
+                    pipeline.Execute(container, variables).GetAwaiter().GetResult();
```
APErebus, line 69: dumb q, but _could_ this method be made `async` to not need the `GetAwaiter().GetResult()`?

source/Calamari.Common/CalamariFlavourProgram.cs
```diff
@@ -134,6 +144,12 @@ protected virtual void ConfigureContainer(ContainerBuilder builder, CommonOption
                 .Where(t => ((CommandAttribute)Attribute.GetCustomAttribute(t, typeof(CommandAttribute))).Name
                     .Equals(options.Command, StringComparison.OrdinalIgnoreCase))
                 .Named<ICommand>(t => ((CommandAttribute)Attribute.GetCustomAttribute(t, typeof(CommandAttribute))).Name);
+            
+            // Register Behaviors
+            builder.RegisterAssemblyTypes(assemblies)
+                   .Where(t => t.IsAssignableTo<IBehaviour>() && !t.IsAbstract)
+                   .AsSelf()
+                   .InstancePerDependency();
```
APErebus, line 146: 🤔 will this register other behaviours we don't expected? Or does `Calamari` not have any dll's that'll contain `IBehaviour`

zentron:
No issue with the code itself, but although I can see what its doing, you're moving/copying the logic in `CalamariFlavourProgramAsync` into `CalamariProgram`, but I perhaps lack the context of why?

source/Calamari/Program.cs
Jtango18: Not a dumb q. I was curious about handling this. I'm still not clear why we have an async and not async version of `CalamariFlavourProgram`.  Given this is an override of the base class I don't "believe" there's a clean way to make it Async.

source/Calamari.Common/CalamariFlavourProgram.cs
Jtango18: I didn't get any pulled in in my testing. And a quick search of the 60(ish) implemnetations suggest they are all in different projects/assemblies.

Jtango18:
> No issue with the code itself, but although I can see what its doing, you're moving/copying the logic in `CalamariFlavourProgramAsync` into `CalamariProgram`, but I perhaps lack the context of why?

I was keeping the Target discovery inline with other implementations which leverage Behaviours and PipelineCommands. This enables that. 

Per our Slack discussions I figured a bit more out regarding the difference between PipelineCommands and Commands and they seemed the most appropriate choice for this kind of action (Target Discovery).

source/Calamari/Program.cs
APErebus: kk

source/Calamari.Common/CalamariFlavourProgram.cs
APErebus: kk, ta

APErebus (approved):

source/Calamari/Program.cs
```diff
@@ -57,6 +59,23 @@ public int Execute(params string[] args)
 
         protected override int ResolveAndExecuteCommand(IContainer container, CommonOptions options)
         {
+            // Handle Pipeline commands such as Target Discovery
+            if (container.IsRegisteredWithName<PipelineCommand>(options.Command))
```
sathvikkumar-octo, line 63: I think I read this wrong initially but how come the pipeline stuff isn't in the `CalamariFlavourProgram` base class?

Jtango18: I don't know the answer to this. It loosk to me like at some point `CalamariFlavourAsync` was added as a 2nd type of Calamari program, rather than just updating what was already there to support async/await. It feels, to me, like they should be recombined into 1 Calamari to rule them all - but that's probably a larger body of work down the track

PR #1936, opened 2026-05-14
CommitToGit should strip non-wildcarded section of the input path prior to copy

rain-on:
As per the upload-artefact operations, the non-wildcarded segment of the input-path should stripped prior to copying.

i.e. only the directory structure from wildcard onwards is maintained.

eg. **/* would results in the WHOLE directory structure being maintained.

The only difference between this and the upload-artefact is that a FULLY-SCOPED filename will be copied flat to the destination path.
Otherwise, the specific file would be required to maintain the same pathing in the destination as in the input.


tleed5 (approved):
Everything looks okay to me, with my limited context anyway. Definitely get a second set of eyes on this before merging though

PR #1939, opened 2026-05-14
Update CommitToGit contract to be able to accept either username and password or ssh

rain-on:


source/Calamari.Contracts/Git/GitCredentials.cs
```diff
@@ -0,0 +1,21 @@
...
+
+public interface IGitCredentialDto
+{
+    string Type { get; }
+    string Url { get; }
+    string Name { get; }
+}
+
+// UsernamePasswordGitCredentialDto - could rename, but not worth altering the API
+public record GitUsernameAndPasswordCredentialDto(string Name, string Url, string Username, string Password) : IGitCredentialDto
```
eddymoulton: I was thinking this name (it's in the comment above)

```suggestion
public record UsernamePasswordGitCredentialDto(string Name, string Url, string Username, string Password) : IGitCredentialDto
```

SSH eqv would be `SshKeyGitCredentialDto`

source/Calamari.Contracts/Git/GitCredentials.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    public const string DiscriminatorValue = "UsernamePassword";
+    public string Type => DiscriminatorValue;
+}
+
+public record GitSshKeyAndKnownHostsDto(string Name, string Url)
```
eddymoulton: I don't think the known hosts adds much - if we have it, it's part of the whole mechanism of the sshkey

rain-on: in terms of naming? Yeah, you're probably right.

source/Calamari.Contracts/Git/GitCredentials.cs
```diff
@@ -0,0 +1,20 @@
...
+
+public record UsernamePasswordGitCredentialDto(string Name, string Url, string Username, string Password) : IGitCredentialDto
+{
+    public const string DiscriminatorValue = "UsernamePassword";
+    public string Type => DiscriminatorValue;
+}
+
+public record SshKeyGitCredentialDto(string Name, string Url)
+{
+    public const string DiscriminatorValue = "SSHKeyAndKnownHosts";
```
eddymoulton: I've been using this - matches with `UsernamePassword` style naming

```suggestion
    public const string DiscriminatorValue = "SshKey";
```

source/Calamari/CommitToGit/CommitToGitConfigFactory.cs
```diff
@@ -34,14 +36,15 @@ public CommitToGitRepositorySettings CreateRepositoryConfig(RunningDeployment de
...
-                                                                       properties.GitCredential.Username,
-                                                                       properties.GitCredential.Password,
-                                                                       uriAsString,
-                                                                       GitReference.CreateFromString(gitReferenceAsString)),
-            commitParameters,
-            variables.Get(SpecialVariables.Action.Git.DestinationPath));
+            var connection = properties.GitCredential switch
+                             {
+                                 UsernamePasswordGitCredentialDto usernamePassword => new HttpsGitConnection(usernamePassword.Username, usernamePassword.Password, uriAsString, GitReference.CreateFromString(gitReferenceAsString)),
+                                 _ => throw new UnsupportedFeatureException("Yeah nah"),
```
eddymoulton: I would suggest a better exception here

source/Calamari/CommitToGit/GitCredentialDtoJsonConverter.cs
```diff
@@ -0,0 +1,38 @@
...
+        => throw new NotSupportedException();
+
+    public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+    {
+        var obj = JObject.Load(reader);
+        var type = obj["Type"]?.Value<string>();
+
+        return type switch
+        {
+            null or UsernamePasswordGitCredentialDto.DiscriminatorValue
```
eddymoulton: We can not handle null here, becuase there is no backwards compatibility concern

eddymoulton (approved):
Works for me, with a few tweaks

PR #1941, opened 2026-05-14
pin v9 of System.Security.Cryptography.Pkcs

Jtango18:
Build Warnings related to conflicting versions of PkCS Assembly.

CleanShot 2026-05-14 at 16 47 16@2x

This pins the higher version in Calamari.Shared and Calamari..Common to resolve.

### DIFF
--- a/source/Calamari.Common/Calamari.Common.csproj
+++ b/source/Calamari.Common/Calamari.Common.csproj
@@ -16,6 +16,7 @@
     <ItemGroup>
         <PackageReference Include="NuGet.Commands" Version="7.0.3" />
         <PackageReference Include="NuGet.Packaging" Version="7.4.0-octopus-release.17001" />
+        <PackageReference Include="System.Security.Cryptography.Pkcs" Version="9.0.0" />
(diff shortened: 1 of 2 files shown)


Jtango18:
This is a temporary workaorund until https://linear.app/octopus/issue/BFT-307/overhaul-forked-nuget-package-publishing-and-dependencies is actioned

source/Calamari.Common/Calamari.Common.csproj
```diff
@@ -16,6 +16,7 @@
     <ItemGroup>
         <PackageReference Include="NuGet.Commands" Version="7.0.3" />
         <PackageReference Include="NuGet.Packaging" Version="7.4.0-octopus-release.17001" />
+        <PackageReference Include="System.Security.Cryptography.Pkcs" Version="9.0.0" />
```
liam-mackie, line 19: Nit: if this was a permanent change that we had to stay on top of, it might have been nice to have a comment here. Given the temporary nature, it's fine :)

liam-mackie (approved):
LGTM!

PR #1945, opened 2026-05-15
fix: Resolve Obsolete warnings in Calamari.Common

Jtango18:
Remove some more Build log noise by updating methods flagged as Obsolete in Calamari.Common


flin-8 (approved):
LGTM

PR #1947, opened 2026-05-18
Add AppliedResources output variable

akavalchuk:
Adds output variable with applied resources so it would be possible to move verification into a separate step.
For kubernates steps we've already had identifiers, so it's only a matter of setting the variable.
For helm step identifiers received be calling `helm get manifest <release> --revision <N>` after the upgrade completes.

### DIFF
--- a/source/Calamari.Tests/KubernetesFixtures/Commands/Executors/KustomizeExecutorTests.cs
+++ b/source/Calamari.Tests/KubernetesFixtures/Commands/Executors/KustomizeExecutorTests.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using System.Threading.Tasks;
 using Calamari.Common.Commands;
 using Calamari.Common.Features.Processes;
+using Calamari.Common.FeatureToggles;
 using Calamari.Common.Plumbing.FileSystem;
 using Calamari.Common.Plumbing.Variables;
 using Calamari.Kubernetes;
@@ -14,6 +15,7 @@ using Calamari.Kubernetes.ResourceStatus.Resources;
 using Calamari.Testing.Helpers;
 using Calamari.Tests.Fixtures.Integration.FileSystem;
 using FluentAssertions;
+using Newtonsoft.Json;
 using NSubstitute;
 using NSubstitute.ClearExtensions;
 using NUnit.Framework;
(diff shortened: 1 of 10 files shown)


source/Calamari/Kubernetes/Commands/Executors/AppliedResourcesOutputHelper.cs
```diff
@@ -0,0 +1,37 @@
...
+                r.GroupVersionKind.Group,
+                r.GroupVersionKind.Version,
+                r.GroupVersionKind.Kind,
+                r.Name,
+                r.Namespace
+            }).ToArray();
+
+            var json = JsonConvert.SerializeObject(resourceList);
+
+            log.SetOutputVariable("AppliedResources", json, deployment.Variables);
```
scme0: Is there a place we can set this as a const? I think we might do it for other output variables?

source/Calamari/Kubernetes/Conventions/Helm/HelmUpgradeExecutor.cs
```diff
@@ -57,9 +62,35 @@ public void ExecuteHelmUpgrade(RunningDeployment deployment,
...
+                    log.Verbose($"Helm manifest for {releaseName} revision {revisionNumber} is empty, skipping applied resources output variable.");
+                    return;
+                }
+
+                var resources = ManifestParser.GetResourcesFromManifest(manifest, namespaceResolver, deployment.Variables, log);
+                AppliedResourcesOutputHelper.SetAppliedResourcesOutputVariable(log, deployment, resources);
+            }
+            catch (Exception ex)
+            {
+                log.Warn($"Failed to set applied resources output variable: {ex.Message}");
```
scme0: What do we expect to fail here? Just getting the manifest or getting the resources from manifest?

source/Calamari.Tests/KubernetesFixtures/Commands/Executors/KustomizeExecutorTests.cs
```diff
@@ -299,6 +301,59 @@ public async Task CommandLineReturnsNonZeroCode_ReturnsFalseToIndicateFailure()
...
+            {
+                new { Group = "", Version = "", Kind = "", Name = "", Namespace = "" }
+            });
+            deserializedResources.Should().HaveCount(2);
+            deserializedResources.Should().Contain(r => r.Kind == "Deployment" && r.Name == "basic-deployment" && r.Namespace == "dev");
+            deserializedResources.Should().Contain(r => r.Kind == "Service" && r.Name == "basic-service" && r.Namespace == "dev");
+        }
+
+        [Test]
+        public async Task DoesNotSetAppliedResourcesOutputVariable_WhenFeatureToggleIsDisabled()
```
scme0, line 408: I guess it doesn't really hurt but because the feature toggle guard is in the SetAppliedResourcesOutputVariable method, we don't really need this test as well?

source/Calamari.Tests/KubernetesFixtures/Conventions/Helm/HelmUpgradeExecutorTests.cs
```diff
@@ -0,0 +1,301 @@
...
+
+            // Assert
+            var appliedResourcesJson = variables.Get("AppliedResources");
+            appliedResourcesJson.Should().BeNull();
+            
+            commandLineRunner.DidNotReceive().Execute(Arg.Is<CommandLineInvocation>(i => i.Arguments.Contains("get manifest")));
+        }
+
+        [Test]
+        public void DoesNotCallGetManifest_WhenFeatureToggleIsDisabled()
```
scme0: This test is basically testing one part of the test above `DoesNotSetAppliedResourcesOutputVariable_WhenFeatureToggleIsDisabled()` - maybe we can just remove this one?

source/Calamari.Tests/KubernetesFixtures/Conventions/Helm/HelmUpgradeExecutorTests.cs
```diff
@@ -0,0 +1,301 @@
...
+
+            // Act
+            executor.ExecuteHelmUpgrade(deployment, ReleaseName, RevisionNumber, installCompletedCts, installErrorCts);
+
+            // Assert
+            commandLineRunner.DidNotReceive().Execute(Arg.Is<CommandLineInvocation>(i => i.Arguments.Contains("get manifest")));
+        }
+
+        [Test]
+        public void CallsGetManifest_WhenFeatureToggleIsEnabled()
```
scme0: Similar to my comment above, should we just add the assertion into the `SetsAppliedResourcesOutputVariable_WhenFeatureToggleIsEnabled` test?

source/Calamari.Tests/KubernetesFixtures/Conventions/Helm/HelmUpgradeExecutorTests.cs
```diff
@@ -0,0 +1,301 @@
...
+            // Assert
+            commandLineRunner.Received().Execute(Arg.Is<CommandLineInvocation>(i => 
+                i.Arguments.Contains("get") && 
+                i.Arguments.Contains("manifest") && 
+                i.Arguments.Contains(ReleaseName) &&
+                i.Arguments.Contains($"--revision {RevisionNumber}")));
+        }
+
+        [Test]
+        public void LogsWarningAndContinues_WhenGetManifestFails()
```
scme0, line 123: I wonder if we should fail on this - The verification step will probably fail if it doesn't get a valid output variable value.

scme0 (approved):
I think this is looking good and I've pre-approved. 🚀 But please take a look at the comments and let me know if something doesn't make sense.

PR #1949, opened 2026-05-20
Use Octopus.LibGit2Sharp

eddymoulton:
Updates to use our fork of LibGit2Sharp that includes the libssh2 support.

No functional changes, but I've added a test to ensure the error message is nice for users who don't have a correct version of OpenSSL installed on their linux target.

Test category changes are for clarity, once this is merged I'll follow up and remove `LinuxSpecificTestingWithoutOpenSsl3`


APErebus (approved):
LGTM!

PR #1950, opened 2026-05-20
chore: Bump System.Text.Json installs for v9.0.16

Jtango18:

Later version is required to avoid conflicts with AWS `Amazon.CDK.Lib` in upcoming SPF Deprecation PR

### DIFF
--- a/build/_build.csproj
+++ b/build/_build.csproj
@@ -32,6 +32,7 @@
         <PackageReference Include="Octopus.1Password.Sdk" Version="0.0.127" />
         <PackageReference Include="AzureSignTool" Version="6.0.1" ExcludeAssets="all" />
         <PackageReference Include="Serilog.Extensions.Logging" Version="8.0.0" />
+        <PackageReference Include="System.Text.Json" Version="9.0.16" />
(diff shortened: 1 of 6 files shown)


build/_build.csproj
```diff
@@ -32,6 +32,7 @@
         <PackageReference Include="Octopus.1Password.Sdk" Version="0.0.127" />
         <PackageReference Include="AzureSignTool" Version="6.0.1" ExcludeAssets="all" />
         <PackageReference Include="Serilog.Extensions.Logging" Version="8.0.0" />
+        <PackageReference Include="System.Text.Json" Version="9.0.16" />
```
sathvikkumar-octo, line 35: Are we using it in the Nuke build?

APErebus (approved):
Same question as @sathvikkumar-octo , but rest is 👍

build/_build.csproj
Jtango18: It is an implicit reference and was using 9.0.0 which was the third version present, so figure best to pin it to the same major and keep disparate versions at 2

PR #1951, opened 2026-05-20
Add Calamari.Contracts types for ECS steps

sathvikkumar-octo:
As part of some rework to ECS steps I am adding the contracts types for the JSON properties we will be sending.

This PR does not consume any of these types.


Jtango18 (approved):
Awesome pass...we'll see if it's a first pass or if I absolutely cook something when consuming it :-P

PR #1954, opened 2026-05-20
feature: Add Newtonsoft serialization settings for contract sharing

Jtango18:
With contracts being shared across Octopus/Calamari I thoguht it might be useful to share the serialization settings to that they could be used for (de)Serialization on both sides of the wire to ensure consistency.

Had a reservation about taking a dep on Newtonsoft for the Contract library and potential future issues of them getting out of sync - to be discussed.


Jtango18:
Pausing For Now

PR #1955, opened 2026-05-20
feature: Add extension method to Variables collection for Deserilizing objects

Jtango18:
With increased use of Calamari.Contracts, we want to be able to retrieve a variable and have it deserialise as a strongly typed object.

Provides an extension method for handing this.


source/Calamari.Contracts/Serialization/JsonSerializationSettings.cs
```diff
@@ -0,0 +1,16 @@
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
+
+namespace Octopus.Calamari.Contracts.Serialization;
+
+public static class JsonSerializationSettings
```
zentron: Doesnt look like this one is used anywhere? Since the extension method above, which has its own SerializationSettings is currently in Calamari.Common we presumably dont need to PackageReference in csproj below?
Or should the serialization above be using this serialization settings?

Jtango18: Ah, that shouldn't be there. It's a remnant of something I started doing yesterday. We should chat, sharing Serialization settings is an interesting challenge.

Jtango18: Resolved in https://github.com/OctopusDeploy/Calamari/pull/1955/commits/16f41f08b7f1d87440d7e6b9c33ed0a54b7d5998

zentron (approved):

PR #1958, opened 2026-05-21
Add known hosts handling

eddymoulton:
Adds checking of known hosts when using git ssh authentication

Relates to MD-1831


source/Calamari/ArgoCD/Git/SshKnownHostsCertificateCheck.cs
```diff
@@ -0,0 +1,157 @@
...
+
+        return sshCertificate;
+    }
+
+    internal static Result CheckKnownHosts(
+        ILookup<string, SshKnownHost> byHost,
+        string hostname,
+        byte[] presentedSha256)
+    {
+        var lookupHost = StripPort(hostname);
```
APErebus: Are we not given the port in the hostname to validate?

APErebus (approved):
LGTM! Just 1 query for my own knowledge

source/Calamari/ArgoCD/Git/SshKnownHostsCertificateCheck.cs
eddymoulton: Very good point, I was under the impression that we needed this - but your work on known hosts has disproven that.

I'll remove this.

PR #1959, opened 2026-05-22
ensure that the environment variables collection is never null

benPearce1:

Related to https://github.com/OctopusDeploy/Issues/issues/10040


source/Calamari.Common/Features/Scripting/ScriptEngine.cs
```diff
@@ -41,11 +44,13 @@ public ScriptSyntax[] GetSupportedTypes()
             return ScriptSyntaxHelper.GetPreferenceOrderedScriptSyntaxesForEnvironment();
         }
 
+        public CommandResult Execute(Script script, IVariables variables, ICommandLineRunner commandLineRunner) => Execute(script, variables, commandLineRunner, []);
```
benPearce1, line 47: Workaround the fact that you cannot set a default value to anything other than `null`

mik-ky (approved):

PR #1961, opened 2026-05-25
CommitToGit prevents copies escaping package being copied

rain-on:
During testing it was found that a user _could_ specify an inputpath for a given source which escaped the source's root directory - meaning the copy operation would/could pull in files from anywhere in the worker's filesystem.

This change checks if an input source defines a path which is within its directory structure before executing the copy 

### DIFF
--- a/source/Calamari.Tests/CommitToGitCommandTest.cs
+++ b/source/Calamari.Tests/CommitToGitCommandTest.cs
@@ -402,6 +402,44 @@ public class CommitToGitCommandTest
     }
 
     [Test]
+    [TestCase("../../../etc/passwd", TestName = "Package_RejectsInputPath_StartingWithDotDot")]
+    [TestCase("/etc/passwd", TestName = "Package_RejectsInputPath_StartingWithSlash")]
+    [TestCase("configs/../../etc/passwd", TestName = "Package_RejectsInputPath_ContainingDotDotSegment")]
+    public void RejectsPackageInputFilePathsThatEscapeSourceDirectory(string maliciousInputPath)
+    {
+        const string packageReferenceName = "my-configs";
+        const string destinationPath = "output-dir";
+
+        var zipPath = CreateZipWithEntry(packageReferenceName, "configs/settings.json", "{}");
+
+        var inputFileSources = $"[{{\"Type\":\"Package\",\"PackageId\":\"{packageReferenceName}\",\"PackageName\":\"{packageReferenceName}\",\"InputFilePaths\":[\"{maliciousInputPath}\"],\"DestinationSubFolder\":\"\"}}]";
+        variables.AddRange([
+            new CalamariExecutionVariable(PackageVariables.IndexedPackageId(packageReferenceName), packageReferenceName, false),
+            new CalamariExecutionVariable(PackageVariables.IndexedOriginalPath(packageReferenceName), zipPath, false),
+            new CalamariExecutionVariable(PackageVariables.IndexedExtract(packageReferenceName), "True", false),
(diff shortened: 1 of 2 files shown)


octonautcal (approved):

source/Calamari.Tests/CommitToGitCommandTest.cs
```diff
@@ -401,6 +401,48 @@ public void OnlyCopiesFilesMatchingInputPathsFromGitDependencyIntoGitRepository(
...
+    [TestCase("C:\\Windows\\System32\\config", TestName = "Package_RejectsInputPath_WindowsAbsolutePath")]
+    [TestCase("\\\\server\\share\\secret", TestName = "Package_RejectsInputPath_UncPath")]
+    public void RejectsPackageInputFilePathsThatEscapeSourceDirectory(string maliciousInputPath)
+    {
+        const string packageReferenceName = "my-configs";
+        const string destinationPath = "output-dir";
+
+        var zipPath = CreateZipWithEntry(packageReferenceName, "configs/settings.json", "{}");
+
+        var inputFileSources = $"[{{\"Type\":\"Package\",\"PackageId\":\"{packageReferenceName}\",\"PackageName\":\"{packageReferenceName}\",\"InputFilePaths\":[\"{maliciousInputPath}\"],\"DestinationSubFolder\":\"\"}}]";
```
eddymoulton: nit: Feel free to throw away, but I find I prefer if we can format the input data for humans and then serialize it for the test. Obviously, one more thing to go wrong so I'm ok with how you've got it now.

eddymoulton (approved):

source/Calamari/Commands/CommitToGitCommand.cs
```diff
@@ -207,6 +209,23 @@ IEnumerable<IConvention> BuildSubstituteAndCopyInputFilesConventions(
...
+        // UNC paths (\\server\share) and drive-letter paths (C:\...) on all platforms.
+        var normalized = inputPath.Replace('\\', '/');
+
+        if (Path.IsPathRooted(normalized) || (normalized.Length >= 2 && normalized[1] == ':'))
+            throw new CommandException($"InputFilePath '{inputPath}' is not allowed because it is an absolute path. Input paths must be relative to the package or git repository source.");
+
+        var segments = normalized.Split('/');
+        if (segments.Any(s => s == ".."))
+            throw new CommandException($"InputFilePath '{inputPath}' is not allowed because it contains a '..' path segment that could reference files outside the input source.");
+    }
```
flin-8, line 222: I wonder if it's safer to evaluate the absolute path, then compare the absolute path with our current path (instead of looking for tokens)

rain-on: its probably more precise as well ....

source/Calamari/Commands/CommitToGitCommand.cs
```diff
@@ -207,6 +209,18 @@ IEnumerable<IConvention> BuildSubstituteAndCopyInputFilesConventions(
         }
     }
 
+    static void EnsureInputPathIsSafe(string sourceDir, string inputPath)
+    {
+        if (string.IsNullOrEmpty(inputPath))
+            return;
+
+        var sourceDirFull = Path.GetFullPath(sourceDir).TrimEnd(Path.DirectorySeparatorChar) + Path.DirectorySeparatorChar;
```
flin-8: I would call these `absoluteSourcePath`/`fullSourcePath` and `absoluteInputSubPath`/`fullInputSubPath`

PR #1963, opened 2026-05-26
preserve inherited AWS_REGION when Terraform step has no region

### LINKED ISSUE #8337: Add Support for Containers to Inherit AWS Web Identities for Non-AWS Steps
# The enhancement
For Step Templates that aren't AWS Steps, using the AWS Account option "Execute using the AWS service role for an EC2 instance" doesn't allow for EKS to use the Instance Role.

This functionality was added in https://github.com/OctopusDeploy/Calamari/pull/961 however it isn't working for all step templates that use the AWS Account.

## The Need
It's currently considered best practise to configure credentials on the container via AWS Web Identities, and containers can be used for running a variety of step types.

## Solution
Add support for inheriting AWS web identities in containers to all step types, e.g. `Apply Terraform` and `Run a Script` Steps

## Links
[Internal Link - Slack](https://octopusdeploy.slack.com/archives/CNHBHV2BX/p1694578009851649)

NickJosevski:
Discussion on [slack](https://octopusdeploy.slack.com/archives/CNHBHV2BX/p1695006795522649?thread_ts=1694578009.851649&cid=CNHBHV2BX).

## Details

__The actual bug is the region overlay, not IRSA (_IAM Roles for Service Accounts_).__ When a Terraform step runs on an EKS Pod worker with no `Octopus.Action.Aws.Region` configured, `AwsEnvironmentGeneration.PopulateCommonSettings` writes `EnvironmentVars["AWS_REGION"] = null` and `["AWS_DEFAULT_REGION"] = null`. `SilentProcessRunner` then overlays that dict onto `ProcessStartInfo.EnvironmentVariables` when spawning `terraform` — and in .NET, a null dict value means "remove this key from the child env."

The flip side: the EKS Pod's IRSA-injected `AWS_REGION` (e.g. `ap-southeast-2`) that Calamari itself inherited from the Pod gets wiped for the `terraform` child process. The Terraform AWS provider then hard-errors before any credential resolution kicks in.

Bug report shape: "AWS Script step works but Terraform step fails on EKS for non-obvious reasons." Worth being honest about why it looked non-obvious — same parent env, different tolerance. The AWS CLI falls back to global STS and reads from `~/.aws/config` when region is missing; the Terraform AWS provider doesn't tolerate it at all.

## Before

`Error: invalid AWS Region: with provider["registry.terraform.io/hashicorp/aws"]`

## After

1. Only assign region into `EnvironmentVars` __when it is actually set.__
2. When the step has no region, leave the inherited parent env alone.

## Testing

Local repro.

The IRSA / web-identity path through `PopulateKeysFromWebRole` (added in #961) itself works correctly today — it was the region-overlay step downstream of it that broke the env handed to `terraform`.

Closes [#8337](https://github.com/OctopusDeploy/Issues/issues/8337).

Does this change require a corresponding Server Change?
- No

### DIFF
--- a/source/Calamari.CloudAccounts/AwsEnvironmentGeneration.cs
+++ b/source/Calamari.CloudAccounts/AwsEnvironmentGeneration.cs
@@ -174,8 +174,17 @@ namespace Calamari.CloudAccounts
         /// </summary>
         void PopulateCommonSettings()
         {
-            EnvironmentVars["AWS_DEFAULT_REGION"] = region;
-            EnvironmentVars["AWS_REGION"] = region;
+            // When the step has no Octopus.Action.Aws.Region, `region` is null. Writing null into
+            // EnvironmentVars used to wipe the parent process's AWS_REGION in any spawned child
+            // (because SilentProcessRunner overlays the dict on ProcessStartInfo.EnvironmentVariables,
+            // where null means 'remove the var'). For Terraform steps on EKS Pods with IRSA, this
+            // erased the Pod's inherited region and broke `terraform plan` with 'invalid AWS Region:'.
+            // See Issues #8337.
+            if (!string.IsNullOrWhiteSpace(region))
+            {
+                EnvironmentVars["AWS_DEFAULT_REGION"] = region;
+                EnvironmentVars["AWS_REGION"] = region;
+            }
         }
 
         /// <summary>
--- a/source/Calamari.Tests/AWS/AwsEnvironmentGenerationFixture.cs
+++ b/source/Calamari.Tests/AWS/AwsEnvironmentGenerationFixture.cs
@@ -103,5 +103,126 @@ namespace Calamari.Tests.AWS
             var envGeneration = new AwsEnvironmentGeneration(Substitute.For<ILog>(), variables);
             envGeneration.AwsRegion.SystemName.Should().Be("aws-global");
         }
+
+        // Repros the Terraform-step-on-EKS scenario from Issues #8337
+        // worker Pod has IRSA env vars set, step is configured with UseInstanceRole=true and no explicit account.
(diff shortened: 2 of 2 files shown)


NickJosevski:
Live EKS verification, with Claude's help to set up the Terrafrom:

**Reproduction harness** (Terraform module + Pod manifest) 
1.  - provisioned a real EKS 1.34 cluster with IRSA, 
2.  - ran a Pod under a ServiceAccount annotated with an IAM role,
3.  - executed six scenarios isolating the credential / region dimensions. 

Bundled Calamari at the time of verification was 2026.2.x via the OctopusDeploy server's BundledCalamariVersion.

**Tests:**

1. aws sts get-caller-identity (sanity) PASS
2. aws sts assume-role-with-web-identity (SDK path) PASS
3. terraform plan with Pod env intact PASS
4. terraform plan, Calamari-style env, region SET PASS
5. terraform plan, Calamari-style env, region ABSENT,
  - web-identity vars left in place FAIL: invalid AWS Region
6. terraform plan, Calamari-style env, region ABSENT,
 - web-identity vars stripped FAIL: invalid AWS Region

Tests 5 and 6 failing identically (with the same "invalid AWS Region:" error from the provider, before credential resolution) confirmed that:
- the bug is region absence, not credential confusion
- the AWS provider doesn't even reach credential resolution without a region
- the fix is purely about preserving the region; web-identity vars are not a factor


Tests 1-4 passing also rule out: 
- misconfigured IRSA, stale OIDC token, AWS SDK version regression, and bundled Calamari's STS path'
 
The change in PR #1872 around AwsRegion fallback to aws-global doesn't help here because terraform's provider doesn't go through that getter.

zentron (approved):

PR #1964, opened 2026-05-26
Change indentation of switch statements

eddymoulton:
The configured indentation of switch expressions is pretty heavily indented. When you take into account some of our long variable names it gets pretty wasteful.

We don't have a heap of these in the codebase and a lot are already intented like this. Aligning the config allows us to auto format more easily.

## Before

```
var result = value switch
                     {
                         true => true,
                         false => false
                     };
```

## After

```
var result = value switch
{
    true => true,
    false => false
};
```


flin-8 (approved):
LGTM

PR #1966, opened 2026-05-26
Add Nuke target to build docker image for each Calamari flavour

APErebus:
As part of a project to have the Kubernetes agent to load Calamari directly from a docker image volume, we need to start publishing Calamari as a docker image

Related to MD-2016


build/Build.Docker.cs
```diff
@@ -0,0 +1,82 @@
...
+using JetBrains.Annotations;
+using Nuke.Common.Tools.Docker;
+using Nuke.Common.Tools.PowerShell;
+
+namespace Calamari.Build;
+
+public partial class Build
+{
+    [PublicAPI]
+    Target BuildDockerImages =>
```
rain-on, line 11: should this have dependencies on the zip files having been created?

build/Build.Docker.cs
```diff
@@ -0,0 +1,82 @@
...
+                           var dockerBuildPlatform = string.Join(",", supportedPlatforms);
+
+                           var dockerFile = KnownPaths.DockerDirectory / "Dockerfile";
+
+                           foreach (var flavour in flavours)
+                           {
+                               Logging.InBlock(flavour, () =>
+                                                        {
+                                                            var compressedArtifactPath = KnownPaths.OutputsDirectory / $"{flavour}.zip";
+                                                            compressedArtifactPath.UncompressTo(KnownPaths.OutputsDirectory / flavour);
```
rain-on: nit: break out KnownPaths.OutputsDirectory / flavour to a varaible?

build/Build.Docker.cs
APErebus: it runs in a different TC build config due to needing `docker`, which the agent that the normal calamari is built on doesn't have (also it's a windows agent and this needs a Linux agent)

build/Build.Docker.cs
```diff
@@ -0,0 +1,82 @@
...
+                                                                                                         .EnableLoad();
+
+                                                                                              return settings;
+                                                                                          });
+
+                                                            var sanitizedTag = tag.Replace("/", "-").Replace(":", ".");
+                                                            var outputFile = KnownPaths.PublishDirectory / $"{sanitizedTag}.tar";
+
+                                                            //create the publish directory
+                                                            if (!Directory.Exists(KnownPaths.PublishDirectory))
```
rain-on: not sure if the "exists" check is needed - CreateDirectory is meant to be smart enough to handle existing paths

APErebus: yeah, I misread the documentation. Will remove

build/Build.Docker.cs
APErebus: Updated in a4f35b96f

build/Build.Docker.cs
APErebus: Updated in a4f35b96f

build/Build.Docker.cs
```diff
@@ -0,0 +1,82 @@
...
+                                                                Directory.CreateDirectory(KnownPaths.PublishDirectory);
+                                                            }
+
+                                                            //save the docker image to a tar file
+                                                            DockerTasks.DockerImageSave(_ => _
+                                                                                             .SetImages(tag)
+                                                                                             .SetOutput(outputFile));
+
+                                                            //compress with gzip
+                                                            PowerShellTasks.PowerShell(_ => _
```
rain-on, line 66: can we not use an inbuilt wrapper of gzip rather than calling out?

rain-on (approved):

build/Build.Docker.cs
APErebus: afaik, there isn't a built-in way to perform gzip compression on an existing `.tar` file. The Nuke compression tasks seem to require a directory, rather than a single file

PR #1967, opened 2026-05-26
CommitToGit: Handle missing destination Path

rain-on:
Turns out Server will remove empty strings from properties/variables - given the destinationPath is (now) allowed to be empty, Calamari needs to expect that it may be missing.

If it is missing, Calamari should assume an empty string.


flin-8 (approved):
LGTM , just add comments in the test & code to explain why we expect nulls

PR #1973, opened 2026-05-28
Use file creation time for GetNearestPackages when version is non-SemVer

NickJosevski:
## Summary

Follow-up to [OctopusDeploy/OctopusDeploy#43750](https://github.com/OctopusDeploy/OctopusDeploy/pull/43750) which adds a `MostRecentlyPublished` ordering strategy for Channel Package Version Rules.

`PackageStore.GetNearestPackages` is called when a package is not already cached on a Tentacle. It finds up to 5 locally cached packages to offer as delta bases for compressed transfer.

**Before:** The method filtered with `version.CompareTo(target) <= 0` and ordered by version descending — both operations are unreliable for non-SemVer version strings (Docker tags, build numbers, date-stamps). Calamari would find no useful delta candidates, causing the full package to be downloaded instead of a delta.

**After:** When `version.Format != VersionFormat.Semver`, the method skips the version filter and instead orders all cached packages for the same package ID by file creation time descending. The most recently cached package is the best delta candidate.

SemVer behaviour is **unchanged**.

## Changes

- `Calamari.Shared/Integration/Certificates/FileSystem/PackageStore.cs` — branch on `version.Format` in `GetNearestPackages`
- `Calamari.Tests/Fixtures/Integration/FileSystem/PackageStoreFixture.cs` — new test with Docker tag versions and deterministic creation times

## Test plan
- [x] Existing `FiltersPackagesWithHigherVersion`, `OldFormatsAreIgnored`, `IgnoresInvalidFiles` tests unchanged
- [x] New `NonSemVerVersionsReturnPackagesOrderedByCreationTime` test verifies creation-time ordering for Docker tags
- [x] `dotnet build` clean

### DIFF
--- a/source/Calamari.Common/Plumbing/FileSystem/CalamariPhysicalFileSystem.cs
+++ b/source/Calamari.Common/Plumbing/FileSystem/CalamariPhysicalFileSystem.cs
@@ -693,6 +693,11 @@ namespace Calamari.Common.Plumbing.FileSystem
             return File.GetCreationTime(filePath);
         }
 
+        public DateTime GetLastWriteTime(string filePath)
+        {
+            return File.GetLastWriteTime(filePath);
+        }
+
         public string GetFileName(string filePath)
         {
             return new FileInfo(filePath).Name;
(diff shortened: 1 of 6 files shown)


NickJosevski:
superseded

PR #1978, opened 2026-05-29
Block SSH on windows

eddymoulton:
Windows uses WinCNG for it's libssh2 crypto backend which is finicky about which SSH keys it accepts, and the failure mode is to hang indefinitely.

For now, we've [documented that Windows doesn't work](https://github.com/OctopusDeploy/docs/pull/3190) and we'll fail fast with a message when users try. If it's worthwhile, we'll come back and figure out how to best check for the key type in use and fail if it's unsupported.

Relates to MD-1957

Failure for windows worker
image


rain-on (approved):
LGTM

PR #1979, opened 2026-05-31
Handle the merge forward from 2026.1 with no changes.

rain-on:
Two changes were backported from main to 2026.1:
* Resolving issues with unstructured text-tags for step-based helm values (introduced in 2026.2.258)
* Forcing basic auth when using libgit2sharp (2026.2.393)

I.e. these changes already existed in 2026.2, and thus the merge forward is a no-op.


eddymoulton (approved):
Description please

PR #1980, opened 2026-06-01
Fix not handling nested groups in GitLab repo URL's

APErebus:
The current GitLab pull request client did not handle nested sub-groups. We now just handle the entire path

Relates to MD-1892

### DIFF
--- a/source/Calamari/ArgoCD/Git/PullRequests/Vendors/GitLab/GitLabPullRequestClient.cs
+++ b/source/Calamari/ArgoCD/Git/PullRequests/Vendors/GitLab/GitLabPullRequestClient.cs
@@ -19,7 +19,8 @@ namespace Calamari.ArgoCD.Git.PullRequests.Vendors.GitLab
             this.baseUrl = baseUrl;
             
             var parts = repositoryConnection.Uri.Value.ExtractPropertiesFromUrlPath();
-            projectPath = $"{parts[^2]}/{parts[^1]}";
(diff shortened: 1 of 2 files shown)


source/Calamari/ArgoCD/Git/PullRequests/Vendors/GitLab/GitLabPullRequestClient.cs
```diff
@@ -19,7 +19,8 @@ public GitLabPullRequestClient(GitLabClient gitLabClient, IHttpsGitConnection re
             this.baseUrl = baseUrl;
             
             var parts = repositoryConnection.Uri.Value.ExtractPropertiesFromUrlPath();
-            projectPath = $"{parts[^2]}/{parts[^1]}";
+            //projects can have nested sub-projects, so we assume the entire path is the project path
+            projectPath =string.Join('/', parts);
```
flin-8, line 23: Do we need to extract then rejoin again? Can we just take the Uri's path?

flin-8 (approved):
LGTM + minor comment

source/Calamari/ArgoCD/Git/PullRequests/Vendors/GitLab/GitLabPullRequestClient.cs
APErebus: We could, but we'd need to do the same logic as the `Extract...` which would be to trim any leading `/` and trip the `.git` suffix. I figured this was fine and consistent

PR #1985, opened 2026-06-02
chmod the binary to executable before building docker image

APErebus:
When these docker images are loaded into Kubernetes agent script pods, they are loaded as readonly file systems, which means the permissions cannot be changed and we need the `Calamari`/`Calamari.XXX` binary to be executable.

Related to MD-2016


rain-on (approved):
LGTM

PR #1981, opened 2026-06-01
Add a docker credential helper to avoid scary security warnings

matt-richardson:
## Summary

In #1542, I implemented a docker credential helper to solve this the scary warning:

```
Download package mcr.microsoft.com/dotnet/sdk v8.0 directly from Docker Registry v1 'Docker Hub' at https://index.docker.io
WARNING! Your credentials are stored unencrypted in 'octo-docker-configs/config.json'. 
Configure a credential helper to remove this warning. See 
https://docs.docker.com/go/credential-store/ 
Login Succeeded 
docker pull mcr.microsoft.com/dotnet/sdk:8.0 
8.0: Pulling from dotnet/sdk 
Digest: sha256:7fd287cec03e027e39be13fb06ca1347f1338c7a6ab4ad305d94d91852693f70 
Status: Image is up to date for mcr.microsoft.com/dotnet/sdk:8.0 
mcr.microsoft.com/dotnet/sdk:8.0 
Staged mcr.microsoft.com/dotnet/sdk version 8.0 (807.973 MB) from Docker Registry v1 'Docker Hub' at https://index.docker.io. 
```

Previously it was implemented as a Calamari sub command, and ended up with quite a lot of messy changes through the startup logging + plumbing.

## Results

This PR approaches it as a standalone, self-contained, trimmed executable `docker-credential-octopus` that Docker invokes directly.

This all exists behind the feature toggle (`calamari-use-docker-credential-helper`).

-----
:warning: Does this change require a corresponding Server Change?
* yes, it needs a feature toggle slug added

:warning: If so - please add a "Requires Server Change" label to this PR!
:done:

PR #1990, opened 2026-06-04
chmod the docker-credential helper binary

APErebus:
We are packaging a new binary, `docker-credential-octopus` (introduced in #1981)

However, when loading calamari as a Kubernetes image volume as a read-only filesystem, this needs to be made executable.

We now perform a `chmod +x` on an array of binary names

image

image

### DIFF
--- a/build/Build.Docker.cs
+++ b/build/Build.Docker.cs
@@ -12,6 +12,12 @@ public partial class Build
         d =>
             d.Executes(() =>
                        {
+                           if (OperatingSystem.IsWindows())
+                           {
+                               Log.Error("The Docker images can not be built on Windows");
+                               return;
+                           }
+
                            var flavours = GetCalamariFlavours();
                            string[] supportedPlatforms = ["linux/amd64", "linux/arm64", "linux/arm"];
                            var dockerBuildPlatform = string.Join(",", supportedPlatforms);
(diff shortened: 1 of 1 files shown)


APErebus:
I _think_ I'm going to need to do this for dotnet-script as well, I'm just testing

APErebus:
> I _think_ I'm going to need to do this for dotnet-script as well, I'm just testing

We may be ok as we execute dotnet-script via `dotnet`, so it might be fine as it's not executing the binary directory from the read-only FS. Will still need to test

build/Build.Docker.cs
```diff
@@ -32,13 +32,20 @@ public partial class Build
...
-                                                                PowerShellTasks.PowerShell(_ => _
-                                                                                                .EnableNoProfile()
-                                                                                                .SetCommand($"chmod +x '{flavourFolder / platformFolder / flavour}'"));
+                                                                // change the native binaries to be executable in each platform
+                                                                foreach (var binary in binariesToChmodX)
+                                                                {
+                                                                    PowerShellTasks.PowerShell(_ => _
+                                                                                                    .EnableNoProfile()
+                                                                                                    .SetCommand($"chmod +x '{flavourFolder / platformFolder / binary}'"));
```
matt-richardson: Any particular reason you're not using `File.SetUnixFileMode(path, mode)`?

Would be quicker (and cleaner).

build/Build.Docker.cs
```diff
@@ -32,13 +32,20 @@ public partial class Build
                                                             Directory.Move(flavourFolder / "linux-x64",
                                                                 flavourFolder / "linux-amd64");
                                                             Log.Information("Renamed 'linux-x64' folder to 'linux-amd64'");
+
+                                                            //a known list of binaries to chmod +x
```
matt-richardson: not sure you need this comment :)

matt-richardson (approved):
:simples:

A couple of pretty minor suggestions, but looks like it does what it says on the tin 👍

build/Build.Docker.cs
APErebus: Nope. Didn't know it existed until today 😅

Will change

APErebus: Updated in [744a42d](https://github.com/OctopusDeploy/Calamari/pull/1990/commits/744a42d4fc7b0e4e0890098d464e8fe90c9446f8)

build/Build.Docker.cs
APErebus: Removed in [744a42d](https://github.com/OctopusDeploy/Calamari/pull/1990/commits/744a42d4fc7b0e4e0890098d464e8fe90c9446f8)

build/Build.Docker.cs
```diff
@@ -32,13 +38,24 @@ public partial class Build
...
-                                                                                                .EnableNoProfile()
-                                                                                                .SetCommand($"chmod +x '{flavourFolder / platformFolder / flavour}'"));
+                                                                
+                                                                // change the native binaries to be executable in each platform
+                                                                foreach (var binary in binariesToChmodX)
+                                                                {
+                                                                    string binaryPath = flavourFolder / platformFolder / binary;
+                                                                    
+                                                                    //Not all binaries exist in all flavours (The extra win check skips the CA1416 issues)
+                                                                    if (File.Exists(binaryPath) && !OperatingSystem.IsWindows())
```
APErebus, line 54: I didn't realised we only bundled the credential helper with the `Calamari` flavour... so rather than do more complex _stuff_, we just check that the binary exists before chmod'ing

build/Build.Docker.cs
```diff
@@ -12,6 +12,12 @@ public partial class Build
         d =>
             d.Executes(() =>
                        {
+                           if (OperatingSystem.IsWindows())
+                           {
+                               Log.Error("The Docker images can not be built on Windows");
+                               return;
+                           }
```
APErebus, line 19: Top guard so it's clear this can't be run on Windows

PR #1994, opened 2026-06-05
Fix opaque NullReferenceException when a NuGet V3 package version is missing (FD-440)

droyad:
## Background

Deploying with a package set to **download on the target**, from a NuGet **V3** feed, for a **version that doesn't exist on the feed**, failed with an opaque `NullReferenceException` — no hint that the real problem is a missing/unpublished version:

```
System.NullReferenceException: Object reference not set to an instance of an object.
   at NuGet.Commands.SourceRepositoryDependencyProvider.GetPackageDownloaderAsync(...)
   at Calamari.Integration.Packages.NuGet.NuGetV3LibDownloader.DownloadPackage(...)
   ...
   at Calamari.Commands.DownloadPackageCommand.Execute(...)
```

Customer-facing context and the end-to-end product repro are in **FD-440**.

The problem is described in [this PR](https://github.com/OctopusDeploy/NuGet.Client/pull/25) for NuGet.Client. But since we don't want to add more things ot it, and the likelyhood of this getting fixed upstream in a timely manner are low, this PR avoids the problem.

## Verification

Reproduced and fixed end-to-end on a local Octopus Server: a `Run a Script` step running on a worker pool, referencing `Newtonsoft.Json` version `9999.9.9` (which doesn't exist) from a NuGet V3 feed (`https://api.nuget.org/v3/index.json`), with the package acquired on the worker.

**Before (stock Calamari)** — bare NRE wrapped by the retry message:

```
Error | The package Newtonsoft.Json version 9999.9.9 could not be downloaded from the external feed
        'https://api.nuget.org/v3/index.json' after making 5 attempts over 101s...
Error | System.NullReferenceException: Object reference not set to an instance of an object.
Error |   at NuGet.Commands.SourceRepositoryDependencyProvider.GetPackageDownloaderAsync(...)
Error |   at Calamari...NuGetV3LibDownloader.DownloadPackage(...) NuGetV3LibDownloader.cs:line 39
```

**After (this change)** — clear, actionable message, no `NullReferenceException` anywhere in the log:

```
Verbose | Attempt 1 of 5: Package Newtonsoft.Json version 9999.9.9 was not found on the NuGet feed
          'https://api.nuget.org/v3/index.json'. Make sure the package version has been pushed to the feed.
```

Automated coverage: integration test `GivesActionableErrorWhenV3FeedIsMissingTheRequestedVersion` downloads a missing version from `api.nuget.org` — fails with the exact NRE above without this change, passes with it.

**On the retries:** the download still retries all attempts before the deployment fails — left as-is intentionally. A version that's absent right now may legitimately be published to the feed mid-window (e.g. a lagging CI push), so retrying is the desired behaviour. This change only improves the *message*; it doesn't alter the retry policy.

Fixes FD-440

### DIFF
--- a/source/Calamari.Shared/Integration/Packages/NuGet/NuGetV3LibDownloader.cs
+++ b/source/Calamari.Shared/Integration/Packages/NuGet/NuGetV3LibDownloader.cs
@@ -8,7 +8,6 @@ using NuGet.Protocol.Core.Types;
 using System.IO;
 using System.Threading.Tasks;
 using Calamari.Common.Plumbing.Extensions;
-using NuGet.Commands;
 using NuGet.Packaging.Core;
 using Octopus.Versioning;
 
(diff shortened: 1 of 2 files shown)


source/Calamari.Shared/Integration/Packages/NuGet/NuGetV3LibDownloader.cs
```diff
@@ -28,17 +27,32 @@ public static void DownloadPackage(string packageId, IVersion version, Uri feedU
...
+
+                var packageIdentity = new PackageIdentity(packageId, version.ToNuGetVersion());
+
                 string targetTempNupkg = Path.Combine(targetPath, Path.GetRandomFileName());
-                var packageDownloader =  providers.GetPackageDownloaderAsync(new PackageIdentity(packageId, version.ToNuGetVersion()), sourceCacheContext, logger, CancellationToken.None)
-                                                      .GetAwaiter()
-                                                      .GetResult();
+                var packageDownloader = findPackageByIdResource.GetPackageDownloaderAsync(packageIdentity, sourceCacheContext, logger, CancellationToken.None)
+                                                               .GetAwaiter()
+                                                               .GetResult();
```
droyad, line 57: The change is essentially pulling the core of the method shown in [this PR](https://github.com/OctopusDeploy/NuGet.Client/pull/25/changes) inline and adding the null check here - this assumes the linked PR doesn't get merged.

source/Calamari.Shared/Integration/Packages/NuGet/NuGetV3LibDownloader.cs
```diff
@@ -28,17 +27,32 @@ public static void DownloadPackage(string packageId, IVersion version, Uri feedU
...
+                                                              .GetAwaiter()
+                                                              .GetResult();
+
+                var packageIdentity = new PackageIdentity(packageId, version.ToNuGetVersion());
+
                 string targetTempNupkg = Path.Combine(targetPath, Path.GetRandomFileName());
-                var packageDownloader =  providers.GetPackageDownloaderAsync(new PackageIdentity(packageId, version.ToNuGetVersion()), sourceCacheContext, logger, CancellationToken.None)
-                                                      .GetAwaiter()
-                                                      .GetResult();
+                var packageDownloader = findPackageByIdResource.GetPackageDownloaderAsync(packageIdentity, sourceCacheContext, logger, CancellationToken.None)
```
bec-callow-oct, line 55: I noticed `sourceRepository.GetResourceAsync` [can return null](https://github.com/OctopusDeploy/NuGet.Client/blob/5c17618c5a0ae31bb20232cf3de979b341ba5ab6/src/NuGet.Core/NuGet.Protocol/SourceRepository.cs#L149), so we might need null handling for `findPackageByIdResource` as well

bec-callow-oct (changes requested):

source/Calamari.Shared/Integration/Packages/NuGet/NuGetV3LibDownloader.cs
droyad: Confirmed — `GetResourceAsync<T>` returns null when no provider can supply the resource for the feed (rather than throwing). Added a guard for `findPackageByIdResource` in 93a2a2cc4: it now throws an actionable error naming the feed, so that path no longer dereferences null and resurfaces the NRE this PR is removing.

bec-callow-oct (approved):

PR #1998, opened 2026-06-09
Fix CA2200: preserve stack trace on rethrow

matt-richardson:
## What

Changes the rethrow in `CalamariPhysicalFileSystem.MoveFile`'s `UnauthorizedAccessException` handler from `throw unauthorizedAccessException;` to a bare `throw;`.

## Why

`throw unauthorizedAccessException;` reset the exception's stack trace to the catch block, discarding where the `UnauthorizedAccessException` actually originated (the `File.Replace`/`File.Copy` call). 

A bare `throw;` re-throws the in-flight exception with its original stack trace intact — same exception type and message, better diagnostics.

## Impact

- Clears the CA2200 warning (the entire override/rethrow-correctness category in the solution).
- No behavioural change beyond the preserved stack trace; the caught variable is still referenced for logging, so no unused-variable warning.

### DIFF
--- a/source/Calamari.Common/Plumbing/FileSystem/CalamariPhysicalFileSystem.cs
+++ b/source/Calamari.Common/Plumbing/FileSystem/CalamariPhysicalFileSystem.cs
@@ -529,7 +529,7 @@ namespace Calamari.Common.Plumbing.FileSystem
                 LogFileAccess(temporaryReplacement);
                 LogFileAccess(backup);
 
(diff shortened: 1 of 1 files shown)


Jtango18 (changes requested):
We need to discuss whether this is actually a valid solution to the problem it's attempting to solve. I have some concerns that need to be discussed.

Jtango18 (approved):
Re looking at this I think it's probably good to go

PR #2001, opened 2026-06-09
Don't include the dotnet-script.*.zip in built packages

APErebus:
We are unintentionally building the `dotnet-script.*.zip` source zip file into the released packages. This is ~7mb in size and completely unnecessary as we unzip it anyway

### DIFF
--- a/source/Calamari.Scripting/Calamari.Scripting.csproj
+++ b/source/Calamari.Scripting/Calamari.Scripting.csproj
@@ -10,18 +10,6 @@
         <TargetFramework>net8.0</TargetFramework>
         <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     </PropertyGroup>
-
-    <ItemGroup>
-        <Content Include="DotnetScript/*.*">
-            <LinkBase>DotnetScript</LinkBase>
-            <Pack>true</Pack>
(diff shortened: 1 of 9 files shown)


source/IncludeDotNetScript.targets
APErebus, line 1: Consolidated all the same bits of the project files into a single `.targets` file we can share

source/Calamari.Scripting/Calamari.Scripting.csproj
```diff
@@ -10,18 +10,6 @@
...
-        <Content Include="DotnetScript/*.*">
-            <LinkBase>DotnetScript</LinkBase>
-            <Pack>true</Pack>
-            <Visible>false</Visible>
-            <PackagePath>contentFiles/any/any/DotnetScript/</PackagePath>
-            <PackageCopyToOutput>true</PackageCopyToOutput>
-            <PackageFlatten>false</PackageFlatten>
-            <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-        </Content>
-    </ItemGroup>
```
APErebus, line 24: This is where we were including it, but each build where we need it just sources it directly from this project source folder anyway

zentron (approved):

PR #2004, opened 2026-06-10
Migrate to Contract classes for TargetDiscovery

Jtango18:

Tipped off by the comment

```
    /// <summary>
    /// This type and the types it uses are duplicated here from Octopus.Core, because:
    /// a) There is currently no existing project to place code shared between server and Calamari, and
    /// b) We expect a bunch of stuff in the Sashimi/Calamari space to be refactored back into the OctopusDeploy solution soon.
```

This copies the Target Discovery files from Server and puts them in Calamari Contracts and replaces existing executions.

Ideally Server would be updated to leverage these instead of it's in-built classes too


APErebus (approved):
Looks good!	I'm assuming the Context is serialized into a variable as JSON and sent across

PR #2005, opened 2026-06-10
Remove lsof command diagnostic

adam-mccoy:
This reverts #1683.

My understanding is that the diagnostic information provided by `lsof` hasn't proven useful in diagnosing the underlying issue. Executing this command produces hundreds of megabytes of output in task logs, the handling of which impacts test execution performance. If it's not providing value then it'd be best to remove it.


xwipeoutx (approved):
Approving - this hasn't come up on our team in my working memory, and Luke is on leave.  We can revert this easily enough if we need to.

APErebus (approved):

PR #2006, opened 2026-06-10
Change AzureWebAppHealthCheck tests to use static web app

APErebus:
These tests only created azure web apps to be able to check they are there 🙄 

So now just look for a static, known web app


flin-8 (approved):
LGTM

PR #2008, opened 2026-06-10
fix(KustomizeUpdater): non `docker.io` image registries were ignored for image updates

oleksandr-codefresh:
Fix for issue:
Update Argo CD Application Image Tags for kustomize apps does commit any changes during deployment if **image pushed to non-default registry** ( `docker.io` )

```yaml
images:
  - name: vplakydatestregistry.azurecr.io/xeonalex/guesbook-unprivileged
    newTag: "latest"
```

### DIFF
--- a/source/Calamari.Tests/ArgoCD/KustomizeUpdaterTests.cs
+++ b/source/Calamari.Tests/ArgoCD/KustomizeUpdaterTests.cs
@@ -293,6 +293,44 @@ spec:
         }
 
         [Test]
+        public void Process_WithImageOnNonDefaultRegistry_ProducesPatchSoChangesAreCommitted()
+        {
+            // Regression test: an image whose registry differs from the default (e.g. GAR/GCR/ECR)
+            // must still produce a JSON patch. Previously the recorded image reference had its
+            // registry stripped, so CreateJsonPatch re-parsed it, defaulted the registry to
+            // docker.io, failed to match, and returned no patch — meaning HasChanges() was false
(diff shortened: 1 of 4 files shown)


rain-on:
That test is great, exposes the problem nicely 🙏 

Having looked at the code, I'm actually wondering if the real problem is ImageReplacementResult.updatedImageReferences.

Originally this field was just used for sticking into the Git commit message, and so a string was appropriate.
However, its not a _FULL_ representation of what was changed in the file-update, as it strips out the registry data - i.e. the problem we're seeing is due to this loss of information.

I _THINK_ I'd recommend we update the KustomizeImageReplacer to appear:
image

This would ensure we don't strip the registry data - this would also solve this problem, and probably in a nicer way ... I think.

If we went down this road, we'd probably need to apply this to all ImageReplacers, not just Kustomize.

If you wanted to go to an ever better solution - I'd change the updatedImageReferences and alreadyUpToDateImages  from strings to ContainerImageReference - then we're passing back TYPED data, which makes it much safer to use in "business logic" (Rather than plain strings).

eg:
```public class ImageReplacementResult(string updatedContents, HashSet<ContainerImageReference> updatedImageReferences, HashSet<ContainerImageReference> alreadyUpToDateImages, HashSet<string>? unrecognisedKinds = null)```


Final thought - we probably want to apply this to 2026.2 release branch, then let TeamCity merge it forward onto main.

oleksandr-codefresh:
reopened under https://github.com/OctopusDeploy/Calamari/pull/2011/changes

PR #2009, opened 2026-06-11
Change AzureAppService TargetDiscovery integration test to use static resources

APErebus:
These tests could take up to a minute as we were creating a new web-app each time, which when we are just making sure they _exist_ is a bit pointless.

There is now new static web app resources that have the correct tags that get discovered.


flin-8 (approved):
LGTM

PR #2014, opened 2026-06-14
Script Step to support git-repos as dependencies

rain-on:
Update the RunScript Command to stage the contents of any git-repositories specified in the step.

This allows users to specify a git repo as a dependency for a script step (then reference the files by git-repo name).

This can safely be enabled without server-side changes.

If a user creates a script step with git-repos via API or TF - these will not be unpacked due to missing feature toggle (on server side).


APErebus (approved):
This does look suspiciously simple 🤔 

LGTM!

zentron (approved):

PR #2015, opened 2026-06-15
Enable Nullable Ref types on All projects

Jtango18:

Just to see what happens for the time being


zentron:
Nice, this was something I was looking forward to as we moved of full framework.

PR #2027, opened 2026-06-18
Fail AI agent step on deterministic Claude Code failures

sathvikkumar-octo:
Per ADR-004, fail the step on any of: a non-zero CLI exit, a non-success terminal status (is_error or subtype != success), or a populated permission_denials list.

The processor now captures the final result event and a pure ClaudeAgentOutcomeEvaluator decides the outcome in the runner.

As part of testing this I discovered that in built tools appear to be allowed even if not called out in the allowed tools list but tackling that separately.

Resolves MD-2077

### DIFF
--- a/source/Calamari.AiAgent/ClaudeCodeBehaviour/ClaudeCodeCliRunner.cs
+++ b/source/Calamari.AiAgent/ClaudeCodeBehaviour/ClaudeCodeCliRunner.cs
@@ -8,7 +8,6 @@ using System.Text.Json;
 using System.Text.Json.Serialization;
 using System.Threading;
 using System.Threading.Tasks;
-using Calamari.Common.Commands;
 using Calamari.Common.Plumbing.Logging;
 
 namespace Calamari.AiAgent.ClaudeCodeBehaviour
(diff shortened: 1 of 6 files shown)


zentron:
Looks like I get a failure which is a good sign
image

source/Calamari.AiAgent/ClaudeCodeBehaviour/ClaudeCodeCliRunner.cs
```diff
@@ -68,7 +65,9 @@ public async Task<string> RunAsync(ClaudeCommandArgsBuilder argsBuilder,
                 log.NewOctopusArtifact(movedFilePath, "claude-agent-verbose.log", fileInfo.Length);
             }
 
-            return stdoutTask.Result.ToString();
+            new ClaudeAgentOutcomeEvaluator().EnsureSuccessful(process.ExitCode, streamProcessor.Result);
```
zentron: Any time we have a stateless class with one method, I figure it makes more sense to just make it a static method\class, we arent using any of the benifits of an instance class, and its presumably still simple enough we dont want to pull out via dependency injection

zentron (approved):

PR #2028, opened 2026-06-19
Throw when package contents paths are invalid

bec-callow-oct:

This PR closes a security risk identified [here](https://linear.app/octopus/issue/FD-504/sev-2-medium-or-malicious-nuget-package-being-used-to-achieve-path). Review the issue or Linear ticket for details.

Fixes FD-504
Fixes https://github.com/OctopusDeploy/OctopusDeploy/issues/44460

Tested with both valid and invalid packages and can demonstrate the fix locally

### DIFF
--- a/source/Calamari.Common/Features/Packages/NuGet/NupkgExtractor.cs
+++ b/source/Calamari.Common/Features/Packages/NuGet/NupkgExtractor.cs
@@ -52,6 +52,8 @@ namespace Calamari.Common.Features.Packages.NuGet
                     if (IsExcludedPath(unescapedKey))
                         continue;
 
+                    PackageExtractorUtils.ThrowIfPathTraversalAttempted(unescapedKey, directory);
+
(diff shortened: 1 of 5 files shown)


source/Calamari.Common/Features/Packages/NuGet/NupkgExtractor.cs
```diff
@@ -52,6 +52,8 @@ public int Extract(string packageFile, string directory)
                     if (IsExcludedPath(unescapedKey))
                         continue;
 
+                    PackageExtractorUtils.ThrowIfPathTraversalAttempted(unescapedKey, directory);
+
                     var targetDirectory = Path.Combine(directory, Path.GetDirectoryName(unescapedKey) ?? string.Empty);
```
octopus-hideaki, line 57: Is it worth raising with whoever owns Calamari to set up a convention to safely combine paths? I can see new additions to the codebase introducing new vectors for path traversal.

Outside the scope of the PR.

octopus-hideaki (approved):
Looks good. I am curious how we could prevent more cases like this elsewhere.

PR #2030, opened 2026-06-21
Separate Azure App Service external-dependency tests from unit tests

zentron:
## Background

The `Calamari.AzureAppService.Tests` suite mixed fast, deterministic checks of Calamari's own logic with slow tests that authenticate against and provision **real Azure**. Interleaved, the whole suite effectively needed Azure credentials even for logic that doesn't.

This branch separates them, working towards a simple rule: **if a test doesn't need Azure it should be a unit test; if it does, it should be a thin, clearly-marked smoke test.**

## What changed

- Add `TestCategory.ExternalCloudIntegration` and tag the real-cloud base classes, so CI can run the credential-free suite separately from the live-Azure one (commit 1).
- Move the cloud-hitting fixtures into an `ExternalCloudIntegration/` folder so the split is visible in the tree.
- Extract Azure calls behind seams (`IAzureWebAppDiscoverer`, `IAzureAppServiceContainerConfigurer`) so target-discovery and container-deploy logic can be unit-tested with mocks.
- Add credential-free unit fixtures for tag-matching, container-deploy logic, and package-provider routing.
- Remove redundant cloud tests (each explained inline): async-polling duplicates, WAR/JAR deploys, the misnamed parallel-deploy file-lock test, and the Windows container OS duplicate.
- Consolidate the test base classes into a 3-level hierarchy (shared auth → provisioned vs static resources).

Net: the credential-free suite grows, the real-cloud suite shrinks to genuine smoke tests.

## Testing

- Credential-free suite passes locally: `dotnet test --filter "TestCategory!=ExternalCloudIntegration"` → 30/30.
- :warning: The `ExternalCloudIntegration` fixtures were **not** run this branch (no Azure creds in dev) — verified by compilation and review only. The split `OneTimeSetUp` in the consolidated base classes (base authenticates, derived provisions/looks-up) wants one real CI run to confirm ordering.

## How to review

Two commits, in order: the first only categorises; the second does the refactor. Production change is small and mostly mechanical (the two seams + their DI registration in `Program.cs`); the rest is test moves. I've left inline comments on each removal/consolidation.

:warning: No Server change required (Calamari-internal test/infra refactor, no contract change).


source/Calamari.AzureAppService.Tests/ExternalCloudIntegration/AppServiceBehaviourFixture.cs
```diff
@@ -16,30 +16,29 @@
...
 {
+    // The two nested fixtures cover two distinct Azure deployment targets - a Web App and a (Linux) Function
+    // App - rather than two operating systems. Calamari runs the same zip-deploy code path for both; the split
+    // exists because the Function App is a different resource type (function runtime, run-from-package), not
+    // because the OS differs.
     public class AppServiceBehaviorFixture
     {
         [TestFixture]
-        public class WhenUsingAWindowsDotNetAppService : AppServiceIntegrationTest
+        public class WhenDeployingToAWebAzureApp : AzureAppServiceWithProvisionedResourcesTestBase
```
zentron, line 36: Three real-cloud tests were removed from this fixture, none of which exercised distinct Calamari logic:
- **WAR/JAR deploys** — both route through the same `JavaPackageProvider`, differing only by Kudu upload URL; that routing is now unit-tested in `PackageProviderFactoryFixture`.
- **`*_WithAsyncDeploymentAndPolling`** — behavioural duplicates. The async path runs whenever `SupportsAsynchronousDeployment` is true (zip/nuget), so the baseline tests already cover `UploadZipAndPollAsync`; the feature-toggle arrange was a no-op (read toggles, write them straight back).
- **`DeployToTwoTargetsInParallel_Succeeds`** — guarded package-file share mode (a generic Calamari staging concern, not Azure-specific) via a full cloud deploy that asserted only `Outcome==Successful`.

source/Calamari.AzureAppService.Tests/PackageProviderFactoryFixture.cs
```diff
@@ -0,0 +1,39 @@
...
+using Calamari.Common.Plumbing.Variables;
+using Calamari.Testing.Helpers;
+using FluentAssertions;
+using NSubstitute;
+using NUnit.Framework;
+
+namespace Calamari.AzureAppService.Tests;
+
+[TestFixture]
+public class PackageProviderFactoryFixture
```
zentron, line 14: Replacement coverage for the removed WAR/JAR cloud deploy tests. The extension -> packaging strategy / Kudu URL / sync-async mapping is the only Calamari-owned decision in an otherwise Azure-bound deploy, so it's unit-tested here rather than exercised through real deploys of each package type.

source/Calamari.AzureAppService.Tests/ExternalCloudIntegration/AzureAppServiceDeployContainerBehaviourFixture.cs
```diff
@@ -13,169 +13,26 @@
...
-                vars.Add(SpecialVariables.Action.Package.PackageId, "e2eteam/sample-apiserver");
-                vars.Add(SpecialVariables.Action.Package.Image, "e2eteam/sample-apiserver:1.17");
-                vars.Add(SpecialVariables.Action.Package.PackageVersion, "1.17");
-                vars.Add(SpecialVariables.Action.Azure.DeploymentType, "Container");
-            }
-        }
-
-        [TestFixture]
-        public class WhenUsingALinuxAppService : AppServiceIntegrationTest
+        public class WhenUsingALinuxAzureAppService : AzureAppServiceWithProvisionedResourcesTestBase
```
zentron, line 35: Collapsed from Windows + Linux to Linux only. The Windows/Linux `FxVersion` branch is unit-tested both ways in `AzureAppServiceContainerDeployBehaviourUnitTestFixture`, so one OS is enough to confirm the real-cloud ARM round-trip. Trade-off: the Windows `FxVersion` PATCH is no longer hit against real Azure (covered by the unit test's Windows branch).

source/Calamari.AzureAppService.Tests/AzureAppServiceContainerDeployBehaviourUnitTestFixture.cs
```diff
@@ -0,0 +1,131 @@
...
+using Calamari.Testing.Helpers;
+using FluentAssertions;
+using NSubstitute;
+using NUnit.Framework;
+using AccountVariables = Calamari.AzureAppService.Azure.AccountVariables;
+
+namespace Calamari.AzureAppService.Tests;
+
+[TestFixture]
+public class AzureAppServiceContainerDeployBehaviourUnitTestFixture
```
zentron, line 19: New unit fixture covering the OS-branch, registry-settings and slot-targeting logic that previously needed four real-cloud container tests. Uses a mocked `IAzureAppServiceContainerConfigurer`; the cloud fixture now keeps a single Linux smoke test.

source/Calamari.AzureAppService.Tests/ExternalCloudIntegration/AzureAppServiceTestBase.cs
```diff
@@ -0,0 +1,85 @@
...
+namespace Calamari.AzureAppService.Tests.ExternalCloudIntegration
+{
+    /// <summary>
+    /// Shared base for fixtures that talk to real Azure. Authenticates with the test service principal and
+    /// creates the <see cref="ArmClient" /> in a OneTimeSetUp that runs before any derived OneTimeSetUp, so
+    /// subclasses can create or look up their resources against the established connection. Carries the
+    /// ExternalCloudIntegration category, which all descendants inherit.
+    /// </summary>
+    [Category(TestCategory.ExternalCloudIntegration)]
+    public abstract class AzureAppServiceTestBase
```
zentron, line 25: Consolidated root base: the auth + `ArmClient` setup that was duplicated across the two former base classes now lives here once, with subclasses adding only their resource lifecycle (provisioned vs static). Relies on NUnit running this base `OneTimeSetUp` before the derived one — that ordering is the main thing worth confirming in a real-cloud CI run.

source/Calamari.AzureAppService/Azure/IAzureWebAppDiscoverer.cs
```diff
@@ -0,0 +1,41 @@
...
+using Calamari.CloudAccounts;
+
+namespace Calamari.AzureAppService.Azure
+{
+    /// <summary>
+    /// Enumerates the Azure web apps and slots visible to an account. This is the single point at
+    /// which target discovery talks to Azure; mocking it lets the tag-matching and target-creation
+    /// logic in <see cref="Behaviors.TargetDiscoveryBehaviour" /> be tested without a real Azure connection.
+    /// </summary>
+    public interface IAzureWebAppDiscoverer
```
zentron, line 14: One of two seams introduced this PR (the other is `IAzureAppServiceContainerConfigurer`). Extracting the single Azure call behind an interface is what lets the surrounding tag-matching / slot / service-message logic be unit-tested with a mock instead of a live subscription.

source/Calamari.AzureAppService.Tests/ExternalCloudIntegration/TargetDiscoveryBehaviourWithStaticResourcesTestFixture.cs
```diff
@@ -0,0 +1,31 @@
...
+// The tag-matching, slot-handling and service-message logic that consumes these resources is
+// covered without Azure in TargetDiscoveryBehaviourUnitTestFixture via a mocked IAzureWebAppDiscoverer.
+[TestFixture]
+public class TargetDiscoveryBehaviourWithStaticResourcesTestFixture : AzureAppServiceWithStaticResourcesTestBase
+{
+    // https://portal.azure.com/#@octopusdeploy.onmicrosoft.com/resource/subscriptions/cf21dc34-73dc-4d7d-bd86-041884e0bc75/resourcegroups/calamari-testing-static-rg/providers/Microsoft.Web/sites/calamari-testing-static-target-discovery/appServices
+    const string WebAppName = "calamari-testing-static-target-discovery";
+
+    [Test]
+    public async Task DiscoverWebAppsAndSlots_ReturnsTheStaticTestWebApp()
```
zentron, line 19: What remains of the old 237-line integration fixture: a single smoke test that the real Azure Resource Graph query + auth still work. All the tag-matching/slot/service-message scenarios it used to run against Azure now live in `TargetDiscoveryBehaviourUnitTestFixture`.

source/Calamari.AzureAppService.Tests/Calamari.AzureAppService.Tests.csproj
```diff
@@ -28,13 +28,8 @@
...
   <ItemGroup>
-    <None Update="sample.1.0.0.war">
-      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
-    </None>
-    <None Update="sample4-1.0.0.jar">
-      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
-    </None>
-    <None Update="functionapp.1.0.0.zip">
+    <None Update="ExternalCloudIntegration/functionapp.1.0.0.zip">
+      <Link>functionapp.1.0.0.zip</Link>
```
zentron, line 32: `functionapp.1.0.0.zip` moved into the `ExternalCloudIntegration/` folder alongside the only test that uses it; the `<Link>` keeps it copying to the output root so the test still resolves it via the assembly directory. The `sample.1.0.0.war` / `sample4-1.0.0.jar` entries were removed — those binaries were orphaned when the WAR/JAR cloud tests were deleted.

zentron (commented):
Inline notes on the test moves — pointing out why things were removed, consolidated, or pushed down to unit tests, since a red diff doesn't carry that on its own.

PR #2031, opened 2026-06-21
Add settings for claude sandbox

eddymoulton:
Related to MD-2116
Related to MD-2135

Changes from many separate properties to a single json blob for sandbox settings, configured via server.

This matches the strategy we're doing for permissions configuration, where users can enter the values in json via the UI, which more closely matches how they use Claude locally + is more flexible.

We remove a lot of defaults in this change, instead we can set these on the frontend. This also gives users more flexibility.

While I'm here, I renamed Srt references to SandboxRuntime (requested by Liam)


source/Calamari.AiAgent/ClaudeCodeBehaviour/InvokeClaudeCodeBehaviour.cs
```diff
@@ -89,13 +89,13 @@ public async Task Execute(RunningDeployment context)
             case SandboxMode.Bash when RuntimeInformation.IsOSPlatform(OSPlatform.Windows):
                 throw new CommandException($"Sandbox mode '{sandboxMode}' is not supported on Windows workers; use 'None' or run on Linux/macOS.");
             case SandboxMode.Bash:
-                BashSandboxSettingsWriter.Write(workingDir, variables);
+                argsBuilder.WithBashSettingsPath(SandboxSettingsWriter.WriteBashSettings(workingDir, variables));
                 break;
-            case SandboxMode.Srt when RuntimeInformation.IsOSPlatform(OSPlatform.Windows):
+            case SandboxMode.SandboxRuntime when RuntimeInformation.IsOSPlatform(OSPlatform.Windows):
                 throw new CommandException($"Sandbox mode '{sandboxMode}' is not supported on Windows workers; use 'None' or run on Linux/macOS.");
```
liam-mackie: Nit: we don't officially support MacOS do we?
```suggestion
                throw new CommandException($"Sandbox mode '{sandboxMode}' is not supported on Windows workers; use 'None' or run on Linux.");
```

source/Calamari.AiAgent/ClaudeCodeBehaviour/SandboxRuntimeVersionGuard.cs
```diff
@@ -8,7 +8,7 @@ namespace Calamari.AiAgent.ClaudeCodeBehaviour;
 
 // sandbox-runtime must be >= 0.0.55 because older versions fail open on network access,
 // allowing sandboxed processes to make outbound network connections they should not be able to.
-public static class SrtVersionGuard
+public static class SandboxRuntimeVersionGuard
 {
     static readonly Version Minimum = new(0, 0, 55);
     static readonly Regex SemVer = new(@"(\d+)\.(\d+)\.(\d+)", RegexOptions.Compiled);
```
liam-mackie: Super nit: this really feels like something we already know how to do and have the tools to do without using regex, but it's already in so 🤷

source/Calamari.AiAgent/ClaudeCodeBehaviour/SandboxSettingsWriter.cs
```diff
@@ -0,0 +1,40 @@
+using System.IO;
+using Calamari.Common.Commands;
+using Calamari.Common.Plumbing.Variables;
+
+namespace Calamari.AiAgent.ClaudeCodeBehaviour;
+
+public static class SandboxSettingsWriter
+{
+    const string ClaudeDirName = ".claude";
```
liam-mackie, line 9: Should this be something we put at a top-level static somewhere instead? I've used this exact pattern too 😓

liam-mackie (approved):
Nits but lgtm

source/Calamari.AiAgent/ClaudeCodeBehaviour/SandboxSettingsWriter.cs
eddymoulton: I'll let you move it in your PR as a clean up

PR #2039, opened 2026-06-24
Claude: per model cost usage

flin-8:
Send per-model usage figures to Server so we can break down the usage

Relates to [MD-2089: Audit records are created with enough information to understand the step](https://linear.app/octopus/issue/MD-2089/audit-records-are-created-with-enough-information-to-understand-the)

### DIFF
--- a/source/Calamari.AiAgent/ClaudeCodeBehaviour/ClaudeCodeStreamProcessor.cs
+++ b/source/Calamari.AiAgent/ClaudeCodeBehaviour/ClaudeCodeStreamProcessor.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Text;
 using System.Text.Json;
 using Calamari.AiAgent.ClaudeCodeBehaviour.JsonResponseModels;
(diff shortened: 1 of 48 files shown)


source/Calamari.AiAgent/ClaudeCodeBehaviour/ClaudeCodeStreamProcessor.cs
```diff
@@ -234,6 +235,8 @@ void HandleResultEvent(ResultStreamEvent evt)
                 log.Info(evt.Result);
             }
 
+            log.Info($"AI Agent Usage — Cost: ${evt.CostUsd} USD (total: ${evt.TotalCostUsd}), Duration: {evt.DurationMs}ms, Turns: {evt.NumTurns}");
```
eddymoulton: This is inconsequential, but did we want to stick with the specific naming since we're touching this anyway?

```suggestion
            log.Info($"Claude Code Usage — Cost: ${evt.CostUsd} USD (total: ${evt.TotalCostUsd}), Duration: {evt.DurationMs}ms, Turns: {evt.NumTurns}");
```

eddymoulton (approved):

source/Calamari.AiAgent/ClaudeCodeBehaviour/ClaudeCodeStreamProcessor.cs
flin-8: Nice pickup

PR #2045, opened 2026-06-26
Let agent signal step failure via a tagged block

zentron:
## Background

The AI agent step runs the Claude CLI, which always exits `0` — even when the agent semantically failed at the task. That left a gap: when a workflow author writes "fail the deployment if the smoke test doesn't pass", the agent could detect the condition but had no way to make Octopus mark the step as failed. `ClaudeAgentOutcomeEvaluator` only inspected the process exit code and the CLI's structured result (`is_error`, `subtype`, permission denials), none of which capture an *intentional* failure.

## Results

Adds a deterministic agent→managed-code failure signal:

- **New `octopus-fail-deployment` skill** — instructs the agent, when the user expressed a failure condition that's been met, to emit an `<octopus-task-failed>…</octopus-task-failed>` block with an operator-facing reason. Absence of the block = success (unchanged default).
- **`ClaudeAgentOutcomeEvaluator`** — scans the final result text for the block and throws a `CommandException` with the captured reason (generic fallback when empty), checked before the generic CLI-status checks so an intentional failure surfaces a clear message.

Design notes:
- Uses a **stdout tag** rather than a marker file, so it works even when the agent is sandboxed without write permissions.
- The **required closing tag doubles as a completeness guard** — a truncated, unclosed block won't match, so it can't be mistaken for a deliberate failure.
- Tag wording is task-neutral (not deployment-specific) so it reads correctly for runbooks too.

No Server change required — the failure propagates through the existing non-zero-exit path.

## Testing

`ClaudeAgentOutcomeEvaluatorFixture` — 14 unit tests passing, including reason capture, multi-line reasons, empty/self-closing blocks, a block embedded in larger output, precedence over a non-success subtype, and that an unclosed (truncated) block does **not** fail the step.

## How to review

Core logic is the regex + check in `ClaudeAgentOutcomeEvaluator.cs`; the skill markdown is the agent-facing contract. The rest is test coverage.

Resolves: [#MD-2151](https://linear.app/octopus/issue/MD-2151/allow-users-to-explicitly-request-the-step-to-fail)

### DIFF
--- a/source/Calamari.AiAgent/ClaudeCodeBehaviour/ClaudeAgentOutcomeEvaluator.cs
+++ b/source/Calamari.AiAgent/ClaudeCodeBehaviour/ClaudeAgentOutcomeEvaluator.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Linq;
+using System.Text.RegularExpressions;
 using Calamari.AiAgent.ClaudeCodeBehaviour.JsonResponseModels;
 using Calamari.Common.Commands;
 
@@ -7,6 +8,14 @@ namespace Calamari.AiAgent.ClaudeCodeBehaviour
 {
     public static class ClaudeAgentOutcomeEvaluator
     {
+        // The agent emits this tagged block (see the octopus-fail-deployment skill) when a user-specified
+        // failure condition has been met. The CLI still exits 0, so this is the only way to detect an intentional
+        // failure from the outside. The inner text is the operator-facing reason; requiring the closing tag also
+        // confirms the agent finished writing the block rather than being truncated mid-message.
+        static readonly Regex FailureSignal = new(
+            @"<octopus-task-failed\s*(?:/>|>(?<reason>.*?)</octopus-task-failed>)",
+            RegexOptions.Singleline | RegexOptions.IgnoreCase | RegexOptions.Compiled);
+
         public static void EnsureSuccessful(int exitCode, ResultStreamEvent? result)
         {
             if (exitCode != 0)
@@ -20,6 +29,17 @@ namespace Calamari.AiAgent.ClaudeCodeBehaviour
             if (result == null)
             {
                 return;
+            }
+
+            // An intentional, user-requested failure takes precedence: the run is otherwise "successful" at the
+            // CLI level, so check the agent's signal before the generic CLI-status checks to surface a clear reason.
+            if (result.Result is { } text && FailureSignal.Match(text) is { Success: true } match)
+            {
+                var reason = match.Groups["reason"].Value.Trim();
+                throw new CommandException(
+                    string.IsNullOrEmpty(reason)
+                        ? "The agent signalled that the step should fail."
+                        : $"The agent signalled that the step should fail: {reason}");
             }
 
             if (result.IsError == true || !"success".Equals(result.Subtype, StringComparison.OrdinalIgnoreCase))
--- a/source/Calamari.AiAgent/ClaudeCodeBehaviour/ClaudeAgentOutcomeEvaluator.cs
+++ b/source/Calamari.AiAgent/ClaudeCodeBehaviour/ClaudeAgentOutcomeEvaluator.cs
@@ -8,11 +8,9 @@ namespace Calamari.AiAgent.ClaudeCodeBehaviour
 {
     public static class ClaudeAgentOutcomeEvaluator
     {
-        // The agent emits this tagged block (see the octopus-fail-deployment skill) when a user-specified
-        // failure condition has been met. The CLI still exits 0, so this is the only way to detect an intentional
-        // failure from the outside. The inner text is the operator-facing reason. We accept either a self-closing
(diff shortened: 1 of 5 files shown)


source/Calamari.Common/Plumbing/Pipeline/PipelineCommand.cs
```diff
@@ -67,7 +70,7 @@ public async Task Execute(ILifetimeScope lifetimeScope, IVariables variables)
             }
             catch (Exception installException)
             {
-                Console.Error.WriteLine("Running rollback behaviours...");
+                log.Verbose("Running rollback behaviours...");
```
zentron, line 73: This was annoying as hell. There is no rollback behaviour taking place, so why make it so prominent.

source/Calamari.Common/Plumbing/Pipeline/PipelineCommand.cs
```diff
@@ -67,7 +70,7 @@ public async Task Execute(ILifetimeScope lifetimeScope, IVariables variables)
...
```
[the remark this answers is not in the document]
zentron: This felt wrong as an error, looking at history I can see no reason why it was listed as an error before.

source/Calamari.Common/Plumbing/Pipeline/PipelineCommand.cs
```diff
@@ -78,7 +81,7 @@ public async Task Execute(ILifetimeScope lifetimeScope, IVariables variables)
                 }
                 catch (Exception rollbackException)
                 {
-                    Console.Error.WriteLine(rollbackException);
```
zentron, line 81: We should generally not be writing to Console in Calamari as it makes testing more difficult

source/Calamari.AiAgent/ClaudeCodeBehaviour/ClaudeAgentOutcomeEvaluator.cs
```diff
@@ -1,27 +1,45 @@
...
 
 namespace Calamari.AiAgent.ClaudeCodeBehaviour
 {
     public static class ClaudeAgentOutcomeEvaluator
     {
+        // The agent emits this tagged block (see the octopus-fail-deployment skill) when a user-specified
+        // failure condition has been met. The CLI still exits 0, so this is the only way to detect an intentional
+        // failure from the outside. The inner text is the operator-facing reason. We accept either a self-closing
+        // <octopus-task-failed/> (a complete signal with no reason) or a paired block ending in </octopus-task-failed>;
+        // an opening tag without its closing tag is treated as a truncated message and ignored.
```
eddymoulton: Wordy

```suggestion
        // The agent emits this tagged block when a user-specified failure condition has been met
        // (see the octopus-fail-deployment skill)
        // Matches either a self-closing <octopus-task-failed/> or a paired block ending in </octopus-task-failed>, capturing the contents ass the reason.
```

source/Calamari.AiAgent/ClaudeCodeBehaviour/DefaultContext/Skills/octopus-fail-deployment.md
```diff
@@ -0,0 +1,31 @@
...
+For example:
+
+<octopus-task-failed>
+Smoke test returned HTTP 500 from /health after 3 retries.
+</octopus-task-failed>
+
+## Rules
+
+- Emit the block **only** when the user expressed a failure condition AND you have determined it is met. If the condition was not met, say nothing special and let the step succeed.
+- Always write a **complete** block — either a paired block ending in `</octopus-task-failed>` or a self-closing `<octopus-task-failed/>`. A closed tag is how Octopus confirms the message is whole — if you open the block but stop before closing it, the failure will not be detected, so finish the block before ending your turn.
```
eddymoulton, line 26: RE: self closing block

I would be tempted to remove that as an option. Makes this whole thing more consistent and the matching regex simpler.

I'm a bit concerned about removing flexibility that Claude might like to take advantage of however.
I'll leave it with you to decide if you think cutting that down makes things simpler enough to warrant the change.

eddymoulton (approved):
Couple of nits, but looks good

source/Calamari.AiAgent/ClaudeCodeBehaviour/DefaultContext/Skills/octopus-fail-deployment.md
zentron: Interesting you say that. I originally had it not allow closing tags but then Claude pointed out it might make it more likely to use it when there happens to be no specific reason to apply. 
Ill leave it in for now, but if it creates any false positives/negatives we can reconsider.

eddymoulton: > Claude pointed out it might make it more likely to use it when there happens to be no specific reason to apply

That's reason enough if Claude thinks it might be a problem without it.

PR #2046, opened 2026-06-28
Throw when transcript exceeds a certain size

flin-8:
Adding a guard for now so we don't send huge transcripts and overload Server's storage

Resolves [MD-2102: Implement a hard limit to prevent runaway storage costs](https://linear.app/octopus/issue/MD-2102/implement-a-hard-limit-to-prevent-runaway-storage-costs)

### DIFF
--- a/source/Calamari.AiAgent/ClaudeCodeBehaviour/ClaudeCodeCliRunner.cs
+++ b/source/Calamari.AiAgent/ClaudeCodeBehaviour/ClaudeCodeCliRunner.cs
@@ -60,6 +60,8 @@ public class ClaudeCodeCliRunner(ILog log)
 
         if (File.Exists(verboseLogPath))
         {
+            ThrowIfTranscriptTooBig(verboseLogPath);
+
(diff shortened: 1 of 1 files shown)


source/Calamari.AiAgent/ClaudeCodeBehaviour/ClaudeCodeCliRunner.cs
```diff
@@ -60,6 +60,8 @@ public async Task<string> RunAsync(ClaudeCommandArgsBuilder argsBuilder,
 
         if (File.Exists(verboseLogPath))
         {
+            ThrowIfTranscriptTooBig(verboseLogPath);
```
zentron, line 63: This throws once the task has completed already right? 
Wonder if we can keep track of this as we append the transcript and throw mid-execution, just to avoid this being a way to hide something malicious

zentron (changes requested):

source/Calamari.AiAgent/ClaudeCodeBehaviour/ClaudeCodeCliRunner.cs
flin-8: Keep track as in audit this?

PR #2057, opened 2026-07-02
Fix: resolve $ref value files for Helm-typed Argo CD ref sources

HuyPhanNguyen:
# Background

The Update Argo CD Image Tags step updates image tags in a multi-source Argo CD Application's Helm value files. When a Helm source references a value file in another source using Argo CD's `$ref` syntax (for example `$app/chart/values-b2b-dev.yaml`), the referenced file lives in the ref source's repository. The step only resolved that reference when the ref source was typed Directory. Argo CD types a ref source as Helm whenever the ref's path contains a `Chart.yaml`, and in that case the referenced value file was skipped: the image tag never changed and the deployment reported "Nothing to update".

# Results

The step now resolves `$ref`-prefixed value files for a ref source regardless of the source type Argo CD reports.

Fixes FD-530

## Before

For a multi-source app with a chart source referencing `$app/chart/values-b2b-dev.yaml` and an `app` ref source whose path contained a `Chart.yaml` (so Argo CD typed it Helm), the step cloned the ref source's repository and did nothing. The referenced file kept its old image tag and the application reported "Nothing to update".

## After

The same app updates the image tag in the referenced file and commits it. A source carrying a ref is routed to `RefUpdater` whether Argo CD typed it Helm or Directory, so the `$ref` reference is resolved either way.

## Implementation

`ApplicationSourceUpdater.CreateSpecificUpdater` now checks `Source.Ref != null` before `SourceType`, so any ref source uses `RefUpdater` (which resolves `$ref` value files). Sources without a ref are unchanged, and Directory + ref already used `RefUpdater`, so the only behaviour change is that a Helm-typed ref source now resolves its referenced files. This matches `RefUpdater`'s existing contract, which updates only referenced files for a ref source and directs users to a separate source for files under the path.

Added test `RefSourceTypedAsHelm_StillResolvesRefAndUpdatesReferencedFile`.

Note on local test runs: nine existing file-content tests in `UpdateArgoCDAppImagesInstallConventionHelmTests` fail on Windows on a trailing-whitespace assertion in `AssertFileContents`. This is a pre-existing line-ending artifact unrelated to this change, confirmed by running the class on `main` with these changes stashed (the same nine fail). They pass on CI.

# Reducing risk

## Automated Tests
- [x] Run full chain build (PRs targeting `main` `release/*` branches only)

# Pre-requisites

- [ ] I have read [How we use GitHub Issues](https://github.com/OctopusDeploy/Issues/blob/main/docs/CONTRIBUTING.internal.md) for help deciding when and where it's appropriate to make an issue.
- [ ] I have considered informing or consulting the right people, according to the [ownership map](https://whimsical.com/ownership-map-NzbiD4HJyvhC9jNJNfS6TG).
- [ ] I have considered the [appropriate target version for this PR](https://octopushq.atlassian.net/wiki/spaces/RND/pages/1145896982/Which+versions+should+I+patch#Process).
- [ ] I have considered appropriate testing for my change.
  - Automated testing / Exploratory testing / Nothing required?
- [ ] I have considered manually testing my changes on a [branch instance](https://octopushq.atlassian.net/wiki/spaces/RND/pages/1262256152/Use+a+Core+Feature+Branch+Instance) to check correctness, stability and performance on Octopus Cloud.
- [ ] I have considered safety nets to reduce any time-to-recovery for my change.

### DIFF
--- a/source/Calamari.Tests/ArgoCD/Commands/Conventions/UpdateArgoCDAppImagesInstallConventionHelmTests.cs
+++ b/source/Calamari.Tests/ArgoCD/Commands/Conventions/UpdateArgoCDAppImagesInstallConventionHelmTests.cs
@@ -1359,6 +1359,95 @@ service:
         }
 
         [Test]
+        public void RefSourceTypedAsHelm_StillResolvesRefAndUpdatesReferencedFile()
+        {
+            // FD-530: when a ref source's `path` contains a Chart.yaml, ArgoCD reports its
+            // SourceType as Helm rather than Directory. The $ref-referenced values file must
+            // still be updated - it must not be skipped just because the ref source is Helm-typed.
+            var updater = CreateConvention();
+            var variables = new CalamariVariables
+            {
+                [PackageVariables.IndexedImage("nginx")] = "index.docker.io/nginx:1.27.1",
+                [PackageVariables.IndexedPackagePurpose("nginx")] = "DockerImageReference",
+                [ProjectVariables.Slug] = ProjectSlug,
+                [DeploymentEnvironment.Slug] = EnvironmentSlug,
+            };
+            var runningDeployment = new RunningDeployment(null, variables);
+            runningDeployment.CurrentDirectoryProvider = DeploymentWorkingDirectory.StagingDirectory;
+            runningDeployment.StagingDirectory = tempDirectory;
+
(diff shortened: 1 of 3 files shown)


flin-8 (approved):
LGTM

PR #2060, opened 2026-07-06
Update injection check prompt

liam-mackie:
This PR fixes a bug where the anthropic api key was being set to the octostache replacement variable, and confused as a legitimate secret.


eddymoulton (approved):

eddymoulton:
Fixes MD-2270

PR #2062, opened 2026-07-06
Add variable to disable the Docker credential helper per action

matt-richardson:
## Summary

_Follow-up to #2059 and #2061._

Adds a new escape-hatch variable `Octopus.Action.Package.DisableDockerCredentialHelper`. When set to `true`, forces Docker/OCI acquisition down the plain `docker login` path.

This is a per-action escape hatch for runtimes where the credential helper doesn't work today (notably podman 3.x workers (see [Issues#10128](https://github.com/OctopusDeploy/Issues/issues/10128))) — users can opt a specific action out via a variable.

## Behaviour

The credential helper runs only when:
- the toggle is on, and
- the variable is not set. 

The variable does not disable `docker login` itself — login still happens, via the old code path (that logs the plaintext warning).

### DIFF
--- a/source/Calamari.Shared/Integration/Packages/Download/DockerImagePackageDownloader.cs
+++ b/source/Calamari.Shared/Integration/Packages/Download/DockerImagePackageDownloader.cs
@@ -1,18 +1,17 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Reflection;
 using System.Runtime.InteropServices;
 using Calamari.Common.Commands;
 using Calamari.Common.Features.Packages;
 using Calamari.Common.Features.Processes;
 using Calamari.Common.Features.Scripting;
 using Calamari.Common.FeatureToggles;
-using Calamari.Common.Features.Scripts;
 using Calamari.Common.Plumbing;
 using Calamari.Common.Plumbing.FileSystem;
 using Calamari.Common.Plumbing.Logging;
 using Calamari.Common.Plumbing.Variables;
+using Calamari.Deployment;
 using Newtonsoft.Json.Linq;
 using Octopus.Versioning;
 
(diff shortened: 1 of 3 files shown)


source/Calamari.Shared/Integration/Packages/Download/DockerImagePackageDownloader.cs
```diff
@@ -60,7 +59,7 @@ public DockerImagePackageDownloader(IScriptEngine scriptEngine,
             this.variables = variables;
             this.log = log;
             this.feedLoginDetailsProviderFactory = feedLoginDetailsProviderFactory;
-            this.useCredentialHelper = OctopusFeatureToggles.UseDockerCredentialHelperFeatureToggle.IsEnabled(variables);
+            this.useCredentialHelper = OctopusFeatureToggles.UseDockerCredentialHelperFeatureToggle.IsEnabled(variables) && !variables.GetFlag(SpecialVariables.Package.DisableDockerCredentialHelper, false);
```
zentron, line 62: whats the plans with the feature toggle? Are we able to roll it out? 
If not it should probably become an explicit setting

zentron (changes requested):
Looking for some guidance on the future of this capability.

When I see toggles combined with variable flags something feels off.

source/Calamari.Shared/Integration/Packages/Download/DockerImagePackageDownloader.cs
matt-richardson: The feature toggle is at 100%, but a customer using an old version of podman (3.x) has hit a bug, and doesn't seem to be able to easily update to the version of podman that would fix it.

Specifically, podman 4.x, in conjunction with #2059 works as it should.

Without version sniffing, this is the best "get-out-of-jail" card that I could think of.

zentron (approved):
Given toggle will soon go then this change LGTM

PR #2063, opened 2026-07-07
Claude step: Fix bug with SRT where the claude settings flag was passed to SRT

flin-8:
When running Claude inside SRT, the deployment fails because SRT is trying to read Claude's settings file. This PR calls Claude using the `-c` flag

CleanShot 2026-07-07 at 11 22 15@2x

Related to [MD-2265: SRT sandbox errors without helpful message](https://linear.app/octopus/issue/MD-2265/srt-sandbox-errors-without-helpful-message)

### DIFF
--- a/source/Calamari.AiAgent.Tests/ClaudeCodeBehaviour/ClaudeCodeProcessStartInfoFixture.cs
+++ b/source/Calamari.AiAgent.Tests/ClaudeCodeBehaviour/ClaudeCodeProcessStartInfoFixture.cs
@@ -50,12 +50,15 @@ public class ClaudeCodeProcessStartInfoFixture
     [Test]
     public void ResolveInvocation_SandboxRuntimeMode_WrapsClaudeWithSrt()
     {
-        var builder = MinimalBuilder().WithSandboxMode(SandboxMode.SandboxRuntime).WithSandboxRuntimeSettingsPath(TestSandboxRuntimeSettingsPath);
+        var builder = MinimalBuilder().WithSandboxMode(SandboxMode.SandboxRuntime).WithSandboxRuntimeSettingsPath(TestSandboxRuntimeSettingsPath).WithPrompt("look! a \"quote\"");
 
         var (fileName, arguments) = ClaudeCodeProcessStartInfo.ResolveInvocation(builder);
 
         fileName.Should().Be("srt");
-        arguments.Should().StartWith($"--settings {TestSandboxRuntimeSettingsPath} claude --model");
+        // The claude invocation is passed as a single escaped argument to srt's -c flag.
+        arguments.Should().StartWith($"--settings {TestSandboxRuntimeSettingsPath} -c \"claude --model");
+        arguments.Should().Contain("-p \\\"look! a \\\\\\\"quote\\\\\\\"\\\"");
+        arguments.Should().EndWith("\"");
     }
 
     [Test]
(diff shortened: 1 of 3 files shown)


source/Calamari.AiAgent/ClaudeCodeBehaviour/ClaudeCommandArgsBuilder.cs
```diff
@@ -153,7 +153,7 @@ public string Build()
         return args.ToString();
     }
 
-    static string EscapeArg(string arg)
+    public static string EscapeArg(string arg)
```
flin-8, line 156: Extract to another class?

source/Calamari.AiAgent.Tests/ClaudeCodeBehaviour/ClaudeCodeProcessStartInfoFixture.cs
```diff
@@ -50,12 +50,15 @@ public void ResolveInvocation_BashMode_PassesSettingsFlag()
...
-        var builder = MinimalBuilder().WithSandboxMode(SandboxMode.SandboxRuntime).WithSandboxRuntimeSettingsPath(TestSandboxRuntimeSettingsPath);
+        var builder = MinimalBuilder().WithSandboxMode(SandboxMode.SandboxRuntime).WithSandboxRuntimeSettingsPath(TestSandboxRuntimeSettingsPath).WithPrompt("look! a \"quote\"");
 
         var (fileName, arguments) = ClaudeCodeProcessStartInfo.ResolveInvocation(builder);
 
         fileName.Should().Be("srt");
-        arguments.Should().StartWith($"--settings {TestSandboxRuntimeSettingsPath} claude --model");
+        // The claude invocation is passed as a single escaped argument to srt's -c flag.
+        arguments.Should().StartWith($"--settings {TestSandboxRuntimeSettingsPath} -c \"claude --model");
+        arguments.Should().Contain("-p \\\"look! a \\\\\\\"quote\\\\\\\"\\\"");
```
eddymoulton: This is a hilarious amount of escapes 😂 

It might be clearer if we write it as a raw string literal (`"""`), but it's not a big deal.

source/Calamari.AiAgent/ClaudeCodeBehaviour/ClaudeCommandArgsBuilder.cs
eddymoulton: Alternatively, we could escape internally and call something like `var claudeArgs = argsBuilder.BuildEscaped();`.

Doesn't over the binary path - but that is probably safer and/or can be escaped elsewhere?

eddymoulton (approved):
You could change the way you've done it (see the comment in the args builder) but this is good enough to go in.

source/Calamari.AiAgent.Tests/ClaudeCodeBehaviour/ClaudeCodeProcessStartInfoFixture.cs
flin-8: yeah raw looks better 😂

source/Calamari.AiAgent/ClaudeCodeBehaviour/ClaudeCommandArgsBuilder.cs
flin-8: I don't think the args builder should know how to escape the arguments (that's up to the caller).

I'd lean towards reusing this guy (in a later PR): https://github.com/OctopusDeploy/Calamari/blob/083324127461770e35c1bd9494d5f2c0e5bf8228/source/Calamari.Common/Features/Processes/CommandLine.cs#L193

PR #2065, opened 2026-07-08
Fix bug where cost always shows zero for Haiku models if prompt injection guard is ON

flin-8:
Server expects one service message per model type (unique DB key).

Prompt injection was sending its own service message, so the main CLI invocation's usage stats would be dropped if its model was the same.

This PR accumulates the usages and only writes it at the end.

Relates to [MD-2256: Cost always shows zero for Haiku models](https://linear.app/octopus/issue/MD-2256/cost-always-shows-zero-for-haiku-models)

### DIFF
--- a/source/Calamari.AiAgent/ClaudeCodeBehaviour/ClaudeCodeUsageReporter.cs
+++ b/source/Calamari.AiAgent/ClaudeCodeBehaviour/ClaudeCodeUsageReporter.cs
@@ -1,0 +1,77 @@
+using System.Collections.Generic;
+using System.Text.Json;
+using Calamari.Common.Plumbing.Logging;
+using Calamari.Common.Plumbing.ServiceMessages;
+using Octopus.Calamari.Contracts.ClaudeCode;
+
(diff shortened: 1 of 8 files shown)


source/Calamari.AiAgent/ClaudeCodeBehaviour/ClaudeCodeUsageReporter.cs
```diff
@@ -0,0 +1,91 @@
+using System.Collections.Generic;
+using System.Text.Json;
+using Calamari.Common.Plumbing.Logging;
+using Calamari.Common.Plumbing.ServiceMessages;
+using Octopus.Calamari.Contracts.ClaudeCode;
+
+namespace Calamari.AiAgent.ClaudeCodeBehaviour;
+
+/// <summary>
```
liam-mackie: Nit: These comments are pretty overdone, would prefer removing some unnecessary summaries.

liam-mackie (approved):
LGTM, minor nit on comments, but the bones are good.

PR #2067, opened 2026-07-09
Tests: Retry azure app service in different region

flin-8:
Right now we randomize the region so a re-run might succeed. Why not just retry in the same go?


APErebus (approved):
let's give it a go!

PR #2071, opened 2026-07-16
Add support for CloudFormationTemplate Parameter Overrides

zentron:
# Background
CloudFormationTemplates currently don't allow passing through parameter overrides directly and instead a parameter json file must be provided. This is unwieldy when just a few simple parameters need to be supplied and you dont want to have to create an Octopus specific template file.

## Result
A new `Octopus.Action.Aws.CloudFormationTemplateParameterOverrides`  variable can be provided as a key-value array of properties to be passed down into the final invocation.


source/Calamari.Aws/Integration/CloudFormation/Templates/CloudFormationParameterMerge.cs
```diff
@@ -0,0 +1,35 @@
+using System.Collections.Generic;
+using System.Linq;
+using Amazon.CloudFormation.Model;
```
liam-mackie, line 4:
```suggestion
using Amazon.CloudFormation.Model;
using System;
```
You'll need to use System to properly use `ArgumentException`

source/Calamari.Aws/Integration/CloudFormation/Templates/CloudFormationParameterMerge.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+            primary ??= Enumerable.Empty<Parameter>();
+            overrides ??= Enumerable.Empty<Parameter>();
+
+            var merged = new Dictionary<string, Parameter>();
+
+            foreach (var parameter in primary)
+            {
+                if (string.IsNullOrWhiteSpace(parameter?.ParameterKey))
+                    throw new ArgumentException("CloudFormation parameter key must not be null or empty.", nameof(primary));
```
liam-mackie: Nit: consider changing the argument error to throw with a message that clearly indicates this is the user's issue, not an Octopus issue

liam-mackie (approved):
LGTM, minor nits but should be fine!

source/Calamari.Aws/Integration/CloudFormation/Templates/CloudFormationParameterMerge.cs
liam-mackie: Never mind, you already did!

PR #2072, opened 2026-07-16
Robe/remove old tests

zentron:
Octopus for some time has required the use of Powershell 5, so we can assume in our tests that it is present.

Similarly we have long since deprecated the use of AzureRM powershell cmdlets, so removing its presence from a test. A later change will remove the code that even checks for it in the context scripts.


source/Calamari.AzureScripting.Tests/AzurePowershellCommandFixture.cs
```diff
@@ -33,13 +33,11 @@ public async Task Setup()
 
         [Test]
         [WindowsTest]
-        [RequiresPowerShell5OrAbove]
         public async Task ExecuteAnInlineWindowsPowerShellScript()
         {
             var psScript = @"
 $ErrorActionPreference = 'Continue'
 az --version
-Get-AzEnvironment
```
APErebus, line 42: We do still support the newer az cmdlets? Are we sure we want to remove these?

zentron: Good catch, I may have been a little over eager when I started removing the older usage of `Get-AzureEnvironment`. Perhaps that may just need the upgrade

source/Calamari.AzureResourceGroup.Tests/Calamari.AzureResourceGroup.Tests.csproj
```diff
@@ -32,4 +32,6 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
   </ItemGroup>
+
+  <Import Project="$(MSBuildProjectDirectory)/../IncludeDotNetScript.targets"/>
```
zentron, line 36: As a result of some of the project moves it looks like this file is sometimes not available during a test

APErebus (approved):

PR #2073, opened 2026-07-16
Robe/consolidate claude

zentron:
Now that we shell directly into claude code cli rather than using libraries, there is no need to seperate out the execution from the main calamari exe.


source/Calamari/Program.cs
```diff
@@ -40,7 +38,7 @@ namespace Calamari
 {
     public class Program : CalamariFlavourProgram
     {
-        protected Program(ILog log) : base(log)
+        public Program(ILog log) : base(log)
```
eddymoulton: Seems like an odd change? Do you know why it's needed?

eddymoulton (approved):
Quick question about one line, but otherwise good.

I'll ping @APErebus in slack as well, he had to make a change for the calamari docker image so we should check that we don't need to unmake a change too.

source/Calamari/Program.cs
zentron: I think that was a holdover from some other test changes. I think that should be fine to revert 👍

PR #2078, opened 2026-07-21
Fix reported KOS stabilisation issues

liam-mackie:
When KOS was running against a deployment with a HPA, there were 2 reported issues:

1. We never stabilised, due to HPAs constantly adjusting replica counts
2. The check took a long time (3min+) to run on clusters with many resources

This change aligns most workload-level resources with the logic that GitOps engine uses, as they have worked out more edge cases than we have. This leads to the checks being more reliable, and we also match KOS with KLOS, which implements the GitOps engine entirely.

We also add a cache for child objects to each check (invalidated for each check run) so that we don't list _all_ resources in the cluster multiple times for any given check, which caused slow check speeds.


tleed5 (approved):
Ship it :shipit:

PR #2082, opened 2026-07-30
ThrowIfPathTraversalAttempted handle linux better

NickJosevski:
ThrowIfPathTraversalAttempted compared the resolved destination against the extraction root using OrdinalIgnoreCase. On a case-sensitive filesystem that accepts ../../work/APP/evil as being "inside" a root of /work/App, when it actually escapes it.

Ordinal on Linux, OrdinalIgnoreCase on Windows and macOS. This mirrors how the BCL itself compares paths (PathInternal.StringComparison).

Verified on linux/amd64: /tmp/Octopus/WORK and /tmp/Octopus/Work are confirmed distinct directories

### DIFF
--- a/source/Calamari.Common/Features/Packages/PackageExtractorUtils.cs
+++ b/source/Calamari.Common/Features/Packages/PackageExtractorUtils.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.IO;
+using Calamari.Common.Plumbing;
 using Calamari.Common.Plumbing.Logging;
 using Polly;
 using Polly.Retry;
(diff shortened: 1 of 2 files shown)


APErebus (changes requested):
There is a lot of unnecessary code changes. The actual change should be to always set the relevant stringcomparison based on the environment OS.

That can be done as a single line check in the existing `ThrowIfPathTraversalAttempted`, rather than all the extra methods and stuff

NickJosevski:
@APErebus simpler now

APErebus (approved):

PR #2087, opened 2026-07-30
Update SharpCompress to 0.49.1 and re-arm its vulnerability alerting (SF-1864)

NickJosevski:
Resolves [SF-1864](https://linear.app/octopus/issue/SF-1864/update-sharpcompress-version-in-calamari).

CVE advisory's says `<= 0.47.4` range impacted. 

0.49.1 is where Octopus Server is at

0.49.0 flipped `WriterOptions.LeaveStreamOpen` from `true` to `false`, so a writer now closes the stream it was handed. 

Remove `<NoWarn>NU1902</NoWarn>`, so we get the warnings again and decide then if we want it back, or we fix right away.

## Was not vulnerable

- vulnerable path is the **archive-level** `IArchive.WriteToDirectory()`. Calamari iterates entries itself and calls the **per-entry** APIs — the latter named in the advisory as already guarded.
- `ThrowIfPathTraversalAttempted` independently resolves and bounds-checks every entry key, directory entries included, before anything is written.
- The TAR symlink escalation is blocked because `PackageExtractionOptions` supplies a `WriteSymbolicLink` handler that **refuses** to create links.

## Automated tests

`ExtractBlocksZipSlipViaDirectoryEntry` crafts a zip containing a traversing *directory* entry and asserts the extractor throws and nothing lands outside the root. It **passes on 0.37.2 too** — so it evidences that *Calamari's own guard* blocks this vector.

## How to review this PR

**Do not merge on green CI alone.** #1913 passed Calamari's full suite, merged, then broke Server unpacking a consolidated package (`An item with the same key has already been added` in `ExtractCalamariPackage`). This is the one area of the repo where the suite is a known-inadequate gate.

Verification done so far:
- [x] CI green across all legs (as #2086, same commit).
- [x] **Shipped linux-x64 artifact verified** — all six archive formats extract correctly from the packaged self-contained binary on a clean Linux image, with a corrupt-archive negative control and a forced-limit positive control proving `ArchiveFactory.OpenArchive` works on 0.49.1. See the [verification comment](#issuecomment-5129337781).

Before this leaves draft:
- [ ] **Exercise the GitHub package downloader path with credentials** — `GitHubPackageDownloader.DeNestContents` is the one shipping site the 0.49.0 `LeaveStreamOpen` flip lands on, and it is the only genuinely uncovered path. Best done via a deployment sourced from a GitHub repository feed.


NickJosevski:
## Verification against the shipped linux-x64 artifact

Ran the packaged Calamari — not `dotnet test` output — on a clean Linux image, because the tests build a different artifact from what ships.

**Setup:** published `source/Calamari` with the real settings (`-r linux-x64 --self-contained`), confirmed the output carries `SharpCompress.dll` **0.49.1**, and ran it under `mcr.microsoft.com/dotnet/runtime-deps:8.0` (linux/amd64). No SDK in the image, so this exercises the self-contained bundle rather than SDK-resolved assemblies. The binary reports commit `4f909e58ab92…`, matching this branch.

### Every supported format extracts correctly

| Format | Result |
|---|---|
| `.zip` | PASS — content + nested directory verified |
| `.tar` | PASS |
| `.tar.gz` | PASS |
| `.tar.bz2` | PASS |
| `.nupkg` | PASS |
| `.whl` | PASS (extracts to `1.0.0-py3-none-any/` — the wheel tag becomes the version) |

Driven via `deploy-package --package=<file>`, asserting on extracted file content including a nested directory, not just exit codes.

### Controls, so the above isn't vacuous

**Negative control** — a deliberately corrupt archive was correctly rejected:

```
SharpCompress.Common.ArchiveException: Failed to locate the Zip Header
```

That exception type confirms the shipped 0.49.1 assembly is doing the work, not some fallback path.

**Positive control for the highest-risk API.** The three decorators that call `ArchiveFactory.OpenArchive()` — SharpCompress's runtime format-sniffing entry point, and the API most exposed by an 11-minor-version jump — turned out to be **opt-in** behind `Octopus.Calamari.ArchiveLimits.*`, so an ordinary extraction never touches them. Re-ran with limits enabled:

- Realistic thresholds → all four formats extracted with **zero exceptions**.
- `MaximumUncompressedSize` forced to `5` → extraction correctly aborted with `ArchiveLimitException`, `files_extracted=0`, on both `.zip` and `.tar.gz`.

So `ArchiveFactory.OpenArchive` works on 0.49.1 in the shipped bundle, and the archive-limit protections still fire.

### Correction to earlier reasoning in this thread

I previously argued the packaged-artifact check mattered because Calamari ships **trimmed**, and trimming plus a library upgrade could silently remove a reflection-reachable path. **That is wrong.** Reading `build/Build.PackageCalamariProjects.cs`, the main Calamari flavours publish with `EnableSelfContained()` only — `PublishSingleFile` and `PublishTrimmed` apply solely to `docker-credential-octopus`, which has zero SharpCompress references.

There is therefore **no trim risk** for this change. The check retained value for a different reason (it runs the real shipped assemblies on real Linux), but the original justification does not hold.

### Still not covered

- **`GitHubPackageDownloader.DeNestContents`** — the one shipping site the 0.49.0 `LeaveStreamOpen` default flip actually lands on. Its fixtures need GitHub credentials (403 locally). Best exercised by a deployment sourced from a **GitHub repository feed**; assessed by code reading only so far.
- ARM RIDs (CI covers Windows and x64 Linux).
- Large or pathological archives — the ~6s vs ~500ms extraction perf question from the original discussion remains open and unbenchmarked.

Given the trimming concern evaporated, a full local Server run is now **optional** for this change rather than important. The genuinely uncovered path is the GitHub feed one.

source/Calamari.Common/Features/Packages/TarBzipPackageExtractor.cs
```diff
@@ -17,7 +17,7 @@ public TarBzipPackageExtractor(ILog log)
 
         protected override Stream GetCompressionStream(Stream stream)
         {
-            return new BZip2Stream(stream, CompressionMode.Decompress, false);
+            return BZip2Stream.Create(stream, CompressionMode.Decompress, false, false);
```
YuKitsune: Nit: Use named arguments to avoid ambiguity:

```suggestion
            return BZip2Stream.Create(stream, CompressionMode.Decompress, decompressConcatenated: false, leaveOpen: false);
```

YuKitsune:
I've run this locally and tested the GitHub path. Works as expected!
I'm running on macOS, so that should also cover the ARM path you mentioned.

> Large or pathological archives.

Can you add some more context to this?

NickJosevski:
close the gap  `GitHubPackageDownloader.DeNestContents` is the one shipping site the 0.49.0 `LeaveStreamOpen` default flip is there

did `BZip2Stream.Create` nit

## "large or pathological archives"


**1. Performance, which is a pre-existing open question.** In the [original thread](https://octopusdeploy.slack.com/archives/C03SG0LFJHX/p1777933912502139) Orion noted extracting from the ConsolidatedPackage took ~6s where ~500ms would be expected, and that there's no benchmark anywhere in the repo to measure against.

That was never resolved.
The fixtures use tiny archives — a couple of files, a few hundred bytes — so **this PR neither improves nor regresses that, and provides no evidence either way.**

**2. Adversarial inputs, where we have partial coverage.**
The archive-limit decorators (`EnforceDecompressionLimitDecorator`, `EnforceCompressionRatioDecorator`) exist for zip bombs and I did verify they still fire on 0.49.1 — forcing `MaximumUncompressedSize` to 5 bytes correctly aborted with `ArchiveLimitException`.

But they're **opt-in** behind `Octopus.Calamari.ArchiveLimits.Enabled`

NickJosevski:
I'll tidy up what claude posted later

YuKitsune (approved):

PR #2088, opened 2026-07-30
Go to NSubstitute 4.2.2, why not later, let me check...

NickJosevski:
All nine test projects now reference NSubstitute 4.2.2. It was 2.0.3


NickJosevski:
Better to go to 6.0.0 see here: https://github.com/OctopusDeploy/Calamari/pull/2090

PR #2092, opened 2026-08-01
Pin the bundled .NET runtime so it is a deliberate choice

NickJosevski:
## Background

Calamari publishes self-contained. Every artifact carries its own copy of the .NET runtime. Customer vulnerability scanners report against that bundled runtime.

Nothing in the repo chose the runtime version. The version came from whichever runtime pack the build agent's SDK happened to ship. Two builds of the same commit on differently patched agents would ship different runtimes.

## Results

The bundled runtime version is now a single constant in `build/BundledRuntime.cs`. The constant applies at both self-contained publish sites and at restore.

Restore needs the property too. Publish runs with `--no-restore`. Without the pinned pack downloaded during restore, publish fails with NETSDK1112 on any agent that has not cached the pack. `--no-restore` is kept. A comment there records that it prevents `project.assets.json` contention across parallel per-RID publishes.

## Reducing risk

- Verified on a cold cache in both directions. The NETSDK1112 failure was reproduced before the fix was confirmed.
- Verified the pin overrides the SDK default. SDK 8.0.421 defaults to runtime 8.0.27. Published output carries 8.0.29.
- 82 extraction tests pass on a machine whose newest installed runtime is 8.0.27.

## How to review this PR

The constant needs bumping whenever .NET ships a security patch. A stale value silently ships a vulnerable runtime. Worth a Renovate rule after Renovate is re-enabled.

### DIFF
--- a/build/Build.PackageCalamariProjects.cs
+++ b/build/Build.PackageCalamariProjects.cs
@@ -114,6 +114,9 @@ public partial class Build
                                                   .SetVersion(NugetVersion.Value)
                                                   .SetInformationalVersion(OctoVersionInfo.Value?.InformationalVersion)
                                                   .EnableSelfContained()
+                                                  // Pin the bundled runtime so it is a deliberate choice rather than a
+                                                  // property of whichever SDK the agent has. See BundledRuntime.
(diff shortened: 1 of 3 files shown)


NickJosevski:
Not sure about this choice...

PR #2101, opened 2026-08-05
Fall back to USEast1 aws region when region is not set

grace-rehn:

# Background

AWS OIDC accounts stopped working for Kubernetes API targets on EKS after the AWS SDK v4 upgrade in 2026.2, with health checks and deployments both failing with `Cannot find supported authentication schema`. A Kubernetes API target has nowhere to set an AWS region, so Calamari falls back to the `aws-global` region which fails for unsigned STS requests.

# Results

- Fixes health checks and deployments for Kubernetes API targets using AWS OIDC accounts, and ECS targets with a blank region.
- Falls back to `us-east-1` instead of `aws-global`. A configured region still takes precedence.
- No variables are contributed, so per-step regions are untouched.
- Verified on an EKS cluster in `ap-southeast-2` with no region set: health check, raw YAML deployment, access-key account and assume-role.

Fixes HPY-1542

## Before

Screenshot 2026-08-05 at 10 35 13 AM

## After

Screenshot 2026-08-05 at 10 57 58 AM

### DIFF
--- a/source/Calamari.CloudAccounts/AwsEnvironmentGeneration.cs
+++ b/source/Calamari.CloudAccounts/AwsEnvironmentGeneration.cs
@@ -136,11 +136,11 @@ namespace Calamari.CloudAccounts
         {
             get
             {
-                // "aws-global" routes to the global STS endpoint (sts.amazonaws.com), preserving SDK v3 behaviour.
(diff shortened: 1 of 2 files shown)


HuyPhanNguyen (approved):
LGTM

PR #2104, opened 2026-08-05
Add set environment state helper functions

geofflamrock:
## Context

We are adding support for setting environment state within a deployment/runbook run. State will be persisted across deployments for the project/environment/(tenant) and will primarily be used across provisioning, deployment and deprovisioning of ephemeral environments.

State will be set using a new service message in the format `set-environmentstate -key "{key}" -value "{value}" -sensitive "{sensitive}" -type "{type}".

Different types of state will be supported: 
- `State`: General state to be persisted.
- `Url`: Url to be associated and shown with the environment.

## Changes

Adds helper functions to PowerShell and Bash for setting environment state, mirroring the functions for setting output variables.

- PowerShell: `Set-EnvironmentState -Key {key} -Value {value} -Sensitive`.
- Bash: `set_environmentstate {key} {value} [-sensitive]`.

These functions only support setting general state and don't allow to send the type, we'll be adding more functions for URLs separately.

## Reviewing and testing

Automated tests have been added, we'll need to pair and test this with the work in Octopus Server to handle service messages.

### DIFF
--- a/source/Calamari.Common/Features/Scripting/Bash/Bootstrap.sh
+++ b/source/Calamari.Common/Features/Scripting/Bash/Bootstrap.sh
@@ -113,6 +113,37 @@ function set_octopusvariable
   echo $MESSAGE
 }
 
+#	---------------------------------------------------------------------------
+# Function for setting environment state
+#   Accepts 3 arguments:
+#     string: value of the key of the environment state
+#     string: value of the value of the environment state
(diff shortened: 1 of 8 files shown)


yangguoif (approved):
🎉 woohoo, looks great! 

Just need to note that we probably want to hold off the merge / publish until the change in Octopus server is rolled out.

adam-mccoy (approved):

PR #2118, opened 2026-08-06
build.sh SDK bootstrap tells you what's actually missing

NickJosevski:
## Background

`build.sh` uses the machine's `dotnet` if `dotnet --version` succeeds, and otherwise downloads an
SDK. That fallback shells out to **curl** and **perl**.
Neither is guaranteed in our Linux containers, so you get:

```
./build.sh: line 27: perl: command not found
Process exited with code 127
```

## Changes

1. **Log why the bootstrap ran**  => on error list SDKs
2. **Use `wget` when no `curl`** ==> or says if can't find either
3. **Use `sed` when no `perl`**

## Testing

| Environment | Before | After |
|---|---|---|
| `amazonlinux:2` (curl, no perl) | `perl: command not found`, exit 127 | bootstrap completes |
| `debian:12` + wget (no curl) | `curl: command not found`, exit 127 | bootstrap completes |
| neither curl nor wget | `curl: command not found`, exit 127 | `Unable to download … - neither curl nor wget is available`, exit 1 |
| SDK 8 present, `global.json` pins `10.0.302` | exit 127 from a missing tool | `'dotnet --version' failed - global.json likely pins an SDK that is not installed (found: 8.0.423)` |

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -24,17 +24,53 @@ export DOTNET_MULTILEVEL_LOOKUP=0
 ###########################################################################
 
 function FirstJsonValue {
-    perl -nle 'print $1 if m{"'"$1"'": "([^"]+)",?}' <<< "${@:2}"
+    # Stock Nuke shells out to perl. The Amazon Linux execution container does not ship it, so fall
+    # back to sed rather than dying with "perl: command not found".
+    if command -v perl &>/dev/null; then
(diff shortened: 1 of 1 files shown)


zentron (approved):
nice, ive seen that once or twice before but not spent enough time looking at it. Thanks ~Claude~ @NickJosevski

PR #2127, opened 2026-08-10
[DO NOT MERGE] CI vehicle — net10 superset + dotnet-script 2.0.1

NickJosevski:
Not for review or merge. Sibling of #2125; exists so TeamCity publishes a second pre-release for A/B testing on a pre-production cloud instance.

Identical to #2125 (net10 superset, includes `DOTNET_ROLL_FORWARD=Major`) plus one commit: **bundled dotnet-script 1.6.0 → 2.0.1**.

## Why the bump is needed as well as roll-forward

Roll-forward gets dotnet-script *launching* on a net10-only machine. It does not make it work. Once launched, `#r "nuget: <framework-provided package>"` fails:

```
Error loading lock file .../net10.0/obj/project.assets.json :
  Cannot compare the value of a token type 'Number' to text.
The lock file ... does not contain a runtime target.
```

.NET 10 SDK package pruning emits **NU1510** — the same warning already suppressed in `Calamari.Common.csproj`. dotnet-script 1.6.0 bundles NuGet.ProjectModel **6.10.0**, which cannot parse the `project.assets.json` the .NET 10 SDK writes when that pruning applies.

> **Correction.** An earlier revision of this description claimed the specific trigger was a numeric `"warningLevel"` in the assets file's top-level `logs` array. That is **wrong** — I tested `RestSharp 110.2.0`, which produces an assets file with the identical shape (a `logs` entry with `"warningLevel": 1`, for NU1902) and restores **fine** on net10. So the numeric `warningLevel` is not the trigger. What is reproducible, and is what this PR rests on, is the behaviour below: framework-provided packages fail on 6.10.0 and pass on 6.14.3. The precise field that breaks the 6.10.0 parser has not been isolated.

2.0.1 bundles NuGet.ProjectModel **6.14.3** (upstream dotnet-script#791) and parses it. Verified against a net10-only runtime:

```
1.6.0 + roll-forward  → Cannot compare the value of a token type 'Number' to text
2.0.1 + roll-forward  → runtime : .NET 10.0.10 / OK : {"ok":true}
2.0.1, no roll-forward → You must install or update .NET to run this application
```

The published 2.0.1 zip is still `tfm: net8.0` requiring `Microsoft.NETCore.App 8.0.0`, so this does **not** replace the roll-forward change — both are required.

## Behaviour changes to watch in pre-production

Both are upstream 2.0 defaults, left unpinned deliberately so testing exercises what would actually ship:

- **Roslyn 4.11.0 → 5.0.0-2.final** (prerelease). `ScriptCompiler` hardcodes `LanguageVersion.Preview`, so the language ceiling available to customer scripts moves with it.
- **Isolated assembly load context becomes the default.** `--isolated-load-context` was removed and inverted to `--disable-isolated-load-context`; Calamari passes neither.

:warning: Does this change require a corresponding Server Change? No — CI vehicle only.


source/Calamari.Scripting/DotnetScript/dotnet-script.1.6.0.zip
NickJosevski, line 1: here

PR #2131, opened 2026-08-11
Upgrade Autofac and remove the dead net462 reference assemblies

NickJosevski:
| Commit | What |
|---|---|
| Remove the dead .NET Framework reference assemblies | `Microsoft.NETFramework.ReferenceAssemblies.net462`, unused |
| Upgrade Autofac 4.8.0 → 9.3.1 | No code changes needed |

`ContainerBuilder.Update()` is the headline Autofac 5 removal and was never used here.

**Neither change is required for .NET 10.** Autofac 4.8.0 builds with no warnings and resolves correctly on `net10.0` — measured. This is hygiene: 4.8.0 is from 2018 and only ships `net45` and `netstandard1.1` assets. Doing it while Calamari still targets net8 means a green net8 build proves the bump in isolation, ahead of the target-framework flip.

-one new warning is `CS8714` at `Calamari.Common/Plumbing/Pipeline/Resolver.cs:17` — Autofac 9's `Resolve<TService>` carries a `notnull` constraint that `TBehaviour` doesn't satisfy. `Calamari.Common` doesn't treat warnings as errors, so it builds.

Yes server will reference this when shipped.

### DIFF
--- a/source/Calamari.Common/Calamari.Common.csproj
+++ b/source/Calamari.Common/Calamari.Common.csproj
@@ -19,7 +19,7 @@
         <PackageReference Include="System.Security.Cryptography.Pkcs" Version="9.0.0" />
         <PackageReference Include="System.Text.Encoding.CodePages" Version="8.0.0" />
         <PackageReference Include="System.Threading.AccessControl" Version="4.3.0" />
-        <PackageReference Include="Autofac" Version="4.8.0" />
+        <PackageReference Include="Autofac" Version="9.3.1" />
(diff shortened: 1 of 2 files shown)


zentron (approved):
Feels pretty low risk.

Our usage of Autofac in Calamari tends to rely on mainly simplistic registrations, we generally dont weave much magic with multiple lifetimescopes, and the whole process is pretty short lived. If testing didnt turn anything up, I think this is safe to go out despire the major bump

PR #2147, opened 2026-08-20
Fix memory growth

liam-mackie:
This PR fixes the 2 main reasons we were seeing memory growth when running resource monitoring in environments with many resources:
1. We stopped caching the raw JSON and started caching the parse. This way, we aren't copying and retaining as much data.
2. All of the resources we made held its JObject in a field, and that JObject was still a node inside the list it came from, so we were retaining a bunch (I'm pretty sure because of `JToken.Parent`). The json is now a constructor parameter rather than a field, so once the objects are made we can let go of the list properly.

### DIFF
--- a/source/Calamari/Kubernetes/Integration/CaptureCommandOutput.cs
+++ b/source/Calamari/Kubernetes/Integration/CaptureCommandOutput.cs
@@ -14,39 +14,21 @@ namespace Calamari.Kubernetes.Integration
 
     public class CaptureCommandOutput : ICommandInvocationOutputSink, ICommandOutput
     {
-        readonly List<Message> messages = new List<Message>();
-        Message[] snapshot;
+        private readonly List<Message> messages = new List<Message>();
+        public Message[] Messages => messages.ToArray();
 
-        public Message[] Messages => snapshot ??= messages.ToArray();
-
-        public IEnumerable<string> InfoLogs
-        {
-            get
-            {
-                foreach (var message in Messages)
-                {
-                    if (message.Level == Level.Info)
-                        yield return message.Text;
-                }
-            }
-        }
+        public IEnumerable<string> InfoLogs => Messages.Where(m => m.Level == Level.Info).Select(m => m.Text).ToArray();
 
         public string MergeInfoLogs() => string.Join(Environment.NewLine, InfoLogs);
 
         public void WriteInfo(string line)
         {
-            Add(new Message(Level.Info, line));
+            messages.Add(new Message(Level.Info, line));
         }
 
         public void WriteError(string line)
         {
-            Add(new Message(Level.Error, line));
-        }
-
-        void Add(Message message)
-        {
-            messages.Add(message);
-            snapshot = null;
+            messages.Add(new Message(Level.Error, line));
         }
     }
 
--- a/source/Calamari/Kubernetes/Integration/CaptureCommandOutput.cs
+++ b/source/Calamari/Kubernetes/Integration/CaptureCommandOutput.cs
@@ -14,21 +14,39 @@ namespace Calamari.Kubernetes.Integration
 
     public class CaptureCommandOutput : ICommandInvocationOutputSink, ICommandOutput
     {
-        private readonly List<Message> messages = new List<Message>();
-        public Message[] Messages => messages.ToArray();
+        readonly List<Message> messages = new List<Message>();
+        Message[] snapshot;
 
-        public IEnumerable<string> InfoLogs => Messages.Where(m => m.Level == Level.Info).Select(m => m.Text).ToArray();
+        public Message[] Messages => snapshot ??= messages.ToArray();
+
+        public IEnumerable<string> InfoLogs
+        {
+            get
+            {
+                foreach (var message in Messages)
+                {
+                    if (message.Level == Level.Info)
+                        yield return message.Text;
+                }
+            }
+        }
 
         public string MergeInfoLogs() => string.Join(Environment.NewLine, InfoLogs);
 
         public void WriteInfo(string line)
         {
-            messages.Add(new Message(Level.Info, line));
+            Add(new Message(Level.Info, line));
         }
 
         public void WriteError(string line)
         {
-            messages.Add(new Message(Level.Error, line));
+            Add(new Message(Level.Error, line));
+        }
+
+        void Add(Message message)
+        {
+            messages.Add(message);
+            snapshot = null;
         }
     }
 
--- a/source/Calamari/Kubernetes/ResourceStatus/KubectlGet.cs
+++ b/source/Calamari/Kubernetes/ResourceStatus/KubectlGet.cs
@@ -1,7 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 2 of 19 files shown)


source/Calamari/Kubernetes/Integration/CaptureCommandOutput.cs
```diff
@@ -14,21 +14,39 @@ public interface ICommandOutput
 
     public class CaptureCommandOutput : ICommandInvocationOutputSink, ICommandOutput
     {
-        private readonly List<Message> messages = new List<Message>();
-        public Message[] Messages => messages.ToArray();
+        readonly List<Message> messages = new List<Message>();
```
flin-8: why's this class changing?

liam-mackie: Originally, I was trying to cache this so that all the `ToArray()`'s copying a bunch, but it turns out the real fix is to just only read the command output once lmao. I'll pull this change out.

flin-8: ok, what about resource retriever etc.? can they be undone too?

source/Calamari/Kubernetes/ResourceStatus/Resources/ResourceFactory.cs
```diff
@@ -32,7 +32,7 @@ public static class ResourceFactory
         public static IEnumerable<Resource> FromListJson(string json, Options options)
         {
             var listResponse = JObject.Parse(json);
-            return listResponse.SelectTokens("$.items[*]").Select(item => FromJObject((JObject)item, options));
+            return listResponse.SelectTokens("$.items[*]").Select(item => FromJObject((JObject)item, options)).ToList();
```
flin-8, line 35: 👍

source/Calamari/Kubernetes/Integration/CaptureCommandOutput.cs
liam-mackie: > ok, what about resource retriever etc.? can they be undone too?

Sorry, to be clear, I meant that it's the real fix _for this specific class of memory growth_. We still need to include changes to the resource retriever to cache the parsed JSON.

source/Calamari/Kubernetes/ResourceStatus/Resources/ResourceFactory.cs
flin-8: It'll be clearer if the method returns a readonlycollection

source/Calamari/Kubernetes/ResourceStatus/KubectlGet.cs
```diff
@@ -38,26 +38,46 @@ public KubectlGetResult AllResources(ResourceGroupVersionKind groupVersionKind,
 
         static KubectlGetResult ProcessResult(CommandResultWithOutput commandResult)
         {
-            return new KubectlGetResult(
-                commandResult.Output.InfoLogs.Join(string.Empty),
-                commandResult.Output.Messages.Select(msg => $"{msg.Level}: {msg.Text}").ToList(),
-                commandResult.Result.ExitCode);
+            // these payloads are a whole namespace of one kind, so we avoid copying them
```
flin-8, line 41: Does this class need to change?

source/Calamari/Kubernetes/ResourceStatus/ResourceRetriever.cs
```diff
@@ -50,7 +50,7 @@ public ResourceRetriever(IKubectlGet kubectlGet, ILog log)
         /// <inheritdoc />
         public IEnumerable<ResourceRetrieverResult> GetAllOwnedResources(IEnumerable<ResourceIdentifier> resourceIdentifiers, IKubectl kubectl, Options options)
         {
-            var childResourceCache = new Dictionary<(ResourceGroupVersionKind, string), KubectlGetResult>();
+            var childResourceCache = new Dictionary<(ResourceGroupVersionKind, string), ChildResourceLookup>();
```
flin-8, line 53: Does this class need to change?

source/Calamari/Kubernetes/ResourceStatus/KubectlGet.cs
```diff
@@ -38,26 +38,46 @@ public KubectlGetResult AllResources(ResourceGroupVersionKind groupVersionKind,
...
+
+            return new KubectlGetResult(resourceJson.ToString(), messages, commandResult.Result.ExitCode);
         }
     }
 
     public class KubectlGetResult
     {
+        readonly Message[] messages;
+        readonly IList<string> formattedOutput;
+
```
flin-8, line 59: The first constructor is only called by tests - we should delete it.

source/Calamari/Kubernetes/ResourceStatus/ResourceRetriever.cs
```diff
@@ -165,15 +176,53 @@ static void LogKubectlErrorIfFailed(KubectlGetResult getResult, Options options,
...
+
+            public static ChildResourceLookup From(IEnumerable<Resource> resources)
             {
-                foreach (var line in getResult.RawOutput)
+                var byOwnerUid = new Dictionary<string, List<Resource>>();
+
+                foreach (var resource in resources)
                 {
-                    log.Verbose(line);
+                    foreach (var ownerUid in resource.OwnerUids ?? Enumerable.Empty<string>())
```
flin-8, line 204: Take first parent or go through all?

flin-8: I guess it's not changing behaviour anyway 👍

flin-8 (approved):
LGTM

Just checking they are on 2026.3?

zentron (approved):

source/Calamari/Kubernetes/ResourceStatus/KubectlGet.cs
liam-mackie: I'll consider it in a follow-up, but I'll merge for now :)

source/Calamari/Kubernetes/ResourceStatus/ResourceRetriever.cs
liam-mackie: Nope - we already checked all parents :)

source/Calamari/Kubernetes/ResourceStatus/ResourceRetriever.cs
liam-mackie: Yep :(

source/Calamari/Kubernetes/ResourceStatus/KubectlGet.cs
liam-mackie: Yep :(

source/Calamari/Kubernetes/ResourceStatus/Resources/ResourceFactory.cs
liam-mackie: Didn't see the point changing the API - it's not technically read-only, but I _could_ return a List. Leaving for now, will follow up in future maybe.

PR #2149, opened 2026-08-24
Report CloudFormation throttling as a rate limit, not an unknown error

NickJosevski:
No server change required.

# Background

`DescribeCloudFormationChangeSetConvention` only special-cases `AccessDenied`; everything else falls through to `UnknownException`. So AWS throttling reports as:

```
Error: Calamari.Aws.Exceptions.UnknownException: An unrecognized exception was thrown
       while describing the CloudFormation change set.
Error:  ---> Amazon.CloudFormation.AmazonCloudFormationException: Rate exceeded
       ... ~60 more lines of SDK stack trace ...
```

Throttling isn't unrecognized — it's documented, usually transient, and has a clear user action. `UnknownException` also derives from plain `Exception` rather than `CommandException`, which is why the log gets the stack dump.

# Results

Throttling now throws `RateExceededException` (deriving from `CommandException`, matching `PermissionException`) with a one-line explanation instead of a stack trace.

# How to review this PR

**The duplicated error-code list.** The SDK has this exact set as `RetryPolicy.ThrottlingErrorCodes`, but it's an *instance* property — using it doesn't compile (`CS0120`), and `IsThrottlingError` is likewise an instance method. Hence the local copy, with HTTP 429 as a backstop. 

**Scope.** Only the describe-changeset path. The same `UnknownException` catch-all exists in other CloudFormation conventions.

Adds no retries; PR #2148 raises the SDK retry budget, which is what should reduce how often this happens

### DIFF
--- a/source/Calamari.Aws/Deployment/Conventions/DescribeCloudFormationChangeSetConvention.cs
+++ b/source/Calamari.Aws/Deployment/Conventions/DescribeCloudFormationChangeSetConvention.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Net;
 using System.Threading.Tasks;
 using Amazon.CloudFormation;
 using Amazon.CloudFormation.Model;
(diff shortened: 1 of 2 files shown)


flin-8 (approved):
LGTM

NickJosevski:
For context on "the SDK already retries throttled calls": Standard retry mode backs off **exponentially with jitter**, capped at **20s** per attempt, before this branch is ever reached.

So by the time a user sees `RateExceededException`, AWS has already been retried several times over tens of seconds — which is why the message says the throttling is transient rather than telling them to retry immediately.

PR #2153, opened 2026-09-07
Add ArgoCD Step annotation support

liam-mackie:
This PR adds support for an optional step annotation for ArgoCD applications. By default, with no step annotations, any step may act on the application/source.


source/Calamari/ArgoCD/DeploymentScope.cs
```diff
@@ -5,12 +5,17 @@
 
 namespace Calamari.ArgoCD;
 
-public record DeploymentScope(ProjectSlug Project, EnvironmentSlug Environment, TenantSlug? Tenant)
+public record DeploymentScope(ProjectSlug Project, EnvironmentSlug Environment, TenantSlug? Tenant, StepSlug? Step = null)
```
flin-8: Can the step ever be null?

liam-mackie: Originally, the `step` scope was behind a toggle, so it could be, but I spoke to Rob and we decided against a feature toggle (it would have been super confusing to roll forward/back, and isn't really a reversible decision). I don't _think_ there's any cases where it's nullable now, so I'll look at pulling that back out.

liam-mackie: Taking a look, there's a possibility of the step slug being null _only_ if the process was persisted before slugs, and never edited and saved again. Since Argo steps got released _after_ slugs, this is indeed a nonexistent problem :D

zentron (approved):

PR #2154, opened 2026-09-08
Run gzip directly in BuildDockerImages instead of through pwsh

NickJosevski:
[Build: Docker images](https://build.octopushq.com/buildConfiguration/TeamModernDeployments_Calamari_VNext_Build_BuildDockerImages/24122436) failing... but why now...

`BuildDockerImages` launched `pwsh` for one thing — running `gzip` on the OCI tar. The agent's `pwsh` is a dotnet global tool needing .NET 10, while `build.sh` bootstraps a .NET 8-only SDK into `.nuke/temp/dotnet-unix` and puts it on `PATH`:

```
App: /root/.dotnet/tools/pwsh
Framework: 'Microsoft.NETCore.App', version '10.0.0' (x64)
.NET location: .../.nuke/temp/dotnet-unix
The following frameworks were found:
  8.0.30
```

**Why it started failing on 2026-09-07:** the `nautilus-linux` agent image changed between the last green build (24081109, Sep 3) and the first red one (24122436, Sep 7) — bash 5.1.4 -> 5.2.21, and the system SDK it reports went 6.0.428 -> 10.0.400. Both logs show the same `./build.sh BuildDockerImages` and the same net8 SDK bootstrap, `BuildDockerImages.xml` is untouched in the settings diff between the two builds, and no Calamari commit landed in that window. The agent's `pwsh` is the only variable that moved.

So: call `gzip` directly. This was the build's only use of `PowerShellTasks`.

Also stops a silent failure — `pwsh -Command` exits 0 regardless of the native exit code, so a failed gzip used to surface later as a missing artifact in `PublishArtifacts`. A Nuke `Tool` asserts a zero exit code.

Verified locally `/usr/bin/gzip` from `PATH`, `-k` still keeps the `.tar` next to the `.gz`, and a forced failure throws `ProcessException: Process 'gzip' exited with code 1`.

### DIFF
--- a/build/Build.Docker.cs
+++ b/build/Build.Docker.cs
@@ -1,12 +1,15 @@
 ﻿using Calamari.Build.Utilities;
 using JetBrains.Annotations;
+using Nuke.Common.Tooling;
 using Nuke.Common.Tools.Docker;
-using Nuke.Common.Tools.PowerShell;
 
 namespace Calamari.Build;
 
 public partial class Build
 {
+    //Resolved from PATH so a missing gzip fails by name, rather than as a mystery exit code
+    static Tool Gzip => ToolResolver.GetPathTool("gzip");
+
     [PublicAPI]
     Target BuildDockerImages =>
         d =>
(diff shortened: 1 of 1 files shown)


APErebus (approved):
That is a much better way of invoking it. I didn't have enough Nuke knowledge to know how to invoke tools directly

PR #2164, opened 2026-09-17
Stop testing bash scripts on Windows

APErebus:
Retargeted to `2026.3`

It doesn't make sense to test bash on Windows as we don't support that in the product


flin-8 (approved):
LGTM

PR #2165, opened 2026-09-17
Update WindowsFxVersion for Azure App Service tests

APErebus:
We are getting errors on the LTS branches `2026.1` and `2026.2`

> The parameter WindowsFxVersion has an invalid value. The specified Windows Containers image is not compatible with Azure App Service. We currently support Windows Container images with tag suffixes from windowsservercore-ltsc2019 to windowsservercore-ltsc2022 or from nanoserver-1809 to nanoserver-ltsc2022.

I think I'll need to make the same change in `2026.3` manually as it was moved around.

I chose not to do this on the `2026.1` branch as this is EOL very soon and I don't anticipate any required Calamari changes before EOL


flin-8 (approved):
LGTM

PR #2166, opened 2026-09-17
Run gzip directly in BuildDockerImages instead of through pwsh

APErebus:
Backports #2154 as the LTS branch suffers from the same issue


flin-8 (approved):
LGTM

